PR #281, opened 2017-06-16
Approximate total derivatives of Group

Kenneth-T-Moore:
1. User can specify that a Group's derivatives be approximated (only by finite difference at present). The resulting Jacobian is used in all derivative calculation above the group's scope. Newton solver at the top level of the group bypasses this to prevent infinite recursion.
2. Fixed a bug where a linear solver's linearize was called instead of the nonlinear solver's linearize.
3. An unnecessary layer of tuples in "of" and "wrt" in `comput_total_derivs` was removed.


kmarsteller:
Did this PR get messed up by the repo going public?

Kenneth-T-Moore:
Yeah, it lost its source repo, but I am withdrawing it. This stuff will come in with the CS pull request.

kmarsteller:
sorry!

PR #296, opened 2017-07-10
Case Recording Alpha

kmarsteller:


openmdao/core/driver.py
```diff
@@ -1,10 +1,13 @@
...
-
 import numpy as np
 
+from openmdao.devtools.problem_viewer.problem_viewer import _get_viewer_data
+from openmdao.utils.record_util import create_local_meta
 from openmdao.utils.options_dictionary import OptionsDictionary
+from openmdao.recorders.recording_manager import RecordingManager
+from openmdao.recorders.recording_iteration_stack import Recording
+
+from six import iteritems
```
Kenneth-T-Moore: Pretty sure this should be the earliest import here, following PEP guidelines

openmdao/core/driver.py
```diff
@@ -29,12 +32,22 @@ class Driver(object):
         Contains all objective info.
     _responses : dict
         Contains all response info.
+    _rec_mgr : <RecordingManager>
+        object that manages all recorders added to this driver
+    _model_viewer_data : dict
+        structure of model, used to make n2 diagram.
+    iter_count : int
```
Kenneth-T-Moore: should do these before the private attributes. Also, begin with upper case.

openmdao/core/driver.py
```diff
@@ -79,20 +113,42 @@ def _setup_driver(self, problem):
...
+        self._rec_mgr.record_metadata(self)
+
+    def get_design_var_values(self, filter=None):
         """
         Return the design variable values.
 
         This is called to gather the initial design variable state.
 
+        Parameters
+        ----------
```
Kenneth-T-Moore: So, the driver recorder uses this method? I guess that is okay, as long as the user is aware that this is the driver-scaled value. Still, that is probalby appropriate for recording on the driver. (same for cons and objs)

openmdao/core/driver.py
```diff
@@ -376,3 +473,21 @@ def get_req_procs(self, model):
...
+
+    def record_iteration(self):
+        """
+        Record an iteration of the current Driver.
+        """
+        metadata = create_local_meta(self._get_name())
+        self._rec_mgr.record_iteration(self, metadata)
+
+    def _get_name(self):
+        """
```
Kenneth-T-Moore: Why a method instead of an attribute? I guess it's no big deal though.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -351,20 +360,24 @@ def _objfunc(self, dv_dict):
         """
         model = self._problem.model
         fail = 0
+        metadata = self.metadata
 
         try:
             for name in self._indep_list:
                 self.set_design_var(name, dv_dict[name])
 
-            # print("Setting DV")
```
Kenneth-T-Moore: I'd really like my debug statements to stay in until we are fully stable. I use them all the time.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -376,9 +389,6 @@ def _objfunc(self, dv_dict):
             fail = 1
             func_dict = {}
 
-        # print("Functions calculated")
-        # print(dv_dict)
```
Kenneth-T-Moore: These ones too, please.

openmdao/recorders/base_case_reader.py
```diff
@@ -0,0 +1,44 @@
...
+    filename : str
+        The name of the file from which to instantiate the case reader.
+
+    Attributes
+    ----------
+    format_version : int
+        An integer representation of the format version in the recorded file.
+    filename : str
+        The name of the file from which the recorded cases are to be loaded.
+    parameters : dict
```
Kenneth-T-Moore: These lasthree attributes don't match the ones in initialize. Also, missing 5 others.

openmdao/recorders/base_recorder.py
```diff
@@ -0,0 +1,294 @@
...
+        Tells recorder whether to record the relative error of a Solver.
+    options['record_solver_output'] :  bool(False)
+        Tells recorder whether to record the output of a Solver.
+    options['record_solver_derivatives'] :  bool(False)
+        Tells recorder whether to record the derivatives of a Solver.
+    options['includes'] :  list of strings("*")
+        Patterns for variables to include in recording.
+    options['excludes'] :  list of strings('')
+        Patterns for variables to exclude in recording (processed after includes).
+    """
```
Kenneth-T-Moore: docstring missing all attributes

openmdao/recorders/case.py
```diff
@@ -0,0 +1,221 @@
+"""
+A Case class.
+"""
+
+
```
Kenneth-T-Moore: Not sure how Case and its children passed lint. Params and attributes don't match the code in __init__.

openmdao/recorders/case.py
```diff
@@ -0,0 +1,221 @@
...
+                                         timestamp, success, msg)
+
+        self.inputs = inputs[0] if inputs.dtype.names else None
+        self.outputs = outputs[0] if outputs.dtype.names else None
+        self.residuals = residuals[0] if residuals.dtype.names else None
+
+
+class SolverCase(Case):
+    """
+    Case wraps the data from a single iteration of a recording to make it more easily accessible.
```
Kenneth-T-Moore: Descriptions are cut and paste from `case`, so need some work to describe SolverCase, etc.

openmdao/solvers/nonlinear/nonlinear_runonce.py
```diff
@@ -33,17 +30,21 @@ def solve(self):
...
-            system._transfer('nonlinear', 'fwd')
-            for subsys in system._subsystems_myproc:
-                subsys._solve_nonlinear()
-            system._check_reconf_update()
-        # If this is not a parallel group, transfer for each subsystem just prior to running it.
-        else:
-            for isub, subsys in enumerate(system._subsystems_myproc):
-                system._transfer('nonlinear', 'fwd', isub)
-                subsys._solve_nonlinear()
+        from openmdao.recorders.recording_iteration_stack import Recording
```
Kenneth-T-Moore: I *think* you can move this import up to the top without having a import cycle.

openmdao/solvers/linear/linear_runonce.py
```diff
@@ -44,7 +44,11 @@ def solve(self, vec_names, mode):
         for vec_name in self._vec_names:
             self._rhs_vecs[vec_name] = b_vecs[vec_name]._clone()
 
-        # Single iteration of GS
-        self._iter_execute()
+        with Recording('LinearRunOnce', 0, self) as rec:
+            # Single iteration of GS
```
Kenneth-T-Moore: This feels less obtrusive with the manager, so I approve

openmdao/drivers/pyoptsparse_driver.py
kmarsteller: oh, no problem...I thought herb had put these in

openmdao/recorders/case.py
kmarsteller: I didn't review them closely enough after getting them from Herb today, but you're right, they are not correct.

openmdao/core/driver.py
kmarsteller: Yes, they should.  begin what with uppercase?

kmarsteller: Oh, the comment.  yeah, got it.

kmarsteller: And I am adding periods at the end.  Fixed locally.

openmdao/core/driver.py
kmarsteller: To clarify, you mean six should be first, right?  If so, got it.

openmdao/drivers/pyoptsparse_driver.py
kmarsteller: fixed locally

openmdao/core/driver.py
kmarsteller: Can switch if you'd prefer.

openmdao/recorders/base_case_reader.py
kmarsteller: I'm guessing this was a cut and paste Herb did and forgot to fix.  I'm fixing them all now.

openmdao/core/system.py
```diff
@@ -2565,3 +2608,23 @@ def _list_states(self):
...
+        """
+        self._rec_mgr.append(recorder)
+
+    def record_iteration(self):
+        """
+        Record an iteration of the current System.
+        """
+        metadata = create_local_meta(self.pathname)
+        # Send the calling method name into record_iteration, e.g. 'solve_nonlinear'.
+        self._rec_mgr.record_iteration(self, metadata, method=inspect.stack()[2][3])
```
kmarsteller: I *think* we put this in before we were keeping track of our own stack for the iter coord...so we may be able to get this info from our own stack now?  If not, do you know of any other ways to get the calling method name?

naylor-b: We should be able to keep track of the calling method using our own stack.  I suppose we can fix this later when we start concentrating on performance optimization.  There are currently a couple of other places (add_input/add_output) where we're doing it that will also have to be fixed.  Just to give you an idea of how slow this is, I profiled propulsor.py, and inspect.stack(), which is called once for each add_input and add_output, is taking up nearly 30% of the total execution time.

openmdao/core/driver.py
```diff
@@ -79,20 +113,42 @@ def _setup_driver(self, problem):
...
         """
         Return the design variable values.
 
         This is called to gather the initial design variable state.
 
+        Parameters
+        ----------
+        filter : list
+            List of desvar names used by recorders to
+            filter by includes/excludes.
```
kmarsteller: This may be poorly worded.  The filter list, when it arrives here, is already the result of earlier processing the includes and excludes lists.  So I guess at this point it's just a list of what we want.  I could re-word these to make that clearer?

naylor-b: Yeah, I guess that's what I was getting at.  My comment could have been clearer.  I was basically just suggesting to get rid of the '/excludes'.

openmdao/core/driver.py
kmarsteller: Yes, before we record, we're allowing the users to filter design vars (cons and objs too), using include/excludes.  So before we record, we grab the values of only the ones that they want.  Do you feel like we need to do something with the scaling here before we record?  (We record scaling factors in driver metadata, for what that's worth.)

openmdao/solvers/nonlinear/nonlinear_runonce.py
kmarsteller: I recall it being there because it caused a circular, but I will try it again in case that is old info.  Also, any suggestions for re-arranging the imports are welcome!

openmdao/core/system.py
kmarsteller: May as well fix it now, at least my instance of it, if it's easy... I'll take a look today.

openmdao/recorders/base_recorder.py
```diff
@@ -0,0 +1,294 @@
...
+        # Compute the inclusion/exclusion lists
+
+        if (isinstance(object_requesting_recording, System)):
+            myinputs = myoutputs = myresiduals = set()
+            incl = self.options['includes']
+            excl = self.options['excludes']
+
+            if self.options['record_inputs']:
+                if object_requesting_recording._inputs:
+                    myinputs = [n for n in object_requesting_recording._inputs._names
```
kmarsteller: That is almost certainly just a mistake, don't remember a reason to convert.  Thanks.

openmdao/drivers/pyoptsparse_driver.py
kmarsteller: Wait, did you edit your original to say to take the print statements OUT? 😄

openmdao/core/system.py
naylor-b: BTW, you call it in 3 different places.  :)

openmdao/solvers/nonlinear/nonlinear_runonce.py
kmarsteller: @Kenneth-T-Moore this was correct, I remember now that I moved this into its own module to prevent the circular import...this one managed to evade being put at the top of file.  Fixed.

openmdao/recorders/base_recorder.py
```diff
@@ -0,0 +1,294 @@
...
+        self._counter = 0
+
+        # Deprecated options here, but need to preserve backward compatibility if possible.
+        if self.options['record_params']:
+            warn_deprecation("record_params is deprecated, please use record_inputs.")
+            # set option to what the user intended.
+            self.options['record_inputs'] = True
+
+        if self.options['record_unknowns']:
+            warn_deprecation("record_ is deprecated, please use record_inputs.")
```
swryan: message should say `record_outputs`

openmdao/recorders/base_recorder.py
```diff
@@ -0,0 +1,294 @@
...
+            # set option to what the user intended.
+            self.options['record_outputs'] = True
+
+        if self.options['record_resids']:
+            warn_deprecation("record_params is deprecated, please use record_inputs.")
```
swryan: message should say `record_residuals`

openmdao/recorders/recording_iteration_stack.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Do things after the code inside the 'with Recording' block.
+        """
+        # Determine if recording is justified.
+        do_recording = not iter_get_norm_on_call_stack() and not \
+            compute_total_derivs_on_call_stack()
+
+        if do_recording:
+            from openmdao.solvers.solver import Solver
+            if isinstance(self.object_requesting_recording, Solver):
```
swryan: might make sense to do this `isinstance` check only once in `__init__`

openmdao/docs/features/recording/basic_recording.rst
```diff
@@ -0,0 +1,205 @@
...
+++++++++++++++++++++++++++++
+
+.. code-block:: console
+
+    def test_simple_driver_recording(self):
+        if OPT is None:
+            raise unittest.SkipTest("pyoptsparse is not installed")
+
+        if OPTIMIZER is None:
+            raise unittest.SkipTest("pyoptsparse is not providing SNOPT or SLSQP")
```
kmarsteller: Yes...currently, it's failing because it has to create a database and query it during the docbuild, I figured I'd put in a story to make that work later.

openmdao/recorders/base_recorder.py
kmarsteller: hahaha caught me in a copy/paste!  Thanks

openmdao/recorders/base_recorder.py
kmarsteller: Double trouble!  Fixed on my branch.  Thanks.

openmdao/recorders/recording_iteration_stack.py
kmarsteller: Done.  Thanks!

openmdao/core/system.py
kmarsteller: Just realized I forgot to put openmdao.recorders in the setup.py files.  Fixed that locally.

openmdao/core/driver.py
```diff
@@ -194,15 +270,29 @@ def get_constraint_values(self, ctype='all', lintype='all'):
             Default is 'all'. Optionally return just the linear constraints
             with 'linear' or the nonlinear constraints with 'nonlinear'.
 
+        filter : list
+            List of objective names used by recorders.
```
Kenneth-T-Moore: For this method in the doctstring, objective --> constraint

openmdao/core/system.py
```diff
@@ -2605,15 +2618,6 @@ def _solve_linear(self, vec_names, mode):
             list of names of the right-hand-side vectors.
         mode : str
             'fwd' or 'rev'.
-
```
Kenneth-T-Moore: This method is always overriden, and the versions of it in subclasses have these returns. Don't we want the parent class stubs to have the fully documented docstring args and returns? I don't know the answer, but what does eveyone think.

openmdao/core/driver.py
kmarsteller: ARGGGH, stupid copy/paste errors!  Fixed locally.

openmdao/core/system.py
kmarsteller: Did I change this?  Or are you just asking in general?

Kenneth-T-Moore: Well I mean you deleted the "Returns" part of the docstring, but i am arguing that it should be restored because the child overrides return that stuff.

kmarsteller: OH, the diff I'm seeing looked like a blank line got deleted.  I'll restore it.  I'm not sure why that would have been on my radar to delete.  Likely an error.

PR #299, opened 2017-07-12
Complex Step

Kenneth-T-Moore:
1. Finite Difference approximation of total derivatives for groups.
2. Complex Step approximation of partial derivatives for components.
3. Complex Step approximation of total derivatives for groups.


openmdao/approximation_schemes/complex_step.py
```diff
@@ -0,0 +1,213 @@
...
+from collections import namedtuple
+from itertools import groupby
+from six.moves import range
+
+from openmdao.approximation_schemes.approximation_scheme import ApproximationScheme
+from openmdao.utils.name_maps import abs_key2rel_key
+
+
+DEFAULT_CS_OPTIONS = {
+    'step': 1e-6,
```
hwangjt: The default step size should be much smaller, say 1e-15, because 1e-15 will always be as accurate as or more accurate than 1e-6

openmdao/approximation_schemes/complex_step.py
```diff
@@ -0,0 +1,213 @@
...
+class ComplexStep(ApproximationScheme):
+    r"""
+    Approximation scheme using complex step to calculate derivatives.
+
+    For example, using the 'forward' form with a step size of 'h' will approximate the derivative in
+    the following way:
+
+    .. math::
+
+        f'(x) = \frac{f(x+h) - f(x)}{h} + O(h).
```
hwangjt: This docstring should be updated for CS

openmdao/vectors/petsc_vector.py
```diff
@@ -107,9 +115,16 @@ def _initialize_data(self, root_vector):
         super(PETScVector, self)._initialize_data(root_vector)
 
         self._petsc = {}
+        self._imag_petsc = {}
```
hwangjt: Are you sure we need to create another PETSc Vec object for the imaginary component? I think we can instantiate one, and both the real and imag parts can share it for transfers (by assigning [PETScVec].array = real_array/imag_array). This would save us the memory overhead of the additional PETSc Vector objects

openmdao/core/system.py
```diff
@@ -835,26 +861,33 @@ def _setup_vectors(self, root_vectors, excl_out, excl_in, resize=False):
...
             Whether to resize the root vectors - i.e, because this system is initiating a reconf.
+        alloc_complex : bool
+            Whether to allocate any imaginary storage to perform complex step. Default is False.
         """
         self._vectors = vectors = {'input': {}, 'output': {}, 'residual': {}}
         self._excluded_vars_out = excl_out
         self._excluded_vars_in = excl_in
 
+        # Check for complex step to set vectors up appropriately.
+        if 'cs' in self._approx_schemes or alloc_complex:
```
hwangjt: I say we force the user to declare alloc_complex before hand, and if they don't but 'cs' is in approx_schemes, we should throw an error. Otherwise, we may run into ugly situations with reconfigurability in which a subgroup or subcomponent adds a new CS approximation, where the root vector is not declared to be complex

openmdao/core/system.py
```diff
@@ -468,7 +484,17 @@ def _get_root_vectors(self, vector_class, initial):
...
+
+                    # Check for complex step to set vectors up appropriately.
+                    # If any subsystem needs complex step, then we need to allocate it everywhere.
+                    alloc_complex = 'cs' in self._approx_schemes
+                    for sub in self.system_iter(include_self=True, recurse=True):
+                        if alloc_complex:
+                            break
+                        alloc_complex = 'cs' in sub._approx_schemes
+
+                    root_vectors[key][vec_name] = vector_class(vec_name, type_, self,
```
hwangjt: We would only ever want to declare set alloc_complex to True for the `vecname='nonlinear'` vector, right? (unless we are going to support complex step around linear solvers

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: Fixed

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: Fixed

openmdao/core/system.py
Kenneth-T-Moore: Yes, that is true. Fixed.

openmdao/approximation_schemes/complex_step.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Parameters
+        ----------
+        approx_tuple : tuple(str, str, dict)
+            A given approximated derivative (of, wrt, options)
+
+        Returns
+        -------
+        tuple(str, str, float, int, str)
+            Sorting key (wrt, form, step_size, order, step_calc)
```
dwmunster: Docstring should be updated to reflect changes in key entries.

openmdao/core/group.py
```diff
@@ -1178,12 +1182,133 @@ def _linearize(self, do_nl=True, do_ln=True):
...
+                        ivc.append(src)
+
+            with self.jacobian_context() as J:
+                for key in product(of, wrt.union(of)):
+                    meta_changes = {
+                        'method': method,
+                    }
+                    if key[0] == key[1]:
+                        size = self._outputs._views_flat[key[0]].shape[0]
+                        meta_changes['value'] = np.eye(size)
```
dwmunster: Would we want the sparse specification here?
```
meta_changes['rows'] = np.arange(size)
meta_changes['cols'] = np.arange(size)
meta_changes['value'] = np.ones(size)
```

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: Yeah, probably better to do that, so fixed.

openmdao/core/group.py
Kenneth-T-Moore: Yeah, probably better to do that, so fixed.

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: Fixed

openmdao/vectors/petsc_vector.py
Kenneth-T-Moore: I am not sure. Does setting `.data` on the petsc object work collectively for distributed vectors? I think i'd rather put up a story to investigate this.

hwangjt: It doesn't work collectively, but the we just have to set the `.array` attribute on every processor. I'm certain this works though; I did this in MAUD when looping over columns of the PETSc 'multi-vectors', assigning each column to the `.array` attribute. I'm fine with it if you want to add a story and get this pull request in as is

openmdao/core/system.py
Kenneth-T-Moore: Done, with test.

openmdao/vectors/petsc_vector.py
Kenneth-T-Moore: I made an initial attempt at this, and got failures, so something wasn't working right. Decided to put up a story for a later revisit.

PR #300, opened 2017-07-14
Support for left preconditioning in the Petsc KSP solver.

Kenneth-T-Moore:


openmdao/solvers/linear/petsc_ksp.py
```diff
@@ -214,6 +214,9 @@ def _declare_options(self):
                              desc='Number of iterations between restarts. Larger values increase '
                              'iteration cost, but may be necessary for convergence')
 
+        self.options.declare('precon_side', default='right', values=['left', 'right'],
```
dwmunster: Any reason to prefer right preconditioning over left? If I recall correctly, more KSP solvers support only left than support only right.

Kenneth-T-Moore: I think mainly because our default is fgmres, which only uses right.

dwmunster: So it would seem. Sounds like a good reason to me.

PR #304, opened 2017-07-19
Part 1 of parallel derivative support

naylor-b:
- parallel derivs can now be specified by adding 'rhs_group' arg to add_design_var, add_constraint or add_objective.
- rhs vecs are currently all full size (no relevance reduction)
- moved function to compute the system graph into Group
- small improvement to call tracing
- got rid of calls to inspect.stack() in add_input, add_output
- a few other misc performance tweaks


openmdao/core/problem.py
```diff
@@ -759,90 +760,167 @@ def _compute_total_derivs(self, of=None, wrt=None, return_format='flat_dict',
...
+                        sz = sizes[rank, in_var_idx]
+                        if sz > 0:
+                            loc_size = sz
+                            break
+
+                voi_info[input_name] = (dinputs, doutputs, irange, loc_size, start, end, dup)
+
+            loc_idxs = defaultdict(lambda: -1)
+
+            # at this point, we know that for all vars in the current
```
hwangjt: I don't follow the logic here why the number of indices is the same for all variables in the current group. It's OK though, because the behavior will change when mtx-mtx is implemented anyway. At that point, each 'vec_name' will have its own matrix with the number of columns given by the number of indices, and the number of indices will not need to be the same for all variables in a group (since it is not the same, in general). Just want to make sure that's what you had in mind for the follow-up story as well; if not, let's discuss

openmdao/core/problem.py
```diff
@@ -759,90 +760,167 @@ def _compute_total_derivs(self, of=None, wrt=None, return_format='flat_dict',
                         ikey = old_input_list[icount]
                         totals[okey][ikey] = -approx_jac[output_name, input_name]
 
+            return totals
+
         # Solve for derivs using linear solver.
-        else:
+        voi_lists = defaultdict(list)
+        v2rhs_name = {}
```
hwangjt: Could you use a more descriptive variable name or add some more comments here? This part of the algorithm/code is very tricky, so it would help to be more verbose than usual with the comments here

hwangjt: To clarify, I was referring to `v2rhs_name` specifically, but more comments would be welcome for `voi_lists` and `old_input_list` as well

openmdao/core/component.py
```diff
@@ -410,7 +408,7 @@ def add_output(self, name, val=1.0, shape=None, units=None, res_units=None, desc
         dict
             metadata for added variable
         """
-        if inspect.stack()[1][3] == '__init__':
+        if self._static_mode:
```
naylor-b: actually we're ok here because IndepVarComps.add_output just stores the args and then later, during setup, calls the base class version of add_output.

hwangjt: Oh, that's right

openmdao/core/problem.py
naylor-b: The logic behind that is that we're potentially making collective MPI calls inside of model.solve_linear, so in order to keep everything in sync we have to iterate the same number of times in all of the procs.

hwangjt: Makes sense; I see that's the only way to make parallel derivatives work without mtx-mtx. I'm guessing the number of times we iterate is the maximum number of indices among all the vecs in this group

openmdao/core/problem.py
naylor-b: OK, I've added some additional comments.  Hope that helps.

hwangjt: thanks!

PR #306, opened 2017-07-24
Feature: check_total_derivatives

Kenneth-T-Moore:
1. Method `check_total_derivatives` added to Problem. This method checks the derivatives of the driver constraints and objectives with respect to desvars, and compares them to fd or cs as chosen. It can also be used to compare user selected 'of' and 'wrt.
2. Output from `check_partials` converted over to use logger.INFO.
3. Added a utility module for logging utilities. It contains a method to return a default logger that writes to stdout.

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -580,6 +580,25 @@ class Problem(object):
 
         return partials_data
 
+    def check_total_derivatives(self, of=None, wrt=None, out_stream=sys.stdout,
+                                compact_print=False):
+        """
+        Check total derivatives for the model vs. finite difference.
+
+        Parameters
+        ----------
+        of : list of variable name strings or None
+            Variables whose derivatives will be computed. Default is None, which
+            uses the driver's objectives and constraints.
+        wrt : list of variable name strings or None
+            Variables with respect to which the derivatives will be computed.
+            Default is None, which uses the driver's desvars.
+        out_stream : file_like
+            Where to send human readable output. Default is sys.stdout. Set to None to suppress.
+        compact_print : bool
+            Set to True to just print the essentials, one line per unknown-param pair.
+        """
+
     def compute_total_derivs(self, of=None, wrt=None, return_format='flat_dict'):
         """
         Compute derivatives of desired quantities with respect to desired inputs.
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -5,7 +5,6 @@ from __future__ import division
 from collections import OrderedDict, defaultdict, namedtuple
 from itertools import product
 import logging
-import sys
 
 from six import iteritems, iterkeys
 from six.moves import range
@@ -23,6 +22,7 @@ from openmdao.core.indepvarcomp import IndepVarComp
 from openmdao.error_checking.check_config import check_config
 
 from openmdao.utils.general_utils import warn_deprecation
+from openmdao.utils.logger_utils import get_default_logger
 from openmdao.utils.mpi import MPI, FakeComm
 from openmdao.utils.graph_utils import all_connected_edges
 from openmdao.vectors.default_vector import DefaultVector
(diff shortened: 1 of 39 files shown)


openmdao/core/problem.py
```diff
@@ -33,6 +35,10 @@
 ErrorTuple = namedtuple('ErrorTuple', ['forward', 'reverse', 'forward_reverse'])
 MagnitudeTuple = namedtuple('MagnitudeTuple', ['forward', 'reverse', 'fd'])
 
+# when setup is called multiple times, we need this to prevent adding
+# another handler to the config_check logger each time (if logger arg to check_config is None)
```
swryan: this comment needs updated for this module

openmdao/core/problem.py
```diff
@@ -345,6 +351,18 @@ def check_partials(self, out_stream=sys.stdout, comps=None, compact_print=False,
...
+            global _set_logger
+            if _set_logger is None:
+                logger = logging.getLogger("check_total_derivatives")
+                _set_logger = logger
+                console = logging.StreamHandler(sys.stdout)
+                console.setLevel(logging.INFO)
+                logger.addHandler(console)
+            else:
+                logger = _set_logger
+
```
swryan: this code is duplicated.. might make it a module level function

openmdao/core/problem.py
```diff
@@ -1013,40 +1172,46 @@ def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out
...
+                if not totals:
+                    logger.info('    Raw Reverse Derivative (Jfor)\n')
+                    logger.info(str(reverse))
+                    logger.info('')
+
+                logger.info('    Raw FD Derivative (Jfd)\n')
+                logger.info(str(fd))
+                logger.info('')
+
+                logger.info(' -' * 30)
```
swryan: I wonder if it would make sense to move some of this code to another file.. since half of the `problem.py` module now seems to consist of "check derivatives" logic, maybe it warrants it's own module similar to `check_config`.

openmdao/core/problem.py
Kenneth-T-Moore: Fixed

openmdao/core/problem.py
Kenneth-T-Moore: Actually, not quite duplicated, except I forgot to change the logger name when I copied it. I don't mind the duplication, but if we remove it, we also need to remove the code in check_config. That code is slightly different in that it specifies a custom format for the messages that are logged. Perhaps we need a new story for a unified logging manager.

swryan: if the only diff is the logger name, that's just an arg to the function.. it is quite a bit of duplication IMO

openmdao/core/problem.py
Kenneth-T-Moore: Well, I am not that worried about it, and half of the code (check_partials) is probably moving into Group or System sometime anyway. If there is enough of a demand for splitting out the check codes into a separate module, then it can be done as a separate story.

openmdao/core/problem.py
Kenneth-T-Moore: Done

openmdao/core/problem.py
```diff
@@ -583,11 +586,100 @@ def check_partials(self, out_stream=sys.stdout, comps=None, compact_print=False,
...
         partials_data = {comp_name: dict(outer) for comp_name, outer in iteritems(partials_data)}
 
-        _assemble_derivative_data(partials_data, rel_err_tol, abs_err_tol, out_stream,
+        logging.getLogger().setLevel(logging.INFO)
+        _assemble_derivative_data(partials_data, rel_err_tol, abs_err_tol, logger,
                                   compact_print, comps, global_options)
 
         return partials_data
 
+    def check_total_derivatives(self, of=None, wrt=None, logger=None, compact_print=False,
```
naylor-b: You should edit the README.md file and remove 'Total Derivatives checking' from the list of things we don't support.

Kenneth-T-Moore: done

PR #309, opened 2017-07-27
AnalysisError feature: Part 2 (MERGE Case Recording First so I can handle the conflicts)

Kenneth-T-Moore:
1 The ArmijoGoldstein line-search now backtracks if an AnalysisError is raised. An option was added to turn this off, but it is on by default.
2 Cleaned up a few analysis-errors related tests that were skipped.


openmdao/solvers/linesearch/backtracking.py
```diff
@@ -191,7 +196,16 @@ def _iter_initialize(self):
...
 
-        norm = self._iter_get_norm()
+        try:
+            norm = self._iter_get_norm()
+
+        except AnalysisError as err:
+            if self.options['retry_on_analysis_error']:
+                self._analysis_error_raised = True
+            else:
+                raise AnalysisError(str(err))
```
naylor-b: If we re-raise the exception in this way, we'll lose the traceback information from the original exception.  Instead we should do something like we did in clippy:

from six import reraise

  .. in your except block...
    exc = sys.exc_info()
    reraise(exc[0], exc[1], exc[2])


If you want to replace the exception with a new one but keep the original traceback, you can replace
exc[1] with the new exception object.

Kenneth-T-Moore: Fixed

openmdao/solvers/linesearch/backtracking.py
```diff
@@ -198,7 +205,18 @@ def _iter_initialize(self):
...
-        norm = self._iter_get_norm()
+        try:
+            norm = self._iter_get_norm()
+
+        except AnalysisError as err:
+            if self.options['retry_on_analysis_error']:
+                self._analysis_error_raised = True
+            else:
+                exc = sys.exc_info()
+                reraise(exc[0], exc[1], exc[2])
```
JustinSGray: could just do `reraise(*exc)`. Slightly cleaner

Kenneth-T-Moore: Fixed

PR #310, opened 2017-07-28
Added a balance comp which is optionally vectorized.

robfalck:
BalanceComp makes it easier for novice users to define simple implicit equations by giving them an ImplicitComponent to which they connect the LHS and RHS of a residual equation:

resid = mult * LHS - RHS

When abs(RHS) >= 2, the residual equation is automatically scaled by (1/RHS).  When abs(RHS) < 2, the residual equation is scaled by a paraboloid that avoids the singularity at RHS=0 yet still provides smoothness at the interface of RHS=2.

New implicit relationships are added with `add_balance(name, eq_units=None, mult_val=1.0, **kwargs)`, where name is the name of the new implicit state variable.  For instance, adding `name=x` automatically creates a new output named `x`, as well as inputs named `mult:x`, `lhs:x`, and `rhs:x`.  `mult:x` is assumed to have no units, while units for `lhs:x` and `rhs:x` are optionally provided with the `eq_units` argument.

All other options in `**kwargs` are forwarded to the `add_output` call at setup.  The implicit state may be declared a vector by providing an initial value `val=np.ones(N)` where N is some positive integer.

If left unconnected, the multiplier will have a default value of `mult_val`.  It must be compatible with the shape of the implicit state.

### DIFF
--- a/openmdao/components/balance_comp.py
+++ b/openmdao/components/balance_comp.py
@@ -1,0 +1,230 @@
+"""Define the BalanceComp class."""
+
+from __future__ import print_function, division, absolute_import
+
+from numbers import Number
+from six import iteritems
+
+import numpy as np
+
+from openmdao.core.implicitcomponent import ImplicitComponent
+
+
+class BalanceComp(ImplicitComponent):
+    """
+    A simple equation balance for solving implicit equations.
+    """
+
+    def __init__(self, name=None, eq_units=None, **kwargs):
+        r"""
+        Initialize a BalanceComp, optionally creating a new implicit state variable.
+
+        The BalanceComp is a bit like IndepVarComp in that it allows for the
+        creation of one or more implicit state variables, and computes the residuals
+        for those variables based on the following equation.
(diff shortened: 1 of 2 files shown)


openmdao/components/balance_comp.py
```diff
@@ -0,0 +1,230 @@
...
+        The BalanceComp is a bit like IndepVarComp in that it allows for the
+        creation of one or more implicit state variables, and computes the residuals
+        for those variables based on the following equation.
+
+        .. math::
+
+          f_{mult}(x,...) \times f_{lhs}(x,...) = f_{rhs}(x,...)
+
+        Where :math:`f_{lhs}` represents the left-hand-side of the equation,
+        :math:`f_{lhs}` represents the right-hand-side, and :math:`f_{mult}`
```
swryan: typo, should be `f_{rhs}`

robfalck: thanks

PR #312, opened 2017-07-28
Part 1 of setup refactor.

Kenneth-T-Moore:
1. New group method `configure` where users can make changes to their subsystems. These are recursed in reverse.
2. Moved things around so that all groups and components have their `setup` called under `setup_procs`.


openmdao/core/group.py
```diff
@@ -56,16 +56,51 @@ def __init__(self, **kwargs):
...
+        pass
+
+    def configure(self):
+        """
+        Configure this group to assign children settings.
+
+        This method may optionally be overidden by your Group's method.
+
+        You may only use this method to change settings on your children subsystems. This includes
+        setting solvers in cases where you want to override the defaults.
```
JustinSGray: I think we should include some detail about timing here. 

"You can assume that the full hierarchy below your level has been instantiated and has already called its own configure methods."

openmdao/core/tests/test_problem.py
```diff
@@ -535,5 +535,108 @@ def test_relevance(self):
...
+
+
+        top = Problem()
+        top.model = Super()
+
+        top.setup(check=False)
+
+        top['sub.comp.a'] = 1.5576015661428098
+        top.run_model()
+        assert_rel_error(self, top['sub.comp.x'], -0.5)
```
JustinSGray: should you explicitly test for the solve class here as well. Testing for convergence seems a round-about-way to know this worked here.

JustinSGray: also, the test could be a bit faster if we didn't actually have to run the solver. Since we're not testing solve behavior here, its not necessary and would even, I think make a better more unit-like test if we didn't run the solver

Kenneth-T-Moore: Fixed

openmdao/core/group.py
Kenneth-T-Moore: Fixed

PR #313, opened 2017-07-31
First round of Case Recording improvements:

kmarsteller:
*Added recording to ScipyOptimizer
*Fixed a bug in System recording for multidimensional arrays; added a test for that.
*Moved some functionality that was not SQL-related out of sqlite_recorder and into the base_recorder.
*Fixed a few small things that Ken had asked for in his review earlier this morning.
*Docstring updates.
*A few things from Herb's effort to parallelize recording.

### DIFF
--- a/openmdao/recorders/recording_manager.py
+++ b/openmdao/recorders/recording_manager.py
@@ -6,6 +6,7 @@ from openmdao.recorders.recording_iteration_stack import recording_iteration_sta
 
 from openmdao.utils.mpi import MPI
 
+
 class RecordingManager(object):
     """
     Object that routes function calls to all attached recorders.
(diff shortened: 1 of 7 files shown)


openmdao/recorders/recording_manager.py
```diff
@@ -22,7 +24,12 @@ def __init__(self):
         init.
         """
         self._recorders = []
-        self.rank = 0
+        self._has_serial_recorders = False
```
Kenneth-T-Moore: Leftover from something? It is undocumented and never used.

hschilling:
Ken, sorry some of the early parallel recording code was mixed in with what was pushed up. It will be used !

openmdao/recorders/recording_manager.py
hschilling: Ken, sorry some of the early parallel recording code was mixed in with what was pushed up. It will be used !

PR #320, opened 2017-08-03
Initialize the recording stack upon init of Problem

hschilling:


openmdao/core/problem.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, model=None, comm=None, use_ref_vector=True, root=None):
 
         self._mode = None  # mode is assigned in setup()
 
+        recording_iteration_stack = []
+
```
Kenneth-T-Moore: So, I did try this when I found the test failure problem, and it didn't fix it for me. Do you maybe have access to a linux system that you can test it on?

hschilling: I could try it on the OpenMDAO cluster, right? Will take me a while to setup the environment there.

kmarsteller: @hschilling : see discussion in #case-recorders

hschilling: So I assume I should hold off on this since it's not crucial?

Kenneth-T-Moore: No, just merge it. I think there is a story too.

PR #321, opened 2017-08-03
support for matrix-matrix

naylor-b:
- still no relevance reduction
-more tests needed, especially MPI tests


hwangjt:
Along with the parallel tests, it would also be good to add support for and tests with apply_multi_linear and solve_multi_linear as well, in a future pull request

PR #322, opened 2017-08-03
OpenMDAO Visualization Server Recorder

ryanfarr01:
Created recorder that sends recorded data as JSON to the HTTP server running. Should look _very_ similar to the SQLITE recorder (including tests)

### DIFF
--- a/openmdao/api.py
+++ b/openmdao/api.py
@@ -67,7 +67,7 @@ from openmdao.utils.options_dictionary import OptionsDictionary
 
 # Recorders
 from openmdao.recorders.sqlite_recorder import SqliteRecorder
-from openmdao.recorders.openmdao_server_recorder import OpenMDAOServerRecorder
+from openmdao.recorders.web_recorder import WebRecorder
 
 # set up tracing or memory profiling if env vars are set.
 import os
--- a/openmdao/api.py
+++ b/openmdao/api.py
@@ -67,6 +67,7 @@ from openmdao.utils.options_dictionary import OptionsDictionary
 
 # Recorders
 from openmdao.recorders.sqlite_recorder import SqliteRecorder
+from openmdao.recorders.openmdao_server_recorder import OpenMDAOServerRecorder
 
 # set up tracing or memory profiling if env vars are set.
 import os
--- a/openmdao/api.py
+++ b/openmdao/api.py
@@ -42,8 +42,10 @@ from openmdao.surrogate_models.surrogate_model import SurrogateModel, \
 
 # Vectors
 from openmdao.vectors.default_vector import DefaultVector
+from openmdao.vectors.default_multi_vector import DefaultMultiVector
 try:
     from openmdao.vectors.petsc_vector import PETScVector
+    from openmdao.vectors.petsc_multi_vector import PETScMultiVector
 except ImportError:
     PETScVector = None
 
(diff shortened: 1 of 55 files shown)


openmdao/recorders/openmdao_server_recorder.py
```diff
@@ -0,0 +1,451 @@
...
+
+        case_data = json.dumps(case_data_dict)
+
+        # Post case and get Case ID
+        case_request = requests.post(self._endpoint, data=case_data, headers=self._headers)
+        response = case_request.json()
+        if response['status'] != 'Failed':
+            self._case_id = str(response['case_id'])
+            if not suppress_output:
+                print("Visualization at: http://openmdao.org/visualization/case/" + self._case_id)
```
ryanfarr01: Printing the URL is temporary while I set up a UI for requesting tokens and listing recordings on the web server

openmdao/api.py
```diff
@@ -67,6 +67,7 @@
 
 # Recorders
 from openmdao.recorders.sqlite_recorder import SqliteRecorder
+from openmdao.recorders.openmdao_server_recorder import OpenMDAOServerRecorder
```
JustinSGray: can we call this "WebRecorder"?

openmdao/recorders/openmdao_server_recorder.py
```diff
@@ -0,0 +1,458 @@
...
+        if isinstance(object_requesting_recording, Driver):
+            self.record_iteration_driver(object_requesting_recording, metadata)
+
+        elif isinstance(object_requesting_recording, System):
+            self.record_iteration_system(object_requesting_recording, metadata)
+
+        elif isinstance(object_requesting_recording, Solver):
+            self.record_iteration_solver(object_requesting_recording, metadata, **kwargs)
+        else:
+            raise ValueError("Recorders must be attached to Drivers, Systems, or Solvers.")
```
swryan: I assume you want to merge out and use base class methods now

JustinSGray: do you mean because of https://github.com/OpenMDAO/blue/pull/328?

openmdao/recorders/web_recorder.py
```diff
@@ -456,3 +406,9 @@ def convert_to_list(self, obj):
             return []
         else:
             return obj
+
+    def close(self):
+        """
+        Close.
+        """
+        pass
```
swryan: you don't need close anymore do you?

ryanfarr01: It looks like `close` actually is necessary (for the SQLITE recorder it's what closes the connection to the DB). I'm going to add `close` back into the BaseRecorder and force subclasses to implement that method

kmarsteller: The base class version that was there before WAS vestigial... the sqlite_recorder version disconnects from the DB...so yes, I should have, instead of deleting the base version, make it throw a not implemented?

kmarsteller: But it seems like other recorders (like Web Recorder) has no use for `close()`?

ryanfarr01: This should be addressed in the most recent commit, where I took `close()` out of the server recorder altogether and simply put in `pass` as the base implementation of `close`. Thus, it can be implemented by the subclass if needed, but isn't required.

PR #330, opened 2017-08-09
Initial sub-solve feature for Newton solver

shamsheersc19:
Adding an initial_sub_solve feature to Newton solver to allow one cycle of sub-solves at the start of the problem (not repeated at each optimization iteration), also added a test.

### DIFF
--- a/openmdao/solvers/nonlinear/newton.py
+++ b/openmdao/solvers/nonlinear/newton.py
@@ -85,6 +85,8 @@ class NewtonSolver(NonlinearSolver):
                              desc='Set to True to turn on sub-solvers (Hybrid Newton).')
         self.options.declare('max_sub_solves', type_=int, default=10,
                              desc='Maximum number of subsystem solves.')
+        self.options.declare('initial_sub_solve', type_=bool, default=False,
+                             desc='Set to True to cycle through sub-solvers once at the start of the problem.')
         self.supports['gradients'] = True
 
     def _setup_solvers(self, system, depth):
@@ -190,6 +192,19 @@ class NewtonSolver(NonlinearSolver):
         # Disable local fd
         approx_status = system._owns_approx_jac
         system._owns_approx_jac = False
+        
+        # This cycles through the solvers of the subsystems once at the start of the problem
+        if do_subsolve == False and self.options['initial_sub_solve'] == True:
+
+            for isub, subsys in enumerate(system._subsystems_allprocs):
+                system._transfer('nonlinear', 'fwd', isub)
+
+                if subsys in system._subsystems_myproc:
+                    subsys._solve_nonlinear()
+
+            system._apply_nonlinear()
+            print("Did initial sub-solve")
+            self.options['initial_sub_solve'] = False
 
         # Hybrid newton support.
         with Recording('Newton_subsolve', 0, self):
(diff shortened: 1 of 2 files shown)


openmdao/solvers/nonlinear/newton.py
```diff
@@ -85,6 +85,8 @@ def _declare_options(self):
                              desc='Set to True to turn on sub-solvers (Hybrid Newton).')
         self.options.declare('max_sub_solves', type_=int, default=10,
                              desc='Maximum number of subsystem solves.')
+        self.options.declare('initial_sub_solve', type_=bool, default=False,
```
Kenneth-T-Moore: If I understand this correctly, this option means that we only subsolve the very first time this solver runs, and then thereafter, we don't perform any subsolves?

openmdao/solvers/nonlinear/newton.py
```diff
@@ -190,6 +192,25 @@ def _iter_execute(self):
...
+
+            system._apply_nonlinear()
+            print("Did initial sub-solve")
+            self.options['initial_sub_solve'] = False
+        
+            if do_subsolve == True:
+                print("WARNING: Both solve_subsystems and initial_sub_solve are set to True.",
+                "Setting solve_subsystems to False and carrying out one initial sub-solve only.")
+                self.options['solve_subsystems'] = False
+                do_subsolve = False
```
Kenneth-T-Moore: I think we don't need this whole block; just use the current sub_solve code (which is integrated in with iprint and recording too.) We can modify the do_subsolve to always return false after the initial execution of the solver. Probably just need to save an execution counter.

Kenneth-T-Moore: Also, are there cases where we might want to run a few iterations of sub solves the first time we execute, and then run with no subsolves thereafter? I think this can be integrated so that all three options work together.

openmdao/solvers/nonlinear/newton.py
```diff
@@ -190,6 +192,25 @@ def _iter_execute(self):
         # Disable local fd
         approx_status = system._owns_approx_jac
         system._owns_approx_jac = False
+        
+        # This cycles through the solvers of the subsystems once at the start of the problem
+        if self.options['initial_sub_solve'] == True:
```
hwangjt: I think it's unintuitive behavior for the code to set initial_sub_solve to False after running it once. That means that when calling run_model multiple times, initial_sub_solve would only be used the first time, which is not what I would expect as a user. This should either be explained in the options documentation, or we find an alternate implementation.
That would be to override `__call__`, put the `initial_sub_solve` code in there, then call super's `__call__` or _run_iterator at the end. This way multiple consecutive calls of run_model would all run the initial_sub_solve

Kenneth-T-Moore: I agree, and think instead we use three options that work together:

solve_subsystems -- Turns on sub solves
max_sub_solves -- Max number of iterations of sub solve per execution
initial_sub_solve_only -- Only do sub solves the first time it executes, then no more.

then we can use the existing code, modify the `do_subsolve` logic a little bit, and just save the exec count

shamsheersc19: Thanks for the feedback!

JustinSGray:
I don't think any new options are needed at all. If you turn on `solve_subsystems`, then it should be expected that the solver will do a solve pass before it even computes the first residual. That is essentially the solve for the 0'th iteration. Then you can control the number of additional sub-solves with the `max_sub_solves` argument. If you set it to 1, you get just that very first one.

shamsheersc19:
I think I misunderstood the purpose of max_sub_solves. I thought ultimately the goal would be for it to behave like maxiters, where the iteration counter is reset every optimization iteration. But it looks like just setting max_sub_solves to 1 achieves what I need. Closing pull request.

PR #331, opened 2017-08-09
Setup refactor parts 2 and 3

Kenneth-T-Moore:
1. Setup split into two phases, with the split right before vector creation.
2. User initial condition sets on Problem are cached and processed after the second phase of setup.

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -237,13 +237,7 @@ class Problem(object):
         Set all initial conditions that have been saved in cache after setup.
         """
         for name, value in iteritems(self._initial_condition_cache):
-            if name in self.model._outputs:
-                self.model._outputs[name] = value
-            elif name in self.model._inputs:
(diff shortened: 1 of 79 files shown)


openmdao/core/problem.py
```diff
@@ -147,13 +220,33 @@ def __setitem__(self, name, value):
...
+            else:
+                msg = 'Variable name "{}" not found.'
+                raise KeyError(msg.format(name))
+
+    def _set_initial_conditions(self):
+        """
+        Set all initial conditions that have been saved in cache after setup.
+        """
+        for name, value in iteritems(self._initial_condition_cache):
+            if name in self.model._outputs:
```
naylor-b: To avoid duplicating the logic in this loop, couldn't you just call self[name] = value?

Kenneth-T-Moore: I would have to set the setup_status to 2 a little earlier to do so, but I could do it.

PR #335, opened 2017-08-10
Added capability for Drivers to record in parallel

hschilling:
This has not been tested with many complicated models. Probably some more code deduplication that could be done but would like to get this on master so people use it and test it.


openmdao/recorders/base_recorder.py
```diff
@@ -387,6 +392,71 @@ def record_iteration(self, object_requesting_recording, metadata, **kwargs):
...
+        if self.options['record_constraints']:
+            if self._filtered_driver:
+                self._constraints_values = \
+                    {name: constraints[name] for name in self._filtered_driver['con']}
+            else:
+                self._constraints_values = constraints
+        else:
+            self._constraints_values = None
+
     def record_iteration_driver(self, object_requesting_recording, metadata):
```
kmarsteller: Currently it looks like there is a `record_iteration_driver` and a `record_iteration_driver_passing_vars` in base_recorder... the "routing" function `record_iteration` still has a call to `record_iteration_driver` in it.  But the two functions appear to be very similar.  Should the new one just replace the old one, and also just assume its old name?

openmdao/recorders/recording_manager.py
```diff
@@ -63,9 +100,82 @@ def startup(self, object_requesting_recording):
         """
         Run startup on each recorder in the manager.
         """
+        # Will only add parallel code for Drivers. Use the old method for System and Solver
+        from openmdao.core.driver import Driver
```
kmarsteller: (answer: yes it did cause a circular, I tried it.)

PR #341, opened 2017-08-15
A little refactoring in solver, particularly Newton

Kenneth-T-Moore:
1. Removed the "apply" calls from `_get_iter_norm` and put them in their own method.
2. Rearranged order of events in Newton so that the initial subsolve is done in `_iter_initialize`, and any subsequent subsolves are after the linear solve in the main loop. This way, initial norms are always on the solved state.

### DIFF
--- a/openmdao/solvers/nonlinear/newton.py
+++ b/openmdao/solvers/nonlinear/newton.py
@@ -130,14 +130,9 @@ class NewtonSolver(NonlinearSolver):
         if self.linesearch is not None:
             self.linesearch._set_solver_print(level=level, type_=type_)
 
-    def _iter_get_norm(self):
+    def _run_apply(self):
         """
-        Return the norm of the residual.
-
-        Returns
-        -------
-        float
-            norm.
+        Run the the apply_nonlinear method on the system.
         """
         recording_iteration_stack.append(('_iter_get_norm', 0))
 
@@ -153,8 +148,6 @@ class NewtonSolver(NonlinearSolver):
 
         # Enable local fd
         system._owns_approx_jac = approx_status
-
-        return system._residuals.get_norm()
 
     def _linearize_children(self):
         """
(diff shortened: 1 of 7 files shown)


openmdao/solvers/solver.py
```diff
@@ -365,12 +372,21 @@ def _iter_initialize(self):
...
         else:
             norm = 1.0
         norm0 = norm if norm != 0.0 else 1.0
         return norm0, norm
 
+    def _run_apply(self):
+        """
+        Run the the apply_nonlinear method on the system.
+        """
+        recording_iteration_stack.append(('_run_apply', 0))
```
naylor-b: Is it necessary to push '_run_apply' onto the iteration stack when there is no recording associated with _run_apply and we already do recording inside of _apply_nonlinear (and in _apply_linear)?

Kenneth-T-Moore: This used to be `get_iter_norm`, and I renamed it to `run_apply` because I moved that code. This tag on the stack is used to prevent recording during this phase, so If I removed this, it would presumably print some extra cases, though I can only say this based on design intent, and not on viewing any case output.

naylor-b: Ah, I see.

openmdao/solvers/nonlinear/newton.py
```diff
@@ -130,14 +130,9 @@ def _set_solver_print(self, level=2, type_='all'):
...
         """
-        Return the norm of the residual.
-
-        Returns
-        -------
-        float
-            norm.
+        Run the the apply_nonlinear method on the system.
         """
         recording_iteration_stack.append(('_iter_get_norm', 0))
```
dwmunster: Shouldn't this be `'_run_apply'`?

Kenneth-T-Moore: No, it does this: `system._apply_nonlinear()`.

dwmunster: I thought all the instances of the `'_iter_get_norm'` recording tag were being replaced by the `'_run_apply'`?

Kenneth-T-Moore: OH, i thought you were talking about the docstring.  Fixed.

PR #343, opened 2017-08-17
Guess nonlinear needs to behave like a NonlinearRunOnce iteration and do data-passing between components after each call to guess. 

Kenneth-T-Moore:


openmdao/core/group.py
```diff
@@ -1273,6 +1273,7 @@ def _solve_nonlinear(self):
         # TODO: It is pointless to run this ahead of non-iterative solvers.
         for sub in self.system_iter(recurse=True):
             if hasattr(sub, 'guess_nonlinear'):
+                self._transfer('nonlinear', 'fwd')
```
hwangjt: You probably want to use `self._transfer('nonlinear', 'fwd', isub)` here
e.g., https://github.com/OpenMDAO/blue/blob/master/openmdao/solvers/nonlinear/nonlinear_block_gs.py#L37

Kenneth-T-Moore: Not sure how to find the correct isub, since we are iterating over a recursive list of subsystems. Is it stored somewhere keyed by pathname?

hwangjt: Ah, right. The way I'd suggest implementing this would be to call `_guess_nonlinear`, instead of iterating at all. Then, `_guess_nonlinear` would have to be a method on group and component. Group's `_guess_nonlinear` would just call its children's `_guess_nonlinear`, and component's `_guess_nonlinear` would call `guess_nonlinear` in an unscaled context

Kenneth-T-Moore: Yeah, that's the only way around it, to assure that we scatter the minimum amount of data that we need.

openmdao/core/group.py
```diff
@@ -1270,17 +1271,25 @@ def _solve_nonlinear(self):
...
         # We need to call this early enough so that any solver that needs initial guesses has
         # them.
-        # TODO: It is pointless to run this ahead of non-iterative solvers.
-        for sub in self.system_iter(recurse=True):
-            if hasattr(sub, 'guess_nonlinear'):
-                with sub._unscaled_context(outputs=[sub._outputs], residuals=[sub._residuals]):
-                    sub.guess_nonlinear(sub._inputs, sub._outputs, sub._residuals)
+        # TODO: It is pointless to run this ahead of non-iterative solvers, but multi-level
+        # models seem to ned it.
+        self._guess_nonlinear()
```
hwangjt: Does this mean all groups run a guess_nonlinear iteration at the start of each solve_nonlinear?

hwangjt: I realize that nothing would happen if no subsystem implements `guess_nonilnear`, but just worried about the overhead of the extra function calls

Kenneth-T-Moore: I am not quite finished with the changes from before. But yeah, right now all groups start with an iteration of solve_nonlinear, but I am going to try to limit that to Newton now. (and Brent or Broyden when we have them.)

hwangjt: Ah, I see

Kenneth-T-Moore: Ok, everything pushed up and ready to look at.

openmdao/core/group.py
```diff
@@ -1267,20 +1268,22 @@ def _solve_nonlinear(self):
...
 
         return result
 
+    def _guess_nonlinear(self):
+        """
+        Provide initial guess for states.
+        """
+        from openmdao.api import ImplicitComponent
+        for isub, sub in enumerate(self._subsystems_myproc):
+            if overrides_method('guess_nonlinear', sub, ImplicitComponent) or \
```
naylor-b: this check has some overhead, so I'm not sure we want to be doing it all of the time.  I would suggest adding a boolean flag to System, maybe _has_guess or something, default it to false, and in one of the setup methods in ImplicitComponent, do the check and set the flag.  That way we won't have to do the import of ImplicitComponent here either.  I guess in Group you might want to set the flag to True, although it wouldn't be hard to determine if any of the Group's children had the flag set to True and if not, leave it False.

openmdao/core/implicitcomponent.py
```diff
@@ -71,6 +71,13 @@ def _solve_nonlinear(self):
             else:
                 return result
 
+    def _guess_nonlinear(self):
+        """
+        Provide initial guess for states.
+        """
+        with self._unscaled_context(outputs=[self._outputs], residuals=[self._residuals]):
+            self.guess_nonlinear(self._inputs, self._outputs, self._residuals)
```
naylor-b: shouldn't we do the check here (maybe using the flag I mentioned earlier) in order to avoid unnecessary scaling operations when the system doesn't provide a guess?

Kenneth-T-Moore: I do the check in Group, so that it never calls into here if the system doesn't provide a guess.

openmdao/core/group.py
Kenneth-T-Moore: That is true, and it might save us some unnecessary group scatters too.

openmdao/core/implicitcomponent.py
naylor-b: I was looking at the call coming from newton, but I guess self._system in newton will always be a Group then?

Kenneth-T-Moore: Technically you could put Newton on a component, but I don't know how that would work unless you manually call the solve in your solve_nonlinear, and if you do that, you will probably end up inside of nested calls of the context manager ... don't know if we check for that and prevent multiplying by the units scales again. So, not sure it even works in practice.

Kenneth-T-Moore: I guess I take some of that back. We do automatially call the nonlinear_solver on component if it is there.

openmdao/core/group.py
Kenneth-T-Moore: Fixed

openmdao/core/group.py
```diff
@@ -1267,20 +1270,21 @@ def _solve_nonlinear(self):
...
         with Recording(name + '._solve_nonlinear', self.iter_count, self):
             result = self._nonlinear_solver.solve()
 
         return result
 
+    def _guess_nonlinear(self):
+        """
+        Provide initial guess for states.
+        """
+        from openmdao.api import ImplicitComponent
```
naylor-b: You don't need this import any more

openmdao/core/implicitcomponent.py
```diff
@@ -26,6 +26,15 @@ def __init__(self, **kwargs):
         """
         super(ImplicitComponent, self).__init__(**kwargs)
 
+        if overrides_method('guess_nonlinear', self, ImplicitComponent):
```
naylor-b: just curious why the overrides_method check for guess_nonlinear is in __init__ but the one for apply_linear is in configure...

openmdao/core/group.py
Kenneth-T-Moore: Fixed

openmdao/core/implicitcomponent.py
Kenneth-T-Moore: I need to recurse from the bottom up so that groups get their flag set, and this method goes from bottom up.

openmdao/core/group.py
```diff
@@ -1267,20 +1270,20 @@ def _solve_nonlinear(self):
...
             result = self._nonlinear_solver.solve()
 
         return result
 
+    def _guess_nonlinear(self):
+        """
+        Provide initial guess for states.
+        """
+        for isub, sub in enumerate(self._subsystems_myproc):
+            if self._has_guess:
```
naylor-b: I would put the if self._has_guess check outside of the loop, and then maybe put an 'if sub._has_guess' inside the loop.

Kenneth-T-Moore: Done

PR #345, opened 2017-08-24
New docs structure

kmarsteller:
Plenty of work needs to be done on content.  This pull is all about structure and organization.
Going from a "throw everything in one doc" approach to a modular, every dir has an index, most files are "bite-sized" approach.  Changed lots of structure and naming, fixed build problems, refined the source-doc-building, did a little bit to unify capitalization and formatting.

To view what these docs look like when built, check: 
http://blue-kgm.readthedocs.io/en/latest/index.html


kmarsteller:
Whoops.  Failing a couple tests.  Will investigate.

PR #346, opened 2017-08-24
Cleanup of iprint stuff

Kenneth-T-Moore:


openmdao/solvers/solver.py
```diff
@@ -23,13 +23,48 @@ class SolverInfo(object):
...
         Prefix to prepend during this iprint.
+    stack : List
+        List of strings; strings are popped and appended as needed.
     """
 
     def __init__(self):
         """
         Initialize.
         """
         self.prefix = ""
```
naylor-b: As far as I can tell, the prefix string simply mirrors what's in the stack, i.e., you can produce the prefix by just doing a ''.join(stack), so I would get rid of prefix.

Kenneth-T-Moore: The prefix is a cache for the stack, so that we don't have to do that .join every time we want to get the prefix. I know that string operations aren't slow, but similarly an extra string of characters isn't a large amount of RAM.

naylor-b: Yeah, the RAM isn't an issue.  I just don't like having two data attributes when we only need one.  I'm not seeing places where caching prefix is going to make much difference, so I would opt for the shorter, cleaner code.  Without caching prefix, you don't really even need those append_solver, append_subsolver, etc. functions.  Getting rid of those extra function calls would probably save you more time than you're saving by caching prefix.  We can always benchmark it later though.  I won't hold this up.

Kenneth-T-Moore: I don't see how I'd be able to get rid of `append_solver` etc. Moving those to functions cleaned up code in the solvers, and unified the printing behavior, so that if i add a new solver with subsolver or precon, I can't accidentally use a different indentation format. Also makes it easier to change it all.  If I get rid of prefix, I will need to add a method get_prefix to use in the solvers because doing a bunch of joins there is unclear.

PR #347, opened 2017-08-25
For aitken

shamsheersc19:
Added Aitken's relaxation for the NLBGS solver. 
Added a test for it.
This also includes some commits by hwangjt (dot product for vectors).


openmdao/solvers/nonlinear/nonlinear_block_gs.py
```diff
@@ -26,18 +26,95 @@ def _setup_solvers(self, system, depth):
...
+        Perform any necessary pre-processing operations.
+
+        Returns
+        -------
+        float
+            initial error.
+        float
+            error at the first iteration.
+        """
+        self._aitken_work1 = self._system._outputs._clone()
```
Kenneth-T-Moore: All of the new code here and in `_iter_execute` should be under an `if self.options['use_aitken'] is True` block. Don't want to be cloning the output vector and doing all the other calculation if we aren't using aitken.

openmdao/vectors/default_vector.py
```diff
@@ -426,6 +426,26 @@ def set_const(self, val):
...
             data[:] = val
 
+    def dot(self, vec):
+        """
+        Compute the dot product of the real parts of the current vec and the incoming vec.
+
+        Parameters
+        ----------
+        vec : <Vector>
+            the vector whose values self is set to.
```
dwmunster: It looks like this comment may have been copied from `set_vec` without being updated.

openmdao/vectors/vector.py
```diff
@@ -491,6 +491,19 @@ def set_const(self, val):
...
+    def dot(self, vec):
+        """
+        Compute the dot product of the current vec and the incoming vec.
+
+        Must be implemented by the subclass.
+
+        Parameters
+        ----------
+        vec : <Vector>
+            the vector whose values self is set to.
```
dwmunster: It looks like this comment may have been copied from `set_vec` without being updated.

openmdao/vectors/default_vector.py
```diff
@@ -426,6 +426,26 @@ def set_const(self, val):
...
+
+        Returns
+        -------
+        float
+            The computed dot product value.
+        """
+        global_sum = 0
+        for set_name, data in iteritems(self._data):
+            global_sum += np.sum(self._data[set_name] * vec._data[set_name])
```
dwmunster: This will create a temporary array of the same size as the data. Why not use `np.dot` or `np.einsum`?

openmdao/solvers/nonlinear/nonlinear_block_gs.py
shamsheersc19: Thanks, I have added the IF statements.

openmdao/vectors/default_vector.py
shamsheersc19: Corrections have been made with hwangjt's inputs. np.sum has been replaced with np.dot. Thanks.

PR #356, opened 2017-09-06
Fixed a couple of DirectSolver related problems

Kenneth-T-Moore:
1. Added `mode` to argument of `_linearize` so that solvers like DirectSolver can set up their linear solve with the requested mode.
2. Fixed DirectSolver to also work in rev mode: prior implementation was unfinished but also never tested because of bug 1.

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -23,16 +23,10 @@ class DirectSolver(LinearSolver):
 
     SOLVER = 'LN: Direct'
 
-    def _linearize(self, mode='fwd'):
+    def _linearize(self):
         """
         Perform factorization.
-
-        Parameters
-        ----------
-        mode : str
-            Mode for derivative calculation, default is fwd.
         """
-        self._mode = mode
         system = self._system
 
         if system._owns_assembled_jac or system._views_assembled_jac:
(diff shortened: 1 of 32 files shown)


openmdao/solvers/linear/direct.py
```diff
@@ -59,7 +65,10 @@ def _linearize(self):
             eye = np.eye(nmtx)
             mtx = np.empty((nmtx, nmtx))
             for i in range(nmtx):
-                self._mat_vec(eye[:, i], mtx[:, i])
+                if self._mode == 'fwd':
```
JustinSGray: I don't understand why this is necessary. looping of columns or rows of identity gives the same results regardless. mtx is just an empty matrix. There should not be any difference.

openmdao/solvers/linear/direct.py
```diff
@@ -82,16 +91,20 @@ def _mat_vec(self, in_vec, out_vec):
         system = self._system
         mode = self._mode
 
-        # assign x and b vectors based on fwd mode
-        x_vec = system._vectors['output'][vec_name]
-        b_vec = system._vectors['residual'][vec_name]
+        # assign x and b vectors based on mode
```
JustinSGray: I don't understand why this is necessary. You should be able to assemble the jacobian via fwd products regardless. 

Why do we need mode?

Kenneth-T-Moore: After discussion with Justin, we decided to always assemble the Jac in forward mode, like Newton, so the change to the _linearize API is unneeded.

PR #363, opened 2017-09-14
adding high-level purposes to index docs; creating ref labels

kmarsteller:


kmarsteller:
Small story (https://www.pivotaltracker.com/story/show/150599227) that I broke out of my current branch.  Feel free to make edits or suggest crosslinks, and I'll add them.

PR #364, opened 2017-09-14
API transition guide - side by side comparison of openmdao Alpha vs. Beta for selected features.

Kenneth-T-Moore:
1. Added a new Sphinx extension to produce the side by side comparison.
2. Gave get_source_code_of_class_or_method an option to remove or not remove the docstrings.
3. Started API transition guide.


openmdao/docs/_utils/docutil.py
```diff
@@ -191,10 +191,11 @@ def remove_initial_empty_lines_from_source(source):
 # pylint: disable=C0103
 
 
-def get_source_code_of_class_or_method(class_or_method_path):
-    '''The function to be called a the custom Sphinx directive code
+def get_source_code_of_class_or_method(class_or_method_path, remove_docstring=True):
+    """
+    The function to be called a the custom Sphinx directive code
```
kmarsteller: This comment is poorly worded.  Oh wait, you inherited that comment.

openmdao/docs/api_translation.rst
```diff
@@ -0,0 +1,363 @@
+.. _`api_translation`:
```
kmarsteller: The only comment I have here is that in the current re-structure all the docs directly beneath our top-level index.rst have their own directories (e.g. docs/user_guide),  with index.rst, docs, and other related items inside.  This doc kind of violates that structure, but then again, it doesn't NEED a directory of sub-docs, I suppose?  In theory, each section could be a sub-doc in an index, to keep things in @JustinSGray 's "bite-size" docs directive?  Anyway, just something to think about.

kmarsteller: (Note, the above re-structure is not something you need to do...just thinking aloud, and if it's decided to do it, I could take care of it after it goes in.)

openmdao/docs/_utils/docutil.py
Kenneth-T-Moore: Fixed the comment. We really should have policed the style and pep stuff while these where being written.

openmdao/docs/api_translation.rst
Kenneth-T-Moore: Too late. already put it in its own directory.

openmdao/docs/index.rst
```diff
@@ -10,6 +10,7 @@ OpenMDAO Documentation
     user_guide/index.rst
     feature_reference/index.rst
     examples/index.rst
+    api_translation/index.rst
```
kmarsteller: Thanks for the fixes...the one thing missing...I see the old version of the doc being "removed" but don't see the index.rst being "added"

Kenneth-T-Moore: Fixed

PR #366, opened 2017-09-14
fix for instance based profile viewer to make it work with larger models

naylor-b:
- make sure to reinstall openmdao after pulling this, because the iprofview console script now points to a different place than before
- also made a small update to the distributed component feature doc to explain when distributed = True is necessary.
- updated the call tracing docs.


openmdao/devtools/iprofile_app/iprofile_app.py
```diff
@@ -0,0 +1,206 @@
...
+    return thread
+
+def _parent_key(d):
+    parts = d['id'].rsplit('-', 1)
+    if len(parts) == 1:
+        return ''
+    return parts[0]
+
+def stratify(call_data, sortby='time'):
+    """
```
Kenneth-T-Moore: Maybe being pedantic here, but since this tool is in our feature reference, it is user-facing, so we really should follow all of our PEP/etc coding standards for docstrings and everything else.

naylor-b: The only user facing functions for any of the tracers/profilers are setup(), start() and stop(), and those have the proper docstrings.  I did go through and make a bunch of stuff private and added some docstrings here and there.

PR #369, opened 2017-09-18
Moved guess_nonlinear so that it is executed before any subsolves in Newton

Kenneth-T-Moore:


openmdao/recorders/recording_manager.py
```diff
@@ -59,7 +59,7 @@ def __iter__(self):
 
     def _gather_vars(self, root, local_vars):
         """
-        Gathers and returns only variables listed in `local_vars` from the `root` System.
+        Gather and returns only variables listed in `local_vars` from the `root` System.
```
naylor-b: returns --> return

Kenneth-T-Moore: Fixed

PR #373, opened 2017-09-21
Fix to DenseJacobians so that they properly obey the in_scope and out_scope that is requested by apply or solve.

Kenneth-T-Moore:
1. Fixed this problem with numpy masks. 
2. The mask is cached so that it doesn't recompute it. This requires the storage of a couple extra numpy bool arrays of the same length as the inputs vector at the level the DenseJacobain is applied.
3. Also got rid of another skipped test we won't be needing.
4. DenseJacobian will not become the default for Components at this time, but may in the future.


openmdao/jacobians/assembled_jacobian.py
```diff
@@ -327,14 +333,90 @@ def _apply(self, d_inputs, d_outputs, d_residuals, mode):
         with system._unscaled_context(
                 outputs=[d_outputs], residuals=[d_residuals]):
             if mode == 'fwd':
-                d_residuals.iadd_data(int_mtx._prod(d_outputs.get_data(), mode, int_ranges))
-                if ext_mtx is not None:
-                    d_residuals.iadd_data(ext_mtx._prod(d_inputs.get_data(), mode, None))
+                if len(d_outputs._names) > 0 and len(d_residuals._names) > 0:
```
naylor-b: I'd recommend just using 'if d_outputs._names and d_residuals._names:' instead of testing if len(x) > 0.  It's more concise and faster as well.

openmdao/jacobians/assembled_jacobian.py
```diff
@@ -327,14 +333,90 @@ def _apply(self, d_inputs, d_outputs, d_residuals, mode):
...
+        Parameters
+        ----------
+        d_inputs : Vector
+            The inputs linear vector.
+        cache_key : tuple
+            Hashable unique key, from d_inputs._names
+        ext_mtx : Matrix
+            External matrix
+        """
+        masked = [name for name in d_inputs._views if name not in cache_key]
```
naylor-b: This loop should be flipped since checking containment in d_inputs_views (a dict) will be much faster than checking containment in cache_key (a tuple).

openmdao/jacobians/assembled_jacobian.py
```diff
@@ -327,14 +333,90 @@ def _apply(self, d_inputs, d_outputs, d_residuals, mode):
...
+                        # Mask need to be applied to ext_mtx so that we can ignore multiplication
+                        # by certain columns.
+                        mask = np.zeros(ext_mtx._matrix.T.shape, dtype=np.bool)
+                        mask[mask_cols, :] = True
+                        masked_mtx = np.ma.array(ext_mtx._matrix, mask=mask, fill_value=0.0)
+
+                        masked_product = np.ma.multiply(masked_mtx.T, dresids).flatten()
+
+                        for set_name, data in iteritems(d_inputs._data):
+                            data += np.ma.filled(masked_product)
```
hwangjt: It doesn't look like this would work with multiple varsets. If that's the case, we should put in a TODO to generalize this later for running with multiple varsets

openmdao/jacobians/assembled_jacobian.py
Kenneth-T-Moore: Fixed

openmdao/jacobians/assembled_jacobian.py
Kenneth-T-Moore: Fixed

openmdao/jacobians/assembled_jacobian.py
Kenneth-T-Moore: I will add a note for this and other advanced features.

PR #377, opened 2017-09-28
Some derivatives API Changes plus improved check features.

Kenneth-T-Moore:
API CHANGES
-----------------
1. The default `dependent` status for subjacobians has been changed from dependent to independent. This means that all derivatives are assumed to be zero unless they are explicitly declared. You will have to make changes to all components that have derivatives.
2. If you try to set an undeclared derivative pair on a jacobian, you will get a keyerror.
3. The `approx_partials` method has been removed. You can use `declare_partials` with the same arguments to declare finite difference or complex step.
4. The `compute_total_derivs` method is now called `compute_totals`.
5. The `check_total_derivs` method is now called `check_totals`.
6. The `check_partial_derivs` method is now called `check_partials`.
7. The `approx_total_derivs` method is now called `approx_totals`.

NEW FEATURES
-------------------
1. You can now pass in a dictionary of global FD options when calling `check_partials` and the approximation will use those settings.
2. Complex Step is supported in `check_partials` now.
3. You may change the options for individual variables in `check_partials` by calling `set_check_partial_options` on a component.
4. The precedence for check options is component > call arguments > global defaults


openmdao/core/component.py
```diff
@@ -615,12 +586,110 @@ def declare_partials(self, of, wrt, dependent=True,
         val : float or ndarray of float or scipy.sparse
             Value of subjacobian.  If rows and cols are not None, this will
             contain the values found at each (row, col) location in the subjac.
+        method : str
+            The type of approximation that should be used. Valid options include:
+                - 'fd': Finite Difference, 'cs': Complex Step
```
naylor-b: should this line mention 'exact'?

openmdao/core/component.py
```diff
@@ -615,12 +586,110 @@ def declare_partials(self, of, wrt, dependent=True,
...
+            Finite difference form for check, can be "forward", "central", or "backward". Leave
+            undeclared to keep unchanged from previous or default value.
+        step : float
+            Step size for finite difference check. Leave undeclared to keep unchanged from previous
+            or default value.
+        step_calc : str
+            Type of step calculation for check, can be "abs" for absolute (default) or "rel" for
+            relative.  Leave undeclared to keep unchanged from previous or default value.
+        """
+        supported_methods = {'fd': FiniteDifference,
```
naylor-b: no need for a dict here because you don't use the values. Could be a set, or since it's so short, maybe just a tuple.

openmdao/core/component.py
```diff
@@ -615,12 +586,110 @@ def declare_partials(self, of, wrt, dependent=True,
...
+        -------
+        dict(wrt : (options))
+            Dictionary keyed by name with tuples of options (method, form, step, step_calc)
+        """
+        opts = {}
+        outs = list(self._var_allprocs_prom2abs_list['output'].keys())
+        ins = list(self._var_allprocs_prom2abs_list['input'].keys())
+        for wrt_list, method, form, step, step_calc in self._declared_partial_checks:
+            for pattern in wrt_list:
+                wrt_matches = set(find_matches(pattern, outs + ins))
```
naylor-b: no need to create a set here since you're only using it for iteration an not lookups.

openmdao/test_suite/components/exec_comp_for_test.py
```diff
@@ -54,6 +54,10 @@ def __init__(self, exprs, nl_delay=0.01, lin_delay=0.01,
         self.fails = fails
         self.fail_hard = fail_hard
 
+    def setup(self):
```
naylor-b: shouldn't need to define setup since the base class already declares the partials.

openmdao/core/component.py
Kenneth-T-Moore: fixed

openmdao/core/component.py
Kenneth-T-Moore: fixed

openmdao/core/component.py
Kenneth-T-Moore: fixed

openmdao/test_suite/components/exec_comp_for_test.py
Kenneth-T-Moore: fixed

openmdao/core/component.py
```diff
@@ -615,12 +586,110 @@ def declare_partials(self, of, wrt, dependent=True,
...
+            Finite difference form for check, can be "forward", "central", or "backward". Leave
+            undeclared to keep unchanged from previous or default value.
+        step : float
+            Step size for finite difference check. Leave undeclared to keep unchanged from previous
+            or default value.
+        step_calc : str
+            Type of step calculation for check, can be "abs" for absolute (default) or "rel" for
+            relative.  Leave undeclared to keep unchanged from previous or default value.
+        """
+        supported_methods = ('fd', 'cs', 'exact')
```
JustinSGray: we allow them to "check" their partials with an exact reference? Would anyone ever use anything but FD or CS?

Kenneth-T-Moore: Fixed

openmdao/core/component.py
```diff
@@ -615,12 +586,110 @@ def declare_partials(self, of, wrt, dependent=True,
         val : float or ndarray of float or scipy.sparse
             Value of subjacobian.  If rows and cols are not None, this will
             contain the values found at each (row, col) location in the subjac.
+        method : str
+            The type of approximation that should be used. Valid options include:
+                - 'fd': Finite Difference, 'cs': Complex Step, 'exact': use the component
```
swryan: suggest 'compute' as more descriptive than `exact` (and semantically links to the `compute_partials` method)

JustinSGray: but you also 'compute' them via FD or CS.

swryan: the framework does in those cases, versus the user implementing the `compute` method, right? so `user` maybe...  just seems like `exact` doesn't really capture the idea that the user has to implement the `compute` function...  /shrug

openmdao/core/problem.py
```diff
@@ -642,12 +638,12 @@ def check_partials(self, logger=None, comps=None, compact_print=False,
 
                             # Testing for pairs that are not dependent so that we suppress printing
                             # them unless the fd is non zero. Note: subjacs_info is empty for
-                            # undeclared partials on implicit components.
+                            # undeclared partials, which is the default behvaior now.
```
swryan: typo "behvaior"... also, the "now" is not really necessary

openmdao/core/tests/test_getset_vars.py
```diff
@@ -47,15 +47,8 @@ def test_no_promotion(self):
...
-
-            # d(output)/d(input)
-            jac['c.y', 'c.x'] = 5.0
-            self.assertEqual(jac['c.y', 'c.x'], 5.0)
-
-            # d(output)/d(output)
-            jac['c.y', 'c.y'] = 5.0
-            self.assertEqual(jac['c.y', 'c.y'], 5.0)
+        # Removed part of test where we set values into the jacobian willy-nilly. You can only set
+        # declared values now.
```
swryan: will need to clean up references to old code at some point

openmdao/core/problem.py
Kenneth-T-Moore: will fix on justin's PR

PR #379, opened 2017-09-29
some speedups

naylor-b:
- added caching for subjac iter keys for DictionaryJacobian and AssembledJacobian
- eliminated unnecessary vector cloning during derivative approximation
- eliminated unnecessary overhead in ExplicitComponent._linearize when component doesn't compute partials.
- changed some internal vector indices to a slice when there is only a single varset.


openmdao/core/explicitcomponent.py
```diff
@@ -36,6 +38,7 @@ def __init__(self, **kwargs):
         super(ExplicitComponent, self).__init__(**kwargs)
 
         self._inst_functs = {name: getattr(self, name, None) for name in _inst_functs}
+        self._has_derivs = overrides_method('compute_partials', self, ExplicitComponent)
```
Kenneth-T-Moore: What if they define `compute_jacvec_product` but not `compute_partials`?

Kenneth-T-Moore: Ok, looks like it doesn't matter for the places where it is used.

naylor-b: I changed the name of the flag to _has_compute_partials to make things clearer.

PR #385, opened 2017-10-03
clean up some deprecations in benchmarks

swryan:


benchmark/benchmark_manycomps.py
```diff
@@ -4,23 +4,24 @@
...
 from openmdao.test_suite.build4test import DynComp, create_dyncomps
 
+
 class BM(unittest.TestCase):
     """Setup of models with lots of components"""
 
     def benchmark_100(self):
-        p = Problem(root=Group())
-        create_dyncomps(p.root, 100, 10, 10, 5)
+        p = Problem()
```
Kenneth-T-Moore: Wonder if we need a way to "run all" on the benchmark suite before we pull-request.

swryan: `testflo -bv`

Kenneth-T-Moore: oh yeah, the old benchflow

PR #386, opened 2017-10-03
Added append option for SqliteRecorder and by default the recorder file is deleted if it exists

hschilling:


openmdao/recorders/sqlite_recorder.py
```diff
@@ -56,7 +57,7 @@ class SqliteRecorder(BaseRecorder):
         Sqlite3 system cursor via the con.
     """
 
-    def __init__(self, out):
+    def __init__(self, filepath, append=False):
         """
```
Kenneth-T-Moore: need to add these to Parameters in the class docstring

kmarsteller:
So the story was basically that people were tired of removing the old file?  But someday, maybe we want append power?

PR #390, opened 2017-10-04
Metadata API changes and added a feature doc for metadata

hwangjt:
The API is documented in the new feature doc, but here's a summary of the API changes:
- The required argument is now gone. It was redundant because it is required if default is not given and it's not required if default wasn't specified
- The default value of `default` is now an instance of a dummy Null class, so if the default was assumed to be None in the past, the metadata must be declared with an explicit, `declare=None`

### DIFF
--- a/openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
+++ b/openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
@@ -1,0 +1,78 @@
+.. _component_metadata:
+
+********************************************
+Component metadata (arguments to components)
+********************************************
+
+Components, whether explicit or implicit, define inputs and outputs and how to map inputs to outputs.
+Often, there are incidental parameters that affect the behavior of the component
+but are not considered input variables in the sense that their values are
+computed as an output of another component.
+
+For convenience, OpenMDAO provides a way of declaring these parameters, which we call *metadata*.
+They are declared in the 'initialize' method of the user's component,
+and the values are typically passed in upon instantiation of the component.
+Metadata can be declared along with their default values and various checks for validity,
+such as a list of acceptable values or types.
+The full list of options can be found in the method signature at the bottom of this page.
+
+A simple example
+----------------
+
+A common use of metadata is to specify the shape or size of the component's input and output variables.
+
+.. embed-code::
+    openmdao.utils.tests.test_options_dictionary_feature.VectorDoublingComp
+
+.. embed-test::
+    openmdao.utils.tests.test_options_dictionary_feature.TestOptionsDictionaryFeature.test_simple
+
+In this example, 'size' is required; the first time we try to get 'size',
+we would have got an error if we
+(1) did not pass in 'size' when instantiating 'VectorDoublingComp' and
+(2) did not set its value in the code for VectorDoublingComp.
+
+Providing default values
+------------------------
+
+One reason why using metadata is convenient is that a default value can be specified,
+allowing passing the value in during component instantiation to be optional.
+
+.. embed-code::
+    openmdao.utils.tests.test_options_dictionary_feature.LinearCombinationComp
+
+.. embed-test::
+    openmdao.utils.tests.test_options_dictionary_feature.TestOptionsDictionaryFeature.test_with_default
+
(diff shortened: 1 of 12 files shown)


openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
```diff
@@ -0,0 +1,82 @@
...
+
+The primary job of a component, whether explicit or implicit, is to define inputs and outputs,
+as well as the mapping that computes the outputs given the inputs.
+Often, however, there are incidental parameters that affect the behavior of the component
+but are not considered input variables in the sense of being computed as an output of another component.
+
+For convenience, OpenMDAO provides a way of declaring these parameters, which are referred to as *metadata*.
+They are declared in the 'initialize' method of the user's component,
+and the values are typically passed in upon instantiation of the component.
+Alternatively, the values can be set at any time inside or outside of the component.
```
JustinSGray: what does "inside or outside" the component mean? This sentence could be made more clear

openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
```diff
@@ -0,0 +1,82 @@
...
+********************************************
+Component metadata (arguments to components)
+********************************************
+
+The primary job of a component, whether explicit or implicit, is to define inputs and outputs,
+as well as the mapping that computes the outputs given the inputs.
+Often, however, there are incidental parameters that affect the behavior of the component
+but are not considered input variables in the sense of being computed as an output of another component.
+
+For convenience, OpenMDAO provides a way of declaring these parameters, which are referred to as *metadata*.
```
JustinSGray: I think some clarification is needed here, since we're telling them to do something very non-standard to declare init-args. 

Perhaps its just the "for convenience" that I don't like.

openmdao/utils/tests/test_options_dictionary_feature.py
```diff
@@ -0,0 +1,112 @@
...
+        self.declare_partials('y', 'x', val=self.metadata['a'])
+
+    def compute(self, inputs, outputs):
+        outputs['y'] = self.metadata['a'] * inputs['x'] + self.metadata['b']
+
+
+class UnitaryFunctionComp(ExplicitComponent):
+
+    def initialize(self):
+        from types import FunctionType
```
JustinSGray: Its weird to import FunctionType inside the method definition like this. Its obviouly valid code, but its not how I would instruct a user to do it. 

Im guessing you did this so the import would show up in the embed? Could we move the class definition into its own file instead? Then you can embed the file with the import at the top. We'll need the import for component as well anyway.

openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
```diff
@@ -0,0 +1,82 @@
...
+    openmdao.utils.tests.test_options_dictionary_feature.UnitaryFunctionComp
+
+.. embed-test::
+    openmdao.utils.tests.test_options_dictionary_feature.TestOptionsDictionaryFeature.test_values_and_types
+
+In this example, it is convenient to specify a list of acceptable values of the 'func' metadata,
+but also to provide a valid type if it is not one of these acceptable values.
+
+
+Method Signature
```
JustinSGray: Method Signature should come first, before usage examples

openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
hwangjt: Elaborated on how metadata can be set by passing in during instantiation

openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
hwangjt: Rephrased

openmdao/utils/tests/test_options_dictionary_feature.py
hwangjt: The classes have been moved into separate files

openmdao/docs/feature_reference/core_features/defining_components/metadata.rst
hwangjt: done

PR #393, opened 2017-10-10
Overhaul of embed_test directive.

Kenneth-T-Moore:
1. Modified embed_test so that code is run under `exec` instead of writing it to a file and executing with subprocess. The change only affects serial tests; an mpi test still executes the old way.
2. All embedded feature tests now must include import statements. Those that do not will show failure. All existing feature tests were edited to include them.
3. Brought back the "no-split'" option to keep all the output together.
4. Fixed various issues in a some tests output.


openmdao/docs/Makefile
```diff
@@ -41,7 +41,7 @@ clean:
 	rm -rf _srcdocs/*
 	rm -rf tags
 	rm -rf tmp
-	rm -f feature_docs_unit_test_db.sqlite
+	# rm -f feature_docs_unit_test_db.sqlite
```
naylor-b: was commenting out the db a debugging thing or is it intentional?

swryan: I'd suggest just deleting this stuff if it's no longer needed

openmdao/docs/_utils/create_feature_docs_unit_test_db.py
```diff
@@ -31,166 +31,164 @@
...
-                                title = test_doc_numpy['Summary'][0]
-                                unit_test_source, run_outputs, skipped, failed = \
-                                    get_unit_test_source_and_run_outputs(method_path)
-                                cur.execute(
-                                    'insert into {tn} values (?,?,?,?,?)'.format(tn=table_name),
-                                    (method_path, feature, title, unit_test_source, run_outputs))
-
-conn.commit()
-conn.close()
+#import importlib
```
swryan: just delete the file?  it's always there in the repo if it's needed again...

openmdao/docs/Makefile
Kenneth-T-Moore: That's from the old numpydocs stuff that used to search through all the tests and find the ones we marked as features. I will take Steve's suggestion and delete the lines and code associated with that.

openmdao/docs/_utils/create_feature_docs_unit_test_db.py
Kenneth-T-Moore: Ok, I will delete that old stuff we don't use.

PR #394, opened 2017-10-11
autodetection of input and/or output scaling

naylor-b:
- If none of ref, ref0, res_ref are set, we skip output scaling.
- if there is no unit conversion and ref/ref0/res_ref are not set on a connected output, we skip input scaling
- there will be a couple of follow-on PRs to this one.  One will keep scalar scalers as scalar to save memory, and the other will turn scaling vectors into scaling arrays that live alongside data in the other Vector objects in order to avoid the setup overhead of creating so many Vector objects.

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -706,7 +706,7 @@ class Group(System):
                         needs_input_scaling = np.any(res_ref != 1.0)
 
                 # if units are defined and different, we need input scaling
-                if needs_input_scaling or (in_units and in_units != out_units):
+                if needs_input_scaling or (in_units and out_units and in_units != out_units):
                     self._has_input_scaling = True
 
         # Now that both implicit & explicit connections have been added,
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -658,6 +658,12 @@ class Group(System):
         else:
             path_len = len(pathname) + 1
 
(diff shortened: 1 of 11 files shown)


openmdao/devtools/run_test.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Run individual test(s).
+    """
+
+    sys.path.append('.')
+    testspec = sys.argv[1]
+
+    parts = testspec.split(':')
+    if len(parts) != 2:
+        print('Usage: run_test my_file_path:my_test_case.test_func_name')
```
swryan: the Usage statement doesn't seem to cover both use cases per the docstring

naylor-b: fixed

openmdao/core/group.py
```diff
@@ -669,12 +681,40 @@ def _setup_connections(self, recurse=True):
...
+
+                if not needs_input_scaling:
+                    res_ref = allprocs_abs2meta_out[abs_out]['res_ref']
+                    if np.isscalar(res_ref):
+                        needs_input_scaling = res_ref != 1.0
+                    else:
+                        needs_input_scaling = np.any(res_ref != 1.0)
+
+                # if units are defined and different, we need input scaling
+                if needs_input_scaling or (in_units and out_units and in_units != out_units):
```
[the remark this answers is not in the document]

PR #397, opened 2017-10-12
Add recording of non VOI system vars to driver recording

hschilling:


.travis.yml
```diff
@@ -1,6 +1,7 @@
 branches:
   only:
   - master
+  - add_system_vars_to_driver_recording
```
kmarsteller: This should not remain in our real .travis.yml

openmdao/recorders/base_recorder.py
```diff
@@ -247,11 +257,38 @@ def startup(self, object_requesting_recording):
...
+                # If MPI, and on rank 0, need to gather up all the variables
+                #    even those not local to rank 0
+                if MPI:
+                    all_vars = root.comm.gather(mysystem_outputs, root=0)
+                    if MPI.COMM_WORLD.rank == 0:
+                        mysystem_outputs = all_vars[-1]
+                        for d in all_vars[:-1]:
+                            mysystem_outputs.update(d)
+                # Filter out only the ones we want
+                # mysystem_outputs = {n for n in mysystem_outputs
```
kmarsteller: Did we want the filtering commented out, or is this done elsewhere now?

.travis.yml
hschilling: fixed and pushed up again

openmdao/recorders/base_recorder.py
hschilling: fixed and pushed up

PR #406, opened 2017-10-18
view_model produces single html file and added an embed option

ryanfarr01:


openmdao/devtools/problem_viewer/problem_viewer.py
```diff
@@ -180,33 +181,104 @@ def view_model(problem_or_filename, outfile='visualization', show_browser=True):
...
-        f.write('var modelData = %s' % model_data)
-
+    vis_dir = os.path.join(code_dir, "visualization")
+    libs_dir = os.path.join(vis_dir, "libs")
+    src_dir = os.path.join(vis_dir, "src")
+    style_dir = os.path.join(vis_dir, "style")
+
+    #grab the libraries
+    with open(os.path.join(libs_dir, "awesomplete.js"), "r") as f:
+        awesomplete = "%s" % (f.read())
```
naylor-b: just curious why you're doing the extra step of "%s" % f.read() rather than just doing 'awesomplete = f.read()'

ryanfarr01: Good question! I actually was just doing it to stay consistent with the original implementation. Not sure why that was originally done, but I just tried it with f.read() alone and it seems to work fine. I'll push that up.

PR #411, opened 2017-10-20
OK, here are the tweaks for the 2.0.2 release

kmarsteller:


kmarsteller:
I believe I finally have this working, thanks to some help from @swryan.  Manifest.in was a red herring.  Unfortunately, before I figured out how to test this locally, I did a 2.0.1 version that I now need to replace with 2.0.2.  Sorry.

PR #425, opened 2017-10-27
Fixing the encrypted doc transfer, several other minor things:

kmarsteller:
*Determined that the old key had become deleted on WF, so regenerated a whole new set of keys and env vars, and the docs rsync works again.  Glad I had written up detailed instructions.

*Fixing two problems with the .travis.yml file:
  a. we are now running the tests from down in the doc area instead of from the top
  b. installing with `pip install .` instead of using `-e` to catch packaging errors...and...

*The above change caught a recent packaging error in the new beam example, so I fixed that.

*An update to the README, based on doc naming scheme change.

*Noticed a vestigial set of comments in the Makefile that are no longer true.

*Fixed some of my own docs with minor typos.

*Retroactively adding in release notes for 2.0.1 and 2.0.2 that I forgot.

*Reformatted .travis.yml to be much more readable.

*Merged in with latest.


kmarsteller:
Coverage has been hanging the last 2 times that CI ran, trying once more, with a slight yaml change, to see if it will run this time.

kmarsteller:
closing to do more tests on my own branch

kmarsteller:
This is the new hotness.

kmarsteller:
We are doing the installed package check now, doing coverage on it, processing the coverage file, and @swryan did a change to coveralls so we can pass our modified coverage file with usable paths over to the website, so that the coveralls site will continue to work seamlessly, all file paths correct, all links clickable.

PR #430, opened 2017-10-30
ExternalCode ported from OpenMDAO 1.x

Kenneth-T-Moore:
1. Ported ExternalCode from OpenMDAO 1.x and implemented it as an ExplicitComponent.
2. Minor modification for testing under `embed_test1` so that it can load local files.


openmdao/docs/_utils/docutil.py
```diff
@@ -662,7 +662,13 @@ def get_test_src(method_path):
             save_opts = np.get_printoptions()
             np.set_printoptions(precision=8)
 
+            # Move to the test directory in case there are files to read.
+            save_dir = os.getcwd()
+            os.chdir('/'.join(test_module.__file__.split('/')[:-1]))
+
             exec(code_to_run, {})
+
+            os.chdir(save_dir)
```
naylor-b: maybe wrap the exec in a try-finally block to make sure that no matter what you get back to the original directory.

Kenneth-T-Moore: Fixed

PR #434, opened 2017-11-02
fix for doc warning

naylor-b:


openmdao/utils/options_dictionary.py
```diff
@@ -88,9 +88,9 @@ def declare(self, name, default=_undefined, values=None, types=None, type_=None,
         Declare an option.
 
         The value of the option must satisfy the following:
-        1. If values and not types was given when declaring, value must be in values.
-        2. If types and not values was given when declaring, value must be an instance of a
-           type in types.
+        1. If values only was given when declaring, value must be in values.
+        2. If types only was given when declaring, value must satisfy isinstance(v, types)
```
swryan: this is really nit-picking but I think `v` should be `value`

naylor-b: fixed

PR #437, opened 2017-11-03
Fix the pyoptsparse installation; Change from py3.4 to py3.6; change numpy version to 1.12.

kmarsteller:
Here's a solution  where we grab the pyoptsparse code straight from bitbucket, build it, and it works. It will now work the same for Linux and MacOS (if we ever bring that testing back online), so was able to get rid of the calls for Mac pyoptsparse wheels.

It works on python3.6 and python 2.7; py3.4 machines were eliminated from the build matrix.

Numpy was moved up from 1.11.2 to 1.12 and pinned there, with comments made to remind us WHY we pinned it there.

Once pyoptsparse was actually installed, a pyoptsparse test failed.  So, I also include a couple of @Kenneth-T-Moore-approved changes to the test openmdao/drivers/tests/test_pyoptsparse_driver.py.

Also, skipped tests on this PR drop from 94 to 42, and Coverage rises from 90% to 92% with the installation of pyoptsparse.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ install:
 
 - if [ "$PY" = "3.4" ]; then
     wget https://bitbucket.org/mdolab/pyoptsparse/get/9af3264429b9.zip;
-    tar -xvf 99af3264429b9.zip;
+    unzip 99af3264429b9.zip;
     cd 99af3264429b9;
     python setup.py install;
     cd ..;
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
 - PY=3.6 MPI=1
 - PY=2.7 MPI=
 - PY=3.6 MPI=
-#- PY=3.6 MPI=
 
 addons:
   apt:
@@ -70,7 +69,7 @@ install:
 # keeping numpy pinned at 1.12 because snopt dies with 1.13 because numpy broke callback
 - conda install --yes python=$PY numpy==1.12 scipy nose sphinx mock swig
 
-#trying a manual install of pyoptsparse
+# trying a manual install of pyoptsparse
 - if [ "$PY" = "2.7" ]; then
     pip install mercurial;
     hg clone https://bitbucket.org/mdolab/pyoptsparse;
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -65,21 +66,24 @@ before_install:
...
 
 install:
-- conda install --yes python=$PY numpy==1.11.2 scipy nose sphinx mock swig
-- if [ "$OS" = "Linux" ]; then
-    if [ "$PY" = "2.7" ]; then
-      pip install https://openmdao.org/dists/pyoptsparse-1.0.0-cp27-none-linux_x86_64.whl;
-    elif [ "$PY" = "3.4" ]; then
-      pip install https://openmdao.org/dists/pyoptsparse-1.0.0-py3-none-linux_x86_64.whl;
-    fi
+# keeping numpy pinned at 1.12 because snopt dies with 1.13 because numpy broke callback
```
swryan: technically 'pySNOPT'

.travis.yml
```diff
@@ -65,21 +66,24 @@ before_install:
...
-      pip install https://openmdao.org/dists/pyoptsparse-1.0.0-py2-none-macosx_10_5_x86_64.whl;
-    elif [ "$PY" = "3.4" ]; then
-      pip install https://openmdao.org/dists/pyoptsparse-1.0.0-cp34-none-macosx_10_5_x86_64.whl;
-    fi
+- if [ "$PY" = "3.6" ]; then
+    wget https://bitbucket.org/mdolab/pyoptsparse/get/9af3264429b9.zip;
+    unzip 9af3264429b9.zip;
+    cd mdolab-pyoptsparse-9af3264429b9;
+    python setup.py install;
+    cd ..;
```
swryan: wonder if it would be better to use a this method for both versions...  the clone will mean we are always testing the latest pyoptsparse on py2.7, so I guess we should be explicit that that is what we want

kmarsteller: I could try to do the same setup and do "tip" on both...let me try that and see if it fails miserably?  OR would you rather they are both pinned to a stable commit?

swryan: I do think it would be cleaner to do it one way for both versions of python... either with `tip` or the specific commit, depending on the consensus

kmarsteller: I think we'll use "tip." Just for a few commits, just to see what it feels like.

PR #438, opened 2017-11-03
Removed `system_includes` recording option, de-duplicated driver iteration recordings and updated docs

ryanfarr01:


openmdao/core/driver.py
```diff
@@ -219,39 +222,39 @@ def _setup_driver(self, problem):
...
 
-        if self.options['record_responses']:
+        if rec_responses or incl:
             myresponses = {n for n in self._responses
-                           if check_path(n, incl, excl)}
+                           if check_path(n, incl, excl, rec_responses)}
 
         # get the system_includes that were requested for this Driver recording
-        if sys_incl:
+        if self.recording_options['includes']:
```
swryan: you have a reference to this that you have been using: `incl`

openmdao/core/driver.py
```diff
@@ -621,36 +633,39 @@ def record_iteration(self):
 
         # Get the data to record
         data = {}
-        if self.options['record_desvars']:
+        if self.recording_options['record_desvars'] or self.recording_options['includes']:
```
swryan: you might do a preliminary test on `self.recording_options['includes']` alone to short-circuit the following tests (and not have to repeat the 'includes' portion of each test)

ryanfarr01: Perhaps I'm thinking about this incorrectly, but doing a test on includes before desvars/constraints/objectives would be equivalent to and-ing them together rather than or-ing them together, right? With that said, it definitely is unnecessarily verbose; I should store self.recording_options['includes'] in a temporary variable at the very least.

ryanfarr01: Oh nevermind, scratch that. I see what you were saying

swryan: GJ on understanding my less than clear comment...  it was just a suggestion, no big deal

openmdao/core/driver.py
```diff
@@ -262,6 +265,15 @@ def _setup_driver(self, problem):
                     for d in all_vars[:-1]:
                         mysystem_outputs.update(d)
 
+            # de-deplicate sets by removing from mysystem_outputs
+            #   if it's already in another set
+            # clone mysystem_outputs so we can remove from the set while iterating
+            #   over its values
+            temp_outputs = mysystem_outputs.copy()
```
naylor-b: I think you could replace this loop with:   mysystem_outputs = mysystem_outputs.difference(mydesvars, myobjectives, myconstraints)

openmdao/core/driver.py
```diff
@@ -262,6 +265,15 @@ def _setup_driver(self, problem):
                     for d in all_vars[:-1]:
                         mysystem_outputs.update(d)
 
+            # de-deplicate sets by removing from mysystem_outputs
```
naylor-b: typo: de-deplicate

openmdao/core/driver.py
ryanfarr01: Nifty! Good catch

openmdao/recorders/recording_manager.py
```diff
@@ -1,12 +1,15 @@
...
 import time
+from six import iteritems
 
 import numpy as np
 
-from openmdao.utils.mpi import MPI
+try:
+    from openmdao.utils.mpi import MPI
+except ImportError:
+    MPI = False
```
naylor-b: Later in the file there is an 'if MPI is None' check, so the value of MPI should be set to None here.  None is the value we assign it elsewhere as well whenever MPI is not active or not installed.

PR #444, opened 2017-11-08
update/fix the 'Listing Variables' doc

swryan:
Pivotal #152279808


openmdao/core/tests/test_impl_comp.py
```diff
@@ -531,10 +531,12 @@ def setUp(self):
         group.connect('comp1.a', 'sub.comp2.a')
         group.connect('comp1.b', 'sub.comp2.b')
         group.connect('comp1.c', 'sub.comp2.c')
+
         group.connect('comp1.a', 'sub.comp3.a')
         group.connect('comp1.b', 'sub.comp3.b')
         group.connect('comp1.c', 'sub.comp3.c')
 
+        global prob
```
Kenneth-T-Moore: What's the reason for this? We definitely don't want these 3 statements in the docs.

openmdao/core/tests/test_impl_comp.py
```diff
@@ -544,93 +546,21 @@ def setUp(self):
         prob.run_model()
 
     def test_list_inputs(self):
-        from openmdao.api import Problem
```
Kenneth-T-Moore: These lines that were removed. Are they automatically being prepended back into the doc for the doctest? Every feature code-block is supposed to show the imports and statements leading up to it, so that it can be pasted standalone into an editor. At least that is what I gathered was the standard motivation for these.

openmdao/core/tests/test_impl_comp.py
swryan: `global` statements do not appear in the docs.. not sure what 3 statements you are referring to

openmdao/core/tests/test_impl_comp.py
swryan: I wrote these docs and the previous file_wrap docs such that successive examples don't have to repeat all the setup code for every example.. in this case since each example is a single line of code that seemed excessive.. I invite the docs czars to have a look at the docs as built and weigh in on whether or not this is more clear and preferred.

Kenneth-T-Moore: Well, I guess this one is up to @JustinSGray .

openmdao/core/tests/test_impl_comp.py
Kenneth-T-Moore: I thought this was in a System setup method for some reason. I withdraw.

openmdao/core/tests/test_impl_comp.py
JustinSGray: Im ok with the way steve did it. The intention is clear and a user should have no trouble understanding what they are supposed to do. 

Shorter docs are better, given equal clarity. I think this is just as clear

PR #450, opened 2017-11-14
In Problem.check_partials, changed options to use individual arguments rather than a dict 

hschilling:
so as to be consistent with the check_totals method

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -456,19 +456,20 @@ class Problem(object):
             next to them in output, making them easy to search for. Note at times there may be a
             significant relative error due to a minor absolute error.  Default is 1.0E-6.
         method : str
-            Method, 'fd' for finite difference or 'cs' for complex step.
+            Method, 'fd' for finite difference or 'cs' for complex step. Default is 'fd'.
         step : float
-            Step size for approximation.
+            Step size for approximation. Default is None.
         form : string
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore:
Maybe we need a little table in the docs that lists the actual values for the FD and CS defaults. I don't know if there is an easy way to auto-fill those from the definitions in the code, or if we will have to remember to modify them if we ever change the default values (though we probably won't change them.)

openmdao/core/problem.py
```diff
@@ -453,10 +455,16 @@ def check_partials(self, logger=None, comps=None, compact_print=False,
             Threshold value for relative error.  Errors about this value will have a '*' displayed
             next to them in output, making them easy to search for. Note at times there may be a
             significant relative error due to a minor absolute error.  Default is 1.0E-6.
-        global_options : dict
-            Dictionary of options that are used as default unless overridden by calling
-            set_check_partial_options on a component. Only 'form', 'step', 'step_calc', and
-            'method' can be specified in this way.
+        method : str
+            Method, 'fd' for finite difference or 'cs' for complex step.
```
swryan: need to show default values in the docstring for `method`, `step`, `force_dense` and `suppress_output`?

hschilling: DONE. Thanks

PR #453, opened 2017-11-16
API Change:  BalanceComp now has a "use_mult" argument

robfalck:
… which is False by default.  This will silence unconnected input warnings in the majority of use cases.  

User can now specify a guess_func when adding a balance.  guess_func is a callable in the form f(inputs, residuals) which returns the initial guess of the implicit state variable.  If None, the default initial guess is not overridden.


robfalck:
Latest changes address those @swryan, thanks.

PR #457, opened 2017-11-27
Moving Regular Grid Interpolator From Pycycle to OpenMDAO

thearn:
Justin, Jeff and others suggested that the `RegularGridInterpComp` explicit component should be moved to OpenMDAO, since we're now writing code that depends on it but doesn't involve other things in Pycycle.

Included are tests and docstrings. Might take a bit to get the formatting just right for the docs.

### DIFF
--- a/openmdao/components/regular_grid_interp_comp.py
+++ b/openmdao/components/regular_grid_interp_comp.py
@@ -164,7 +164,7 @@ class RegularGridInterpolator(object):
             which will use the method defined at
             the construction of the interpolation object instance.
 
-        computer_gradients : bool, optional
+        compute_gradients : bool, optional
             If a spline interpolation method is chosen, this determines
             whether gradient calculations should be made and cached.
             Default is True.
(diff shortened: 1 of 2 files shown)


thearn:
Also, the component is a wrapper of a `RegularGridInterpolator` object that actually does the heavy lifting. It's written to be compatible with `scipy.interpolate.RegularGridInterpolator`, but includes no scipy code, so we're not redistributing anything.

openmdao/components/regular_grid_interp_comp.py
```diff
@@ -0,0 +1,646 @@
...
+        xi : ndarray of shape (..., ndim)
+            The coordinates to sample the gridded data at
+
+        method : str, optional
+            The method of interpolation to perform. Supported are 'slinear',
+            'cubic', and 'quintic'. Default is None,
+            which will use the method defined at
+            the construction of the interpolation object instance.
+
+        computer_gradients : bool, optional
```
Kenneth-T-Moore: Probably compute_gradients rather than computer_gradients

thearn:
still working on the tests. the linter is pretty picky.

PR #458, opened 2017-11-27
several functions now callable from command line

naylor-b:
There's now an 'om' command with a number of sub-commands that can be called on a run script. Several of the commands take advantage of the new post-setup hook in Problem that allows commands that require the Problem to be fully set up to be called from the command line instead of having to edit the run script and import and call the function.  The following functions are now supported:

    - view_model
    - view_connections
    - tree
    - summary
    - dump_idxs

The following were also added to the `om` command at jubs' request to avoid having too many console commands at the top level.
    - iprof  (instance based profiling)
    - iprof_totals
    - mem (memory profiling)
    - trace

All of these functions are still available programmatically if people want to use them that way, but it's generally more convenient to use them from the command line .

Other functions we could make available on the command line might be:

    - check_config  (this one I think will work much better from the command line since the config
                                checking output will no longer be mixed in with solver output or other output
                                from the model as it runs.  It will also allow us to put potentially more expensive
                                checks in there since they'll no longer impact model execution time)
    - check_partials
    - check_totals
    - anything else that can be called immediately after final_setup


openmdao/core/problem.py
```diff
@@ -431,6 +436,11 @@ def final_setup(self):
             self._setup_status = 2
             self._set_initial_conditions()
 
+        # check for post-setup hook
+        if Problem._post_setup_func is not None:
+            Problem._post_setup_func(self)
+            sys.exit(0)
```
swryan: would it be more general for the post_setup function to decide whether it's appropriate to exit or not?

openmdao/devtools/iprof_mem.py
```diff
@@ -39,25 +39,17 @@ def _trace_mem_ret(frame, arg, stack, context):
...
-def setup(methods=None):
-    """
-    Setup memory profiling.
-
-    Parameters
-    ----------
-    methods : list of (glob, (classes...)) or None
-        Methods to be profiled, based on glob patterns and isinstance checks.
-    """
+def _setup(options):
```
swryan: I guess we don't require docstrings on private or dev-specific stuff, but it seems like there is a small audience which might fight them helpful...

openmdao/docs/feature_reference/working_with_models/debugging/om_command.rst
```diff
@@ -0,0 +1,120 @@
...
+command.  There are two types of commands available, those that perform some sort of viewing or
+configuration checking on the Problem after its setup is complete, and those that are used to
+collect information about the entire run of the Problem, things like profilers and tracers.
+
+
+Post-setup Commands
+-------------------
+
+The following commands all run immediately after a Problem's final_setup function.  After they
+complete, the program will exit, rather than continuing to the end of the user's run script.
```
swryan: This seems unclear in that the commands themselves don't run after `setup`...  the commands  are actually the start of the process.. they run the script up to the final_setup and then hijack/intercept execution to run the desired function.  (The function runs after final setup, but the "command" embodies the entire process.)

openmdao/docs/feature_reference/working_with_models/debugging/om_command.rst
```diff
@@ -0,0 +1,120 @@
...
+This makes it convenient to view or check the configuration of a model in any run script without
+having to wait around for the entire script to run.
+
```
swryan: I guess I'll have to be the one that says `om` seems like it might be overly terse.  In previous iterations we had an `openmdao` command which didn't seem all that cumbersome and was less likely to collide with other applications (a quick search finds Open Music and Operations Manager using the `om` abbreviation). The user could easily alias to the shorter version if he/she has a posix bent and no competition for the name.

naylor-b: om was suggested by Jubs and I thought it was ok. I'm also ok with openmdao though.

openmdao/core/problem.py
naylor-b: Agreed.  I had actually just made that change before I saw your comment.

openmdao/core/problem.py
```diff
@@ -70,6 +72,8 @@ class Problem(object):
         2 -- The `final_setup` has been run, everything ready to run.
     """
 
+    _post_setup_func = None
```
Kenneth-T-Moore: Is this going to hurt nested problems if you want different post setup hook functions in each?

openmdao/devtools/iprof_mem.py
Kenneth-T-Moore: I am part of this small audience.

openmdao/core/problem.py
naylor-b: There are always things we can do if we choose to support that sort of use case, but I think it'll be very uncommon and we shouldn't make this more complicated unless users actually request that sort of thing.

Kenneth-T-Moore: Ok, but I should note that I was helping John get a nested problem to work right before Thanksgiving.

JustinSGray: nested problems??? for what application?

hwangjt: I used nested problems to enable a bracketed root finding algorithm with a composite residual function

JustinSGray: like the brent method?

hwangjt: yes

PR #461, opened 2017-12-01
clean new commit of doc automation stuff.

kmarsteller:


kmarsteller:
My pull request is finally working...would like it to go in before the release, so we can test the doc automation naming with the 2.1.0 release.

Other than the doc automation stuff, there are two other small things in here:
1. Removal of a no-longer-relevant note from one doc.
2. Addition of documentation on the embed-options directive.

PR #466, opened 2017-12-08
added check and test for multiply connected inputs

naylor-b:


openmdao/core/group.py
```diff
@@ -639,10 +639,26 @@ def _setup_global_connections(self, recurse=True, conns=None):
...
         # then adding contributions from systems above/below, then allgathering.
+        conn_list = list(iteritems(global_abs_in2out))
+        conn_list.extend(iteritems(abs_in2out))
         global_abs_in2out.update(abs_in2out)
 
         for subsys in self._subsystems_myproc:
             global_abs_in2out.update(subsys._conn_global_abs_in2out)
+            conn_list.extend(iteritems(subsys._conn_global_abs_in2out))
+
+        if len(conn_list) > len(global_abs_in2out):
```
Kenneth-T-Moore: Can there ever be a case where those 2 items are equal, but there are still multiply connected imports?

naylor-b: That would only be possible if conn_list didn't contain every connection from global_abs_in2out, but it does, so I think we're good here.

hwangjt: What if we have a single group in our model with multiple components below it and there is an explicit and an implicit connection involving the same input?

naylor-b: If you're talking about an implicit and explicit connection at the same level, we already prevent that because once a var is promoted it no longer exists for use in explicit connections so they get an "Input 'blah.x' does not exist for connection ..."

hwangjt: Right. What if there is an explicit connection made to the promoted name?

naylor-b: Oops, looks like that case wasn't being caught. Good catch.  It's fixed now.

PR #467, opened 2017-12-08
In Directsolver, raise error when matrix is singular.

Kenneth-T-Moore:


openmdao/solvers/linear/tests/test_direct_solver.py
```diff
@@ -101,6 +102,114 @@ def test_rev_mode_bug(self):
...
+        teg.nonlinear_solver.options['max_sub_solves'] = 1
+        teg.nonlinear_solver.options['atol'] = 1e-4
+
+        prob.setup(check=False)
+        prob.set_solver_print(level=0)
+
+        with self.assertRaises(RuntimeError) as cm:
+            prob.run_model()
+
+        expected_msg = "Singular entry found in 'thrust_equilibrium_group' for state/residual 'thrust'."
```
JustinSGray: is it possible to be more precise with this error? Can we tell them whether its a row or a column thats singular?

Kenneth-T-Moore: Well, numpy doesn't give the info, but we could return the norm of the row and column so that the user can decide what to do.

PR #470, opened 2017-12-14
Simultaneous deriv support

naylor-b:
- also fixed a bug that prevented numpy.int32 and numpy.int64 from being used as indices.
- simul derivs work only in fwd mode currently.


openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -1020,7 +1020,7 @@ def setup(self):
 
                 self.declare_partials('*', '*')
 
-                self.fail_deriv = [2, 4]
+                self.fail_deriv = [3, 5]
```
Kenneth-T-Moore: Why did you change which derivs fail on this test?

openmdao/docs/theory_manual/simul_derivs.rst
```diff
@@ -0,0 +1,74 @@
...
+                # design variable
+                'indeps.x': {
+                    0: ([0, 1, 2, 3, 4], [0, 2, 4, 6, 8]),
+                    1: ([0, 1, 2, 3, 4], [1, 3, 5, 7, 9])
+                }
+            }
+        }
+    )
+
+    prob.driver.set_simul_coloring(color_info)
```
hwangjt: Perhaps it would be good to add a feature doc in addition to this theory doc or turn this into a feature doc, since this is a user-facing method. The function signature can be included in the feature doc then

hwangjt: Are these names promoted or absolute names?

openmdao/drivers/tests/test_pyoptsparse_driver.py
naylor-b: I don't remember why I had to do that, but I just reverted the change and the test passes, so I'll get rid of that change.

openmdao/docs/theory_manual/simul_derivs.rst
naylor-b: I will move this to a feature doc and add the function signature.  The names in the coloring data structure are currently absolute.  I could add code to do the mapping from promoted to absolute if that's preferred.

hwangjt: Great. Yes, I think promoted would be easier for users

openmdao/core/driver.py
```diff
@@ -418,14 +439,15 @@ def set_design_var(self, name, value):
         desvar = self._problem.model._outputs._views_flat[name]
         desvar[indices] = value
 
-        # Scale design variable values
-        scaler = meta['scaler']
-        if scaler is not None:
-            desvar[indices] *= 1.0 / scaler
+        if self._has_scaling:
```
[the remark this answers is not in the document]
naylor-b: yeah

openmdao/core/driver.py
```diff
@@ -741,3 +752,30 @@ def _get_name(self):
...
+    def set_simul_coloring(self, simul_info):
+        """
+        Set the coloring for simultaneous derivatives.
+
+        Parameters
+        ----------
+        simul_info : tuple
+            Information about simultaneous coloring for design vars and responses.
+        """
+        if self.supports['simultaneous_derivatives']:
```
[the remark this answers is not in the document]
naylor-b: This was done as a quick and dirty way to make sure someone doesn't try using simul derivs with a driver that doesn't split up linear and nonlinear constraints like pyoptsparse does. We had talked about adding this to the scipy optimizer at some point.

openmdao/core/system.py
```diff
@@ -2315,7 +2339,8 @@ def add_objective(self, name, ref=None, ref0=None, index=None,
         self.add_response(name, type_='obj', scaler=scaler, adder=adder,
                           ref=ref, ref0=ref0, index=index,
                           parallel_deriv_color=parallel_deriv_color,
-                          vectorize_derivs=vectorize_derivs)
+                          vectorize_derivs=vectorize_derivs,
+                          simul_coloring=simul_coloring, simul_map=simul_map)
```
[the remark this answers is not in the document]
naylor-b: I agree that simul_map isn't a very descriptive name, but I don't think simul_io_map makes things any clearer.

PR #472, opened 2017-12-18
adding a citation helper method so folks know what to cite 

JustinSGray:


openmdao/core/problem.py
```diff
@@ -70,6 +82,9 @@ class Problem(object):
         0 -- Newly initialized problem or newly added model.
         1 -- The `setup` method has been called, but vectors not initialized.
         2 -- The `final_setup` has been run, everything ready to run.
+    cite: str
```
hwangjt: `cite: str` should be `cite : str`; listing should be capitalized; `citataions` should be `citations`; `class` should be followed by a period

openmdao/core/driver.py
```diff
@@ -47,6 +47,9 @@ class Driver(object):
         Dictionary with general pyoptsparse options.
     recording_options : <OptionsDictionary>
         Dictionary with driver recording options.
+    cite: str
```
hwangjt: See below comment

openmdao/core/system.py
```diff
@@ -56,6 +56,10 @@ class System(object):
         Int that holds the number of times this system has iterated
         in a recording run.
     #
+    cite: str
```
hwangjt: See above comment

openmdao/solvers/solver.py
```diff
@@ -120,6 +120,9 @@ class Solver(object):
         object that manages all recorders added to this solver
     _solver_info : <SolverInfo>
         Object to store some formatting for iprint that is shared across all solvers.
+    cite: str
```
hwangjt: See above comment

openmdao/utils/find_cite.py
```diff
@@ -0,0 +1,71 @@
...
+    citations : dict
+        the dictionary to add a citation to, if found
+    """
+    if instance.cite:
+        klass = instance.__class__
+        # return klass, cite
+        citations[klass] = instance.cite
+
+
+def find_citations(prob, out_stream=sys.stdout):
```
hwangjt: What about making this an option on `Problem.setup`? That saves the user the work of importing another function. Instead they can just type `prob.setup(print_citations=True)`

hwangjt:
The PETSc paper is probably another good one to add for when PETScVector is used, but that can be in a separate story

JustinSGray:
@hwangjt updated per your comments. Added a petsc citation as well

openmdao/core/problem.py
```diff
@@ -343,7 +361,7 @@ def cleanup(self):
         self.driver.cleanup()
 
     def setup(self, vector_class=DefaultVector, check=True, logger=None, mode='rev',
-              force_alloc_complex=False):
+              force_alloc_complex=False, print_citations=False):
```
naylor-b: Why are we adding another argument to setup here when you've already made this accessible as an 'openmdao' command?  In the unlikely event that a user wants to do this programmatically, they can just do the necessary import and call find_citations(prob).  If we're worried about user convenience, it seems like the 'openmdao cite' command is the most convenient way to access this functionality.

hwangjt: I asked for this change so that for the non-command line way, the user doesn't have to do the extra import. Not a big deal for me either way though

JustinSGray: using the `openmdao` command is pretty easy, though it doesn't yet work if you have a file with multiple problems. Citations are something people are lazy about, so asking them to remember an import is likely to result in less people doing it.

I guess we could push the `openmdao` method and not offer two ways to do it ... 
but I don't really see a huge advantage to the direct function call vs the setup arg, and the setup arg requires one less step for the user

naylor-b: finding the citations is only one of many post-setup functions that are best called by the user from the command line.  Do we want to add all of these as args to the Problem setup command to handle the uncommon use case where a user wants to call them programmatically?  The issue of the openmdao command not working with scripts having multiple problems could be fixed by allowing problems to have a 'name' attribute so that problems could be specified by name from the command line.

Kenneth-T-Moore: Could just make it a method on problem too, instead of an argument.

JustinSGray:
ok. I took out the setup argument and edited the docs to tell people to use the script

openmdao/docs/other/citing.rst
```diff
@@ -5,27 +5,20 @@ How to Cite OpenMDAO
...
-
+    openmdao.test_suite.test_examples.tldr_paraboloid.TestParaboloidTLDR.test_tldr
+    :no-split:
 
 With the `openmdao` command
 ----------------------------------
 
-If you copy the above script into a file called `paraboloid.py` and remove the call to `find_citations`,
+If you copy the above script into a file called `paraboloid.py`,
 then you can get the citations from the command line using the :ref:`openmdao run script<om-command>`.
```
naylor-b: should be 'openmdao cite' instead of 'openmdao run'

JustinSGray: not cite.. but i've changed to `openmdao command-line script`

PR #474, opened 2017-12-21
Having to go through ridiculous gymnastics to get TravisCI to work again...

kmarsteller:


kmarsteller:
I know there's some commented-out stuff in here. I'm just so excited that it works, I don't want to mess with it, plus, we may need it back.

PR #483, opened 2018-01-08
Lintfix

kmarsteller:
So, various directories of source were not being linted for docstring problems for many many months. This fixes that (for now) and fixes all the lint problems that occurred. 

Remaining work for future PRs (putting in stories): 
1.) Change test_lint_docstrings.py to use an excludes_list type of method to find source code, instead of a needs-to-be-updated includes method.
2.) Figure out the docstring problems causing 36 warnings in the sphinx docbuild. (all in classes derived from System)


openmdao/components/balance_comp.py
```diff
@@ -21,8 +21,8 @@ def __init__(self, name=None, eq_units=None, lhs_name=None,
         r"""
         Initialize a BalanceComp, optionally creating a new implicit state variable.
 
-        The BalanceComp is a bit like IndepVarComp in that it allows for the
-        creation of one or more implicit state variables, and computes the residuals
+        The BalanceComp is a bit like IndepVarComp in that it allows for the \
```
naylor-b: are these end of line '\'s actually necessary?

openmdao/components/meta_model_structured.py
```diff
@@ -85,12 +49,56 @@ def _interp_methods():
...
                  fill_value=np.nan, spline_dim_error=True):
-        """Initialize instance of interpolation class."""
+        """
+        Initialize instance of interpolation class.
+
+        Parameters
+        ----------
+        points : tuple of ndarray of float, with shapes (m1, ), ..., (mn, )
+            The points defining the regular grid in n dimensions.
+
```
naylor-b: we're not consistent on our line spacing for parameters.  In most of our existing code we don't put a blank line between each of our parameter descriptions.  I think we should stick with that since our docstrings are long enough as it is.

openmdao/components/multifi_meta_model.py
```diff
@@ -118,7 +124,10 @@ def add_input(self, name, val=1.0, **kwargs):
             For advanced users only. ID or color for this variable, relevant for
             reconfigurability. Default is 0.
         """
-        metadata = super(MultiFiMetaModelUnStructured, self).add_input(name, val, **kwargs)
+        parent = MultiFiMetaModelUnStructured
```
naylor-b: 'parent' isn't the best name to use here.  MultiFiMetaModelUnStructured is the class of self, and super(..., self) gives us the parent class.

openmdao/components/balance_comp.py
kmarsteller: Oh, no, you're right, that was an attempt to get rid of the sphinx warnings, based on other people's reports on the internet. I'll pull those out, they had zero effect.

openmdao/components/meta_model_structured.py
kmarsteller: Agreed.  Let me see how many of these need fixing.

openmdao/components/multifi_meta_model.py
kmarsteller: wait, what? oh, did Ken maybe change this? what should it say?

Kenneth-T-Moore: yeah, just change it to 'name' or 'item' or something

openmdao/components/meta_model_structured.py
kmarsteller: Looks like a lot of these to fix...will have them in soon.

kmarsteller: OK, fixed like 150 of these spacing inconsistencies.

openmdao/components/balance_comp.py
kmarsteller: Fixed.

openmdao/components/multifi_meta_model.py
kmarsteller: I went with "item." Fixed.

kmarsteller:
aw man, looks like there is a similar problem with linting of attributes.  will make another story for that.

kmarsteller:
Anyway, once this CI passes, this should be ready to go, pending any other review comments.

openmdao/components/meta_model_unstructured.py
```diff
@@ -149,6 +166,11 @@ def _setup_vars(self, recurse=True):
         Return our inputs and outputs dictionaries re-keyed to use absolute variable names.
```
swryan: looks like you removed the return value, making this description incorrect?

Kenneth-T-Moore: yeah, Keith, just copy the description from the parent class.

PR #486, opened 2018-01-11
P153623065 Add debug print option drivers

hschilling:
Add some option to let users print info about variables as a Driver iterates

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -60,8 +60,6 @@ class Driver(object):
         Dictionary with general pyoptsparse options.
     recording_options : <OptionsDictionary>
         Dictionary with driver recording options.
-    recording_options : <OptionsDictionary>
-        Dictionary with driver recording options.
     debug_print : <OptionsDictionary>
         Dictionary with debugging printing options.
     cite : str
@@ -839,7 +837,7 @@ class Driver(object):
 
     def _pre_run_model_debug_print(self):
         """
-        Set up metadata for simultaneous derivative solution.
+        Optionally print some debugging information before the model runs.
         """
         if self.debug_print['debug_print']:
             if not MPI or MPI.COMM_WORLD.rank == 0:
@@ -861,7 +859,7 @@ class Driver(object):
 
     def _post_run_model_debug_print(self):
         """
-        Set up metadata for simultaneous derivative solution.
+        Optionally print some debugging information after the model runs.
         """
         if self.debug_print['debug_print']:
             if self.debug_print['debug_print_nl_con']:
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -602,10 +602,8 @@ class Driver(object):
         boolean
             Failure flag; True if failed to converge, False is successful.
         """
-        with Recording(self._get_name(), self.iter_count, self) as rec:
-            self._pre_run_model_debug_print()
+        with RecordingDebugging(self._get_name(), self.iter_count, self) as rec:
             failure_flag = self._problem.model._solve_nonlinear()
-            self._post_run_model_debug_print()
 
         self.iter_count += 1
         return failure_flag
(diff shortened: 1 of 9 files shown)


openmdao/core/driver.py
```diff
@@ -804,3 +836,63 @@ def _setup_simul_coloring(self, mode='fwd'):
                     del dvdict[dv]
                     dv = prom2abs[dv][0]
                     dvdict[dv] = col_dict
+
+    def _pre_run_model_debug_print(self):
+        """
+        Set up metadata for simultaneous derivative solution.
```
Kenneth-T-Moore: I don't think this docstring (or the other one) is right

Kenneth-T-Moore: Also, it occurs to me that you don't really need to record desvars before the model runs. They don't change. You can put all of this code into the _post_rn_model_debug_print, and get rid of this function.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -434,6 +439,8 @@ def _objfunc(self, dv_dict):
                     model._clear_iprint()
                     fail = 1
 
+                self._post_run_model_debug_print()
```
Kenneth-T-Moore: You may not even need to ever call this in the drivers, if you instead call it in the _exit_ function of the Recording context manager.

hschilling: I was wondering about that but Justin told me "Print the des-vars before it calls the model, print the objective, constraints after the model returns"

hschilling: maybe in case the model dies?

Kenneth-T-Moore: Yeah, if it outright dies, then I can see a need for recording them first.

Kenneth-T-Moore: but I still think the pre function and the post function can be a called at the appropriate points in the context manager functions, so you never have to put them in drivers.

openmdao/core/driver.py
```diff
@@ -35,6 +36,18 @@ class Driver(object):
...
+    debug_print['debug_print_desvars'] :  bool(False)
+        If True and the option debug_print is True, print design variables every Driver iteration.
+    debug_print['debug_print_nl_con'] :  bool(False)
+        If True and the option debug_print is True, print nonlinear constraints every
+        Driver iteration.
+    debug_print['debug_print_ln_con'] :  bool(False)
+        If True and the option debug_print is True, print linear constraints every Driver iteration.
+    debug_print['debug_print_objective'] :  bool(False)
+        If True and the option debug_print is True, print objectives every Driver iteration.
+
```
swryan: seems like some redundancy in the naming here..  suggest a more concise: 
```
    debug['print'] :  bool(False)
    debug['print_desvars'] :  bool(False)
    debug['print_nl_con'] :  bool(False)
    debug['print_ln_con'] :  bool(False)
    debug['print_objective'] :  bool(False)
```

Kenneth-T-Moore: Also, kind of weird that you have to turn on two switches to get any debug printing, though maybe that's what Jubs wanted.  You can automatically determine the value for 'debug_print' as a logical 'or' on the other ones.

naylor-b: why not just have one option (self.options['debug_print'])  that defaults to an empty list, and you can set it to ['desvars', 'nl_con', ..]?  Seems like less work than setting all of those individual bools.

JustinSGray: agreed that we shouldn't have two swtiches... I was thinking that wed just turn on 'debug_print` and it would split out everything. 

I likes brets suggestion, but I don't think our options_dict currently allows that kind of multi_enum input does it?

openmdao/drivers/pyoptsparse_driver.py
hschilling: good idea about the context manager. I didn't know if we wanted to hide it, especially in something named "Recording". I suppose we could rename it to something else. "Debugging" ? "RecordingDebugging" ? Or am I worrying over nothing?

openmdao/core/driver.py
hschilling: Sounds good to me with having a list. So just a list of strings?

JustinSGray: a list of strings, but you have to enforce which ones are allowed

naylor-b: yeah a list of strings

hschilling: OK. Looks like I can use the is_valid option to do the enforcing so that works

hschilling: valid strings ['desvars','nl_cons','ln_cons','objs']

Look good?

openmdao/core/driver.py
```diff
@@ -804,3 +854,113 @@ def _setup_simul_coloring(self, mode='fwd'):
...
+class RecordingDebugging(Recording):
+    """
+    A class that acts as a context manager.
+
+    Handles doing the case recording and also the Driver
+    debugging printing.
+    """
+
+    def __init__(self, name, iter_count, recording_requester):
```
naylor-b: \_\_init\_\_ does nothing except call the base class, so why is it defined here?

hschilling: fixed

PR #494, opened 2018-01-23
a couple of speedups to help with x57

naylor-b:
- got rid of a lot of vector cloning
- various small things to speed up DictionaryJacobian._apply
- a couple of simul coloring fixes

### DIFF
--- a/openmdao/jacobians/dictionary_jacobian.py
+++ b/openmdao/jacobians/dictionary_jacobian.py
@@ -85,17 +85,24 @@ class DictionaryJacobian(Jacobian):
 
         fwd = mode == 'fwd'
         system = self._system
-        with system._unscaled_context(
-                outputs=[d_outputs], residuals=[d_residuals]):
+        d_res_names = d_residuals._names
+        d_out_names = d_outputs._names
+        d_inp_names = d_inputs._names
+        res_flat_views = d_residuals._views_flat
+        out_flat_views = d_outputs._views_flat
+        inp_flat_views = d_inputs._views_flat
+        np_add_at = np.add.at
+
+        with system._unscaled_context(outputs=[d_outputs], residuals=[d_residuals]):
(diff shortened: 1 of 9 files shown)


openmdao/jacobians/dictionary_jacobian.py
```diff
@@ -80,64 +80,79 @@ def _apply(self, d_inputs, d_outputs, d_residuals, mode):
...
+        system = self._system
+        d_res_names = d_residuals._names
+        d_out_names = d_outputs._names
+        d_inp_names = d_inputs._names
+        res_flat_views = d_residuals._views_flat
+        out_flat_views = d_outputs._views_flat
+        inp_flat_views = d_inputs._views_flat
+        np_add_at = np.add.at
+
+        with system._unscaled_context(outputs=[d_outputs], residuals=[d_residuals]):
```
Kenneth-T-Moore: You added an unscaling here. Is this needed? How did this code work before if we were supposed to unscale here but did not?

naylor-b: I didn't add an unscaling.  If you look above, it used to be two lines and I made it a single line.

Kenneth-T-Moore: That is a relief.

openmdao/jacobians/dictionary_jacobian.py
```diff
@@ -80,64 +80,79 @@ def _apply(self, d_inputs, d_outputs, d_residuals, mode):
         mode : str
             'fwd' or 'rev'.
         """
+        # avoid circular import
+        from openmdao.core.explicitcomponent import ExplicitComponent
```
hwangjt: I'm not sure how expensive this is, but in case it's not negligible, what about importing ExplicitComponent once in __init__ and saving it as an attribute?

PR #499, opened 2018-01-26
Lintfix

kmarsteller:
This PR includes:
1. Changing lint_docstrings and lint_attributes to use an exclude list rather than an includes list that can (and did) go out-of-date without complaint. 
2. Changing those same two tests to recurse into subdirs and lint those, e.g.  openmdao/solvers/linear or openmdao/surrogate_models/nn_interpolators.
3. Adding in a bunch of docstrings from many of the forgotten directories, both to fix missing documentation for attributes, and for some parameters that were also missing. Contributions by @Kenneth-T-Moore , @hschilling, @dwmunster, and me.
4. Currently on the excludes list is "components," which we are still waiting on some docstrings for, that we will add in next week, and then remove "components" from the exclude list.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,9 +81,8 @@ install:
 - pip install --upgrade pip
 
 # install pyoptsparse
-- wget https://bitbucket.org/mdolab/pyoptsparse/get/tip.zip;
-- unzip tip.zip;
-- cd mdolab-pyoptsparse*;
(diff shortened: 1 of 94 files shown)


kmarsteller:
Looks like there is a problem on Travis in the way I construct package namess that works locally and with Appveyor, but not on Travis for some reason. Investigating.

PR #503, opened 2018-01-30
coloring fix

naylor-b:
- also some performance tweaks for x57
- added dependency for networkx>=2.0


openmdao/core/driver.py
```diff
@@ -837,6 +838,10 @@ def _setup_simul_coloring(self, mode='fwd'):
             raise NotImplementedError("Simultaneous derivatives are currently not supported "
                                       "in 'rev' mode")
 
+        # command line simul_coloring uses this env var to turn pre-existing coloring off
+        if os.environ.get("SIMUL_NO_COLOR"):
```
Kenneth-T-Moore: Do we document this env var anywhere?

naylor-b: it's not for users.  I just use it internally when the command line coloring command is running.

naylor-b: OK I came up with a simpler way to do this that doesn't use an environment variable.  Now it just uses a module level variable in coloring.py.

PR #504, opened 2018-02-02
P153784367 refactor logger embed test

hschilling:
Convert the use of loggers to regular prints everywhere in the code except for check_config.


openmdao/core/system.py
```diff
@@ -1,5 +1,5 @@
 """Define the base System class."""
-from __future__ import division
+from __future__ import division, print_function
```
kmarsteller: Guess we no longer need this print_function import if we aren't using it?

PR #506, opened 2018-02-05
some performance tweaks

naylor-b:
- got rid of unnecessary looping and unnecessary creation of metadata dicts for deriv key pairs when dependent==False.
- got rid of unnecessary dict layer for _var_allprocs_abs2meta and _var_abs2meta.

This results in an approx. 10x improvement (40 sec down to 4 sec) in the setup time for the CFM56 model.

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -728,16 +728,13 @@ class Component(System):
             Value of subjacobian.  If rows and cols are not None, this will
             contain the values found at each (row, col) location in the subjac.
         """
-        if not dependent:
-            return
-
-        if val is not None and not issparse(val):
+        if dependent and val is not None and not issparse(val):
             val = np.atleast_1d(val)
             # np.promote_types  will choose the smallest dtype that can contain both arguments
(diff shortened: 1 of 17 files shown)


openmdao/core/component.py
```diff
@@ -726,6 +728,9 @@ def _declare_partials(self, of, wrt, dependent=True, rows=None, cols=None, val=N
             Value of subjacobian.  If rows and cols are not None, this will
             contain the values found at each (row, col) location in the subjac.
         """
+        if not dependent:
```
Kenneth-T-Moore: Does this mean we can't do:

    self.declare_partials(of=['*'], wrt=['*'])
    self.declare_partials(of=['x'], wrt=['y'], dependent = False)

Or were we never able to do that even before. Plus I guess we have pursued the philosophy of avoiding '*' and always being explicit by listing all names, even though nobody does this in practce.

naylor-b: OK I added a fix to make sure the scenario you mentioned still works.

PR #507, opened 2018-02-05
User can call `keys()` on any vector object to get a list of the variables held in it

swryan:
Pivotal #152300036

### DIFF
--- a/openmdao/vectors/vector.py
+++ b/openmdao/vectors/vector.py
@@ -2,7 +2,7 @@
 from __future__ import division, print_function
 import numpy as np
 
-from six import iteritems
+from six import iteritems, PY3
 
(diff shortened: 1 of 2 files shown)


openmdao/vectors/vector.py
```diff
@@ -292,6 +292,17 @@ def _contains_abs(self, abs_name):
...
+    def keys(self):
+        """
+        Return absolute variable names of variables contained in this vector.
+
+        Returns
+        -------
+        list
+            absolute variable names of variables contained in this vector.
+        """
+        return self._views.keys()
```
naylor-b: keys() should use self._names instead of self._views because the active set of what is contained in the vector changes depending on whether the vector is inside of a matvec_context or not, and that active set is specified by self._names.

PR #509, opened 2018-02-06
Fix for missing stdout from check_partials and check_totals

Kenneth-T-Moore:


openmdao/core/problem.py
```diff
@@ -451,7 +455,7 @@ def final_setup(self):
         if Problem._post_setup_func is not None:
             Problem._post_setup_func(self)
 
-    def check_partials(self, out_stream=None, comps=None, compact_print=False,
+    def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, comps=None, compact_print=False,
```
swryan: looks like `out_stream` and `suppress_output` are redundant?

Kenneth-T-Moore: We could change the API and teach users to specify None for outstream instead of using the suppress_output flag. My recommendation is to reject Herb's story and kick this back to him to finish. This is an emergency fix for me.

PR #513, opened 2018-02-07
Doc review

kmarsteller:
Hitting up the features section this time around.
Also, renaming it from feature_manual and/or feature_reference to just "features"--so you see entirely deleted files in this diff, but they're just having their paths renamed.


openmdao/docs/_utils/generate_sourcedocs.py
```diff
@@ -73,9 +73,11 @@ def generate_docs():
         'recorders',
         'solvers',
         'surrogate_models',
+        'surrogate_models.nn_interpolators',
```
Kenneth-T-Moore: Surrogate entries not alphabetized. Supposes it really doesn't matter, but it looks like you keeping this in order.

kmarsteller: good point. easy fix. will do.

kmarsteller: (Done)

PR #516, opened 2018-02-08
P154975979 fix recording test doc

hschilling:
Convert example code on case recording doc to use embed-test and fix it up so that it is suitable for a doc page


hschilling:
Here is a screenshot of how the test looks in the doc now. 
screen shot 2018-02-08 at 5 26 49 pm

PR #518, opened 2018-02-12
fix to allow CSCJacobian to be used above DenseJacobians

naylor-b:
also...
- misc performance tweaks
- updated the Wing project template to version 6
- updated the tree function to allow display of jacobian types
- added a fix for some cases where multiple inputs from the same component connect to the same source
- added a check setup command to check for duplicate inputs mentioned above


openmdao/jacobians/tests/test_jacobian.py
```diff
@@ -778,5 +756,56 @@ def compute_partials(self, inputs, partials):
...
+        prob.model.add_design_var('indeps.x')
+
+        prob.model.connect('indeps.x', 'G1.C1.x')
+        prob.model.connect('indeps.x', 'G1.C1.y')
+
+        prob.setup(check=False)
+        prob.run_model()
+
+        # before ths fix, this would raise an exception due to an indexing error if CSCJacobian
+        # was used.
```
swryan: there's a typo in the comment, but I wonder if referencing the "this fix" is useful outside the context of this PR. 
maybe the comment should just explain the config that is verified to be valid by this test.

openmdao/solvers/linear/direct.py
```diff
@@ -65,6 +65,45 @@ def format_singluar_error(err, system, mtx):
...
 
 
+def format_singluar_csc_error(system, matrix):
+    """
+    Format a coherent error message when the CSC matrix is singular.
+
+    Parameters
+    ----------
+    system : <System>
+        OpenMDAO system containing the Directsolver.
```
swryan: is it necessary to qualify this as an `OpenMDAO` system?  seems to be clear by context.. 
(don't want to have to add `OpenMDAO` to all our docstrings...)

naylor-b: fixed

openmdao/jacobians/tests/test_jacobian.py
naylor-b: I reworked this test.

PR #520, opened 2018-02-13
Cleanup of armijo goldstein docs plus speedup of doc build.

Kenneth-T-Moore:
1. Cleaned up the Armijo-Goldstein feature doc, which was pointing to the wrong tests, and fixed a couple of other problems there.
2. Sped up doc build significantly by refactoring some code to use string parsing instead of Red Baron, and to call the slower RB functions more sparingly.
3. Removed some extraneous output from the scipyoptimize feature tests.
4. Cleaned up a little wacky formatting in feature tests, now that we have to be a bit more careful with code structure because of string parsing.


openmdao/docs/_utils/docutil.py
```diff
@@ -383,52 +399,63 @@ def insert_output_start_stop_indicators(src):
...
-                r = r.parent
+        for item in print_producing:
+            if item in line:
+                indent = '' * (len(line) - len(line.lstrip()))
 
-            r.insert_after('print(">>>>>%d")\n' % input_block_number)
-            input_block_number += 1
+                # Line continuations are a litle tricky.
+                line = line.rstrip()
+                if line.endswith(',') or line.endswith('\\') or line.endswith('('):
```
naylor-b: This won't catch a number of possible continuations.  You'll probably need to keep track of the number of open and closed parens, square brackets, and curly braces as well to make this more robust.

Kenneth-T-Moore: We don't need to put any of that stuff inside of function calls. Let's write nice readable tests for the poor user.

openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -1483,9 +1483,13 @@ def test_debug_print_option(self):
...
 
+
+@unittest.skipIf(OPT is None or OPTIMIZER is None, "only run if pyoptsparse is installed.")
 class TestPyoptSparseFeature(unittest.TestCase):
 
     def setUp(self):
+        from openmdao.utils.general_utils import set_pyoptsparse_opt
+        OPT, OPTIMIZER = set_pyoptsparse_opt('SNOPT')
         if OPT is None:
             raise unittest.SkipTest("pyoptsparse is not installed")
```
kmarsteller: I would probably change the skip message to say that SNOPT is not available as an optimizer here?

Kenneth-T-Moore: This part of the skip check just looks for pyoptsparse. The second part is what looks for SNOPT.  But, I did take the SLSQP mention out.

kmarsteller: OH, I guess I was confused by the skiptest 5 lines up that I thought was only running this test if pyoptsparse was installed.  Then i thought setting the OPT to SNOPT was checking to see further if SNOPT were installed?

PR #526, opened 2018-02-28
P152080094 in check_partials display, only show reverse dev info for matrix-free comps

hschilling:


openmdao/core/problem.py
```diff
@@ -455,6 +455,20 @@ def final_setup(self):
         if Problem._post_setup_func is not None:
             Problem._post_setup_func(self)
 
+    def _all_components_provide_jacobians(self):
+        """
```
Kenneth-T-Moore: So this method is only called in one place. I wonder if we really need an api method given that.

PR #528, opened 2018-03-02
some tracing improvements

naylor-b:
- fixed a bug in programmatic use of instance based profiling
- added PROC_FILES_DIR env var checking to allow user to specify where USE_PROC_FILES puts MPI stdout files
- added memory tracking to the tracer
- added object tracking (display of objects created in a function and not garbage collected) to the tracer
- created a simple plotting script to take tracer memory tracking output and generate a plot of total memory use vs. time with calls to a specified function overlaid on top of it
- got rid of a couple of circular references


openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -389,6 +389,9 @@ def run(self):
             # optimizers other than pySNOPT may not populate this dict
             pass
 
+        self.opt_prob = None
```
Kenneth-T-Moore: It is possible we don't even need to save either of these in the instance in the first place.

naylor-b: Nothing seemed to be referencing them, so I went ahead and removed them.  Didn't break anything as far as I can tell.

PR #534, opened 2018-03-13
P153285942 check_partials should make it easier to find incorrect subjacs

hschilling:
Practical models can be very large and contain tens, sometimes hundreds of components. Currently, the user has to manually scroll through a lot of output when trying to ensure all the derivatives are correct when running at a new design point.

modify the default behavior to add a * for any row that violates a user specified tolerance. 

It would be helpful to add 2 features to check_partials.

The first is printing the worst subjac at the bottom of the output. Worst is defined by looking at the fwd and rev columns of the relative error (i.e., the 2nd and 3rd last columns) of the compact_print=True output. We should print the component name, then repeat the full row for the worst-case subjac (i.e., output-input pair). This should only occur in the compact_print=True case.

The second is adding an option to show only the incorrect subjacs (according to abs_err_tol and rel_err_tol), called show_only_incorrect. This should be False by default, but when True, it should print only the subjacs found to be incorrect. This applies to both compact_print=True and False.


openmdao/docs/advanced_guide/derivs/partial_derivs_explicit.rst
```diff
@@ -63,7 +63,7 @@ How Do I Know If My Derivatives Are Correct?
 ********************************************
 
 It is really important, if you are going to provide analytic derivatives, that you make sure they are correct.
-Nothing else will screw up the convergence of your analysis or optimization more quickly than bad derivatives.
+It is hard to understate the importance of accurate derivatives in the convergence of analysis and optimization problems.
```
Kenneth-T-Moore: This is the one that should be overstate.

openmdao/core/problem.py
```diff
@@ -1890,40 +1936,65 @@ def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out
...
+                        out_buffer.write('\n')
 
                     if out_stream:
-                        out_stream.write(' -' * 30 + '\n')
+                        out_buffer.write(' -' * 30 + '\n')
+
+                # End of if compact print if/else
+            # End of if not suppress_output
+        # End of for of, wrt in sorted_keys
+    # End of for system in system_list
```
Kenneth-T-Moore: Loop end comment for "system" should be below the next block of code.

openmdao/core/problem.py
```diff
@@ -1890,40 +1936,65 @@ def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out
...
+                "Sub Jacobian with Largest Relative Error: {1} '{2}'".format(*worst_subjac)
+            out_stream.write('\n' + '#' * len(worst_subjac_header) + '\n')
+            out_stream.write("{}\n".format(worst_subjac_header))
+            out_stream.write('#' * len(worst_subjac_header) + '\n')
+            out_stream.write(header + '\n')
+            out_stream.write('-' * len(header) + '\n')
+            out_stream.write(worst_subjac_line + '\n')
 
 
 def _format_if_not_matrix_free(matrix_free, val):
```
Kenneth-T-Moore: I don't think this `quotes` argument is ever used.

hschilling: It is here, right? https://github.com/hschilling/OpenMDAO/blob/6d40866d9bed2fd1c2374cc9c6d7444342d6fe74/openmdao/core/problem.py#L1748

hschilling: as a positional argument.

Kenneth-T-Moore: Oh, I thought pep (or good practice) was that you should always use keyword arguments with the keyword specified, rather than as a positional.

hschilling: probably ! Will fix

openmdao/core/problem.py
```diff
@@ -1890,40 +1936,66 @@ def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out
...
-                        out_stream.write('    Raw FD Derivative (Jfd)\n')
-                        out_stream.write(str(fd) + '\n')
-                        out_stream.write('\n')
+                        out_buffer.write('    Raw FD Derivative (Jfd)\n')
+                        out_buffer.write(str(fd) + '\n')
+                        out_buffer.write('\n')
 
                     if out_stream:
-                        out_stream.write(' -' * 30 + '\n')
+                        out_buffer.write(' -' * 30 + '\n')
```
swryan: could some of these `if out_stream:` blocks be consolidated?  for example, it looks like lines 1956 through 1977 could be handled with a single such conditional rather than five...

openmdao/core/tests/test_check_derivs.py
```diff
@@ -63,28 +63,63 @@ def compute_partials(self, inputs, partials):
...
+class MyCompGoodPartials(ExplicitComponent):
+    def setup(self):
+        self.add_input('x1', 3.0)
+        self.add_input('x2', 5.0)
+        self.add_output('y', 5.5)
+        self.declare_partials(of='*', wrt='*')
 
-class TestProblemCheckPartials(unittest.TestCase):
+    def compute(self, inputs, outputs):
+        """ Doesn't do much. """
```
swryan: I would vote to banish all occurrences of this docstring ;) 
maybe replace it with something like "Calculates y = 3*x1 + 4*x2" in this case...

openmdao/core/problem.py
hschilling: They sure could! There's a story to re-factor these methods. Included in that is getting rid of things like the suppress_output option and things like you are talking about. Really need to reduce the if's and reduce the depths of the blocks.

openmdao/core/tests/test_check_derivs.py
hschilling: I agree that docstring looks bad. The only issue with your proposed replacement is that it is totally obvious from the code what it does so that doesn't look too great either. 

Could we just use the docstring we use for the compute method from ExplicitComponent? 

"Compute outputs given inputs".

swryan: I think the docstring being a more readable, non-coded version of the equation is useful.  Just IMO.

Kenneth-T-Moore: If the compute docstring says anything, it should say something like "Values for compute are correct" to contrast with the derivatives, which are intentionally declared with incorrect values to trigger the desired response in the checks. We don't care that this component provides calculation for the equation `y = 3*x1 + 4*x2`; it is irrelevant.

Further, if we are going to dig into intra-test object docstrings this much, then we should turn on lint checks for the tests and enforce it all rigidly, and put up a story to clean up the tests.

hschilling: I guess I need Justin or John to give their thoughts on this debate.

PR #538, opened 2018-03-16
Support restarting solution from cache for the relevant linear solvers.

Kenneth-T-Moore:
KSP is the only one it wasn't already working for.

### DIFF
--- a/openmdao/solvers/linear/tests/test_petsc_ksp.py
+++ b/openmdao/solvers/linear/tests/test_petsc_ksp.py
@@ -8,7 +8,7 @@ import warnings
 import numpy as np
 
 from openmdao.api import Problem, Group, IndepVarComp, PETScKrylov, LinearBlockGS, DirectSolver, \
-     ExecComp, NewtonSolver, NonlinearBlockGS, PetscKSP
+     ExecComp, NewtonSolver, NonlinearBlockGS, PetscKSP, ImplicitComponent
 from openmdao.test_suite.components.expl_comp_simple import TestExplCompSimpleDense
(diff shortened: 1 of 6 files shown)


openmdao/solvers/linear/tests/test_petsc_ksp.py
```diff
@@ -333,6 +333,84 @@ def test_solve_on_subsystem(self):
         assert_rel_error(self, output[1], g1.expected_solution[0], 3e-15)
         assert_rel_error(self, output[5], g1.expected_solution[1], 3e-15)
 
+    def test_linear_solution_cache(self):
+        # Test derivatives across a converged Sellar model.
+
+        class Comp(ImplicitComponent):
```
JustinSGray: why is this class definition copied in two tests? Pull it out and define it in one place?

Kenneth-T-Moore: How strongly do you feel about the need for a defective component that is used twice to show up in our library of testing components? That library is already packed with stuff, and not very well sorted or documented, to the extent that I had to create a new component for this test because I couldn't find what I needed, though an adequate component may have been there. If it is important, than we need a story to go through our test and reorganize things, getting rid of all duplicate testing components, or generalizing and consolidating them into a well-documented taxonomy of components/systems/problems.

Kenneth-T-Moore:
The only solver that needed a change was the Petsc one, which has a flag that you set to True if the initial condition is non-zero.

openmdao/solvers/linear/tests/test_petsc_ksp.py
Kenneth-T-Moore: but anyway, done, and new code pushed up for re-review.

PR #539, opened 2018-03-19
user can set a solver option to control what happens when a case fails to converge

swryan:
Pivotal #155441150


openmdao/solvers/linear/scipy_iter_solver.py
```diff
@@ -141,8 +141,6 @@ def _mat_vec(self, in_vec):
         scope_out, scope_in = system._get_scope()
         system._apply_linear([vec_name], self._rel_systems, self._mode, scope_out, scope_in)
 
-        # print('in', in_vec)
```
Kenneth-T-Moore: Please stop removing my debug statements here. I use them all the time, including just last week. @channel The debug prints in this story don't show near the granularity that these prints do.

openmdao/solvers/solver.py
```diff
@@ -522,8 +527,64 @@ def record_iteration(self, **kwargs):
...
+    _iter_count : int
+        Number of iterations for the current invocation of the solver.
+    _rec_mgr : <RecordingManager>
+        object that manages all recorders added to this solver
+    _solver_info : <SolverInfo>
+        Object to store some formatting for iprint that is shared across all solvers.
+    _err_cache : dict
+        Dictionary holding input and output vectors at start of iteration, if requested.
+    cite : str
+        Listing of relevant citataions that should be referenced when
```
Kenneth-T-Moore: sp: citations

openmdao/solvers/solver.py
```diff
@@ -537,7 +598,27 @@ def solve(self):
...
+            out_str = "\n# Inputs and outputs at start of iteration '%s':\n" % coord
+            for vec_type, vec in iteritems(self._err_cache):
+                out_str += '\n'
+                out_str += '# %s %ss\n' % (vec._name, vec._typ)
+                for abs_name in sorted(vec._views.keys()):
+                    out_str += '%s = %s\n' % (abs_name, repr(vec._views[abs_name]))
+
+            print(out_str)
+
+            filename = coord.replace('._solve_nonlinear', '') + '.dat'
```
Kenneth-T-Moore: Doesn't the iteration coord have weird symbols like pipes and colons? Not sure if those work on windows filenames.

swryan: yes, it has non-alphanumeric characters.. does not seem to be a problem in testing.  I wanted to keep as close to possible to the iter name, but certainly willing to take a better suggestion.

openmdao/docs/advanced_guide/implicit_comps/defining_icomps.rst
```diff
@@ -107,7 +108,7 @@ The Newton solver's use of that Jacobian information is why we need to declare a
 
 .. embed-code::
     openmdao.test_suite.test_examples.test_circuit_analysis.TestCircuit.test_circuit_plain_newton
-    :layout: code, output
+    :layout: interleave
```
JustinSGray: why did you switch these to interleave?

swryan: to make it easier for the reader to distinguish the solver output and identify which values were which (it was a large block of numbers), and it didn't effect the contiguousness of the code itself, just the print statements.

openmdao/solvers/solver.py
Kenneth-T-Moore: Only suggestion would be to string replace the nonalphanum chars, but if it already works in windows, then don't worry about it.

swryan: you're right.. that would be safer and thus better. done.

.travis.yml
```diff
@@ -128,7 +128,7 @@ deploy:
   skip_cleanup: true
   script:
   # only deploy docs in a build after a PR or merge is accepted
-  - if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  - if [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ "$TRAVIS_REPO_SLUG" = "OpenMDAO/OpenMDAO" ]; then
```
kmarsteller: Adding this to make sure that those of us who do verification via Travis on our own repos @ master,( e.g. kmarsteller/OpenMDAO) don't inadvertently push up our docs versions at pull request time.  Approve.

.travis.yml
```diff
@@ -16,11 +16,11 @@ env:
...
   matrix:
-    - PY=2.7 MPI=1 NUMPY=1.12
-    - PY=3.6 MPI=1 NUMPY=1.12
-    - PY=2.7 MPI= NUMPY=1.12
-    - PY=3.6 MPI= NUMPY=1.12
-    - PY=3.6 MPI=1 NUMPY=1.14
+    - PY=2.7 NUMPY=1.12
+    - PY=2.7 NUMPY=1.12 MPI=1
+    - PY=3.6 NUMPY=1.12
+    - PY=3.6 NUMPY=1.12 MPI=1 UPLOAD_DOCS=1
```
kmarsteller: Realized that we now have two PY 3.6, MPI machines, so the way things were set up, we were pushing the docs twice.  To differentiate, Steve added this new env var.  👍

PR #541, opened 2018-03-21
P155914185 Error msg from asser_check_partials needs to be more clear

hschilling:
We currently get the following: 
```Traceback (most recent call last):
  File "test_usatm1976.py", line 27, in test_derivs
    assert_check_partials(data, atol=1e-6, rtol=1e-6)
  File "/Users/justingray/work/blue/openmdao/utils/assert_utils.py", line 49, in assert_check_partials
    verbose=True)
  File "/Users/justingray/anaconda3/lib/python3.6/site-packages/numpy/testing/utils.py", line 1411, in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)
  File "/Users/justingray/anaconda3/lib/python3.6/site-packages/numpy/testing/utils.py", line 796, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-06
rel error error in partial of Ts wrt alt in component std_1976
(mismatch 66.66666666666666%)
 x: array([  7.483854e-06,   7.483854e-06,   0.000000e+00])
 y: array([ 0.,  0.,  0.])```

I believe thats supposed to be fwd-fd, rev-fd, fd-rev? The message needs to make it more clear what the problem is. Not just spit out 6 numbers

### DIFF
--- a/openmdao/utils/tests/test_assert_utils.py
+++ b/openmdao/utils/tests/test_assert_utils.py
@@ -82,11 +82,20 @@ class TestAssertUtils(unittest.TestCase):
         rtol = 1.e-6
         try:
             assert_check_partials(data, atol, rtol)
-        except AssertionError as err:
+        except ValueError as err:
+            err_string = str(err)
+            self.assertEqual(err_string.count('Assert Check Partials failed for the following Components'), 1)
+            self.assertEqual(err_string.count('1e-06'), 2)
+            self.assertEqual(err_string.count('Component:'), 1)
+            self.assertEqual(err_string.count('< output > wrt < variable >'), 1)
+            self.assertEqual(err_string.count('norm'), 2)
+            self.assertEqual(err_string.count('y wrt x1'), 4)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: openmdao/utils/tests/test_assert_utils.py [hunk 1]
```diff
@@ -82,14 +82,59 @@ def compute_partials(self, inputs, partials):
...
+        prob.run_model()
+
+        data = prob.check_partials(suppress_output=True)
+
+        atol = 1.e-6
+        rtol = 1.e-6
+        try:
+            assert_check_partials(data, atol, rtol)
+        except ValueError as err:
+            print(str(err))
```
swryan: should this just raise the err rather than print it?

hschilling: but then the test fails and the docs don't build unless I am missing a trick to get around that

swryan: oh sorry, I missed that this was the feature  test

PR #561, opened 2018-04-05
Potential fix for formatting issues with numpy >- 1.14

hschilling:


openmdao/utils/general_utils.py
```diff
@@ -410,3 +411,28 @@ def run_model(prob):
...
+
+@contextlib.contextmanager
+def printoptions(*args, **kwargs):
+    """Context manager for setting print options.
+    Set print options for the scope of the `with` block, and restore the old
+    options at the end. See `set_printoptions` for the full description of
+    available options.
+    Examples
+    --------
+    >>> with np.printoptions(precision=2):
```
swryan: no `np.` for the context...

swryan: should we put the version check in here?  and automatically conform to 1.13 formatting?

kmarsteller: If you do make a change for this...can you add the context manager to the test I originally did this for?

naylor-b: I'd rather not bury any hidden formatting stuff in there since this is a general function and will probably eventually go away and be replaced with the verison inside of numpy (which is identical to what we have)

kmarsteller:
Closing this because I took what herb had and expanded upon it while he was in meetings all day.

PR #566, opened 2018-04-09
Fixed name checking in add_input and add_output.

Kenneth-T-Moore:
1. Added a regex check for adding input and output names to a component.
2. Added tests for all missing coverage in add_input and add_output.

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -2,12 +2,14 @@
 
 from __future__ import division
 
-import numpy as np
-from itertools import product
-from six import string_types, iteritems, itervalues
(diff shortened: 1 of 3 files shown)


openmdao/core/component.py
```diff
@@ -25,6 +27,10 @@
                       'exact': None}
 
 
+# regex to check for valid names.
+namecheck_rgx = re.compile('[_a-zA-Z0-9\:]+')
```
naylor-b: This will allow variable names that are just a simple number or just a colon.  I think we should stick to regular pythonic names plus a possible colon.

PR #592, opened 2018-04-24
DOE Driver

swryan:
Pivotal #151705118


openmdao/drivers/doe_driver.py
```diff
@@ -0,0 +1,262 @@
...
+        ----------
+        design_vars : dict
+            Dictionary of design variables for which to generate values.
+
+        Yields
+        ------
+        list
+            list of name, value tuples for the design variables.
+        """
+        return
```
naylor-b: return followed by yield

openmdao/core/driver.py
```diff
@@ -708,66 +734,60 @@ def record_iteration(self):
...
+        if opts['record_constraints']:
+            con_vars = self.get_constraint_values()
         else:
-            objectives = {}
+            con_vars = {}
 
-        if self.recording_options['record_constraints']:
-            constraints = self.get_constraint_values()
+        if opts['record_responses']:
+            # res_vars = self.get_response_values()  # not really working yet
```
Kenneth-T-Moore: I thought that, if anything was ever going to use this "response" stuff, it was the DOE driver. Maybe it is time to strip that out of the API, particularly the recording part, which is already complicated. Doesn't affect this PR though.

openmdao/docs/features/building_blocks/drivers/index.rst
```diff
@@ -9,3 +9,4 @@ Drivers
     scipy_optimize_driver.rst
     pyoptsparse_driver.rst
     genetic_algorithm.rst
+    doe_driver.rst
```
Kenneth-T-Moore: doe_driver.rst appears to be missing from the PR

openmdao/drivers/doe_generators.py
```diff
@@ -0,0 +1,412 @@
...
+        ----------
+        design_vars : dict
+            Dictionary of design variables for which to generate values.
+
+        Yields
+        ------
+        list
+            list of name, value tuples for the design variables.
+        """
+        names = design_vars.keys()
```
naylor-b: names isn't used

openmdao/drivers/doe_generators.py
```diff
@@ -0,0 +1,412 @@
...
+        values = np.zeros((size, self._samples))
+
+        # yield desvar values for doe samples
+        for row in doe:
+            retval = []
+            col = 0
+            var = 0
+            for name, meta in iteritems(design_vars):
+                size = meta['size']
+                val = np.zeros(size)
```
naylor-b: could use np.empty here instead of np.zeros since you set every entry of it later

openmdao/docs/features/building_blocks/drivers/index.rst
swryan: aw crap

openmdao/drivers/doe_driver.py
swryan: This is intentional.. I found this to be the cleanest way to implement an empty iterator that works fro both py2 and py3.  Open to better suggestions.

openmdao/drivers/doe_generators.py
swryan: removed

openmdao/drivers/doe_generators.py
swryan: done

openmdao/drivers/doe_driver.py
naylor-b: could you just return an empty tuple for example, or, if it has to be an iterator, return iter(())?

PR #598, opened 2018-04-26
Component name consistency update

kmarsteller:
This should go in after Ken's Bsplines PR, but I have incorporated his PR in this one, since he was working directly on a few of these files. Conflicts are resolved.


openmdao/api.py
```diff
@@ -11,17 +11,25 @@
 
 # Components
 from openmdao.components.balance_comp import BalanceComp
+from openmdao.components.bsplines_comp import BsplinesComp
 from openmdao.components.cross_product_comp import CrossProductComp
 from openmdao.components.dot_product_comp import DotProductComp
-from openmdao.components.external_code import ExternalCode
 from openmdao.components.exec_comp import ExecComp
+from openmdao.components.external_code_comp import ExternalCode
```
Kenneth-T-Moore: You probably should add tests for the DeprecationWarning for each of these just to prevent potential errors like we had in the past.

kmarsteller: OK, I had tested them all manually, wasn't sure we needed formal tests...will do.

kmarsteller: OH, I suppose I should mention you to get you to see my submitted changes. Let me know if these tests are OK...I hadn't done deprecation tests before, but I went with the strategy of 
1.) Make sure you get the dep warning
2.) Make sure the test still runs and that you still get the right answer
@Kenneth-T-Moore

Kenneth-T-Moore: Well, you went above and beyond the call of duty. All i meant was what we normally do for those, like we did for the ScipyOptimizer deprecation:

    def test_scipyoptimizer_deprecation(self):

        msg = "'ScipyOptimizer' provides backwards compatibility " \
            + "with OpenMDAO <= 2.2 ; use 'ScipyOptimizeDriver' instead."

        # check deprecation on getter
        with warnings.catch_warnings(record=True) as w:
            driver = ScipyOptimizer()

        self.assertEqual(len(w), 1)
        self.assertTrue(issubclass(w[0].category, DeprecationWarning))
        self.assertEqual(str(w[0].message), msg)

PR #602, opened 2018-05-01
Fixed distrib_sqlite_recorder tests and updated driver iteration recording to account for None vars

ryanfarr01:


openmdao/core/driver.py
```diff
@@ -781,11 +781,17 @@ def record_iteration(self):
             sys_vars = self._gather_vars(model, sys_vars)
             in_vars = self._gather_vars(model, in_vars)
 
-        outs = des_vars
-        outs.update(res_vars)
-        outs.update(obj_vars)
-        outs.update(con_vars)
-        outs.update(sys_vars)
+        outs = {}
+        if des_vars is not None:
```
Kenneth-T-Moore: So, these are only none when self._gather_vars returns None, and that happens on Rank 0 only. Might clean this up by putting them all under a single if statement that checks the rank. However, I don't know whether your original story was to make recording work on all ranks, so you may want to think about how this all works together.

PR #604, opened 2018-05-02
P155832732 User can load a case back into a problem

hschilling:
You pick a case from a case-recorder, and pass it into a method on problem as an argument. The problem will pull all the input and output variables it finds in that case back into the model.


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -795,22 +792,360 @@ def test_get_vars(self):
...
+            prob.model._inputs[name] += 1.0
+        for name in prob.model._outputs:
+            prob.model._outputs[name] += 1.0
+
+        # Now load in the case we recorded
+        prob.load_case(case)
+
+        _assert_model_matches_case(case, model)
+
+    def test_feature_load_system_case_for_restart(self):
```
swryan: might want to add a test for when the case being loaded doesn't match the model...  is the error message user-friendly enough?

hschilling: great idea!! We should think about this with all our stories.

hschilling: Thanks for pointing this test out. When I implemented it, I realized I wasn't doing the proper checking when loading cases so this would have been completely missed!

PR #606, opened 2018-05-04
standardize usage of Options (vs Metadata)

swryan:
Pivotal #157251568: Make self.options and self.metadata both point to the same OptionsDict object
Pivotal #156956397: ks_component options and metadata should be consolidated into 'options'
Pivotal #157261933: ExternalCodeComp subclasses can't use metadata


openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -211,9 +211,6 @@ def run(self):
         fwd = problem._mode == 'fwd'
         optimizer = self.options['optimizer']
 
-        # Metadata Setup
-        self.metadata = create_local_meta(optimizer)
```
Kenneth-T-Moore: Since this happens in record_iteration now, you can remove the `create_local_meta` from the imports.

swryan: right... missed that, thanks.

PR #608, opened 2018-05-07
enable setting of Driver options via kwargs, similar to Solver

swryan:
Pivotal #157364704

### DIFF
--- a/openmdao/drivers/tests/test_scipy_optimizer.py
+++ b/openmdao/drivers/tests/test_scipy_optimizer.py
@@ -1096,9 +1096,7 @@ class TestScipyOptimizeDriverFeatures(unittest.TestCase):
         model.add_subsystem('p2', IndepVarComp('y', 50.0), promotes=['*'])
         model.add_subsystem('comp', Paraboloid(), promotes=['*'])
 
-        prob.driver = ScipyOptimizeDriver()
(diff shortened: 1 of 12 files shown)


openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -1098,9 +1096,7 @@ def test_feature_optimizer(self):
         model.add_subsystem('p2', IndepVarComp('y', 50.0), promotes=['*'])
         model.add_subsystem('comp', Paraboloid(), promotes=['*'])
 
-        prob.driver = ScipyOptimizeDriver()
-        prob.driver.options['optimizer'] = 'COBYLA'
-        prob.driver.options['disp'] = True
+        prob.driver = ScipyOptimizeDriver(optimizer='COBYLA')
```
Kenneth-T-Moore: This looks like a feature test, so I think the intent was to keep "disp" set to True here.

swryan: True is the default, and the example was only meant to illustrate setting the optmizer, per the text.

swryan: (so no reason to reinforce the already-default setting)

PR #610, opened 2018-05-07
Create an undeclare for OptionsDictionary

kmarsteller:
Certain classes have misleading documentation of their options, because they don't actually use options from their base. (e.g. `maxiter` is not used in non-iterative solvers).  Provided a way to remove these options, `undeclare`, then used it in two classes.  

A follow-up story is needed for getting `maxiter` removed from LinearRunOnce.

### DIFF
--- a/openmdao/solvers/linear/linear_runonce.py
+++ b/openmdao/solvers/linear/linear_runonce.py
@@ -27,7 +27,6 @@ class LinearRunOnce(LinearBlockGS):
         # Remove unused options from base options.
         self.options.undeclare("atol")
         self.options.undeclare("rtol")
-        # self.options.undeclare("maxiter") # eventually want to undeclare this
 
     def solve(self, vec_names, mode, rel_systems=None):
         """
--- a/openmdao/solvers/linear/linear_runonce.py
+++ b/openmdao/solvers/linear/linear_runonce.py
@@ -13,6 +13,22 @@ class LinearRunOnce(LinearBlockGS):
 
     SOLVER = 'LN: RUNONCE'
 
+    def __init__(self, **kwargs):
+        """
+        Initialize all attributes.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Options dictionary.
+        """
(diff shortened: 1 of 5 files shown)


openmdao/utils/options_dictionary.py
```diff
@@ -148,6 +148,19 @@ def declare(self, name, default=_undefined, values=None, types=None, type_=None,
         if default_provided:
             self._assert_valid(name, default)
 
+    def undeclare(self, name):
+        """
+        Remove entry from the OptionsDictionary, for classes that don't use that option.
+
```
Kenneth-T-Moore: You should add a unit test to `test_options_dictionary` that verifies that removing a key removes it (i.e., it should raise a keyerror if you try to interact with the removed option.)

kmarsteller: Done, @Kenneth-T-Moore ... see what you think.

openmdao/solvers/linear/linear_runonce.py
```diff
@@ -13,6 +13,22 @@ class LinearRunOnce(LinearBlockGS):
...
+        ----------
+        **kwargs : dict
+            Options dictionary.
+        """
+        super(LinearRunOnce, self).__init__(**kwargs)
+
+        # Remove unused options from base options.
+        self.options.undeclare("atol")
+        self.options.undeclare("rtol")
+        # self.options.undeclare("maxiter") # eventually want to undeclare this
```
swryan: why is this "eventually" versus doing it now?

Kenneth-T-Moore: Because it will take some rework on the solver call stack. Might have to rethink some things, and probably should be worked on by someone who is familiar with how the solvers work.

swryan: I guess there's a story for that...  the comment on it's own, though, only serves to confuse (IMO)

kmarsteller: All gone! 😸

PR #611, opened 2018-05-08
PETScVector/PETScTransfer is now used only in groups that own interprocess connections.

naylor-b:
- deprecated the 'vector_class' arg in Problem.setup()
- added optional 'local_vector_class' and 'distributed_vector_class' args to Problem.setup() so that the user can override the default classes if needed.
- moved all transfers classes into their own modules.
- moved setup_transfers from Group to static functions in the PETScTransfer and DefaultTransfer classes.
- moved connection and transfer related functions/data from System to Group.


openmdao/core/component.py
```diff
@@ -104,6 +105,45 @@ def setup(self):
         """
         pass
 
+    def _setup_procs(self, pathname, comm, mode):
+        """
+        Distribute processors and assign pathnames.
+
+        Parameters
```
Kenneth-T-Moore: Interesting. I just moved this to Component too on one of my branches, plus changed the docstring. Yours will get merged first for sure though.

benchmark/benchmark_beam.py
```diff
@@ -29,7 +28,7 @@ def benchmark_beam_slow_np1(self):
                                      num_elements=num_elements, num_cp=num_cp,
                                      num_load_cases=num_load_cases))
 
-        prob.setup(vector_class=PETScVector)
```
Kenneth-T-Moore: So, we don't need to assign the PETScVector anymore when we run in parallel? Wonder if there is a feature doc thata needs updated with that info.

naylor-b: Yeah it's no longer necessary. There was a feature doc that showed how to specify use of PETScVectors for a simple model.  I got rid of that section since the user no longer needs to know about it.

openmdao/core/system.py
```diff
@@ -1564,37 +1407,12 @@ def resolve(to_match, io_types, matches, proms):
 
         return maps
 
-    def _get_scope(self, excl_sub=None):
+    def _get_scope(self):
```
swryan: could use a docstring

openmdao/core/group.py
```diff
@@ -195,12 +352,72 @@ def _setup_procs(self, pathname, comm):
...
+            with self._unscaled_context_all():
+                self.resetup('update')
+
+            # Reset the _reconfigured attribute to False
+            for subsys in self._subsystems_myproc:
+                subsys._reconfigured = False
+
+            self._reconfigured = True
+
+    def _list_states(self):
```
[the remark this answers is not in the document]
naylor-b: I don't disagree.  I simply moved these functions from System to Group but know nothing about them really.  I did look at them a little based on your comment, and it turns out that _list_states_allprocs() isn't correct.  It actually won't lists states from all processes.  In fact I'd expect it to raise an exception when it calls _list_states_allprocs on a non-local subsystem.  Not sure what to do about it here though since it has nothing really to do with this story.  Maybe add a story to fix it?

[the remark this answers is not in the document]
naylor-b: Ok I went ahead and fixed _list_states_allprocs and added a test for it.

openmdao/core/system.py

openmdao/core/group.py
```diff
@@ -107,6 +144,121 @@ def configure(self):
...
+                                dimidxs = [convert_neg(cols[:, i], global_shape_out[i])
+                                           for i in range(cols.shape[1])]
+                                src_indices = np.ravel_multi_index(dimidxs, global_shape_out)
+
+                        # TODO: if either ref or ref0 are not scalar and the output is
+                        # distributed, we need to do a scatter
+                        # to obtain the values needed due to global src_indices
+                        if meta_out['distributed']:
+                            raise RuntimeError("vector scalers with distrib vars "
+                                               "not supported yet.")
```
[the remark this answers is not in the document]

openmdao/core/group.py
```diff
@@ -107,6 +144,121 @@ def configure(self):
         """
         pass
 
+    def _get_scope(self, excl_sub=None):
```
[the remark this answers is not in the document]

PR #612, opened 2018-05-08
Deriv debug print should print the whole derivative value.  Also, more flushing.

Kenneth-T-Moore:
1. Modified debug print so that derivative values are printed out too.
2. Added a couple more buffer flushes.

### DIFF
--- a/openmdao/core/total_jac.py
+++ b/openmdao/core/total_jac.py
@@ -5,6 +5,7 @@ from __future__ import print_function, division
 
 from collections import OrderedDict
 from copy import deepcopy
+import pprint
 from six import iteritems
 from six.moves import zip
 import sys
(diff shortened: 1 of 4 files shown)


openmdao/core/total_jac.py
```diff
@@ -1207,6 +1212,32 @@ def _do_scaling(self, J):
             raise RuntimeError("Derivative scaling by the driver only supports the 'dict' and "
                                "'array' formats at present.")
 
+    def _print_debug_derivatives(self):
+        """
+        Print out the derivatives when debug_print is True.
+        """
```
swryan: this is really just `_print_derivatives`... the "when debug_print is True" part is external to this function...

Kenneth-T-Moore: renamed

PR #618, opened 2018-05-10
Fix signatures of declare_partials and approx_totals to not use **kwargs

kmarsteller:
*Updated signatures to remove kwargs and use the actual named args
*Updated docstrings to reflect new signature
*Verified that documentation builds/is accurate, and all tests pass.

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -629,8 +629,6 @@ class Component(System):
         """
         Declare information about this component's subjacobians.
 
-        NOTE: `DEFAULT_FD_OPTIONS` and `DEFAULT_CS_OPTIONS` can be accesssed via `openmdao.api`.
-
         Parameters
         ----------
         of : str or list of str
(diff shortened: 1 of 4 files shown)


openmdao/core/component.py
```diff
@@ -658,8 +660,14 @@ def declare_partials(self, of, wrt, dependent=True, rows=None, cols=None, val=No
             The type of approximation that should be used. Valid options include:
             'fd': Finite Difference, 'cs': Complex Step, 'exact': use the component
             defined analytic derivatives. Default is 'exact'.
-        **kwargs : dict
-            Keyword arguments for controlling the behavior of the approximation.
+        step : float
+            Step size for approximation. Default is the value of DEFAULT_FD_OPTIONS['step'].
```
swryan: If this docstring is intended for the end user, I wonder if they should be told where to find `DEFAULT_FD_OPTIONS`?

kmarsteller: screen shot 2018-05-10 at 3 59 31 pm
I was wondering about that, but when I built the docs, the args in the sig evaluated to their actual values (see screenshot), so I wasn't sure if I needed to call it out...but I guess if they want to change the defaults, they need to know where they are...I am fine with adding that info.

swryan: yeah, I wonder if it should be in the main text of the docstring rather than blowing up the word count for each of those args.  also wonder if `DEFAULT_FD_OPTIONS` might be more conveniently accessed/reference in `openmdao.api`...

kmarsteller: @naylor-b @Kenneth-T-Moore any thoughts on @swryan's musings?

Kenneth-T-Moore: One thing that will actually change from this is: we currently support a different default stepsize for FD and for CS, but that completely goes away with this pull request, because we explicitly pass in defaults for the stepsize. Maybe the default should go back to None for the default, and we state in the docstring that None means 1e-6 for FD and 1e-15 for CS. I don't see any way to make this convenient for the user (i.e, no "Genetic Optimizer: see Optimizer, Genetic") without hardcoding numbers into the docstring here.

openmdao/core/group.py
```diff
@@ -1733,31 +1732,49 @@ def _linearize(self, do_nl=True, do_ln=True):
         if self._linear_solver is not None and do_ln:
             self._linear_solver._linearize()
 
-    def approx_totals(self, method='fd', **kwargs):
+    def approx_totals(self, method='fd', step=None, form=None, step_calc=None):
         """
         Approximate derivatives for a Group using the specified approximation method.
 
+        NOTE: `DEFAULT_FD_OPTIONS` and `DEFAULT_CS_OPTIONS` can be accessed via `openmdao.api`.
```
kmarsteller: whoops! removing this for next version, since we got rid of this idea.

openmdao/core/problem.py
```diff
@@ -617,13 +617,16 @@ def _all_components_provide_jacobians(self):
...
-                       method='fd', step=None, form=DEFAULT_FD_OPTIONS['form'],
+                       method='fd', step=DEFAULT_FD_OPTIONS['step'],
+                       form=DEFAULT_FD_OPTIONS['form'],
                        step_calc=DEFAULT_FD_OPTIONS['step_calc'],
                        force_dense=True, suppress_output=False,
                        show_only_incorrect=False):
         """
         Check partial derivatives comprehensively for all components in your model.
 
+        NOTE: `DEFAULT_FD_OPTIONS` can be accesssed via `openmdao.api.DEFAULT_FD_OPTIONS`.
```
kmarsteller: pulling this also.

PR #622, opened 2018-05-14
Add a repr to OptionsDictionary

swryan:
Pivotal #157510111
`__repr__` returns the options dict
`__str__` returns a printable version of the options table based on the rST used in docs
the logic to generate the rST has been moved from the docs extension to the class

### DIFF
--- a/openmdao/utils/tests/test_options_dictionary.py
+++ b/openmdao/utils/tests/test_options_dictionary.py
@@ -26,7 +26,6 @@ class TestOptionsDict(unittest.TestCase):
 
         self.assertEqual(self.dict.__repr__(), self.dict._dict)
 
-        print(self.dict.__str__(width=83))
         self.assertEqual(self.dict.__str__(width=83), '\n'.join([
             "========= ============ ================= ===================== ====================",
             "Option    Default      Acceptable Values Acceptable Types      Description         ",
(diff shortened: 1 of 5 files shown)


openmdao/utils/tests/test_options_dictionary.py
```diff
@@ -9,6 +9,30 @@ class TestOptionsDict(unittest.TestCase):
...
+                          desc='This description is ridiculously long and verbose, so much '
+                               'so that it takes up multiple lines in the options table.')
+
+        self.assertEqual(self.dict.__repr__(), self.dict._dict)
+
+        self.assertEqual(self.dict.__str__(width=80), '\n'.join([
+            "========= ============ ================= ================ ======================",
+            "Option    Default      Acceptable Values Acceptable Types Description           ",
+            "========= ============ ================= ================ ======================",
+            "flag      False        N/A               ['bool']                               ",
```
kmarsteller: Only comment is more of a question: does the acceptable values field need to be tested? Seems like all three test options don't use it.  Other than that question, this stuff looks good.

swryan: added a couple more options to the test

kmarsteller:
👍

PR #633, opened 2018-05-21
Added driver option to make recording N2 data optional, updated tests used in "getting metadata" doc

ryanfarr01:
Additionally removed grabbing 'promotions' from _get_viewer_data as it is unused and required non-trivial computation time for large problems.


openmdao/devtools/problem_viewer/problem_viewer.py
```diff
@@ -28,17 +28,6 @@ def _get_tree_dict(system, component_execution_orders, component_execution_index
     tree_dict['name'] = system.name
     tree_dict['type'] = 'subsystem'
 
-    local_prom_dict = OrderedDict()
-
-    from_prom_name = OrderedDict(system._var_allprocs_prom2abs_list['input'])
-    from_prom_name.update(system._var_allprocs_prom2abs_list['output'])
-    for var_prom_name, var_abs_name_list in iteritems(from_prom_name):
```
Kenneth-T-Moore: So it turned out that this data structure wasn't used or needed?

ryanfarr01: It would appear so! I searched OpenMDAO and the visualization codebase and wasn't able to find it used anywhere. My guess is that it was going to be used for the N^2 but ended up being vestigial

PR #635, opened 2018-05-22
Speedup the beam problem using sparse LU

Kenneth-T-Moore:


kmarsteller:
@naylor-b no rush, review this one after you put in your current PR

PR #636, opened 2018-05-22
Warn user if they're recording options which cannot be pickled

ryanfarr01:


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -1000,6 +1000,30 @@ def test_load_driver_cases(self):
...
+        self.prob.setup(check=False)
+        self.prob.run_driver()
+        self.prob.cleanup()
+
+        cr = CaseReader(self.filename)
+
+        d1_options = cr.system_metadata['d1']['component_options']
+
+        # no options should have been recorded for d1
+        self.assertEqual(len(d1_options._dict), 0)
```
swryan: should probably also check for the warning
(secondarily, could this be tested with a simpler model?  the Sellar problem seems overkill...)

PR #637, opened 2018-05-22
assembled jacobian API change

naylor-b:
Assembled jacobians are now associated with linear solvers instead of systems and are no longer settable directly by the user.  Linear solver option 'assemble_jac'  (True/False) specifies use of an assembled jac for that linear solver and the system option 'assembled_jac_type'  ('csc', 'dense') determines the type of the assembled jacobian.


openmdao/core/system.py
```diff
@@ -27,12 +27,18 @@
 from openmdao.utils.array_utils import convert_neg
 from openmdao.utils.record_util import create_local_meta, check_path
 from openmdao.utils.write_outputs import write_outputs
+from openmdao.jacobians.assembled_jacobian import DenseJacobian, CSCJacobian
```
Kenneth-T-Moore: Not sure why you don't combine all 3 imports from here into the first import statement.

naylor-b:
just ran this PR against dymos (updated for API changes) and am getting a failure, so I'm closing this until I sort it out.

PR #648, opened 2018-06-04
Rework how we handle check_partials with complex step

Kenneth-T-Moore:
If we have a model where some components request a check with complex step, but the complex vector wasn't allocated in setup, then do the following:
1. Do finite difference instead of raising an error
2. When done calculating derivatives, the user is informed that this was done, and the list of fd'd components is printed out.


openmdao/core/problem.py
```diff
@@ -1016,6 +1017,12 @@ def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes
         if out_stream == _DEFAULT_OUT_STREAM:
             out_stream = sys.stdout
 
+        if len(comps_could_not_cs) > 0:
+            msg = "The following components requested complex step, but force_alloc_complex " + \
+                "has not been set to True, so finite difference was used:"
+            print(msg)
+            print(list(comps_could_not_cs))
```
swryan: why not make this a `warning`?

Kenneth-T-Moore: Done

PR #657, opened 2018-06-12
P157746717 User can use an ImplicitExternalCode comp from standard library

hschilling:
wrap an external code, but treat it as implicit component. All partials are FD around the apply_nonlinear method.

### DIFF
--- a/openmdao/components/external_code_comp.py
+++ b/openmdao/components/external_code_comp.py
@@ -401,3 +401,21 @@ class ExternalCodeImplicitComp(ImplicitComponent):
             unscaled, dimensional residuals written to via residuals[key]
         """
         self._external_code_runner.run_component()
+
+    def solve_nonlinear(self, inputs, outputs):
+        """
+        Compute outputs given inputs. The model is assumed to be in an unscaled state.
+
+        Parameters
+        ----------
+        inputs : Vector
+            unscaled, dimensional input variables read via inputs[key]
+        outputs : Vector
+            unscaled, dimensional output variables read via outputs[key]
+
+        Returns
+        -------
+        None or bool or (bool, float, float)
+            The bool is the failure flag; and the two floats are absolute and relative error.
+        """
+        self._external_code_runner.run_component()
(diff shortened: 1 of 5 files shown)


openmdao/components/external_code_comp.py
```diff
@@ -401,3 +401,16 @@ def apply_nonlinear(self, inputs, outputs, residuals):
             unscaled, dimensional residuals written to via residuals[key]
         """
         self._external_code_runner.run_component()
+
+    def solve_nonlinear(self, inputs, outputs):
```
Kenneth-T-Moore: I think we need to make clear in the docs that, if the user overrides `apply_nonlinear` but does not override `solve_nonlinear` in their comp, then they are going to get a double run. This is different than with a regular ImplicitComponent, where if you don't define `solve_nonlinear`, then it does nothing, so it might cause some confusiion.

openmdao/components/tests/extcode_mach.py
```diff
@@ -0,0 +1,58 @@
...
+    with open(input_filename, 'r') as input_file:
+        output_or_resids = input_file.readline().strip()
+        area_ratio = float(input_file.readline())
+        if output_or_resids == 'residuals':
+            mach = float(input_file.readline())
+        else: # outputs
+            super_sonic = (input_file.readline().strip() == "True")
+
+    if output_or_resids == 'outputs':
+        mach_output = mach_solve(area_ratio,super_sonic=super_sonic)
```
Kenneth-T-Moore: PEP space after comma

openmdao/components/tests/extcode_mach.py
```diff
@@ -0,0 +1,58 @@
...
+        with open(output_filename, 'w') as output_file:
+            output_file.write('%.16f\n' % mach_output)
+
+    if output_or_resids == 'residuals':
```
Kenneth-T-Moore: make this an "elif"

openmdao/components/external_code_comp.py
JustinSGray: i don't understand this comment. Plus, the user HAS to define apply_nonlinear?

hschilling: Since you know the subject way better than I do, can you give me the actual words to put into the doc?

Kenneth-T-Moore: I posted in General to discuss further.

PR #664, opened 2018-06-14
Experimental - For Review.

Kenneth-T-Moore:
When calculating totals, instead of setting -1 into the input side, here we multiply the resulting column (or row) by -1.


JustinSGray:
for the record, I asked Ken to do this over his objections. All consequences and fall out rest squarely on my shoulders :)

kmarsteller:
So if it's over Ken's objections, squarely on you, do we only need the one review? 😃

Kenneth-T-Moore:
For what it is worth, my objections are fairly mild, if this makes it easier for people to reconcile the implementation with the original equations.

PR #666, opened 2018-06-18
Add a new math component for element-wise vector and matrix addition / subtraction

bbrelje:
In the process of developing a conceptual aircraft design tool in OpenMDAO (to be released eventually) I ended up developing some math components in a similar vein to the dot_product and cross_product components. Would probably be best to contrib these back to OpenMDAO as they may be more generally useful. 

Tests pass and docs build correctly on my machine.

First time making a PR on a large project; I'm definitely open to style feedback apart from the actual code content (i.e. did I do the version control in a non-annoying fashion)


bbrelje:
Sorry about the extra commits. I was running pytest on my machine, not testflo, so I was missing the code review / style checks.

kmarsteller:
Once you pass CI, we will have a bit of a code review. So the comments and suggestions will come then.  Thanks!

openmdao/components/add_subtract_comp.py
```diff
@@ -0,0 +1,235 @@
...
+    def initialize(self):
+        """
+        Declare options.
+
+        Parameters
+        ----------
+        complex : Boolean
+            Set True to enable complex math (e.g. for complex step verification)
+        """
+        self.options.declare('complex', default=False,
```
Kenneth-T-Moore: I think you can get rid of this option, and instead, in `compute` you can query if `outputs._alloc_complex` is True. That way you only allocate complex when you are in a complex step situation. I am away from my computer right now, so will verify this later, but I am pretty sure it works. We should probably document this so that users can use it in their comps (and maybe remove the underscore since it is being used public.)

PR #672, opened 2018-06-26
Added indexing to SQLite DBs and ability to load DB into memory

ryanfarr01:


openmdao/docs/features/recording/getting_values.rst
```diff
@@ -134,3 +134,23 @@ runs of the model.
 .. embed-code::
     openmdao.recorders.tests.test_sqlite_recorder.TestFeatureSqliteRecorder.test_feature_load_system_case_for_restart
     :layout: interleave
+
+Loading a DataBase into Memory
+------------------------------
+
+Every time the `get_case` method is used, the case reader is making a new query
+to the DB (with the exception of recurring requests, which are cached). This doesn't
```
swryan: suggest replacing "DB" with "database" (i.e. not using shorthand in documentation)

openmdao/docs/features/recording/getting_values.rst
```diff
@@ -134,3 +134,23 @@ runs of the model.
...
+with minimal queries.
+
+To use this method, simply create the CaseReader, call the `load_cases` method, and use the
+reader as you normally would.
+
+.. code-block:: console
+
+    cr = CaseReader('cases.sql')
+    cr.load_cases()
+    ...
```
swryan: I wonder if the docs should also mention the availability of the more granular methods, e.g. `cr.driver_cases.load_cases()` I imagine solver and system cases could be unwieldy memory-wise for a large dataset, but it might be useful to just load all the driver cases (and then possibly cherry pick solver and system cases).

ryanfarr01: Good point! I'll add that in

PR #676, opened 2018-06-28
Fixing things/simplifying Appveyor run

kmarsteller:
*Appveyor was running only on 2.7 due to changes made at some point. 3.6 now works also, due to fixing the path to get to Miniconda

* Many lines of code were not even executing or working in any way, so I pared the file down to what works. (we weren't running docs/mpi/petsc/pyoptsparse on Windows anyway, so I removed the dead code that wasn't executing)

* Updated numpy to no longer be pinned, and run latest numpy release

* Had to pin scipy to 1.0.x in order to get one test that uses ScipyKrylov to pass (same problem that is seen in benchmark running of the tests.)

* Removed parametric testing

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ install:
   - conda config --set always_yes yes
   # We need to do this first as other commands may not work with older versions of conda.
   - conda update conda
-  - conda install numpy=1.11.2 scipy nose hdf5 h5py sphinx mock pip --quiet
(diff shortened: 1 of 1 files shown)


kmarsteller:
for the kmarsteller/OpenMDAO testing on Appveyor, it installs the cloned repo into folder "OpenMDAO", for the OpenMDAO/OpenMDAO testing, it installs the cloned repo into folder "blue".
Repo name changing sucks so hard.

PR #679, opened 2018-06-29
Some pyoptsparse optimizers don't run the model at the initial value, so always run those first.

Kenneth-T-Moore:


openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -1567,6 +1569,9 @@ def test_initial_run_SNOPT(self):
         self.assertEqual(comp.visited_points[0], 1.0)
         self.assertNotEqual(comp.visited_points[1], 1.0)
 
+    # This one hangs on Travis for numpy 1.12 and we can't reproduce the error anywhere where we can
+    # debug it, so we're skipping it for numpy 1.12.
```
swryan: this comment should be updated... it's a bug in 1.12 not a hang, right?
also the commit message says 1.14, but the skiptest is checking for 1.13...  assume 1.13 is good

Kenneth-T-Moore: Changed the comment and commit.

PR #680, opened 2018-07-02
P158331772 record final

hschilling:
User can set an option that has case recorder do a final, more comprehensive save of data.


openmdao/core/problem.py
```diff
@@ -531,11 +567,170 @@ def run(self):
...
+
+        names = model._outputs._names
+        views = model._outputs._views
+
+        if MPI:
+            des_vars = model._gather_vars(model, des_vars)
+            obj_vars = model._gather_vars(model, obj_vars)
+            con_vars = model._gather_vars(model, con_vars)
+
+        outs = {}
```
naylor-b: Is there any reason why we couldn't combine des_vars, obj_vars, and con_vars into 'outs' BEFORE we call model._gather_vars, and just do a single MPI gather call using 'outs' rather than 3 different gather calls?  We do this same thing in Driver.record_iteration as well, except there we make 6 calls when it seems like we could get away with just 2.

openmdao/recorders/case.py
```diff
@@ -297,6 +297,44 @@ def __init__(self, filename, counter, iteration_coordinate, timestamp, success,
         self.rel_err = rel_err
 
 
+class ProblemCase(Case):
```
naylor-b: I may be missing something, but it seems like we have a bunch of Case wrappers in this file that don't really do anything but reorder some args to __init__ or maybe set some to None.  Just curious why the arg ordering is different between Case and some of the wrappers, and why we couldn't give some of the Case.__init__ args default values of None to eliminate the need for some of this code.

openmdao/core/problem.py
```diff
@@ -531,11 +567,170 @@ def run(self):
...
+                    mysystem_outputs = all_vars[-1]
+                    for d in all_vars[:-1]:
+                        mysystem_outputs.update(d)
+
+            # de-duplicate mysystem_outputs
+            mysystem_outputs = mysystem_outputs.difference(all_desvars, all_objectives,
+                                                           all_constraints)
+
+        # if rec_inputs:
+        #     root = self.model
```
Kenneth-T-Moore: Why is this section commented out?

openmdao/core/problem.py
```diff
@@ -531,11 +567,170 @@ def run(self):
...
+        #         all_vars = root.comm.gather(myinputs, root=0)
+        #         if MPI.COMM_WORLD.rank == 0:
+        #             myinputs = all_vars[-1]
+        #             for d in all_vars[:-1]:
+        #                 myinputs.update(d)
+
+        if MPI:  # filter based on who owns the variables
+            # TODO Eventually, we think we can get rid of this next check. But to be safe,
+            #       we are leaving it in there.
+            if not model.is_active():
```
Kenneth-T-Moore: Can prob.root ever not be active?

openmdao/core/problem.py
```diff
@@ -531,11 +567,170 @@ def run(self):
...
+           A recorder instance.
+        """
+        self._rec_mgr.append(recorder)
+
     def cleanup(self):
         """
         Clean up resources prior to exit.
         """
         self.driver.cleanup()
+        self._rec_mgr.close()
```
Kenneth-T-Moore: Heads up, you will need to make some changes once Steve's current PR is merged, where close-> cleanup

openmdao/core/tests/test_problem.py
```diff
@@ -1140,6 +1139,8 @@ def setup(self):
         self.assertTrue(isinstance(top.model.sub.nonlinear_solver, NewtonSolver))
         self.assertTrue(isinstance(top.model.sub.linear_solver, ScipyKrylov))
 
+
```
Kenneth-T-Moore: You've added a few extra carriage returns here.

openmdao/core/problem.py
```diff
@@ -531,11 +567,170 @@ def run(self):
...
+            if MPI:
+                all_vars = model.comm.gather(mysystem_outputs, root=0)
+                if MPI.COMM_WORLD.rank == 0:
+                    mysystem_outputs = all_vars[-1]
+                    for d in all_vars[:-1]:
+                        mysystem_outputs.update(d)
+
+            # de-duplicate mysystem_outputs
+            mysystem_outputs = mysystem_outputs.difference(all_desvars, all_objectives,
+                                                           all_constraints)
```
ryanfarr01: May want to wait until tomorrow's discussion to put this in. Excluding all_desvars, all_objectives, and all_constraints in the line above (rather than (mydesvars, myobjectives, and myconstraints, respectively) is why we're running into some unintuitive problems with Driver recording options. Looks like this will run into the same issues and probably require the same solution

openmdao/recorders/case.py
ryanfarr01: I think this is a problem that evolved over the course of a couple PRs. I think `ProblemCase` is the only one that has to put `None` into its `super` call at the moment. But indeed, with re-ordering and a default value for Inputs we could reduce the code/comments for every case type (except SolverCase which stores a couple additional values).

openmdao/core/tests/test_problem.py
hschilling: fixed. Not sure how that happened

openmdao/core/problem.py
hschilling: From Bret: Currently prob.root can't be inactive

openmdao/core/problem.py
hschilling: I put in a chore story to do this

openmdao/recorders/case.py
hschilling: I put in a chore story to do this

PR #696, opened 2018-07-11
fixed indexing for reporting singular errors

Kenneth-T-Moore:
These were off for some of the larger ones reported in pycycle. Luckily now we have _var_sizes cached in system so we don't have to try to sum up sizes anymore.

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -57,11 +57,11 @@ def format_singular_error(err, system, mtx):
 
     n = 0
     varname = "Unknown"
-    for name in system._var_allprocs_abs_names['output']:
-        relname = system._var_abs2prom['output'][name]
-        n += len(system._outputs[relname])
+    varsizes = system._var_sizes['nonlinear']['output']
+    for j, name in enumerate(system._var_allprocs_abs_names['output']):
+        n += varsizes[0][j]
         if loc <= n:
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore:
Yeah, wasn't sure I wanted to put that whole pycycle model in here.

Kenneth-T-Moore:
Though actually, I've got an idea on how I can shrink this down to a smaller test.

Kenneth-T-Moore:
Added tests, and good thing because there was an additional fix.

openmdao/solvers/linear/direct.py
```diff
@@ -57,11 +57,11 @@ def format_singular_error(err, system, mtx):
...
     n = 0
     varname = "Unknown"
-    for name in system._var_allprocs_abs_names['output']:
-        relname = system._var_abs2prom['output'][name]
-        n += len(system._outputs[relname])
-        if loc <= n:
-            varname = relname
+    varsizes = system._var_sizes['nonlinear']['output']
+    for j, name in enumerate(system._var_allprocs_abs_names['output']):
+        n += varsizes[0][j]
```
naylor-b: might want to use owning_rank here instead of 0.  Var might not exist on proc 0.

Kenneth-T-Moore: Modified this to index via owning_rank.

PR #697, opened 2018-07-12
P158161973 model metadata optionally should be recorded

hschilling:
Original story was:

If a recorder is attached to a system object, the metadata for that system and all systems below it should be recorded by default. If the recorder is attached to the driver or root, it should record the metadata for all systems by default.

Followed up by this simplification from Justin:

1) if you record any system meta-data you record it all (regardless of where its saved to) (edited)
2) i think the option should just be called `record_model_metadata` for both systems and driver. The docs can make it clear that this means ALL metadata from all groups/components will be saved

Also fixed the fact that `record_metadata` option on System was never used. Added test for that.

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -141,6 +141,8 @@ class Driver(object):
 
         self.recording_options.declare('record_metadata', types=bool, default=True,
                                        desc='Record metadata')
+        self.recording_options.declare('record_model_metadata', types=bool, default=True,
+                                       desc='Record metadata for all Systems in the model')
         self.recording_options.declare('record_desvars', types=bool, default=True,
                                        desc='Set to True to record design variables at the '
                                             'driver level')
@@ -408,6 +410,11 @@ class Driver(object):
                     from openmdao.devtools.problem_viewer.problem_viewer import _get_viewer_data
                     self._model_viewer_data = _get_viewer_data(problem)
             self._rec_mgr.record_metadata(self)
+
+        # Also record the system metadata to the recorders attached to this Driver
+        if self.recording_options['record_model_metadata']:
+            for sub in model.system_iter(recurse=True, include_self=True):
+                self._rec_mgr.record_metadata(sub)
 
     def _get_voi_val(self, name, meta, remote_vois, unscaled=False, ignore_indices=False):
         """
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -140,7 +140,7 @@ class Driver(object):
         self.recording_options = OptionsDictionary()
 
         self.recording_options.declare('record_metadata', types=bool, default=True,
-                                       desc='Record metadata')
+                                       desc='Record Driver metadata')
         self.recording_options.declare('record_model_metadata', types=bool, default=True,
                                        desc='Record metadata for all Systems in the model')
         self.recording_options.declare('record_desvars', types=bool, default=True,
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -695,7 +695,7 @@ class System(object):
         self._setup_var_sizes(recurse=recurse)
         self._setup_connections(recurse=recurse)
 
-    def _setup_case_recording(self, recurse=True):
+    def _setup_recording(self, recurse=True):
         myinputs = myoutputs = myresiduals = set()
         incl = self.recording_options['includes']
         excl = self.recording_options['excludes']
@@ -726,7 +726,7 @@ class System(object):
         # Recursion
         if recurse:
             for subsys in self._subsystems_myproc:
-                subsys._setup_case_recording(recurse)
+                subsys._setup_recording(recurse)
 
     def _final_setup(self, comm, setup_mode, force_alloc_complex=False):
         """
@@ -785,7 +785,7 @@ class System(object):
         self._setup_partials(recurse=recurse)
         self._setup_jacobians(recurse=recurse)
 
-        self._setup_case_recording(recurse=recurse)
+        self._setup_recording(recurse=recurse)
 
         # If full or reconf setup, reset this system's variables to initial values.
         if setup_mode in ('full', 'reconf'):
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -799,7 +799,7 @@ class System(object):
         # Also, optionally, record to the recorders attached to this System,
         #   the system metadata for all the subsystems
         if self.recording_options['record_model_metadata']:
-            for sub in self.system_iter(recurse=True, include_self=False):
+            for sub in self.system_iter(recurse=True, include_self=True):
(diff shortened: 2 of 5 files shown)


openmdao/core/driver.py
```diff
@@ -141,6 +141,8 @@ def __init__(self, **kwargs):
 
         self.recording_options.declare('record_metadata', types=bool, default=True,
                                        desc='Record metadata')
```
Kenneth-T-Moore: There might be some confusion about what the difference between metadata and model metadata. Wonder if we should be specific on the option name and make it `record_driver_metadata`.  Even if we don't, the desc for it should be more descriptive.

openmdao/core/system.py
```diff
@@ -788,8 +791,16 @@ def _final_setup(self, comm, setup_mode, force_alloc_complex=False):
         if setup_mode in ('full', 'reconf'):
             self.set_initial_values()
 
+        # Tell all subsystems to record their metadata if they have recorders attached
         for sub in self.system_iter(recurse=True, include_self=True):
-            sub._rec_mgr.record_metadata(sub)
+            if sub.recording_options['record_metadata']:
+                sub._rec_mgr.record_metadata(sub)
```
Kenneth-T-Moore: So, doesn't each sub recorder also need to recurse down below it and record subsystem metadata if record_model_metadata is turned on for that sub recorder?

openmdao/recorders/base_recorder.py
```diff
@@ -117,6 +117,33 @@ def record_metadata_driver(self, recording_requester):
...
         raise NotImplementedError()
 
+    def _get_metadata_system(self, recording_requester):
+        # Cannot handle PETScVector yet
+        from openmdao.api import PETScVector
+        if PETScVector and isinstance(recording_requester._outputs, PETScVector):
+            return None, None  # Cannot handle PETScVector yet
+
+        from six import iteritems
+        from openmdao.utils.options_dictionary import OptionsDictionary
```
Kenneth-T-Moore: Seems like these imports can go to the top.

openmdao/recorders/tests/test_sqlite_recorder.py
```diff
@@ -384,6 +384,77 @@ def test_driver_records_metadata(self):
...
+
+        recorder = SqliteRecorder("cases.sql")
+        prob.model.add_recorder(recorder)
+
+        prob.run_model()
+        prob.cleanup()
+
+        cr = CaseReader("cases.sql")
+        self.assertEqual(set(cr.system_metadata.keys()),
+                         set(['root', 'px', 'pz', 'd1', 'd2', 'obj_cmp', 'con_cmp1', 'con_cmp2']))
```
Kenneth-T-Moore: To be safe on all these tests, maybe you want to check the values too?

openmdao/core/driver.py
hschilling: Good point. I agree we could do record_driver_metadata but then we would also, to be consistent, have to use record_system_metadata and record_solver_metadata. Do we want to do that?

I did update the docstring to indicate that this option records Driver metadata.

openmdao/recorders/base_recorder.py
hschilling: Sorry. Had those there while writing it and missed those while cleaning up

openmdao/recorders/tests/test_sqlite_recorder.py
hschilling: We could but a lot of the other tests did not. So if we do this one, shouldn't we do the other ones?

openmdao/core/system.py
hschilling: Yeah, this starts to get complicated. I had it that way for a while. But it seemed like the primary (only?) use case for this was to do the recursing from the root. 

Any thoughts on that?

openmdao/recorders/tests/test_sqlite_recorder.py
Kenneth-T-Moore: To be safe, yes.

openmdao/core/system.py
Kenneth-T-Moore: Well, I'll confess that I don't quite understand the motivation behind some of this. My thought would be to just record metadata for the variables that you are recording in the case recorder, in which case you would just need one flag to turn that on or off.

JustinSGray: the rule is that if you record system metadata from any level, then you record it from the root down. There is no partial recording of this. You get all or nothing.

Kenneth-T-Moore: Okay. The code above doesn't quite do that. It needs to do a nested system_iter so that each sub from every system_iter loops over all systems in the model and sets it to record that metadata too.

    for sub in self.system_iter(recurse=True, include_self=True):
        sub_rec = sub._rec_mgr

        # The system slotted here needs to record metadata at all levels if flag set to True
        if sub.recording_options['record_metadata']:
            for each_system in self.system_iter(recurse=True, include_self=True):
                sub_rec.record_metadata(each_system)

Does this match what we mean here?

hschilling: Justin, so you are saying that if record_model_metadata is True at any level, then record System metadata for all Systems from the root down? Remember record_model_metadata is by default True, so that if you want to prevent recording all System metadata recording, you would have to set that to False for all Systems. I must be missing something!

Even if we make the detault False, then for this code below, because d1 has record_model_metadata set to True, then again, all Systems from root down will be recorded. It seems a little non-intuitive, if that is the case.

```

        prob = SellarProblem()
        prob.setup()

        recorder = SqliteRecorder("cases.sql")
        prob.model.add_recorder(recorder)

        d1 = prob.model.d1
        d1.recording_options['record_model_metadata'] = True
        d1.add_recorder(recorder)

        prob.run_model()
```

Would it be easier if the record_model_metadata option were on the recorder itself?

openmdao/recorders/tests/test_sqlite_recorder.py
hschilling: ok

openmdao/core/system.py
JustinSGray: hrm... you would only have to set it to false anywhere you added a recorder. In 99.9% of times that will either be the root or the top `model` system. 

I think its best to just keep it simple. The option is with all the other options. If you ask for metadata anywhere, you get it all. Default to true

hschilling: So in the Seller example, if we only attach a recorder to d1, and leave record_model_metadata as the default of true, you want all metadata from all the systems in the model recorded to the recorder. Right? 

Have to see how to do that. Subsystems don't have pointers to the model.

JustinSGray: But what you described is what I want. 
if its not workable because of lack of pointers, then I'll accept that you just store all the metadata from that point down.

hschilling: ok. will see what I can do !

PR #698, opened 2018-07-16
Generalized `generate_docs` function to be used for other projects as well as OpenMDAO

johnjasa:


kmarsteller:
Approve, this will help with OAS and future repos.

PR #699, opened 2018-07-18
Support for recording derivatives.

Kenneth-T-Moore:


openmdao/docs/features/recording/getting_derivatives.rst
```diff
@@ -0,0 +1,14 @@
+****************************************
+Getting Derivatives with the Case Reader
+****************************************
+
+If you are running with a driver that computes derivatives, you can also record and read them through the `CaseReader`.
```
kmarsteller: This may be a little picky, but this sentence makes it sound like you record them through the CaseReader.
Maybe "you can also record derivatives, and then subsequently read them through the `CaseReader`.  ?  Read it over and see if you agree. Maybe only need to change it if there are other changes needed to the PR.

openmdao/recorders/sqlite_reader.py
```diff
@@ -127,6 +130,12 @@ def _load(self):
                 self.driver_cases._case_keys = [coord[0] for coord in rows]
                 self.driver_cases.num_cases = len(self.driver_cases._case_keys)
 
+                cur.execute("SELECT iteration_coordinate FROM driver_derivatives ORDER BY id ASC")
```
swryan: per slack discussion this should be wrapped in a try/except for reading legacy databases that do not have this table  (and/or the database version should be updated due to change in schema)

openmdao/docs/features/recording/getting_derivatives.rst
swryan: I agree with Keith's comment that this could use a bit of wordsmithing

openmdao/recorders/base_recorder.py
```diff
@@ -258,6 +258,45 @@ def record_iteration_problem(self, recording_requester, data, metadata):
...
+            Some implementations of record_derivatives need additional args.
+        """
+        if not self._parallel:
+            if MPI and MPI.COMM_WORLD.rank > 0:
+                raise RuntimeError("Non-parallel recorders should not be recording on ranks > 0")
+
+        self._iteration_coordinate = recording_iteration.get_formatted_iteration_coordinate()
+
+        if isinstance(recording_requester, Driver):
+            self.record_derivatives_driver(recording_requester, data, metadata)
```
swryan: seems like there should be an `else raise RuntimeError` here...

Kenneth-T-Moore: I only call this function from a Driver. Probably don't even need the "if", but I was keeping the code in parallel with the other recorders.  I will just take that out to eliminate confusion.

openmdao/recorders/sqlite_reader.py
Kenneth-T-Moore: This will be addressed in the follow-on story https://www.pivotaltracker.com/story/show/159152450

openmdao/recorders/base_recorder.py
swryan: but the user could incorrectly call it from another object, in which case we want it to error, right?

Kenneth-T-Moore: Can you give me a usecase (code example) where a user calls record_derivatives?

Kenneth-T-Moore: Or record_iteration if you are more familiar with that method.  We currently have not coverage for the "else raise RuntimeError" branch on that.

swryan: sorry you're right.. ignore last comment

openmdao/docs/features/recording/getting_derivatives.rst
Kenneth-T-Moore: Reworded this.

PR #701, opened 2018-07-19
Updated sql format version to 2. Higher merge priority.

Kenneth-T-Moore:


openmdao/recorders/sqlite_recorder.py
```diff
@@ -67,7 +67,7 @@ def blob_to_array(blob):
     return np.load(out)
 
 
-format_version = 1
```
swryan: seems like we should have some kind of doc/record of what changed between versions..   maybe just a changelog-like comment here in the code?

PR #705, opened 2018-07-25
Record/read JSON rather than binary numpy arrays

ryanfarr01:
Based on benchmarking using PyCycle's CFM56:

**DB Size:** Increase by 5.36%
**Average Recording Time:** Decrease by 13.656%
**Average Reading Time:** Decrease by 77.45%

Overall, looks like this ended up being a pretty big win


openmdao/recorders/sqlite_recorder.py
```diff
@@ -23,12 +24,14 @@
 """
 SQL case output format version history.
 ---------------------------------------
+3 -- OpenMDAO 2.5
```
Kenneth-T-Moore: I suspect this will also make it in 2.4.

PR #707, opened 2018-07-31
P158975747 user can get scaled or unscaled values from driver cases using case reader

hschilling:
Below is the description from the story. In addition to working the story, this PR also fixes several bugs noticed along the way.

Currently, when recording from a driver things are recorded scaled, but none of the actual scaling factors are recorded, so things can't be returned to an unscaled state.

I think it would be simpler to just record the physical values out of the arrays. Then the case object can always convert between scaled and physical. 

If its easier to record in scaled, thats fine. We just need the information and some methods on case (or on case reader maybe) to control if the values are scaled or physical when you get them from the case


openmdao/recorders/sqlite_reader.py
```diff
@@ -81,14 +84,28 @@ def __init__(self, filename):
...
+                cur.execute(
+                    "SELECT format_version, abs2prom, prom2abs, abs2meta FROM metadata")
             row = cur.fetchone()
             self.format_version = row[0]
             self._abs2prom = None
             self._prom2abs = None
             self._abs2meta = None
+            self._var_settings = None
+
+            if self.format_version == 4:
```
Kenneth-T-Moore: you could future proof this with a >=

hschilling: good point. I did that for the >=3 and then forgot to do the same for my new version.

kmarsteller: should I be waiting for this small change before i merge? @hschilling

hschilling: Sorry, I just forgot to push up. Done.

PR #710, opened 2018-08-02
SimpleGADriver AnalysisError import error fixed

onodip:
In genetic_algorithm_driver AnalysisError import error fixed by adding import statement for AnalysisError. 
Previously SimpleGADriver failed with NameError and did not return the original error from the model if an exception was raised during the execution of run_driver()
In test_genetic_algorithm_driver.py  test_analysis_error() test function added, which checks the type of the exception.

### DIFF
--- a/openmdao/drivers/tests/test_genetic_algorithm_driver.py
+++ b/openmdao/drivers/tests/test_genetic_algorithm_driver.py
@@ -153,13 +153,44 @@ class TestSimpleGA(unittest.TestCase):
         prob['area3'] = 0.0005
         prob.run_driver()
 
-        # Note, GA doesn't do so well with the continunous vars, naturally, so we reduce the space
-        # as much as we can. Objective is stll rather random, but it is close. GA does a great job
+        # Note, GA doesn't do so well with the continuous vars, naturally, so we reduce the space
+        # as much as we can. Objective is still rather random, but it is close. GA does a great job
         # of picking the correct values for the integer desvars though.
         self.assertLess(prob['mass'], 6.0)
         assert_rel_error(self, prob['mat1'], 3, 1e-5)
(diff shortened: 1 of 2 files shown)


openmdao/drivers/tests/test_genetic_algorithm_driver.py
```diff
@@ -153,13 +153,44 @@ def compute(self, inputs, outputs):
...
         assert_rel_error(self, prob['mat2'], 3, 1e-5)
-        #Material 3 can be anything
+        # Material 3 can be anything
+
+    def test_analysis_error(self):
+        np.random.seed(1)
+
+        class ValueErrorComp(ExplicitComponent):
+            """
+            Raises value error. AnalysisError should return the original error.
```
naylor-b: This comment isn't quite right.  A solver can raise an AnalysisError if it fails to converge, or certain components will generate AnalysisErrors, but in the case of ValueErrorComp, it will simply raise its ValueError and no AnalysisError will ever be involved.  I would just remove this comment since it will probably cause confusion.

openmdao/drivers/tests/test_genetic_algorithm_driver.py
```diff
@@ -153,13 +153,44 @@ def compute(self, inputs, outputs):
...
+
+        model.connect('p.x', 'comp.x')
+
+        model.add_design_var('p.x', lower=-5.0, upper=10.0)
+        model.add_objective('comp.f')
+
+        prob.driver = SimpleGADriver(max_gen=75, pop_size=25)
+        prob.driver._randomstate = 1
+        prob.setup(check=False)
+        # prob.run_driver()
```
JustinSGray: how does the error get raised if you don't call run_driver?

PR #713, opened 2018-08-03
JSON Serialization fix

robfalck:
Sometimes the user might specify driver variable options, like `equals` or `indices` as ndarrays.  This breaks the initialization of the SQLiteRecorder, since ndarrays are not JSON serializable.  This fix converts any _var_settings that might be ndarrays to lists upon initialization of the SQLiteRecorder.  

Note that the values are not converted back to ndarray by the SQLiteCaseReader at this time, since we're not saving all possible ndarray attributes in _var_settings and doing so doesn't seem necessary.

Added recorder test that fails due to indices and equals var options being ndarrays, and that passes with this fix.

### DIFF
--- a/openmdao/recorders/sqlite_recorder.py
+++ b/openmdao/recorders/sqlite_recorder.py
@@ -26,6 +26,9 @@ from openmdao.solvers.solver import Solver
 """
 SQL case output format version history.
 ---------------------------------------
+5 -- OpenMDAO 2.4
+    More general handling of ndarray variable settings metadata.  Stores metadata keys which
+    are ndarrays in a separate 'ndarrays' entry to var_settings.
 4 -- OpenMDAO 2.4
     Added variable settings metadata that contains scaling info.
 3 -- OpenMDAO 2.4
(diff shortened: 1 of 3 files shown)


hschilling:
Not sure if this is an issue but shouldn't SqliteReader be updated so that the metadata arrays like lower, upper, and indices be converted to ndarrays just like they were set originally ? SqliteReader returns them as lists now. 

Other than that, it all looks good.

robfalck:
If it doesn't break unit tests then I'm inclined to leave it as is, but we could cache the names of the settings that are ndarrays and reconstruct them.  It gets complicated because some (`indices`, for example) should have a dtype of int while others will be float.  I'm not sure if any of this data will ever change, or if any other ndarray options need to be saved in order to properly reconstruct them.

JustinSGray:
its simpler to leave them as lists, and I don't think it really hurts anything. They don't really need to be arrays until they are actually used in practice inside the framework, but thats not the problem of the case reader. 

I know its a bit wonky, but for now I agree with Rob. We have a lot of little things to fix on case reader and this gets us past a bug that was holding him up without having to increment the case reader version again.

hschilling:
Ok. Makes sense

PR #723, opened 2018-08-12
Update problem_viewer.py

Js775:
1. Added an option [encoding="utf8"] to Ln.226, because there was some 'UnicodeDecodeError' when I was try to generate the N2 diagram using OpenMDAO CommandLine Tools.  Also imported "io" module and used "io.open" method to satisfy the python 2.7 environment instead of "open" method. The original error message was as follow :

    UnicodeDecodeError : 'cp949' codec can't decode byte 0xe2 in position 104: illegal multibyte sequence

 I am using the Windows 10 for Korean language as a development environment. The problem was solved after specifying Unicode as utf-8 via the option. Perhaps someone who has a similar problem with non-English operating systems will be able to solve the problem in the same way.


JustinSGray:
it looks like your update worked on python 3, but not python 2

PR #724, opened 2018-08-12
P159445512 record correct size for arrays so that when reading from case recording files the proper shape is used

hschilling:
In versions of sqlite_recorder writing with format_version >= 3, we are using JSON to store the values and previously, when loading those back in as part of reading the cases, the code did not use the shape of the arrays. So multidimensional arrays weren't loaded back in properly.


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -1049,13 +1050,19 @@ def test_reading_driver_cases_with_indices(self):
...
         model = prob.model
         indeps = model.add_subsystem('indeps', IndepVarComp(), promotes_outputs=['*'])
 
+        # the following were randomly generated using np.random.random(10)*2-1 to randomly
+        # disperse them within a unit circle centered at the origin.
+        # Also converted this array to > 1D array to test that capability of case recording
+        x_vals = np.array([0.55994437, -0.95923447, 0.21798656, -0.02158783, 0.62183717,
+                           0.04007379, 0.46044942, -0.10129622, 0.27720413, -0.37107886]).reshape(
+            (-1, 1))
+
```
swryan: looks like this is unused?

hschilling: Thanks for catching that.  I fixed it.

PR #733, opened 2018-08-16
 add a tooltip showing promoted name of input/output vars to model viewer

swryan:
Pivotal #159672300 : User can see promoted name in n2 diagram
(a similar change will also be made to OVIS)


swryan:
fixed

PR #736, opened 2018-08-21
Complex Step Refactor

Kenneth-T-Moore:
1. Refactor Vectors to contain swappable real/complex arrays rather than a real and imaginary one.
  a. Implementation is much simpler at a possible cost of a little more average memory.
  b. Fixes a bug that would have otherwise required increased complexity in the Vectors
  c. Removes roadblock for complex step around newton solves.
2. Added a System attribute "under_complex_step" that the user can check to find out if he is in a complex step.
  a. Removed class attribute in Vector that used to contain under_complex_step.
3. Added a System attribute "force_alloc_complex" that the user can check to find out if the vector has been force into complex allocation.
4. Set default complex step stepsize to 1e-40.
5. Reworked how the step size is handled in check_totals and check_partials so that the correct default value is used.

### DIFF
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -76,7 +76,7 @@ class ApproximationScheme(object):
         if deriv_type == 'total':
             run_model = system.run_solve_nonlinear
             results_vec = outputs
-        elif deriv_type == 'partial':
+        else:
             run_model = system.run_apply_nonlinear
             results_vec = system._residuals
 
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -79,8 +79,6 @@ class ApproximationScheme(object):
         elif deriv_type == 'partial':
             run_model = system.run_apply_nonlinear
             results_vec = system._residuals
-        else:
-            raise ValueError('deriv_type must be one of "total" or "partial"')
 
         for in_name, idxs, delta in input_deltas:
             if in_name in outputs._views_flat:
--- a/openmdao/approximation_schemes/complex_step.py
+++ b/openmdao/approximation_schemes/complex_step.py
@@ -110,18 +110,14 @@ class ComplexStep(ApproximationScheme):
 
         if deriv_type == 'total':
             current_vec = system._outputs
-        elif deriv_type == 'partial':
+        else:
             current_vec = system._residuals
 
         # Clean vector for results
         results_clone = current_vec._clone(True)
 
         # Turn on complex step.
-        for sub in system.system_iter(include_self=True, recurse=True):
-            sub.under_complex_step = True
-            sub._inputs.set_complex_step_mode(True)
-            sub._outputs.set_complex_step_mode(True)
-            sub._residuals.set_complex_step_mode(True)
+        system._set_complex_step_mode(True)
         results_clone.set_complex_step_mode(True)
 
         # To support driver src_indices, we need to override some checks in Jacobian, but do it
@@ -183,11 +179,7 @@ class ComplexStep(ApproximationScheme):
                     jac._override_checks = False
 
         # Turn off complex step.
-        for sub in system.system_iter(include_self=True, recurse=True):
-            sub.under_complex_step = False
-            sub._inputs.set_complex_step_mode(False)
-            sub._outputs.set_complex_step_mode(False)
-            sub._residuals.set_complex_step_mode(False)
+        system._set_complex_step_mode(False)
 
     def _run_point_complex(self, system, input_deltas, result_clone, deriv_type='partial'):
         """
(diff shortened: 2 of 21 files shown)


openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -79,8 +79,6 @@ def _run_point(self, system, input_deltas, out_tmp, in_tmp, result_array, deriv_
         elif deriv_type == 'partial':
             run_model = system.run_apply_nonlinear
             results_vec = system._residuals
-        else:
-            raise ValueError('deriv_type must be one of "total" or "partial"')
```
swryan: It seems like it would be more robust and wouldn't hurt to leave the ValueError in there... I know this is not normally called by the user, but it's looks like incomplete logic to not handle the case where neither the `if` nor the `elif` is true and just fall through (leaving an opening for future bugs to creep in, e.g. if someone typos the arg)...

openmdao/approximation_schemes/complex_step.py
```diff
@@ -102,23 +102,28 @@ def compute_approximations(self, system, jac=None, deriv_type='partial'):
...
+
         if jac is None:
             jac = system._jacobian
 
         if deriv_type == 'total':
             current_vec = system._outputs
         elif deriv_type == 'partial':
             current_vec = system._residuals
-        else:
-            raise ValueError('deriv_type must be one of "total" or "partial"')
```
swryan: It seems like it would be more robust and wouldn't hurt to leave the ValueError in there... I know this is not normally called by the user, but it's looks like incomplete logic to not handle the case where neither the `if` nor the `elif` is true and just fall through (leaving an opening for future bugs to creep in, e.g. if someone typos the arg)...

openmdao/approximation_schemes/complex_step.py
```diff
@@ -214,26 +221,21 @@ def _run_point_complex(self, system, input_deltas, out_tmp, result_clone, deriv_
         elif deriv_type == 'partial':
             run_model = system.run_apply_nonlinear
             results_vec = system._residuals
-        else:
-            raise ValueError('deriv_type must be one of "total" or "partial"')
```
swryan: It seems like it would be more robust and wouldn't hurt to leave the ValueError in there... I know this is not normally called by the user, but it's looks like incomplete logic to not handle the case where neither the `if` nor the `elif` is true and just fall through (leaving an opening for future bugs to creep in, e.g. if someone typos the arg)...

openmdao/vectors/vector.py
```diff
@@ -620,3 +561,23 @@ def print_variables(self):
...
+    def set_complex_step_mode(self, flag):
+        """
+        Turn on or off complex stepping mode.
+
+        When turned on, the default real ndarray is replaced with a complex ndarray and all
+        pointers are updated to point to it.
+
+        Parameters
+        ----------
+        flag : bool
```
naylor-b: I would suggest changing 'flag' to something a little less generic, maybe 'active' or something like that.

openmdao/approximation_schemes/finite_difference.py
```diff
@@ -160,15 +160,16 @@ def compute_approximations(self, system, jac=None, deriv_type='partial'):
...
+
         if jac is None:
             jac = system._jacobian
 
         if deriv_type == 'total':
             current_vec = system._outputs
         elif deriv_type == 'partial':
             current_vec = system._residuals
-        else:
-            raise ValueError('deriv_type must be one of "total" or "partial"')
```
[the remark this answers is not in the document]
Kenneth-T-Moore: I disagree, but don't feel like fighting it. I will put them back in if you put up a story for writing tests that cover all 4 of these "else" branches.

openmdao/approximation_schemes/approximation_scheme.py
Kenneth-T-Moore: see above

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: see above

openmdao/approximation_schemes/complex_step.py
Kenneth-T-Moore: see above

openmdao/approximation_schemes/finite_difference.py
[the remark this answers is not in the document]
swryan: you disagree that this leaves a hole in the logic and that the code is thus less robust?  an alternative would be to change the `elif` to an `else`, which closes the logic hole and just means you do partial if the arg is anything besides `total` (correctly spelled ;))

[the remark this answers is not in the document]
Kenneth-T-Moore: I disagree that we need to check the inputs on deep private functions.  We don't do an isinstance on jac to make sure it is of class Jacobian or None.  Typoing "total" to "ttoal" will still raise an exception.  We won't get that exception if we enclose it in else: we will get partials instead of totals.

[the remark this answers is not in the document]
Kenneth-T-Moore: I went with your second suggestion though. It is only called for total in one place.

PR #737, opened 2018-08-22
Constrained genetic algorithm

onodip:
Implementation of a penalty method for SimpleGADriver to handle constraints. Both equality and inequality constraints are supported, and the penalty method is described in the docstring of objective_callback().

The "penalty_parameter" and "penalty_exponent" parameteres of the penalty method added to the options of the driver. 

Tests added with a simple consstrained optimization example.

### DIFF
--- a/openmdao/drivers/genetic_algorithm_driver.py
+++ b/openmdao/drivers/genetic_algorithm_driver.py
@@ -253,8 +253,8 @@ class SimpleGADriver(Driver):
 
         Takes into account constraints with a penalty function.
 
-        All constraints are converted to the form of :math:`g(x)_i \leq 0` for inequality constraints and
-        :math:`h(x)_i = 0` for equality constraints.
+        All constraints are converted to the form of :math:`g(x)_i \leq 0` for
+        inequality constraints and :math:`h(x)_i = 0` for equality constraints.
         The constraint vector for inequality constraints is the following:
 
         .. math::
@@ -267,9 +267,10 @@ class SimpleGADriver(Driver):
 
         .. math::
 
-            N_g = \sum_{i=1}^N N_{g_i},  N_h = \sum_{i=1}^N N_{h_i}
+           N_g = \sum_{i=1}^N N_{g_i},  N_h = \sum_{i=1}^N N_{h_i}
 
-        The fitness function is constructed with the penalty parameter :math:`p` and the exponent :math:`\kappa`:
+        The fitness function is constructed with the penalty parameter :math:`p`
+        and the exponent :math:`\kappa`:
 
         .. math::
 
@@ -345,7 +346,8 @@ class SimpleGADriver(Driver):
                         violation = np.absolute(diff)
                     else:
                         violation = np.array([])
-                        msg = 'Specify a number for one of "lower", "upper" or "equals" for constraint "{}"'
+                        msg = ('Specify a number for one of '
+                               '"lower", "upper" or "equals" for constraint "{}"')
                         ValueError(msg.format(name))
                     constraint_violations = np.hstack((constraint_violations, violation))
                 fun = obj + penalty * sum(np.power(constraint_violations, exponent))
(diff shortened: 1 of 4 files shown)


kmarsteller:
Note, we have some PEP tests that you're failing--take a look, and I'm sure they are minor details to fix--line length and docstring syntax.  Try running testflo before you resubmit so you can be sure you got them all.  Thanks!

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -243,6 +251,37 @@ def objective_callback(self, x, icase):
         """
```
Kenneth-T-Moore: Make this a raw string, since it contains control characters.  (r""" )

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -243,6 +251,37 @@ def objective_callback(self, x, icase):
...
+           h = [h_1, h_2  \dots h_N], h_i \in R^{N_{h_i}}
+
+        The number of all constraints:
+
+        .. math::
+
+            N_g = \sum_{i=1}^N N_{g_i},  N_h = \sum_{i=1}^N N_{h_i}
+
+        The fitness function is constructed with the penalty parameter :math:`p` and the exponent :math:`\kappa`:
+
```
Kenneth-T-Moore: The Integration tests are mostly failing due to PEP standard issues. Here a few of these lines are too long.  If you take them down to under 100 it'll be fine.

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -281,6 +321,34 @@ def objective_callback(self, x, icase):
...
+                    if (con['lower'] is not None) and np.isfinite(con['lower']):
+                        diff = val - con['lower']
+                        violation = np.array([0. if d >= 0 else abs(d) for d in diff])
+                    elif (con['upper'] is not None) and np.isfinite(con['upper']):
+                        diff = val - con['upper']
+                        violation = np.array([0. if d <= 0 else abs(d) for d in diff])
+                    elif (con['equals'] is not None) and np.isfinite(con['equals']):
+                        diff = val - con['equals']
+                        violation = np.absolute(diff)
+                    else:
```
Kenneth-T-Moore: Don't think you need the else branch, since the constraint metadata should be checked during add_constraint (and if we don't check for cons without lower|upper|equals there, then we should.)

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -281,6 +322,35 @@ def objective_callback(self, x, icase):
...
+            penalty = self.options['penalty_parameter']
+            exponent = self.options['penalty_exponent']
+
+            if penalty == 0:
+                fun = obj
+            else:
+                constraint_vals = list()
+                constraint_violations = np.array([])
+                for name, val in iteritems(self.get_constraint_values()):
+                    constraint_vals.append(val)
```
Kenneth-T-Moore: Constraint vals is unused anywhere.

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -116,9 +124,9 @@ def _setup_driver(self, problem):
             msg = 'SimpleGADriver currently does not support multiple objectives.'
             raise RuntimeError(msg)
 
-        if len(self._cons) > 0:
-            msg = 'SimpleGADriver currently does not support constraints.'
-            raise RuntimeError(msg)
+        # if len(self._cons) > 0:
```
Kenneth-T-Moore: Can just delete these lines now.

swryan:
might also consider adding a bit to the GA driver documentation (docs/features/building_blocks/drivers/genetic_algorithm.rst)

openmdao/drivers/genetic_algorithm_driver.py
onodip: Thanks, changes done.

PR #741, opened 2018-08-23
P158798354 Change defaults and precedence for recording options in Drivers and Problems

hschilling:
1. Default record_desvars, record_objectives, and record_constraints to True
2. Default Includes to empty
3. Make record_desvars, record_objectives, record_constraints not affect exclusion
4. Excludes should have precedence over all other options in Driver and Problem cases
5. Update docs to better explain recording options and precedence in Driver and Problem
6. Create tests for recording option precedence

#1, #3, and #4 were already done. 

#2 was only done for Drivers

### DIFF
--- a/openmdao/recorders/tests/test_sqlite_reader.py
+++ b/openmdao/recorders/tests/test_sqlite_reader.py
@@ -914,6 +914,10 @@ class TestSqliteCaseReader(unittest.TestCase):
 
 
     def test_recording_option_precedence_driver_cases(self):
+        from openmdao.api import Problem, IndepVarComp, ExecComp, ScipyOptimizeDriver, \
+            SqliteRecorder, CaseReader
+        from openmdao.test_suite.components.paraboloid import Paraboloid
+
         prob = Problem()
         model = prob.model
         model.add_subsystem('p1', IndepVarComp('x', 50.0), promotes=['*'])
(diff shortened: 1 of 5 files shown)


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -912,6 +912,83 @@ def test_load_solver_cases(self):
 
         _assert_model_matches_case(case, model)
 
+
+    def test_recording_option_precedence_driver_cases(self):
+        prob = Problem()
```
Kenneth-T-Moore: Feature tests need all relevant imports in the body of the test. This should fail the doctest without them.

openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -912,6 +912,83 @@ def test_load_solver_cases(self):
...
+        prob.driver.options['disp'] = False
+
+        model.add_design_var('x', lower=-50.0, upper=50.0)
+        model.add_design_var('y', lower=-50.0, upper=50.0)
+        model.add_objective('f_xy')
+        model.add_constraint('c', lower=15.0)
+
+        prob.driver.add_recorder(self.recorder)
+        prob.driver.recording_options['record_desvars'] = True
+        prob.driver.recording_options['includes'] = []  # the new default according to this story
```
swryan: looks like some extraneous comments here (4 occurrences)... 
references to "the new default" and "this story" are temporal and definitely won't make sense to a reader of the docs

PR #746, opened 2018-08-28
add deprecation for 'distributed' property

swryan:


swryan:
looks like needs some further work

PR #747, opened 2018-08-28
 add deprecation for 'distributed' property

swryan:
Pivotal #160059310
Also: 
- Issue a warning instead of erroring out, if `distributed` is True but PETSc is not available
   (fall back to running non-distributed)
- cleaned up a deprecation in `parameterized` tests ('name_func' replaces 'testcase_func_name')
- changed a couple strings in tests to raw strings to address a PY3 warning
- linted test scripts that I touched (just line length and white space issues mainly)

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -106,6 +106,16 @@ class Component(System):
         self._approximated_partials = []
         self._declared_partial_checks = []
 
+    def _declare_options(self):
+        """
+        Declare options before kwargs are processed in the init method.
+        """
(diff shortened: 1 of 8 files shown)


openmdao/core/component.py
```diff
@@ -116,6 +118,34 @@ def _declare_options(self):
                              desc='True if the component has variables that are distributed '
                                   'across multiple processes.')
 
+    @property
```
JustinSGray: wow, thats a lot of code just to add a silly property...

PR #749, opened 2018-08-30
remove extraneous comment

swryan:
this comment was unintentional.. there is no version 5, the cake is a lie.


swryan:
or maddening, depending on your perspective ;)

Issue #689, opened 2018-07-10
Add support for approximation schemes for partials in MetaModelUnStructuredComp


jbussemaker:
Currently, if an output is added to a MetaModelUnStructuredComp component that uses a surrogate model that does not support analytical partials (aka did not override SurrogateModel.linearize), an optimization will always fail with the error that no Jacobian method has been defined (as is the default behavior in SurrogateModel.linearize).

I agree that surrogate models normallize *should* provide analytical gradients, but it might sometimes not be possible, for example during code testing.

To solve this, when adding an output variable, it should be tracked whether the surrogate model supports analytical gradients. This can for example be done using the overrides_method utils function (in openmdao.utils.class_util). Then, the _setup_partials function must be modified to support approximation schemes for the relevant output variable partials.

JustinSGray:
solved by #754. Will make it into version 2.5

PR #754, opened 2018-09-04
P159492538 Modify MetaModelUnstructured to use FD if surrogate model derivatives aren't available. 

hschilling:
Based on github issue https://github.com/OpenMDAO/OpenMDAO/issues/689

Also fixed some typos seen along the way


openmdao/components/tests/test_meta_model_unstructured_comp.py
```diff
@@ -929,5 +927,54 @@ def test_metamodel_deprecated(self):
...
+            def setup(self):
+
+                surrogate = FloatKrigingSurrogate()
+                self.add_input('x', 0.,
+                               training_data=np.linspace(0,10,20))
+                self.add_output('sin_x', 0.,
+                                surrogate=surrogate,
+                                training_data=.5*np.sin(np.linspace(0,10,20)))
+
+                self.declare_partials(of='sin_x', wrt='x', method='fd')
```
Kenneth-T-Moore: The `declare_partials` with method='fd' needs to be called automatically instead of making the user call it.

hschilling:
fixed

openmdao/components/tests/test_meta_model_unstructured_comp.py
```diff
@@ -929,5 +927,52 @@ def test_metamodel_deprecated(self):
         prob.run_model()
         assert_rel_error(self, prob['mm.y1'], 1.5, 1e-2)
 
+    def test_metamodel_use_fd_if_no_surrogate_linearize(self):
```
Kenneth-T-Moore: How does the component 'fd' itself when there isn't any code in the metamodel to declare that all derivs are to be calculated with the fd method? I think this exposes a hole in your test, and it is possible that either the function delete isn't working, or it takes more than just that to clean up some existing derivatives stuff (probably the partials dictionary needs to be cleaned out), but more pressing is fixing the test so that it doesn't give a false pass.

For the second half of the test, can you rebuild the test from scratch with a new Problem instance? That'll clear up any ambiguity about what is going on here.

hschilling: The method delete is definitely working. We should probably talk through this.

openmdao/docs/features/building_blocks/components/metamodelunstructured_comp.rst
```diff
@@ -86,4 +86,24 @@ to be the same size as specified in the ``vec_size`` argument.
...
+Using Surrogates That Not Define Linearize Method
+-------------------------------------------------
+
+In some cases, users might define surrogates but not define a `linearize` method. In this case, the
+`MetaModelUnStructuredComp` derivatives will be computed using finite differences for the output variables that use that
+surrogate. By default, the default values for the finite differencing method will be used.
+
+If the user would like to specify finite differencing options, they can do so by calling the `declare_partials`
+method in the component's `setup` or `configure` methods. This example, which uses a simplified surrogate with no
+`linearize` method and no training, shows `declare_partials` called in `setup`.
```
Kenneth-T-Moore: -> "component's setup method or in a parent group's configure method"

openmdao/components/meta_model_unstructured_comp.py
```diff
@@ -218,12 +221,48 @@ def _setup_partials(self, recurse=True):
...
+                       "options because they were not explicitly declared.\n".format(self.name)
+                msg += "The derivatives computed using the defaults are:\n"
+                for abs_key in non_declared_partials:
+                    msg += "    {}, {}\n".format(*abs_key)
+                simple_warning(msg, RuntimeWarning)
+
+            for out_name, out_shape in self._surrogate_output_names:
+                surrogate = self._metadata(out_name).get('surrogate')
+                if surrogate and not overrides_method('linearize', surrogate, SurrogateModel):
+                    self._approx_partials(of=out_name,
```
Kenneth-T-Moore: If the user declares a partial, and it is in the set of declared_partials, doesn't this code re-declare it by calling self._approx_partials on it? I think you want to exclude the surrogates for which partials have already been declared by checking against `declared_partials` or `non_declared_partials`.   To do this, you will need a test with a really tiny input (so something like f(x) - 1e-9 * sin( 1e-9 * x)), for which the default value will do a horrible job. Then in the test, specify a size that will work for this, like 1e-14.  This test should fail if the default size overrides the requested size. That way you will be certain.

Kenneth-T-Moore: edited the equation for the poor fd sine

Kenneth-T-Moore: Test is actually covered, and fd seems to address my concern natively.

openmdao/docs/features/building_blocks/components/metamodelunstructured_comp.rst
hschilling: fixed

openmdao/docs/features/building_blocks/components/metamodelunstructured_comp.rst
```diff
@@ -86,4 +86,24 @@ to be the same size as specified in the ``vec_size`` argument.
     openmdao.components.tests.test_meta_model_unstructured_comp.MetaModelTestCase.test_metamodel_feature_vector2d
     :layout: code, output
 
+
+Using Surrogates That Not Define Linearize Method
```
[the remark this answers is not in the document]
hschilling: fixed

openmdao/docs/features/building_blocks/components/metamodelunstructured_comp.rst
```diff
@@ -86,4 +86,24 @@ to be the same size as specified in the ``vec_size`` argument.
     openmdao.components.tests.test_meta_model_unstructured_comp.MetaModelTestCase.test_metamodel_feature_vector2d
     :layout: code, output
 
+
+Using Surrogates That Do Not Define Linearize Method
+----------------------------------------------------
+
+In some cases, users might define surrogates but not define a `linearize` method. In this case, the
```
swryan: In general, I think that the docs address the user (the reader) rather than referring to them in the third person.  (e.g. the intro to this section: "MetaModelUnStructuredComp lets you quickly create a component").  Suggest replacing references to "users" here and "the user" below with the pronoun "you" for consistency. (also removing the "user" reference in the embedded test)

hschilling: All comments by swryan addressed

PR #756, opened 2018-09-07
Add initial design values to SimpleGADriver

madsmpedersen:
Add initial design variable values as first individual of first generation in SimpleGADriver

### DIFF
--- a/openmdao/drivers/genetic_algorithm_driver.py
+++ b/openmdao/drivers/genetic_algorithm_driver.py
@@ -219,6 +219,16 @@ class SimpleGADriver(Driver):
             lower_bound[i:j] = meta['lower']
             upper_bound[i:j] = meta['upper']
 
+        # Initial Design Vars
+        x0 = np.empty(count)
+        desvar_vals = self.get_design_var_values()
+        i = 0
+        for name, meta in iteritems(self._designvars):
+            size = meta['size']
+            x0[i:i + size] = desvar_vals[name]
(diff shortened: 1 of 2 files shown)


openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -219,6 +219,16 @@ def run(self):
...
+        # Initial Design Vars
+        x0 = np.empty(count)
+        desvar_vals = self.get_design_var_values()
+        i = 0
+        for name, meta in iteritems(self._designvars):
+            size = meta['size']
+            x0[i:i + size] = desvar_vals[name]
+            i += size
+
+
```
Kenneth-T-Moore: You can move this into the previous loop with the cached indices so that you don't have to sum the sizes again. So it'll look like x0[i:j] = desvar_vals[name].

Looks good otherwise. You might need to remove the extra carriage return here to pass the lint tests.

madsmpedersen:
God point - I have now implemented it.

In the ´test_two_branin_parallel_model´ test, the random population initialization was quite lucky, and therefore this test failed when new_gen[0] was overwritten with the initial independen_variable values.
The test passed if I increased the max_gen value from 10 to 20, but instead I changed the initial independent_variable value to a more lucky value (7.5 to 5)

Kenneth-T-Moore:
@madsmpedersen : Ok, I think that change to the test initial point is fine.

PR #760, opened 2018-09-10
Made the float value used for 'infinite' bounds smaller.

naylor-b:
- new value is 1e30
- created a global variable called openmdao.INF_BOUND for the new 'infinite' value

### DIFF
--- a/openmdao/__init__.py
+++ b/openmdao/__init__.py
@@ -1,1 +1,3 @@
 __version__ = '2.4.0'
+
+INF_BOUND = 1.0E30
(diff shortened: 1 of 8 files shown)


openmdao/__init__.py
```diff
@@ -1 +1,3 @@
 __version__ = '2.4.0'
+
```
Kenneth-T-Moore: Not sure how to feel about putting stuff in the __init__, since we have avoided doing it up to this point. What about someting like an opendmao.settings?

naylor-b: I think putting constants in __init__ is fine.  The problems come when you import things in __init__.

PR #766, opened 2018-09-13
support for simul coloring under MPI

naylor-b:


openmdao/recorders/tests/test_sqlite_recorder.py
```diff
@@ -1298,7 +1298,7 @@ def test_record_solver_linear_block_gs(self):
             'pz.z': [0., 0.],
             'd1.y1': [-9.10908321e-11],
             'd2.y2': [0.],
-            'obj_cmp.obj': [2.03287907e-20],
+            'obj_cmp.obj': [2.28698896e-20],
```
JustinSGray: I guess the number changed a bit?

naylor-b: Yeah, but I wasn't too worried about it since the number is so small.  Do you want me to spend the time to dig into it?

JustinSGray: no. I hadn't noticed that it was 1e-20 at first. Thats totally fine.

PR #769, opened 2018-09-17
added vectorize init arg to ExecComp

naylor-b:
- this change was done to support later PR that uses declared partials sparsity for coloring.
- changed default complex stepsize to 1e-40
- changed default variable value from 0.0 to 1.0

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -44,11 +44,15 @@ class ExecComp(ExplicitComponent):
         List of expressions.
     _codes : list
         List of code objects.
+    _vectorize : bool
+        If True, treat all array/array partials as diagonal if both arrays have size > 1.
+        All arrays with size > 1 must have the same flattened size or an exception will be raised.
     complex_stepsize : double
         Step size used for complex step which is used for derivatives.
+
     """
 
-    def __init__(self, exprs, **kwargs):
+    def __init__(self, exprs, vectorize=False, **kwargs):
         r"""
         Create a <Component> using only an expression string.
 
@@ -120,6 +124,11 @@ class ExecComp(ExplicitComponent):
             standard Python operators, a subset of numpy and scipy functions
             is supported.
 
+        vectorize : bool
+            If True, treat all array/array partials as diagonal if both arrays have size > 1.
+            All arrays with size > 1 must have the same flattened size or an exception will be
+            raised.
+
         **kwargs : dict of named args
             Initial values of variables can be set by setting a named
             arg with the var name.  If the value is a dict it is assumed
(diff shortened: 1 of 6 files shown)


openmdao/components/exec_comp.py
```diff
@@ -131,7 +140,7 @@ def __init__(self, exprs, **kwargs):
         -----
         If a variable has an initial value that is anything other than 0.0,
```
Kenneth-T-Moore: Should this be reworded with the initial value change? "If a variable has an initial value that is anything other than  0.0 ..."

openmdao/test_suite/components/matmultcomp.py
```diff
@@ -45,11 +45,6 @@ def compute(self, inputs, outputs):
         time.sleep(self.sleep_time)
```
Kenneth-T-Moore: Do we even need the MatMultComp anymore?

openmdao/components/exec_comp.py
```diff
@@ -342,16 +367,33 @@ def compute_partials(self, inputs, partials):
...
+                    pwrap[param][idx] += step
+
+                    uwrap = _TmpDict(self._outputs, return_complex=True)
+
+                    # solve with complex param value
+                    self._residuals.set_const(0.0)
+                    self.compute(pwrap, uwrap)
+
+                    for u in out_names:
+                        jval = imag(uwrap[u] / self.complex_stepsize)
```
Kenneth-T-Moore: If uwrap[u] is an ndarry, then multiplying by the scalar (1.0 / complex_stepsize) is way faster than division.

openmdao/components/exec_comp.py
```diff
@@ -342,16 +367,33 @@ def compute_partials(self, inputs, partials):
...
+                        if (u, param) not in partials:  # create the dict entry
+                            partials[(u, param)] = np.zeros((jval.size, psize))
```
Kenneth-T-Moore: Whoa, we never have to create a dictionary entry in the jacobian inside of a component's `compute_partials`. Why do we have to do it in execcomp? (also, does coverage confirm that this gets called?  [both branches of if]

openmdao/test_suite/components/matmultcomp.py
naylor-b: Yeah it's still used for parallel FD testing.

openmdao/components/exec_comp.py
naylor-b: fixed

openmdao/components/exec_comp.py
naylor-b: fixed

openmdao/components/exec_comp.py
naylor-b: fixed

PR #771, opened 2018-09-18
Use setuptools instead of distutils

felixonmars:
install_requires and entry_points are actually attributes of setuptools, so installing with "setup.py install" won't work correctly without the change.


kmarsteller:
build caches deleted before merge to main, here we go.

PR #772, opened 2018-09-19
updated partial jacobian randomization during determination of total jac sparsity

naylor-b:
- also fixed a bug when using assembled jacobians on models having the same source connected to multiple inputs on the same component (dense assembled jac only - we don't allow it for sparse assembled jacs at all).

### DIFF
--- a/openmdao/core/explicitcomponent.py
+++ b/openmdao/core/explicitcomponent.py
@@ -344,7 +344,7 @@ class ExplicitComponent(Component):
         Parameters
         ----------
         jac : Jacobian or None
-            If None, use local jacobian, else use assembled jacobian jac.
+            Ignored.
         sub_do_ln : boolean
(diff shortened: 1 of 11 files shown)


openmdao/core/explicitcomponent.py
```diff
@@ -344,7 +344,7 @@ def _linearize(self, jac=None, sub_do_ln=False):
         Parameters
         ----------
         jac : Jacobian or None
-            If None, use local jacobian, else use assembled jacobian jac.
+            Ignored.
```
JustinSGray: im guessing the argument is needed for implicit comps then? otherwise why not just remove it?

naylor-b: It is used for implicit comps, but even if it weren't we'd still need it because it's needed by groups.

PR #776, opened 2018-09-25
added a setup arg to turn off derivative allocations

naylor-b:
- also updated the 'openmdao mem' memory tracer so it now shows the memory allocations in call tree format.


openmdao/core/problem.py
```diff
@@ -935,6 +938,9 @@ def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes
 
         model = self.model
 
+        if not model._use_derivatives:
```
Kenneth-T-Moore: Probably should add this check to `check_totals` as well.

naylor-b: Calling check_totals will call compute_totals, which raises an exception saying check_totals can't be called because derivs are turned off.

PR #779, opened 2018-10-04
P160160211 sparsity diagram

hschilling:
User can generate a sparsity diagram/heat-map of the FD and Analytic using the data returned by check_partials


openmdao/components/tests/test_bsplines_comp.py
```diff
@@ -7,10 +7,7 @@
 
 import numpy as np
 
-try:
-    import matplotlib
-except ImportError:
-    matplotlib = None
+import matplotlib
```
Kenneth-T-Moore: Don't think we want to list this as a dependency, so you should go back to the skip test here, and add it to the new test. You could probably make the `paratial_deriv_plot` instruct the user to install matplotlib if it detects it didn't import.

PR #780, opened 2018-10-04
support for discrete variable (pass-by-object) data passing

naylor-b:


openmdao/core/component.py
```diff
@@ -273,6 +268,18 @@ def _setup_var_data(self, recurse=True):
...
                 abs2meta[abs_name] = metadata
 
+            for prom_name, val in iteritems(self._var_discrete[type_]):
+                abs_name = prefix + prom_name
+                allprocs_prom2abs_list[type_][prom_name] = [abs_name]
+                self._var_allprocs_discrete[type_][abs_name] = val
+
+        self._var_abs_names = self._var_allprocs_abs_names
+        if self._var_discrete['input'] or self._var_discrete['output']:
+            self._discrete_inputs = _DictValues(self._var_discrete['input'])
```
Kenneth-T-Moore: Is there a docstring for _discrete_inputs and outputs somewhere?

openmdao/core/explicitcomponent.py
```diff
@@ -178,7 +179,11 @@ def _apply_nonlinear(self):
 
                 self._inputs.read_only = True
                 try:
-                    self.compute(self._inputs, outputs)
+                    if self._discrete_inputs or self._discrete_outputs:
```
Kenneth-T-Moore: Isn't there a way to pack the discrete vars into the `inputs` and `outputs` dict-like objects that the user sees, so that the api for the call doesn't need additional arguments?

openmdao/core/group.py
```diff
@@ -398,6 +405,8 @@ def _setup_procs(self, pathname, comm, mode):
         # build a list of local subgroups to speed up later loops
         self._subgroups_myproc = [s for s in self._subsystems_myproc if isinstance(s, Group)]
 
+        self._loc_subsys_map = {s.name: s for s in self._subsystems_myproc}
```
Kenneth-T-Moore: This variable doesn't seem to be documented either.

openmdao/solvers/linear/direct.py
```diff
@@ -131,11 +131,10 @@ def format_nan_error(system, matrix):
     # Because of how we built the matrix, a NaN in a comp cause the whole row to be NaN, so we
     # need to associate each index with a variable.
     varname = []
-    all_vars = system._var_allprocs_abs_names['output']
     varsizes = system._var_sizes['nonlinear']['output']
     for row in rows:
         n = 0
-        for j, name in enumerate(all_vars):
+        for j, name in enumerate(system._var_allprocs_abs_names['output']):
             n += varsizes[system._owning_rank[name]][j]
```
Kenneth-T-Moore: I don't understand this change. Is this more efficient?

openmdao/solvers/nonlinear/newton.py
```diff
@@ -100,6 +100,11 @@ def _setup_solvers(self, system, depth):
         """
         super(NewtonSolver, self)._setup_solvers(system, depth)
 
+        # check for any discrete variables
+        if system._var_allprocs_discrete['output']:
+            raise RuntimeError("System '%s' has a Newton solver and contains discrete outputs %s." %
```
Kenneth-T-Moore: I think newton could work on a system that contains discrete outputs, but only if that output doesn't connect to an input of a component that is in that system. (i.e., it could be pushing a discrete variable to a component in a higher group after this group converges, which would be fine conceptually.)

It would also work for all cases if the sending and receiving component are both under an 'fd' group under the Newton's group.

Kenneth-T-Moore: Also, anything we do to Newton also applies to Broyden, our other solver that uses derivatives.

openmdao/core/explicitcomponent.py
naylor-b: We would have to make some sort of dict wrapper to contain both discrete and non-discrete.  Justin and I discussed this as an option but the decision was made to go with just having a different arg list for components that actually use discrete vars.

openmdao/solvers/nonlinear/newton.py
naylor-b: Jubs says *any* discrete output in the system being solved by Newton/Broyden is invalid, and being under an fd group doesn't make it valid either.

naylor-b: I added a check and a test for broyden.

openmdao/core/group.py

openmdao/core/component.py
naylor-b: fixed

PR #787, opened 2018-10-17
Fix for MPI bug in GA by padding with dummy cases to make them divide evenly into number of procs. Fixes hang on benchmarks.

Kenneth-T-Moore:


openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -562,6 +562,13 @@ def execute_ga(self, x0, vlb, vub, vob, bits, pop_size, max_gen, random_state, P
                 cases = [((item, ii), None) for ii, item in enumerate(x_pop)
                          if np.all(item - vob <= 0)]
 
+                # Pad the cases with some dummy cases to make the cases divisible amongst the procs.
+                # TODO: Add a load balancing option to this driver.
+                extra = len(cases) % comm.size
+                if extra > 0:
+                    for j in range(comm.size - extra):
+                        cases.append(cases[-1])
+
```
swryan: could we get a quick test mirroring the benchmark case?

PR #790, opened 2018-10-22
Fixed incorrect dot product in parallel on PETSc vector classes.

timryanb:


kmarsteller:
@timryanb it looks like you have tripped a couple of our very particular pep 8 tests. If we could trouble you to take a look at those and make a couple of minor changes to whitespace, etc. (as itemized in the failure messages on appveyor), then you will pass our CI and we can approve your PR.  Thanks!

PR #796, opened 2018-10-30
Refactor of CaseReader **Backwards incompatible**

swryan:
Pivotal #160117632

Also: rewrite case_recording docs (Pivotal #160117673)


kmarsteller:
Tests finally passing, just need one more review.

openmdao/recorders/sqlite_recorder.py
```diff
@@ -26,18 +25,20 @@
 
 
 """
-SQL case output format version history.
----------------------------------------
+SQL case database version history.
+----------------------------------
+5 -- OpenMDAO 2.5
```
Kenneth-T-Moore: There is so much here that I can't really tell, but did you add a test that verifies we can read a version 4 recording?

swryan: Good point... I did not.   Will cook up a new test.

swryan: done

PR #798, opened 2018-10-30
Fix deprecation warning in units.py

swryan:
Pivotal #161297707

### DIFF
--- a/openmdao/utils/units.py
+++ b/openmdao/utils/units.py
@@ -12,9 +12,12 @@ Justin Gray.
 from __future__ import division, print_function
 
 
+import sys
+
 import re
 import os.path
 from collections import OrderedDict
+
 from six import iteritems
 from six.moves import zip
 from six.moves.configparser import RawConfigParser as ConfigParser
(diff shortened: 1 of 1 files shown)


openmdao/utils/units.py
```diff
@@ -715,7 +718,13 @@ def import_library(libfilepointer):
     _UNIT_CACHE = {}
     _UNIT_LIB = ConfigParser()
     _UNIT_LIB.optionxform = _do_nothing
-    _UNIT_LIB.readfp(libfilepointer)
+
+    # New in Python 3.2: read_file() replaces readfp().
+    if sys.version_info[0] >= 3 and sys.version_info[1] >= 2:
```
42B:
```python
if sys.version_info >= (3, 2):  # ?
```

openmdao/utils/units.py
```diff
@@ -769,7 +778,13 @@ def update_library(filename):
     try:
         cfg = ConfigParser()
         cfg.optionxform = _do_nothing
-        cfg.readfp(inp)
+
+        # New in Python 3.2: read_file() replaces readfp().
+        if sys.version_info[0] >= 3 and sys.version_info[1] >= 2:
```
42B:
```python
if sys.version_info >= (3, 2):  # ?
```

openmdao/utils/units.py
swryan: better, thanks!

42B: @swryan 

No problem!

PR #800, opened 2018-11-01
New "trust-constr" optimizer in the ScipyOptimizeDriver

onodip:
Added "trust-constr" optimizer to the supported optimizers. 
Implemented a new type of constraint input in `run()` for `scipy.optimize.minimize`, because this algorithm does not support the dictionary format to pass the constraints.
Added tests with equality, inequality and double-sided constraints.

Bounds are not supported yet, because there are unresolved issues with the bounds of the "trust-constr" in SciPy.

Added exception handling to the "fail" attribute of the driver, because apparently  not all optimizers return a success flag in `OptimizeResult`.

### DIFF
--- a/openmdao/drivers/scipy_optimizer.py
+++ b/openmdao/drivers/scipy_optimizer.py
@@ -595,10 +595,10 @@ class ScipyOptimizeDriver(Driver):
 
 def signature_extender(fcn, extra_args):
     """
-    A closure function, which appends extra arguments to the original function call.
-    The first argument is the design vector.
-    The possible extra arguments from the callback of :func:`scipy.optimize.minimize` are not
-    passed to the function.
+    Closure function, which appends extra arguments to the original function call.
+
+    The first argument is the design vector. The possible extra arguments from the callback
+    of :func:`scipy.optimize.minimize` are not passed to the function.
 
     Some algorithms take a sequence of :class:`~scipy.optimize.NonlinearConstraint` as input
     for the constraints. For this class it is not possible to pass additional arguments.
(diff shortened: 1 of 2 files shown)


JustinSGray:
I don't think we'll make 1.1 a hard requirement in the near future. There are a lot of systems where supporting older numpy and scipy makes it easier for users (like HPC systems where they might not have control over it). 

Could you make this feature conditional on the availability of the optimizer? Only add the option if scipy version is 1.1 or greater?

onodip:
Is any plan to calculate approximate Hessian or Hessian vector product in OpenMDAO? For now this optimizer  uses BFGS from the `scipy.optimize` module for the Hessians.

JustinSGray:
No current near-term plans to add in hessian calculations. First theoretical developments would be needed to extend the unified derivative equations to 2nd derivatives. Then a new implementation would be needed to support the new theory. 

We'd welcome any outside contributions on this front, though I would suggest some significant up-front planning before any implementation started. Absent any outside contributions, its not on our current road-map because we don't have any use for it. If one comes up, that might change the priority. Until then... its just something on the distant horizon.

openmdao/drivers/scipy_optimizer.py
```diff
@@ -319,6 +352,12 @@ def run(self):
         else:
             jac = None
 
+        if opt in _hessian_optimizers:
+            from scipy.optimize import BFGS  # FIXME temporary, should be option. (P.O.)
```
JustinSGray: The #FIXME here is related to your question on hessian computation, right? So this is really a TODO in the future if we supported Hessians?

onodip: indeed, it should be a TODO

PR #801, opened 2018-11-02
run feature test in temp dir

swryan:


kmarsteller:
Fail is not related to the PR.

PR #803, opened 2018-11-05
new declaration of two extra COBYLA options in ScipyOptimizeDriver dr…

emrebar:
ScipyOptimize-COBYLA's default first perturbation step for design variables is 1. This is enforced even though the values are outside upper and lower bounds. COBYLA has an option ''rhobeg'' to adjust this. 
I added this and constraint tolerance options to the ScipyOptimizeDriver driver class. So that these are accessible from the "highest level" in the hierarchy i.e. prob.driver.options['initstep'] = 0.1 etc.

### DIFF
--- a/openmdao/drivers/scipy_optimizer.py
+++ b/openmdao/drivers/scipy_optimizer.py
@@ -220,10 +220,10 @@ class ScipyOptimizeDriver(Driver):
         # maxiter and disp get passsed into scipy with all the other options.
         self.opt_settings['maxiter'] = self.options['maxiter']
         self.opt_settings['disp'] = self.options['disp']
+        # COBYLA specific options passed into scipy
         if opt == 'COBYLA':
             self.opt_settings['rhobeg'] = self.options['initstep']
             self.opt_settings['catol'] = self.options['catol_cob']
-
 
         # Size Problem
         nparam = 0
--- a/openmdao/drivers/scipy_optimizer.py
+++ b/openmdao/drivers/scipy_optimizer.py
@@ -134,6 +134,10 @@ class ScipyOptimizeDriver(Driver):
         self.options.declare('dynamic_derivs_repeats', default=3, types=int,
                              desc='Number of compute_totals calls during dynamic computation of '
                                   'simultaneous derivative coloring')
+        self.options.declare('initstep', default=1., types=float,
+                             desc='COBYLA: Reasonable initial changes to the variables')
+        self.options.declare('catol_cob', default=0.0002, types=float,
+                             desc='COBYLA: Tolerance (absolute) for constraint violations')
 
     def _get_name(self):
         """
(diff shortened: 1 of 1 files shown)


openmdao/drivers/scipy_optimizer.py
```diff
@@ -216,6 +220,10 @@ def run(self):
         # maxiter and disp get passsed into scipy with all the other options.
         self.opt_settings['maxiter'] = self.options['maxiter']
         self.opt_settings['disp'] = self.options['disp']
+        if opt == 'COBYLA':
+            self.opt_settings['rhobeg'] = self.options['initstep']
+            self.opt_settings['catol'] = self.options['catol_cob']
```
onodip: The formatting is bad here, delete one of the blank lines.

emrebar: thanks. i corrected this. I cant seem to be able to assign reviewers. (which I assume, has to be done by me not by the developers?)

Kenneth-T-Moore:
Hey, emrebar.  Thanks for the pull request. We have been talking about this some, and are not sure that these options are needed, because there is already an alternate way for the user to set them:

    `prob.driver.opt_settings['rhobeg'] = my_value`

That was the intention when creating the `opt_settings` attribute, because potentially every optimizer can have a ton of settings. We do this in the pyoptsparse driver too, because the fortran optimizers tend to have a lot of settings (SNOPT has dozens.)

I think the one thing gained here is that it internalizes to OpenMDAO some documentation of the settings for Cobyla, though it gets trickier if scipy changes the call attributes (like it did with gmres), so I don't know. But yeah, we are still debating this idea.

emrebar:
> Hey, emrebar. Thanks for the pull request. We have been talking about this some, and are not sure that these options are needed, because there is already an alternate way for the user to set them:
> 
> ```
> `prob.driver.opt_settings['rhobeg'] = my_value`
> ```
> That was the intention when creating the `opt_settings` attribute, because potentially every optimizer can have a ton of settings. We do this in the pyoptsparse driver too, because the fortran optimizers tend to have a lot of settings (SNOPT has dozens.)
> 
> I think the one thing gained here is that it internalizes to OpenMDAO some documentation of the settings for Cobyla, though it gets trickier if scipy changes the call attributes (like it did with gmres), so I don't know. But yeah, we are still debating this idea.

I was not aware of the possibility of accessing this option via driver.opt_settings (which, I now see is clear from the code and also in the pyoptsparse driver manual pages). Then I agree this change is redundant. But maybe it would be  nice to mention this option more explicitly in the user manual. (features-drivers).

Kenneth-T-Moore:
>  Then I agree this change is redundant. But maybe it would be nice to mention this option more explicitly in the user manual. (features-drivers).

Yeah, I think we definitely need to improve the documentation for this feature.

emrebar:
> > Then I agree this change is redundant. But maybe it would be nice to mention this option more explicitly in the user manual. (features-drivers).
> 
> Yeah, I think we definitely need to improve the documentation for this feature.

Thanks a lot for the comments, I am closing the pull request.

PR #804, opened 2018-11-05
Added value for atol to preserve legacy behavior and eliminate warnings

hschilling:


openmdao/solvers/linear/scipy_iter_solver.py
```diff
@@ -241,7 +241,7 @@ def solve(self, vec_names, mode, rel_systems=None):
             self._iter_count = 0
             if solver is gmres:
                 x, info = solver(linop, b_vec._data.copy(), M=M, restart=restart,
-                                 x0=x_vec_combined, maxiter=maxiter, tol=atol,
+                                 x0=x_vec_combined, maxiter=maxiter, tol=atol, atol='legacy',
```
Kenneth-T-Moore: Is "legacy" on option on older scipy (the ones we currently test with)?  If not, then we will need to only do this when the scipy is 1.1.

hschilling: I will check!

PR #808, opened 2018-11-09
Update Dependencies

sanbales:
This pull request is to help with building a conda recipe for **OpenMDAO** that removes the current dependencies that are not required for running **OpenMDAO** and provides them as `extras` that can be installed only if needed.

* Broke apart dependencies not necessary for running **OpenMDAO** into `develop` and `docs` in `setup.py`.
* Rewrote `README.md` and added instructions for installing the new optional dependencies and for testing a local install of **OpenMDAO**.
* Removed `exec` call for defining `__version__` in `setup.py`.
* Removed unnecessary imports in `conf.py`.
* Added missing auto-generated files to `.gitignore`.
* Updated `appveyor.yml` and `.travis.yml` to install optional dependencies and cleaned their installation scripts to make them more concise and consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ openmdao/docs/_srcdocs/index.rst
 openmdao/docs/tags/
 *.out
 doc_plot_*
+coloring.json
+connections.html
+rank*.dat
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ openmdao/docs/_srcdocs/usr/
 openmdao/docs/_srcdocs/dev/
 openmdao/docs/_srcdocs/packages/
 openmdao/docs/_srcdocs/index.rst
-*testflo_report.out
 openmdao/docs/tags/
 *.out
 doc_plot_*
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,9 +92,10 @@ install:
     conda create --yes -n PY$PY python=$PY;
     source $HOME/miniconda/bin/activate PY$PY;
 
-    conda install --yes numpy=$NUMPY scipy=$SCIPY coveralls cython matplotlib swig;
+    conda install --yes numpy=$NUMPY scipy=$SCIPY cython matplotlib swig;
 
     pip install --upgrade pip;
+    pip install git+https://github.com/swryan/coveralls-python@work;
 
     git clone https://github.com/OpenMDAO/pyoptsparse.git;
     cd pyoptsparse;
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,9 @@ install:
     conda install --yes numpy=$NUMPY scipy=$SCIPY cython matplotlib swig;
 
     pip install --upgrade pip;
+    # Install customized packages
     pip install git+https://github.com/swryan/coveralls-python@work;
+    pip install git+https://github.com/OpenMDAO/testflo.git;
 
     git clone https://github.com/OpenMDAO/pyoptsparse.git;
     cd pyoptsparse;
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,45 +79,59 @@ install:
 
 # if we don't have a cached conda environment then build one, otherwise just activate the cached one
 - if [ "$CACHED_ENV" ]; then
-    echo "Using cached environment..."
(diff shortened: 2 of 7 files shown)


JustinSGray:
the dev team has talked about this. We think testflo should be a required dependency, so users can actually run the tests to check their install. In fact, we think the install docs should be modified to include explicit instructions to run testflo. 

As for the other stuff, we're open to this idea... but we question if its really necessary. Could you tell us why you want a narrower set of requires packages?

sanbales:
Thank you for the response, Justin.

_TL;DR: motivation is to keep environments cleaner and leaner for those that are only interesting in using OpenMDAO, and make it easier to create a conda package for OpenMDAO._

The reason for asking for a more specific breakdown is that people can install what they need as they need it.  For someone that just wants to `pip install openmdao` and is not interested in building the docs or running the tests, they "just" need to get `numpy`, `scipy`, and `networkx`.  These are big dependencies, but they are also well supported and you can readily get them from conda and other places.  When you start talking about baron and other things that are only required to build the docs, you start running into some issues.  Our main issue right now is that if you are going to package OpenMDAO with conda, you need to have conda packages for them.  You can still pip install things for running the tests before you build the packages, but when someone wants to `conda install openmdao` you want that to only depend on what you need to run openmdao, not what you need to build it.

The benefit of this approach is that people can still get their development / doc / (something else?) environment easily set up by `pip install openmdao[docs]` or `pip install openmdao[develop]`.  I would suggest that the testing makes sense under the `[develop]` option as it is a critical element of that activity.  Alternatively, if you want to keep testing as it's own thing we could put the testing dependencies under `[test]` and have `[develop]` be for a combination of test and docs.  I would suggest following what other packages are doing, but we are not restricted in how these optional dependencies are organized.  My two cents are that I agree that `testflo` is not a big dependency, but if it is not hard to install it through the optional dependency, why include it for everyone that just wants to use OpenMDAO without running the tests?

We can live with the current approach, but in the spirit of full disclosure, our main and most immediate motivation is that we want to build a conda-forge recipe for OpenMDAO so we can start using it with `anaconda-project` and `constructor`, but having all these `pip` only dependencies it highly complicates things.  Also, most of our users are just interested in just getting the packages required to run OpenMDAO, not necessarily build the docs or running the tests, so having fewer dependencies keeps their environments cleaner and leaner.

I can add instructions for testing to the readme, in comparable level of detail to the ones for building the docs. If you want more in-depth explanations for the docs, I can't make any promises, but would be willing to take a hack at it.

JustinSGray:
i see your argument. On the flip side of it though, if a user installs openmdao but can't run `testflo openmdao` then how do they know its working? 

I guess you might counter that with "they can then pip install testflo, if they want to". 

Is there an alternative approach where the default is a larger install, but you can optionally get a narrower dependency path? If not Im inclined to agree with this PR subject to a few minor revisions and some internal testing.

sanbales:
I see what you mean, but one could argue that by importing `openmdao` and running the examples or your own problems you will know if it is working.  Granted, it could be producing results but failing some of the tests, but as you mention, if you want to be super rigorous about it, you can install the testing dependencies and run them.  I've added instructions in the `README.md` for running the tests.  Maybe went a bit overboard and almost completely overhauled the readme.  Hope it goes without saying, but I won't be offended if you want to roll back those changes.

Regarding the question about having the default install include all the dependencies, but have an option for installing only the "runtime" dependencies, I don't think it is possible with `setuptools`.  There may be a way to do this, but I'm not aware of how we could go about it.

Going back to the first point though, when (if?) we can create a conda recipe for OpenMDAO, it would create conda packages for the various operating systems and numpy/scipy versions, and it would only create them if they pass the tests, so when people install it that way, OpenMDAO will work in the system the same way it worked in the environment that built the package and ran the tests.  Maybe I should say: "they should work", but I'm fairly confident that that is the case, after all that is the strength of `conda`.  I have to defer to @bollwyvl for any insight on this.

swryan:
@sanbales   see PR #814   Thanks!

bollwyvl:
Awesome stuff!

>

sanbales:
Thank you, OpenMDAO team, very much appreciate you accepting these changes.  I will remove this PR.

PR #812, opened 2018-11-23
ScipyOptimizeDriver Hessian as user option

onodip:
* Hessian calculation method can be a user option (with `opt_settings`) for optimizers which support this.
* Test added for setting the Hessian approximation with `opt_settings`
* Put some attributes into the `__init__`, which where not initialized
* Added 'linear' to COBYLA's bound constraints, this key was missing. It did not result in an error only, because COBYLA is not a gradient optimizer (but this feature to turn bounds into constraints could be reused to any optimizer, which supports constraints but not bounds)
* Some minor fixes in the tests and in the driver
* Some docs to the previous 'trust-constr' PR.

### DIFF
--- a/openmdao/drivers/tests/test_scipy_optimizer.py
+++ b/openmdao/drivers/tests/test_scipy_optimizer.py
@@ -903,20 +903,74 @@ class TestScipyOptimizeDriver(unittest.TestCase):
         x0 = np.array([1.2, 0.8, 1.3])
 
         prob = Problem()
+        model = prob.model
         indeps = prob.model.add_subsystem('indeps', IndepVarComp(problem=prob), promotes=['*'])
         indeps.add_output('x', list(x0))
 
         prob.model.add_subsystem('rosen', Rosenbrock(problem=prob), promotes=['*'])
         prob.model.add_subsystem('con', ExecComp('c=sum(x)', x=np.ones(3)), promotes=['*'])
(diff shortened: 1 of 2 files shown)


openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -985,6 +996,52 @@ def __init__(self, problem):
...
+
+    @unittest.skipUnless(LooseVersion(scipy_version) >= LooseVersion("1.1"),
+                         "scipy >= 1.1 is required.")
+    def test_trust_constr_inequality_con(self):
+
+        class Sphere(ExplicitComponent):
+
+            def __init__(self, problem):
+                super(Sphere, self).__init__()
+                self.problem = problem
```
naylor-b: There are multiple instances in this file where the Problem is passed as an __init__ arg to a Component, e.g. Sphere and Rosenbrock.  As far as I can tell, the problem attribute of these components is never used so we should remove the __init__ arg and the attribute anywhere where it currently appears in the file.

onodip:
One test fails in all failed cases:

`test_scipy_iter_solver.py:TestScipyKrylovFeature.test_specify_precon
`
One of the variables is 0, which should not be.
Seems unrelated to the PR.

swryan:
This test is giving us trouble right now..  a team member is working on it.  Definitely not related to your PR.

PR #815, opened 2018-11-28
Three Solver-related Stories.

Kenneth-T-Moore:
1. Removed the return values (fail, abs rel) from the entire openmdao solve/solve_*/run_* hierarchy; AnalysisError should be used to detect and act on errors.
2. Replaced vector deepcopy in the debug printing with a lightweight copy that only deepcopies the views.
3. Added an error message for DirectSolver with csc assembled jacobian for when there a row or column in the matrix is a linear  combination of other rows/cols.

### DIFF
--- a/openmdao/utils/general_utils.py
+++ b/openmdao/utils/general_utils.py
@@ -506,7 +506,7 @@ def pad_name(name, pad_num=10, quotes=False):
             return '{0}'.format(name)
 
 
-def run_model(prob, allow_exception=False):
+def run_model(prob, ignore_exception=False):
     """
     Call `run_model` on problem and capture output.
 
@@ -514,8 +514,8 @@ def run_model(prob, allow_exception=False):
     ----------
     prob : Problem
         an instance of Problem
-    allow_exception : bool
-        Set to True to allow an exception of any kind.
+    ignore_exception : bool
+        Set to True to ignore an exception of any kind.
 
     Returns
     -------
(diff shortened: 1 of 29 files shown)


openmdao/utils/general_utils.py
```diff
@@ -506,14 +506,16 @@ def pad_name(name, pad_num=10, quotes=False):
             return '{0}'.format(name)
 
 
-def run_model(prob):
+def run_model(prob, allow_exception=False):
```
naylor-b: would 'ignore_exception' be clearer here?

openmdao/vectors/vector.py
```diff
@@ -213,6 +214,22 @@ def _clone(self, initialize_views=False):
...
+    def _copy_views(self):
+        """
+        Return a lightweight copy containing just the views.
+
+        Returns
+        -------
+        <Vector>
+            Vector containing _name, _typ, and _views.
+        """
+        vec_cp = self.__class__(self._name, self._kind, self._system, self._root_vector,
```
naylor-b: do we really need to recreate the whole Vector here?  We just need to copy the views of the existing vector, right?  Calling __init__ on the Vector calls _initialize_data() and _initialize_views() which do quite a bit of stuff that seems unnecessary here.

openmdao/utils/general_utils.py
Kenneth-T-Moore: Changed to `ignore_exception`

openmdao/vectors/vector.py
Kenneth-T-Moore: Made a few adjustments so that we just need to save a copy of the views dictionary, then made the suggested change.

PR #817, opened 2018-11-29
eliminate issues with assembled jacobians having overlapping partials

naylor-b:


openmdao/solvers/linear/tests/test_direct_solver.py
```diff
@@ -207,23 +207,6 @@ def test_raise_error_on_singular(self):
 
         self.assertEqual(expected_msg, str(cm.exception))
 
-    def test_raise_error_on_dup_partials(self):
-        prob = Problem()
-        model = prob.model
-
-        model.add_subsystem('des_vars', IndepVarComp('x', 1.0), promotes=['*'])
-        model.add_subsystem('dupcomp', DupPartialsComp())
```
Kenneth-T-Moore: That DupPartialsComp component: is specification like that supported now? The first component I worked in OpenAeroStruct, I accidentally duplicated some source indices while planning out the various stripes in the jacobian. If that's supported now, then I think we should add a case like this, but with purposeful sensible declaration of duplicate indices, and make sure the derivs are right.

naylor-b: We do support repeated src_indices now (and there's a test for it), but I should put back in the check to prevent duplicate indices in rows and cols passed to declare_partials(), because duplicate src_indices might be valid, but duplicate rows/cols indices makes no sense.

naylor-b: ok I put in a check to prevent duplicate subjac indices from being specified via row/cols in declare_partials() and updated the test.

PR #821, opened 2018-12-06
Fix KSComp so compute_partials doesn't rely on info from compute

swryan:
Pivotal #162296529

### DIFF
--- a/openmdao/components/ks_comp.py
+++ b/openmdao/components/ks_comp.py
@@ -184,7 +184,12 @@ class KSComp(ExplicitComponent):
         if opt['lower_flag']:
             con_val = -con_val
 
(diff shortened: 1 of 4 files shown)


openmdao/components/ks_comp.py
```diff
@@ -174,7 +198,18 @@ def compute_partials(self, inputs, partials):
...
+        vec_size = opt['vec_size']
+        width = opt['width']
+
+        derivs = np.empty((vec_size, width))
+
+        con_val = inputs['g'] - opt['upper']
+        if opt['lower_flag']:
+            con_val = -con_val
+
+        for j in range(vec_size):
```
JustinSGray: this seems like it could be vectorized, instead of using the for loop, maybe?

PR #827, opened 2018-12-12
Solver refactor

naylor-b:
- renamed _run_iterator to _solve
- renamed _iter_execute to _single_iteration
- moved duplicated code for nonlinear GS iteration into _gs_iter method
- fixed a bug in CSCMatrix that showed up under older versions of scipy
- in MetaModelStructuredComp, changed warning about scipy version too old from a warning to an exception (since it won't work with the old scipy version)
- removed the `assembled_jac_type` option from ExplicitComponent
- added an 'openmdao call_tree' command to list all 'self' methods called from a given method, including the class where each method is defined.
- added docs for `openmdao call_tree`
- added theory manual entry for solvers, describing types of solvers and a little about writing custom solvers. (this doesn't seem like theory, but not sure where to best put it)


openmdao/components/meta_model_structured_comp.py
```diff
@@ -8,9 +8,12 @@
 
 from scipy import __version__ as scipy_version
 try:
-    from scipy.interpolate._bsplines import make_interp_spline
+    from scipy.interpolate._bsplines import make_interp_spline as _make_interp_spline
 except ImportError:
-    make_interp_spline = False
+    def _make_interp_spline(*args, **kwargs):
```
JustinSGray: I wonder if you will ever keep separate efforts things to separate branches ??? :)

openmdao/core/explicitcomponent.py
```diff
@@ -87,6 +88,18 @@ def _setup_partials(self, recurse=True):
             self._declare_partials(out_name, out_name, rows=arange, cols=arange,
                                    val=np.full(meta['size'], -1.))
 
+    def _setup_jacobians(self, recurse=True):
```
JustinSGray: why did this get added to ExplicitComp?

openmdao/docs/theory_manual/solver_api.rst
```diff
@@ -0,0 +1,87 @@
+.. _theory_solver_api:
+
+****************
+OpenMDAO Solvers
+****************
+
+Every OpenMDAO solver is either a linear solver, inheriting from the :code:`LinearSolver` class, or
+a nonlinear solver, inheriting from the :code:`NonlinearSolver` class. A solver can be either monolithic
+or recursive in behavior.  Monolithic solvers typically construct some sort of linearization of
```
JustinSGray: "Monolithic solvers treat the associated system as a single block. Recursive solvers, as the name suggests, recurse down through their system hierarchy asking each sub-system to operate on itself. "

openmdao/core/explicitcomponent.py
naylor-b: This is just a minor optimization basically.  Since explicit comps don't have linear solvers, the entire setup_jacobians function goes away except for the call to _set_partials_meta.

openmdao/docs/theory_manual/solver_api.rst
naylor-b: docs updated

openmdao/code_review/test_lint_docstrings.py
```diff
@@ -551,9 +551,10 @@ def test_docstrings(self):
                         funcs = []
 
                     for func_name in funcs:
-                        func = getattr(mod, func_name)
-                        self.check_function(dir_name, file_name, func_name,
-                                            func, failures)
+                        if not func_name.startswith('_'):
+                            func = getattr(mod, func_name)
```
Kenneth-T-Moore: Are we sure we want to do this (team?)  The way we use underscore methods, it is nice to have our style of docstring, and if we remove them from the testing, then we will have to rely on catching them in the reviews.

naylor-b: Note that this only applies to module level '_'* functions.  Are you suggesting that even the most private function buried in a module should be forced to have our full, verbose docstring?  That seems like overkill IMHO.  For example, should the following _make_interp_spline function have a full docstring?
```
try:
    from scipy.interpolate._bsplines import make_interp_spline as _make_interp_spline
except ImportError:
    def _make_interp_spline(*args, **kwargs):
        msg = "'MetaModelStructuredComp' requires scipy>=0.19, but the currently" \
              " installed version is %s." % scipy_version
        raise RuntimeError(msg)
```

Kenneth-T-Moore: I don't have a problem with putting a docstring there, but we could support a "lint ignore" directive syntax for this as well.

JustinSGray: i agree with @naylor-b  here. We don't need to force people to lint internal module level private function

PR #845, opened 2019-02-11
Restored values for some colors that were accidentally removed during a previous mod

hschilling:
This caused the display of connections to disappear

### DIFF
--- a/openmdao/docs/advanced_guide/implicit_comps/n2.html
+++ b/openmdao/docs/advanced_guide/implicit_comps/n2.html
@@ -1,5 +1,4 @@
 
-
                       <html>
                       <head>
                         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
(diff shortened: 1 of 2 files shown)


openmdao/docs/advanced_guide/implicit_comps/n2.html
```diff
@@ -2308,7 +2310,7 @@ <h1>OpenMDAO Partition Tree and N<sup>2</sup> diagram.</h1>
 
             text += "<br />self.connect(\"" + unknownName + "\", \"" + paramName + "\")";
         }
-        if(true) {
+        if(false) {
```
Kenneth-T-Moore: Are these `if` blocks frequently turned on/off?  Otherwise you could just delete these, since they are effectively commented out.

hschilling: This file is created via a Python script that flips the true/false based on an option to the Python function. We could make that code more intelligent to remove the "commented out" block but it doesn't like it would be worth it.

PR #846, opened 2019-02-12
NonlinearBlockGS solver convergence

relf:
This PR comes from our need to adapt the convergence criterium of the NonlinearBlockGS solver. The rtol criterium is somehow relaxed by selecting a subset of variables used to check the convergence.

Indeed in our model some output variables may be 'highly' nonlinear and prevent the convergence while we just want their final value as is. This problem is illustrated in the test, where a contrived variable is added to the Sellar MDA; by selecting only the state variables the MDA is converged faster.


JustinSGray:
@relf I understand the core concept behind this new feature, and we've played with an idea similar to this ourselves with the broyden solver. However, we decided to go a different route for broyden and I'm not sure that this is the correct approach here either. 

I have played with your example to make sure I understood totally what was going on. Its a good stress-test case. Thank you for providing it. As you indicated, the `highly_nonlinear` output from `ContrivedSellarDis1` does take longer to converge than the `y1` and `y2` variables. *if* all you cared about was y1/y2 then your fix does work to get a quicker stopping criterion. 

However, it does not change the reality that this highly nonlinear variable is still not fully converged, because it is much more sensitive to `y2` input than the `y1` calculation is. If you have an analysis where you passed `highly_nonlinear` downstream to another component, then you potentially have a problem. If you were FDing across such a model, then solver error would be massive. If you are using analytic derivatives, which work under the assumption that R(u)=0 then you're also introducing a source of error. 

My first inclination here is to address this problem using `res_ref`. if you set `res_ref=1e-2` then you would have 2 orders of magnitude less sensitivity to that output residual. This would allow you to scale that variable to behave the same as the others in a convergence loop, while still ensuring that you consider its residual for the stopping criterion. The problem is that when I tested that out, it clearly had no impact on NLBGS convergence. So it seems that NLBGS is not currently scaling its residual vector before computing the relevant norm. 

So I suggest that correcting the scaling mistake in NLBGS combined with a res_ref on the offending variables is a better solution. What do you think about this alternate path?

relf:
Indeed I was wondering if I could not achieve the same with scaling capabilities of OpenMDAO but I didn't test it. I will be happy to test it once it is fixed. 
Nevertheless another point (which is not shown by my contrived example) is that we use black box models with several tenth of variables and we just promote all of them to get everything wired. Out of this set of variables, just three of them are fed back. It is simply more convenient and straightforward to have the ability to select them to get the quickest convergence. Even if other variables are not fully converged we can still check how bad they are (thanks to <code>list_outputs()</code> power) and be happy with that.
So I still argue that the ability to select convergence variables is a valuable option.
But I can understand you do not want to introduce this interface in the solver for the sake of purity.
Would it be possible then to keep <code>_is_rtol_converged()</code> method (or something equivalent) to be able to subclass NonlinearBlockGS without overriding <code>solve()</code> entirely?

JustinSGray:
I supposed, at an API level we could modify the solvers call stack to use is_rtol_converged instead of get_norm directly. Then you could overwrite that method. At the very least, we'd need to also create an is_atol_converged.  A scondary question is, why you only make this change to NLBGS? why not also to newton, broyden, and NLBJacobi? Really, all the built in solvers should be modified to use these methods then. 

The downside to this is that the solver call stack, which was already fairly complicated, now gets yet another layer in it.

I can appreciate your needs in this case, even if I don't think the approach is safe for general usage. As long as you are careful with your own usage I guess this is acceptable. The beauty of an open source framework is that you can make whatever changes you need to suite your purposes. 

My current inclination is that, unless you saw the feature being useful in all of the solvers, that this just remain as part of a custom solver you distribute with your own models. Even if we added the new API methods, you would still have to sub-class the solvers you cared about. The added code from the solve method in your sub-class seems minor compared to making the already complex solvers more so for the base framework.

relf:
Actually initially I implemented the new options in the NonlinearSolver class but I was a bit reluctant by modifying all the solvers without having your feelings about those changes and second I am not sure to master all the implications of such changes. That is why I scope it down to the NLBGS solver which is the solver we mainly (not to say exclusively) use and even here I think it breaks the encapsulation of vector interface in <code>_iter_get_norm()</code> when I call<code>self._system._residuals._views_flat[name]</code>.
So as you do not see any use case on your side. I will keep the changes in a custom solver and remove NLBGS changes. I did not get if you want me to add a is_atol_converged or if you want to refactor the stuff on your own or if you do not want to make any change (which is fine as well). Thank you to let me know.

JustinSGray:
I was saying that, if we added this to the framework core codes then we would also have to do the `is_atol_converged` method as well. However, since Im reluctant to include that API change into solver base, I would say there is no need. Im going to close this PR for now, but if you start to find a broad need for this functionality in other solvers lets revisit the topic then!

relf:
Ok it is fine as well :wink:

relf:
FYI, the custom NonlinearBlockGS with convergence variables option is now available in [this repo](https://github.com/OneraHub/openmdao_extensions) as <code>RecklessNonlinearBlockGS</code>.

PR #848, opened 2019-02-12
fix xdsm test to handle non-existence of pdflatex

swryan:


openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -518,6 +522,14 @@ def number_label(number, text, alignment):
 
     if writer_name == 'pyxdsm':  # pyXDSM
         x = XDSMWriter()
+
```
swryan: I am leaning against doing the following here... I think it's better handled in pyXDSM and adding this logic would put the burden on us to add some awkward tests... thoughts?

Kenneth-T-Moore: This does seem like pyXDSM's responsibility.

openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -20,6 +20,15 @@
 
 from __future__ import print_function
 
+try:
+    from pyxdsm.XDSM import XDSM
+except ImportError:
+    msg = ('The pyxdsm package should be installed. You can download the package '
+           'from https://github.com/mdolab/pyXDSM')
+    raise ImportError(msg)
```
naylor-b: any chance we could include the pip install command here?  Something like:

    msg = ('The pyxdsm package should be installed. You can download the package '
           'from https://github.com/mdolab/pyXDSM, or install using the pip command: '
           'pip install git+https://github.com/mdolab/pyXDSM.git')

openmdao/devtools/xdsm_viewer/xdsm_writer.py
naylor-b: I also think it's better handled by pyXDSM.

openmdao/devtools/xdsm_viewer/xdsm_writer.py
swryan: done

openmdao/devtools/xdsm_viewer/xdsm_writer.py
swryan: removed

PR #852, opened 2019-02-13
fix condition when reading legacy recordings (not is True for all-zeros)

swryan:
Pivotal #163880271

### DIFF
--- a/openmdao/recorders/case.py
+++ b/openmdao/recorders/case.py
@@ -130,7 +130,7 @@ class Case(object):
                 inputs = json_to_np_array(data['inputs'], abs2meta)
             elif data_format in (1, 2):
                 inputs = blob_to_array(data['inputs'])
-                if not inputs:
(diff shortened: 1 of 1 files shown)


openmdao/recorders/case.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, source, data, prom2abs, abs2prom, abs2meta, voi_meta, data_fo
                 inputs = json_to_np_array(data['inputs'], abs2meta)
             elif data_format in (1, 2):
                 inputs = blob_to_array(data['inputs'])
```
Kenneth-T-Moore: You should add a test to `TestSqliteCaseReaderLegacy`.

swryan: I considered that but seemed to be overkill for an edge case on a several versions old format that nobody outside of us even has used.   Would have to revert to a very old version of openmdao to create a test case. Given that it was debated whether this needed fixed at all, I thought the extra test was not warranted.

PR #854, opened 2019-02-15
Separated tests for pyXDSM and XDSMjs

onodip:
Separated pyXDSM and XDSMjs tests to two classes, so XDSMjs tests are not skipped if pyXDSM is not installed.


onodip:
Now this PR also solves this issue:
https://www.pivotaltracker.com/n/projects/1885757/stories/163921400

PR #855, opened 2019-02-18
sync XDSMjs: fix fontello icons display

relf:
This PR fixes [this bug](https://www.pivotaltracker.com/story/show/163947690)


JustinSGray:
wow, was the problem just a missing icon folder?

relf:
The fontello folder is just for the record and license, I just keep the XDSMjs folder in sync with XDSMjs repo. The bug is fixed because now the fontello.css file embeds the icons (the previous css was not standalone and included a <code>./fontello.woff</code> file which is present in the XDSMjs folder but is missing where the OpenMDAO html is created, hence the squares).

PR #864, opened 2019-02-28
fix jacobian scaling issue when computing bool jac for coloring

naylor-b:


openmdao/utils/coloring.py
```diff
@@ -516,8 +516,7 @@ def _get_bool_jac(prob, repeats=3, tol=1e-15, orders=5, setup=False, run_model=F
                 fullJ += np.abs(J)
         elapsed = time.time() - start_time
 
-    # normalize the full J by dividing by the max value
```
Kenneth-T-Moore: is the comment still accurate?

naylor-b: it's not accurate any more, and it was kind of unnecessary anyway since it was pretty obvious that the next line was dividing the matrix by the max.

openmdao/utils/coloring.py
```diff
@@ -516,8 +516,7 @@ def _get_bool_jac(prob, repeats=3, tol=1e-15, orders=5, setup=False, run_model=F
...
-    fullJ /= np.max(fullJ)
+    fullJ /= repeats
```
swryan: looks like the docstring for the function needs to be updated:
"then dividing by the _max value_, then converting to a boolean array, ..."

naylor-b: fixed

PR #860, opened 2019-02-22
XDSM parallel blocks

onodip:
Components within a parallel group can be shown stacked on the XDSM. On the picture the group of _d1_ and _d2_ is a parallel group.

Some additional tests added.

image

PR #866, opened 2019-02-28
XDSM block styling improvements

onodip:
Implicit components and metamodels now have different color.
In the circuit example below _n1_ and _n2_ are implicit comps.

EDIT:
Now all features in this PR work for both XDSMjs and pyXDSM.

Added features:
- Implicit comps.
- ExecComps and IndepvarComps as Functions
- DOE
- MetaModel

This PR includes the changes from #860.

image


onodip:
@swryan okay, I updated it

openmdao/devtools/tests/test_xdsm_viewer.py
```diff
@@ -18,21 +19,28 @@
...
 
 @unittest.skipUnless(XDSM, "The pyXDSM package is required.")
 class TestPyXDSMViewer(unittest.TestCase):
 
     def setUp(self):
-        self.startdir = os.getcwd()
-        self.tempdir = tempfile.mkdtemp(prefix='TestPyXDSMViewer-')
-        os.chdir(self.tempdir)
+        if not DEBUG:
+            self.startdir = os.getcwd()
```
naylor-b: It's not clear to me why we're not using a temp directory if DEBUG is False.  Even in the DEBUG=False case, the tests are creating local files.  These files should be cleaned up in order to avoid the possibility of tests passing when they shouldn't (due to existence of a file left over from a previous test run even if the current test failed to create it), and also just to avoid cluttering up the file system with a bunch of generated files.

onodip: I use DEBUG=True during development to visually inspect the files (since the tests can only check if the file was created, but not that the diagram is correct). But i can get rid of it, if it is confusing or could be a source of error. Actually I stole the idea [from the problem viewer tests](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/devtools/problem_viewer/tests/test_viewmodeldata.py).

naylor-b: Oops, my bad.  It's totally fine.  Ignore my comment.

PR #868, opened 2019-03-06
XDSM outputs on the right

onodip:
New option to choose if inputs should be on the left or right side. 
For now it is only supported for pyXDSM.

Example with the right side...
image

...and left (as usual):
image


onodip:
@relf is this possible now in XDSMjs?

relf:
@onodip The evolution is not exactly straightforward as the current implementation relies on the assumption of an invisible user component sitting in 0,0. Moreover the [XDSMjs json format](https://github.com/OneraHub/XDSMjs/wiki/XDSMjs-JSON-format) is dedicated to actual MDO content, representation option should be passed through [configuration options](https://github.com/OneraHub/XDSMjs/wiki/XDSMjs-configuration) which are not currently exposed as data attributes (like MDO data are through <code>data-mdo</code>). I will open issues for these and hopefully implement solutions or I can accept PRs as well.

onodip:
@relf the data attributes would be great. I was thinking to make the OpenMDAO XDSMjs a bit more interactive (similar to the N^2 diagram, so the user can change the size, what is shown, etc. The menu bar could be reused from the N^2.

onodip:
For the right side outputs you can have maybe an invisible last component.

relf:
If it is the last component the outputs will be on the right, yes... but the inputs will at the bottom! We do not want that configuration option, do we? 
I have to look... but off the top of my head, the _node on the diagonal_ assumption is a strong one in the current XDSMjs implementation and several things might break.

onodip:
Yeah, in my opinion inputs makes sense only on the top. 
But it could be that for inputs you always connect with _U_ (which is left and top), and there would be the last component _E_ ()

onodip:
Now can be set separately for different comp types (e.g. optimization): 
image

openmdao/devtools/tests/test_xdsm_viewer.py
```diff
@@ -944,6 +974,28 @@ def test_circuit_recurse(self):
...
+
+        prob.setup(check=False)
+        prob.final_setup()
+
+        # Write output
+        write_xdsm(prob, filename=filename, out_format='html', show_browser=False, quiet=QUIET,
+                   output_side='right')
+
+        # Check if file was created
+        self.assertTrue(os.path.isfile('.'.join([filename, 'html'])))
```
swryan: I would suggest testing for the UserWarning here.. it helps to know that the resulting diagram is not intended to show outputs on the right in spite of the logic...
(e.g. https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/components/tests/test_meta_model_structured_comp.py#L1271 )

onodip: Good idea, added.

PR #870, opened 2019-03-09
Problem viewer toolbar

onodip:
The code for the problem viewer toolbar in the template file was replaced with a simple API. Aim was to make it easier to implement new features and to make it reusable for the XDSMjs viewer. No functional change in the problem viewer.


onodip:
So for XDSMjs it would look like something like this:

image

With the first four buttons from the original XDSMjs toolbar. This is just mock-up for now (and not in the PR), buttons are not working.

Issue #874, opened 2019-03-15
Genetic algorithm ValueError with vector design variables


jennirinker:
## Version
I'm testing with 2.5.0, but it's probably in other versions as well

## MWE

```
from openmdao.api import Problem, ScipyOptimizeDriver, ExecComp, IndepVarComp, SimpleGADriver

will_i_live = False
prob = Problem()
indeps = prob.model.add_subsystem('indeps', IndepVarComp())
indeps.add_output('x', [4.0, -4])
prob.model.add_subsystem('paraboloid', ExecComp('f = (x[0]-3.5)**2 + (x[1]+4.2)**2 - 3',
                                                x=[0, 0]))
prob.model.connect('indeps.x', 'paraboloid.x')

if will_i_live: prob.driver = ScipyOptimizeDriver()
else: prob.driver = SimpleGADriver()

prob.model.add_design_var('indeps.x', lower=-10, upper=10)
prob.model.add_objective('paraboloid.f')
prob.setup()
prob.run_driver()
print(prob['indeps.x'])
```

In particular, it's the log_range calculation part that dies:
```

  File "c:\users\rink\git\openmdao\openmdao\core\problem.py", line 556, in run_driver
    return self.driver.run()

  File "c:\users\rink\git\openmdao\openmdao\drivers\genetic_algorithm_driver.py", line 256, in run
    if log_range % 2 > 0:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

swryan:
resolved by PR #875

PR #875, opened 2019-03-15
Resolving Issue #874 (GA driver bug with vector design variables)

jennirinker:
My proposed patch for Issue #874

Two main changes:
- Switch the log_range logic to use masking so it can handle design variables that are vectors
- Added a new test for vector design variables


jennirinker:
@Kenneth-T-Moore My pleasure. Thanks for making such a useful tool.

Quick question: I just realized that there's no GA test for multiobjective optimization, which leaves about 8 lines out of test coverage. Would you like me to add a test to this PR to include that case, or not worry about it?

jennirinker:
Sorry, I had to write the test anyway for local test coverage requirements so I've pushed a new commit. It's basically the same test as before, just split into two objective functions.

Kenneth-T-Moore:
Thanks, @jennirinker  We are definitely lacking multi-objective tests, so your modified test definitely improves our coverage.

PR #878, opened 2019-03-16
XDSM model path bug fix

onodip:
The XDSM writer now works correctly if a model path is given.

This PR solves issue this issue: https://www.pivotaltracker.com/n/projects/1885757/stories/163920286

The tests were also rationalized a bit, to not create the PDF when it is not actually tested. Some similar tests, which used the same model were merged into one.

### DIFF
--- a/openmdao/devtools/xdsm_viewer/xdsm_writer.py
+++ b/openmdao/devtools/xdsm_viewer/xdsm_writer.py
@@ -1007,12 +1007,13 @@ def _get_path(name, sep='.'):
     return name.rsplit(sep, 1)[0]
 
 
-def _make_rel_path(full_path, model_path):
+def _make_rel_path(full_path, model_path, sep='.'):
     # Path will be cut from this character. Length of model path + separator after it.
     # If path does not contain the model path, the full path will be returned.
     if model_path is not None:
-        first_char = len(model_path) + 1
-        if full_path.startswith(model_path):
+        path = model_path + sep  # Add separator character
+        first_char = len(path)
+        if full_path.startswith(path):
             return full_path[first_char:]
         else:
             return full_path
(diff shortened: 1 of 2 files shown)


openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -982,14 +993,33 @@ def _accumulate_connections(conns):
...
+    # Returns path until the last separator in the name
+    return name.rsplit(sep, 1)[0]
+
+
+def _make_rel_path(full_path, model_path):
+    # Path will be cut from this character. Length of model path + separator after it.
+    # If path does not contain the model path, the full path will be returned.
+    if model_path is not None:
+        first_char = len(model_path) + 1
+        if full_path.startswith(model_path):
```
naylor-b: this could give false matches unless you check for startswith(model_path + '.') instead of startswith(model_path)

openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -1075,22 +1105,18 @@ def _prune_connections(conns, model_path=None, sep='.'):
...
-                tgt = tgt[len(model_path):]
-            tgt_path = _format_name(tgt.rsplit(sep, 1)[0])
-
-            if src0.startswith(model_path) and tgt0.startswith(model_path):
+            src = conn['src']
+            src_path = _make_rel_path(src, model_path=model_path)
+            tgt = conn['tgt']
+            tgt_path = _make_rel_path(tgt, model_path=model_path)
+
+            if src.startswith(model_path) and tgt.startswith(model_path):
```
naylor-b: same problem here as mentioned before.  You need to check startswith using model_path + '.' or you can get false matches.

onodip: indeed, nice catch @naylor-b

PR #882, opened 2019-03-18
XDSM from SQL

onodip:
The `write_xdsm()` function can take an SQL file as input to generate the diagram. 
Yet the driver, design variables and responses are not included. (Any easy way to get them from the SQL file or from an `SqliteCaseReader`?)

It also might make sense to change the name of the first argument from `problem` to `source_data` to make it consistent with the N^2 writer.


onodip:
@swryan, yeah that would be a good solution

swryan:
PR #910 addresses these comments

PR #884, opened 2019-03-21
Driver name for recording

relf:
This PR to use driver <code>_get_name</code> call consistently in RecordingDebugging creation and specify scipy or pyoptsparse prefix in corresponding driver names.


relf:
I have to fix case reader tests, before proceeding I was wondering if you are ok with the scipy/pyoptsparse prefixing or if you prefer that I undo that change which can be considered backward incompatible?

JustinSGray:
this is a nice change. It makes it clear what optimizer was used. Thanks!

PR #885, opened 2019-03-21
XDSM target name formatting bugfix

onodip:
Small formatting error corrected. Bug was there only with `include_solver` option turned on.

Before
image

Now:
image


swryan:
Looks like this fix was incorporated into your newer PR, so I think we can close this one and merge the latest.

onodip:
Yes, the newer one includes this fix.

PR #886, opened 2019-03-22
Scaling fix addendum.

Kenneth-T-Moore:
I missed a scaled context we needed around linearize.  Added it here with a test.

Also included some scripts to generate sequence diagrams that show how scaling works for a few easy cases.  These are python scripts that generate svg, and require the `svgwrite` package.


swryan:
I see you added the code for generating the scaled context sequence diagrams to the theory manual directory of the `docs`, but did not add the diagrams themselves or a doc page...  wouldn't it be much more useful (without much more effort) to add a brief page showing the diagrams with a few words of explanation? If not contributing to the docs, these scripts seem more like just `devtools`...

Kenneth-T-Moore:
The diagrams are a work in progress. I don't think we are at a point where we are ready to show them to the public.  They are a little complicated, and I am only showing some very basic operations, so we also probably need to discuss where we want to go with this stuff.  If there is strong enough general objection to having these files here, I can remove them or put them somewhere else.

Kenneth-T-Moore:
Deleted scaling diagrams because the bug fix is important.

PR #887, opened 2019-03-22
Small fix for nlbgs so that the res_ref scales the residual computed from difference in outputs.

Kenneth-T-Moore:
Just needed to add a couple of unscaled contexts.


JustinSGray:
@relf this PR should allow you to use `res_ref` to properly scale your output residuals for problems with highly nonlinear variables that are problematic for convergence

relf:
@Kenneth-T-Moore, @JustinSGray : Thank you, I will try out.

PR #888, opened 2019-03-22
XDSM solver hierarchy

onodip:
Now with the `include_solver` option solvers of subsystems are also included, not only the top level solver. On the picture below both `model` and `cycle` have a solver.

image

Other changes:
- A writer object can be also given as input, which enables more customization for the user.
- Writer specific formatting moved into the classes.

Known issues:
- Numbering of components in pyXDSM can be incorrect, because nested loops are not taken into account in the solver/driver labels currently


onodip:
Hmm, for me the new test does not fail on my computer, just here.

swryan:
Looks like your test failure on CI is due to the default out_format='tex' in `write_xdsm`, which requires `pyxdsm`.  

I ran testflo on travis with the `-vs` option to see the message:
```
The "tex" and "pdf" formats require the pyxdsm package. You can download the package from https://github.com/mdolab/pyXDSM, or install it directly from github using:  pip install git+https://github.com/mdolab/pyXDSM.git
openmdao.devtools.tests.test_xdsm_viewer.py:TestXDSMjsViewer.test_custom_writer ... FAIL (00:00:0.01, 76 MB)
```

onodip:
Thanks @swryan, it was a good clue! Actually it does not even require pyXDSM, but it was looking for the pdflatex executable, because the out_format (which is otherwise ignored with a custom writer) was still `tex`.

openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -285,39 +346,49 @@ def add_system(self, node_name, style, label=None, stack=False, **kwargs):
         if label is None:
             label = node_name
         if stack:  # Parallel block
-            style += '_multi'  # Block will be stacked in XDSMjs, if ends with this string
+            style += self._multi_suffix  # Block will be stacked in XDSMjs, if ends with this string
         dct = {"type": style, "id": self._format_id(node_name), "name": label}
         self.comps.append(dct)
 
-    def add_workflow(self, comp_names=None):
+    def add_workflow(self, solver=None):
```
swryan: nit: the text in the docstring needs to be updated...   if "solver" is None

openmdao/devtools/xdsm_viewer/xdsm_writer.py
```diff
@@ -505,23 +587,100 @@ def add_driver(self, name, label=None, driver_type='Optimization', **kwargs):
             kwargs : dict
                 Keyword args
             """
-            style = _COMPONENT_TYPE_MAP[self._writer_name].get(driver_type, 'Optimization')
-            self.add_system(node_name=name, style=style, label='\\text{%s}' % label, **kwargs)
+            style = self.type_map.get(driver_type, 'Optimization')
+            self.add_system(node_name=name, style=style, label=self._textify(label), **kwargs)
 
-        def add_workflow(self, comp_names=None):
+        def add_workflow(self, solver=None):
```
swryan: nit: the text in the docstring needs to be updated... if "solver" is None

PR #899, opened 2019-03-27
 implement guess_func option to BalanceComp

swryan:
Get rid of the per-state guess_func argument and instead provide a component level guess function hook via init args/options.

This is a BACKWARD-INCOMPATIBLE change.


openmdao/components/balance_comp.py
```diff
@@ -21,8 +22,16 @@ class BalanceComp(ImplicitComponent):
         so everything can be saved until setup is called.
     """
 
-    def __init__(self, name=None, eq_units=None, lhs_name=None,
-                 rhs_name=None, rhs_val=0.0, guess_func=None,
+    def initialize(self):
+        """
+        Declare options.
+        """
+        self.options.declare('guess_func', types=(FunctionType), allow_none=True, default=None,
```
naylor-b: If (FunctionType) is intended to be a one item tuple, it needs a comma inside.  Otherwise I'd just get rid of the parens.

PR #901, opened 2019-03-29
User can pass shape as an argument to ExecComp, when all variables have the same shape 

swryan:
shape, units and vectorize are now proper options to ExecComp

Also:
- Update options to properly show Available Values for bool types

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -22,6 +22,7 @@ _allowed_meta = {'value', 'shape', 'units', 'res_units', 'desc',
 # Names that are not allowed for input or output variables (keywords for options)
 _disallowed_names = {'vectorize', 'units', 'shape'}
 
+
 def array_idx_iter(shape):
     """
     Return an iterator over the indices into a n-dimensional array.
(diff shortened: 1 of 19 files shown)


openmdao/components/exec_comp.py
```diff
@@ -193,18 +223,21 @@ def __init__(self, exprs, vectorize=False, units=None, **kwargs):
...
         self._codes = None
         self._kwargs = kwargs
-        self._vectorize = vectorize
-        self._units = units
 
     def setup(self):
         """
         Set up variable name and metadata lists.
         """
+        if not self._exprs:
```
Kenneth-T-Moore: Can this ever happen? I guess if they pass in `None` as their expression list.  Raising this in `__init__` instead of `setup` might be more helpful to the user.

swryan: This was a change I made (see the new test) to enable calling the constructor with no args, which is necessary for `embed-options` to display the options table.

openmdao/components/exec_comp.py
```diff
@@ -175,14 +205,14 @@ def __init__(self, exprs, vectorize=False, units=None, **kwargs):
                               x={'value': numpy.ones(10,dtype=float),
                                  'units': 'ft'})
         """
-        super(ExecComp, self).__init__()
+        # separate disallowed var names from kwargs, pass them as options to __init__
+        options = {}
+        for name in _disallowed_names:
+            if name in kwargs:
+                options[name] = kwargs[name]
```
naylor-b: I think you could just do 'options[name] = kwargs.pop(name)' and save a line.

PR #914, opened 2019-04-19
MPI bug fix in component class setup_var_sizes.

nbons:
This is the same fix made in PR #904.


JustinSGray:
same kind of fix, but in different place in the code.

PR #917, opened 2019-04-25
Support discrete design variables in the simple GA optimizer and in general drivers.

Kenneth-T-Moore:


openmdao/core/driver.py
```diff
@@ -152,7 +164,7 @@ def __init__(self, **kwargs):
         self.supports.declare('linear_constraints', types=bool, default=False)
         self.supports.declare('two_sided_constraints', types=bool, default=False)
         self.supports.declare('multiple_objectives', types=bool, default=False)
-        self.supports.declare('integer_design_vars', types=bool, default=False)
+        self.supports.declare('integer_design_vars', types=bool, default=True)
```
naylor-b: Just curious why support for integer design vars is True by default.

openmdao/core/driver.py
```diff
@@ -236,6 +248,14 @@ def _setup_driver(self, problem):
             np.any([dv['scaler'] is not None for dv in itervalues(self._designvars)])
         )
 
+        # Determine if any design variables are discrete.
+        self._designvars_discrete = [dv for dv in self._designvars
```
naylor-b: I didn't see anywhere where we actually check to see if a given discrete var is an int, since discrete vars in general can be any python object.

openmdao/core/driver.py
Kenneth-T-Moore: There were a bunch of teats with Driver failing when I put the "supports" check in, but I fixed something else in the meantime, and now they pass with False supports, so I will change it back to False.

openmdao/core/driver.py
Kenneth-T-Moore: We could rename the option "supports_discrete_designvars" to match the implementation.  

Alternatively, I don't know what to do with this. We don't know what the user really wants to put into a discrete variable, and they are free to do whatever they want with it because it has no checking at all.  The only thing we can do is divine some intent on the driver -- if they hook it up as a desvar for a Genetic algorithm, then it is either an integer, an array of integers, or maybe a list/iterator/object that gives integers.

This whole feature probably needs a lot more thought and design.  But the point of the error message is to prevent someone from accidentally using a discrete desvar in their optimizer for the general case.  Further typecheckiing should probably be done in specific drivers that are capable of using them.

Kenneth-T-Moore: I have added some checking on the get.

PR #925, opened 2019-05-08
improved doc search

swryan:
Modified sphinx doc search:
- option to include/exclude source docs (default is to exclude)
- if source docs are included, sort them to the end of the results with a separator
- if source docs are included, only include one result for each filename

### DIFF
--- a/openmdao/docs/_theme/static/searchtools.js
+++ b/openmdao/docs/_theme/static/searchtools.js
@@ -7,6 +7,11 @@
  * :copyright: Copyright 2007-2019 by the Sphinx team, see AUTHORS.
  * :license: BSD, see LICENSE for details.
  *
+ * Modified for OpenMDAO:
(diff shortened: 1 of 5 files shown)


openmdao/docs/_theme/static/searchtools.js
```diff
@@ -0,0 +1,598 @@
...
+        title: 15,
+        partialTitle: 7,
+        // query found in terms
+        term: 5,
+        partialTerm: 2
+    };
+}
+
+if (!splitQuery) {
+    function splitQuery(query) {
```
Kenneth-T-Moore: We talked some today about searching for "cross product" as a single entity. Would it be possible to get that to work by modifying this regex to split on whitespace that isn't enclosed in quotes?

swryan: added support for quoted multi-word search terms

PR #938, opened 2019-05-23
adding comm to Optimization args in pyoptsparse_driver

naylor-b:


swryan:
we need a test for this?

PR #943, opened 2019-05-28
attempt to make failing MPI test give us more info

naylor-b:


swryan:
I think you want to merge out and unskip this test.. then we can put this in and see if we can catch a failure

naylor-b:
OK I merged out and unskipped the test.

PR #950, opened 2019-05-31
discrete vars now show up in view_model

naylor-b:


naylor-b:
I asked @JustinSGray about using a different symbol for discrete vars before I finished the story, and he wanted to wait and possibly do that later.

PR #955, opened 2019-06-07
New error raise and unit test for missing objective when using ScipyOptimizer

DKilkenny:


DKilkenny:
Made addition to pyoptsparse and genetic as well as unit tests for each.

openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -345,6 +345,34 @@ def test_unsupported_equality(self):
...
+
+        with self.assertRaises(Exception) as raises_msg:
+            prob.run_driver()
+
+        exception = raises_msg.exception
+
+        msg = "ScipyOptimizeDriver requires objective to be declared"
+
+        self.assertEqual(exception.args[0], msg)
+
```
Kenneth-T-Moore: Some extra carriage returns in these tests. 2 blank lines should only be at the start of a new class.

DKilkenny: Fixed in latest push

openmdao/core/driver.py
```diff
@@ -9,6 +9,7 @@
 from six import iteritems, itervalues, string_types
 
 import numpy as np
+import re
```
naylor-b: doesn't look like you added any regex here.  Do you need this import?

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -221,6 +221,9 @@ def run(self):
         Pm = self.options['Pm']  # if None, it will be calculated in execute_ga()
         Pc = self.options['Pc']
 
+        # Check for missing objective
```
naylor-b: this comment is unnecessary.  The function name very clearly indicates what's happening.

PR #958, opened 2019-06-12
Diodeissue

garobed1:
I replaced the incorrect Shockley diode equation as written:

`I = Is * np.exp(deltaV / Vt - 1)`

with the correct equation:

`I = Is * (np.exp(deltaV / Vt) - 1)`

in the test suite and scripts directories. I also edited the derivative calculations and the expected errors for the unit test in `test_circuit_analysis_derivs.py`.

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1310,3 +1310,10 @@ class _DictValues(object):
 
     def __len__(self):
         return len(self._dict)
+
+    def items(self):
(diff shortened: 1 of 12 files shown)


swryan:
Could you sync your branch up with master again?

I'm not sure why this PR is showing changes from a previous PR that has already been merged..
Something is not quite right.  Hopefully merging out again will fix it.

PR #959, opened 2019-06-13
handle discrete vars in list_input/list_output functions

swryan:
- also did a bit of refactoring of the code that writes the variable table


openmdao/core/system.py
```diff
@@ -3124,57 +3161,68 @@ def _write_outputs(self, in_or_out, comp_type, outputs, hierarchical, print_arra
...
+                                var_dict[name]['resids'] = \
+                                    np.append(var_dict[name]['resids'],
+                                              proc_vars[name]['resids'])
+
+        # get list of var names in execution order, based on the order subsystems were setup
+        var_list = []
+        in_or_out = 'input' if var_type is 'input' else 'output'
+        for subsys in self._subsystems_allprocs:
+            for n in self._var_allprocs_abs_names[in_or_out]:
+                if n in var_dict and n.startswith(subsys.pathname):
```
naylor-b: You're looping over _subsystems_allprocs here, but subsys.pathname will only be defined properly if a subsystem is local.  subsys.name *will* be properly defined though, so maybe  if n startswith '.'.join((self.pathname, subsys.name)) + '.'  ?

openmdao/core/system.py
```diff
@@ -3124,57 +3161,68 @@ def _write_outputs(self, in_or_out, comp_type, outputs, hierarchical, print_arra
...
+        var_dict = OrderedDict()
+        for name, vals in var_data:
+            var_dict[name] = vals
 
-        # If parallel, gather up the outputs. All procs must call this
+        # If parallel, gather up the vars. All procs must call this
         if MPI:
             # returns a list, one per proc
-            all_dict_of_outputs = self.comm.gather(dict_of_outputs, root=0)
+            all_var_dict = self.comm.gather(var_dict, root=0)
```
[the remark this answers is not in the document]
swryan: done

openmdao/core/system.py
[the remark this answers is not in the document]
swryan: yup.

swryan: done.  needs a test.

openmdao/core/system.py
```diff
@@ -3212,14 +3212,21 @@ def _write_table(self, var_type, var_data, hierarchical, print_arrays, out_strea
...
         for subsys in self._subsystems_allprocs:
-            for n in self._var_allprocs_abs_names[in_or_out]:
-                if n in var_dict and n.startswith(subsys.pathname):
-                    var_list.append(n)
-            for n in self._var_allprocs_discrete[in_or_out]:
-                if n in var_dict and n.startswith(subsys.pathname):
-                    var_list.append(n)
+            # subsys.pathname will only be defined properly if a subsystem is local,
+            # but subsys.name will be properly defined.
+            path = '.'.join((self.pathname, subsys.name)) if self.pathname else subsys.name + '.'
```
naylor-b: I think we need parens around `'.'.join( ... else subsys.name`.  Otherwise we're missing the trailing '.' if the first part of the `if` is True.

swryan: already fixed in the next commit... sorry about that

PR #962, opened 2019-06-17
modify summary command to list nonlinear and linear constraints separately

swryan:


openmdao/test_suite/scripts/circle_opt.py
```diff
@@ -1,73 +1,75 @@
```
Kenneth-T-Moore: Can you modify `circle_opt.py` to use `import openmdao.api as om`?  This is to head off a merge conflict with my branch since there is major structural redesign here.

swryan: Done..  although this script doesn't actually appear in the docs so you wouldn't necessarily have had to modify it.

PR #965, opened 2019-06-18
Removed under_mpirun() and replaced with an attempt to load mpi4py, g…

tadkollar:
…overned by user-set environment variable

### DIFF
--- a/openmdao/utils/mpi.py
+++ b/openmdao/utils/mpi.py
@@ -57,35 +57,28 @@ def use_proc_files():
             ofile = open(os.path.join(working_dir, "%d.out" % MPI.COMM_WORLD.rank), 'wb')
         _redirect_streams(ofile.fileno())
 
-# Determine whether OPENMDAO_REQUIRE_MPI is set. Use 'try' as the default value. 
-use_mpi = 'try'
 
+# Attempt to import mpi4py.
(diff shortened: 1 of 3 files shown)


openmdao/utils/mpi.py
```diff
@@ -57,31 +57,37 @@ def use_proc_files():
             ofile = open(os.path.join(working_dir, "%d.out" % MPI.COMM_WORLD.rank), 'wb')
         _redirect_streams(ofile.fileno())
 
+# Determine whether OPENMDAO_REQUIRE_MPI is set. Use 'try' as the default value. 
+use_mpi = 'try'
+
+if 'OPENMDAO_REQUIRE_MPI' in os.environ:
+    use_mpi = os.environ['OPENMDAO_REQUIRE_MPI']
```
Kenneth-T-Moore: Going forward, do we need to put `OPENMDAO_REQUIRE_MPI` in our environments? If so, we should probably add that requirement to the mpi documentation. (however, I don't think we have any general mpi/parallel documentation...)

tadkollar: It's not required - the way it's implemented, it will try to use MPI by default, but revert back if it can't import mpi4py or MPI.COMM_WORLD.size == 1. The env var gives the user some control over the behavior

PR #970, opened 2019-06-19
Added some clarification to spots where we do driver scaling.

Kenneth-T-Moore:


openmdao/core/driver.py
```diff
@@ -458,8 +458,9 @@ def _get_voi_val(self, name, meta, remote_vois, unscaled=False, ignore_indices=F
         remote_vois : dict
             Dict containing (owning_rank, size) for all remote vois of a particular
             type (design var, constraint, or objective).
-        unscaled : bool
-            Set to True if unscaled (physical) design variables are desired.
+        driver_scaling : bool
+            When True, return values that are scaled by driver scaling that was specified in calls
+            to add_design_var, add_objective, and add_constraint.
```
swryan: I think this does not quite fully clarify things...  those functions are on `System` and I think what you are referring to here are the `scaler` and `adder` args to those functions?

Perhaps it would be more clear if the docstrings for those arguments to the System functions referred to their use by the driver... something like: 
```
        adder : float or ndarray, optional
            Value to add to the model value to get the scaled value used by the driver. 
            Adder is applied before scaler (if any).
        scaler : float or ndarray, optional
            Value to multiply the model value to get the scaled value used by the driver. 
            Scaler is applied after adder (if any).
```
And then this docstring could say something like:
```
        driver_scaling : bool
            When True, return values that are scaled according to the adder and scaler values specified when 
            the add_design_var, add_objective, and add_constraint functions were called on the model.
```
Is that correct?  If it is, it seems to more explicitly explain the relationship and functionality of these args...

Kenneth-T-Moore: It is mostly correct, but you can also specify scaling with ref and ref0. I have modified the docstring to reflect both.

PR #971, opened 2019-06-20
fixed cache_linear_solution example

naylor-b:


openmdao/core/tests/test_feature_cache_linear_solution.py
```diff
@@ -105,11 +105,11 @@ def solve_linear(self, d_outputs, d_residuals, mode):
 
         assert_rel_error(self, p['states'], [-0.26794919, -4.], 1e-6)
 
-        derivs = p.compute_totals(of=['states'], wrt=['a'])
+        derivs = p.driver._compute_totals(of=['states'], wrt=['a'], global_names=False)
```
swryan: I'm a little bothered that this is a feature doc and we are telling the user to call an internal/underscored function.
Is there a more appropriate API-level way to show this? Can you just `run_driver`?

naylor-b: I'll modify the test so that calling run_driver() will do a couple of iterations and print out the cached linear solution as it goes.

naylor-b: OK I updated the example so it just does a run_driver.

PR #973, opened 2019-06-21
Added normalization to multifi_cokriging

Kenneth-T-Moore:
Note: this surrogate is an experimental external contribution and is currently not in the docs. Story is up to finish it out.


Kenneth-T-Moore:
Turning on or off normalization shouldn't matter for these small well-conditioned cases.  The switch to turn it off is mostly so the tests where we set the hyperparameters directly, since they would have different values with normalized data. I didn't want to change the numbers in the test since I don't know their provenance.  I don't expect anyone should ever turn this off unless they are already normalizing their values.

I could add a test with a large random dataset that blows up the choleski solve if you don't normalize it.

PR #976, opened 2019-06-24
Fixed bug in PetscVector norm calculation, which was totally wrong.

Kenneth-T-Moore:
Also fixed a bug in `get_slice_dict` when called on a PetscVector under parallel derivatives.


openmdao/vectors/petsc_vector.py
```diff
@@ -84,7 +120,50 @@ def get_norm(self):
...
             norm of this vector.
         """
-        return self._system.comm.allreduce(np.linalg.norm(self._data))
+        system = self._system
+        comm = system.comm
+        if comm.size > 1:
+            dup_slice = self._dup_slice
+            if dup_slice is None:
+
+                # Here, we are find the indices that are not locally owned so that we can
```
naylor-b: typo:  'Here, we are find ...`

Kenneth-T-Moore: Fixed

PR #977, opened 2019-06-25
made it easier for users to control which checks are done during setup

naylor-b:
- also made output from checks go to a file 'openmdao_checks.out' by default (in addition to stdout), so that the check output won't get lost among all of the other optimization output.

### DIFF
--- a/openmdao/docs/theory_manual/setup_stack.rst
+++ b/openmdao/docs/theory_manual/setup_stack.rst
@@ -13,9 +13,19 @@ to inputs, converging solvers, and calculating derivatives. It also includes set
 communicators.
 
 Setup also performs some level of model checking, mainly for critical errors. More extensive model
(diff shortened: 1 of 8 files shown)


openmdao/docs/theory_manual/setup_stack.rst
```diff
@@ -13,9 +13,19 @@ to inputs, converging solvers, and calculating derivatives. It also includes set
 communicators.
 
 Setup also performs some level of model checking, mainly for critical errors. More extensive model
-checking can be done by setting "check" to True when calling `setup`, or by using the :ref:`openmdao command
+checking can be done by setting "check" when calling `setup`, or by using the :ref:`openmdao command
```
swryan: It looks like `check_config` is also a public method now on Problem...  should that be mentioned?

naylor-b: Done.

PR #981, opened 2019-06-28
P166291448 Add a feature that lets users tag variables

hschilling:
These tags can then be used to filter what gets listed when calling list_inputs and list_outputs


openmdao/core/component.py
```diff
@@ -474,6 +477,9 @@ def add_input(self, name, val=1.0, shape=None, src_indices=None, flat_src_indice
         if units is not None and not valid_units(units):
             raise ValueError("The units '%s' are invalid" % units)
 
+        if tags is not None and not isinstance(tags, (str, list)):
+            raise TypeError('The tags argument should be a str, list, or tuple')
```
Kenneth-T-Moore: Probably should add a test with a tuple, just to be safe.

openmdao/core/indepvarcomp.py
```diff
@@ -48,6 +48,13 @@ def __init__(self, name=None, val=1.0, **kwargs):
         self._indep_external = []
         self._indep_external_discrete = []
 
+        if 'tags' not in kwargs:
+            kwargs['tags'] = []
+        else:
+            if isinstance(kwargs['tags'], str):
+                kwargs['tags'] = [kwargs['tags'], ]
+        kwargs['tags'].append('indep_var_comp')  # Tag all indep var comps this way
```
swryan: since we are tagging a var, not a comp, should this just be `indep_var` or even just `indep`?  @JustinSGray

openmdao/core/tests/test_expl_comp.py
```diff
@@ -671,6 +687,130 @@ def compute(self, inputs, outputs):
...
+        ])
+
+        # Outputs with tag that does not match
+        outputs = prob.model.list_outputs(out_stream=None, tags="tag3")
+        self.assertEqual(sorted(outputs), [])
+
+    def test_feature_simple_var_tags(self):
+        from openmdao.api import Problem, ExplicitComponent
+
+        class RectangleCompWithTags(ExplicitComponent):
```
swryan: it would be nice if we had a more realistic example for the docs demonstrating how/why this feature is useful, rather than just these arbitrary tags...

openmdao/core/component.py
```diff
@@ -492,6 +498,10 @@ def add_input(self, name, val=1.0, shape=None, src_indices=None, flat_src_indice
         metadata['desc'] = desc
         metadata['distributed'] = self.options['distributed']
 
+        if isinstance(tags, str):
+            tags = [tags, ]
+        metadata['tags'] = tags
```
naylor-b: We use tags later as a set, so we should just store them internally as a set to begin with.  Also, turning tags into a set removes the reference to the tags list that gets passed in.  That's a good thing since that list could get modified later outside of this component.

openmdao/core/indepvarcomp.py
naylor-b: I second this suggestion.

openmdao/core/system.py
```diff
@@ -2967,7 +2972,19 @@ def list_inputs(self,
         meta = self._var_abs2meta
         inputs = []
 
+        if isinstance(tags, str):
+            tags = [tags, ]
+
         for var_name, val in iteritems(self._inputs._views):  # This is only over the locals
+
+            # Filter based on tags
+            if tags:
```
naylor-b: Might consider making this filtering based on tags code into a function since it's replicated multiple times.

openmdao/core/indepvarcomp.py
JustinSGray: yeah, should be `indep_var` for sure. We'll have to add the same tag to some dymos stuff ....  of course, for that one I think I need to tag only the first index of an array variable... welp, thats another wrinkle for a future story

PR #994, opened 2019-07-10
added check of var sizes when loading a pre-existing coloring

naylor-b:


openmdao/core/tests/test_coloring.py
```diff
@@ -942,6 +942,193 @@ def test_multipoint_with_coloring(self):
...
+            if fixed:
+                p.driver.use_fixed_coloring()
+
+        indeps = model.add_subsystem('indeps', om.IndepVarComp())
+        for name, sz in zip(wrtnames, sizes):
+            indeps.add_output(name, val=np.ones(sz))
+            model.add_design_var('indeps.' + name)
+
+        for name, sz in zip(ofnames, sizes):
+            model.add_constraint('comp.' + name, lower=0.0)
```
swryan: `sz` is not used

openmdao/utils/coloring.py
```diff
@@ -395,27 +391,75 @@ def _check_config_partial(self, system):
...
+            for i, (my_sz, sz) in enumerate(zip(self._col_var_sizes, wrt_sizes)):
+                if my_sz != sz:
+                    changed_sizes.append(wrt_names[i])
+
+        if changed_sizes:
+            msg.append('The following variables have changed sizes: %s.' % sorted(changed_sizes))
+
+        if len(msg) > 1:
+            msg.append(msg_suffix)
+            raise RuntimeError(' '.join(msg))
```
swryan: Would `'n'.join` be better?
Seeing the long unbroken messages in the tests, wonder if line breaks would make it more readable?

openmdao/core/tests/test_coloring.py
naylor-b: fixed

openmdao/utils/coloring.py
naylor-b: fixed

PR #1005, opened 2019-07-19
fix for a bug discovered in aviary model running under MPI

naylor-b:
- with this fix, the total derivatives now agree between the serial and MPI versions of the aviary model (and the test model).

### DIFF
--- a/openmdao/core/tests/test_aviary_mpi_bug.py
+++ b/openmdao/core/tests/test_aviary_mpi_bug.py
@@ -1,0 +1,152 @@
+from __future__ import print_function, division, absolute_import
+
+import sys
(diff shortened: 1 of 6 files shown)


openmdao/core/tests/test_aviary_mpi_bug.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import print_function, division, absolute_import
+
+import sys
```
Kenneth-T-Moore: Most of these imports aren't used.

naylor-b: fixed

PR #1008, opened 2019-07-26
Implement `get_val` on Case object, similar to function on `Problem`

swryan:
User can call `get_val` for data on CaseReader inputs and outputs to specify a specific unit they want the value returned in


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -1108,6 +1106,60 @@ def test_getitem(self):
...
+        prob = om.Problem(model)
+        prob.setup()
+        prob.run_model()
+        prob.cleanup()
+
+        cr = om.CaseReader(self.filename)
+
+        case = cr.get_case(0)
+
+        assert_rel_error(self, prob.get_val('comp.x'), 77.0, 1e-6)
```
JustinSGray: should this test have used `case.get_val()` instead of `prob.get_val()`?

swryan:
messed up the test..   fixing.

PR #1017, opened 2019-08-05
Added Goldstein condition in line search

onodip:
- User can choose between Armijo or Goldstein stopping condition (Nocedal p. 33-36), Armijo is default. 
- The assumptions for the directional derivative are explained a bit more.
- Bounds are added to some of the user options.
- Simple test added for Goldstein.

The _c_ constant should also have an upper bound of 1 for Armijo and 1/2 for Goldstein, but now higher values are used in some of the tests.

### DIFF
--- a/openmdao/solvers/linesearch/backtracking.py
+++ b/openmdao/solvers/linesearch/backtracking.py
@@ -243,9 +243,7 @@ class ArmijoGoldsteinLS(LinesearchSolver):
         opt['maxiter'] = 5
         opt.declare('c', default=0.1, lower=0.0,
                     desc="Slope parameter for line of sufficient decrease. The "
-                    "larger the step, the more decrease is required to terminate the line search. "
-                    "This parameter is 'c' in: '||res_k|| < ||res_0|| (1 - c * alpha)' for the "
-                    "termination criterion.")
+                    "larger the step, the more decrease is required to terminate the line search.")
         opt.declare(
             'bound_enforcement', default='vector', values=['vector', 'scalar', 'wall'],
             desc="If this is set to 'vector', the entire vector is backtracked together " +
--- a/openmdao/solvers/linesearch/backtracking.py
+++ b/openmdao/solvers/linesearch/backtracking.py
@@ -307,7 +307,7 @@ class ArmijoGoldsteinLS(LinesearchSolver):
         fval : float
             Current line search objective value.
         method : str, optional
-            Methodd to caculate stopping condition. Can be "Armijo" or "Goldstein".
+            Method to calculate stopping condition. Can be "Armijo" or "Goldstein".
 
         Returns
         -------
@@ -322,11 +322,11 @@ class ArmijoGoldsteinLS(LinesearchSolver):
         if method == 'armijo':
             return fval <= fval0 + c1 * alpha * df_dalpha
         elif method == 'goldstein':
-            return fval0 + (1-c1) * alpha * df_dalpha <= fval <= fval0 + c1 * alpha * df_dalpha
+            return fval0 + (1 - c1) * alpha * df_dalpha <= fval <= fval0 + c1 * alpha * df_dalpha
 
     def _update_step_length_parameter(self, rho):
         """
-        Updates the step length parameter by multiplying with the contraction factor.
+        Update the step length parameter by multiplying with the contraction factor.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


openmdao/solvers/linesearch/backtracking.py
```diff
@@ -291,13 +295,54 @@ def _single_iteration(self):
...
+
+        The initial line search objective and the step length parameter are stored in the class
+        instance.
+
+        Parameters
+        ----------
+        fval : float
+            Current line search objective value.
+        method : str, optional
+            Methodd to caculate stopping condition. Can be "Armijo" or "Goldstein".
```
naylor-b: typo: Methodd

openmdao/solvers/linesearch/backtracking.py
```diff
@@ -251,13 +253,15 @@ def _declare_options(self):
...
-        opt.declare('alpha', default=1.0, desc="Initial line search step.")
+        opt.declare('rho', default=0.5, lower=0.0, upper=1.0, desc="Contraction factor.")
+        opt.declare('alpha', default=1.0, lower=0.0, desc="Initial line search step.")
         opt.declare('print_bound_enforce', default=False,
                     desc="Set to True to print out names and values of variables that are pulled "
                     "back to their bounds.")
         opt.declare('retry_on_analysis_error', default=True,
                     desc="Backtrack and retry if an AnalysisError is raised.")
+        opt.declare('method', default='Armijo', values=['Armijo', 'Goldstein'],
+                    desc="Backtrack and retry if an AnalysisError is raised.")
```
Kenneth-T-Moore: Description needs updated.

openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -656,7 +656,7 @@ def test_with_subsolves(self):
         ls = model.nonlinear_solver.linesearch = om.ArmijoGoldsteinLS(bound_enforcement='vector')
 
         # This is pretty bogus, but it ensures that we get a few LS iterations.
-        ls.options['c'] = 100.0
+        ls.options['c'] = 100.0  # FIXME c should be 0 <= c <= 1
```
Kenneth-T-Moore: I will put up a story to fix this test so that `c` can be given an upper value of 1.0.

openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -1054,6 +1054,37 @@ def test_feature_armijo_print_bound_enforce(self):
...
+        top.model.add_subsystem('px', om.IndepVarComp('x', np.ones((3, 1))))
+        top.model.add_subsystem('comp', ImplCompTwoStatesArrays())
+        top.model.connect('px.x', 'comp.x')
+
+        top.model.nonlinear_solver = om.NewtonSolver()
+        top.model.nonlinear_solver.options['maxiter'] = 10
+        top.model.linear_solver = om.ScipyKrylov()
+
+        ls = top.model.nonlinear_solver.linesearch = om.ArmijoGoldsteinLS(bound_enforcement='vector')
+        ls.options['bound_enforcement'] = 'vector'
```
swryan: was there a reason for setting this option twice?  (passing the arg is the same as setting the option directly)

openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -1054,6 +1054,37 @@ def test_feature_armijo_print_bound_enforce(self):
...
+
+        # Test lower bounds: should go to the lower bound and stall
+        top['px.x'] = 2.0
+        top['comp.y'] = 0.
+        top['comp.z'] = 1.6
+        top.run_model()
+
+        for ind in range(3):
+            assert_rel_error(self, top['comp.z'][ind], [1.5], 1e-8)
+        self.assertEqual(ls.options['method'], 'Goldstein')
```
swryan: wondering why you are checking the option value here...
is there some reason to fear that the option would change during the run?

onodip: I removed it. There was a problem with _alpha_ before, but that changed in the class attribute. I guess the options shouldn't be touched anywhere.

openmdao/solvers/linesearch/tests/test_backtracking.py
onodip: Indeed. It was copied from an older test. There were 2 old tests, where this options was set twice, removed from all.

PR #1023, opened 2019-08-08
Fixed list_outputs to print vars when component is the root model

tadkollar:


openmdao/core/system.py
```diff
@@ -3288,16 +3288,26 @@ def _write_table(self, var_type, var_data, hierarchical, print_arrays, out_strea
...
+            # For components with no children, self._subsystems_allprocs is empty.
             for var_name in real_vars:
-                if var_name in var_dict and var_name.startswith(path):
+                if var_name in var_dict:
                     var_list.append(var_name)
+
             for var_name in disc_vars:
-                if var_name in var_dict and var_name.startswith(path):
+                if var_name in var_dict:
                     var_list.append(var_name)
```
Kenneth-T-Moore: Needs to have a test.

openmdao/core/tests/test_root_comp_list_outputs.py
```diff
@@ -0,0 +1,59 @@
...
+        prob.setup()
+
+        prob['comp1.a'] = 1.
+        prob['comp1.b'] = -4.
+        prob['comp1.c'] = 3.
+        prob.run_model()
+
+        prob2 = om.Problem(model=comp1)
+        stream2 = cStringIO()
+        outputs2 = prob2.model.list_outputs(hierarchical=False, out_stream=stream2)
```
Kenneth-T-Moore: I don't think this is a typical use of openmdao. The component you slotted into the prob2 already has some datastructures set up from being used in prob1.  I would create a clean model with a newly instantiated component at the root, call setup on the problem, and test with that.

(NOTE: disregard if Justin indicated that this actually was his use-case).

tadkollar: OK, I'll take a look... it does trigger the problem in unpatched openmdao though.

PR #1027, opened 2019-08-12
Added example of discrete var with implicit comp

hschilling:


openmdao/core/tests/test_discrete.py
```diff
@@ -718,6 +718,59 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
...
+                x = outputs['x']
+
+                partials['x', 'a'] = x ** 2
+                partials['x', 'b'] = x
+                partials['x', 'x'] = 2 * a * x + b
+
+            def guess_nonlinear(self, inputs, outputs, resids, discrete_inputs, discrete_outputs):
+                # Default initial state of zero for x takes us to x=1 solution.
+                # Here we set it to a value that will take us to the x=3 solution.
+                outputs['x'] = 5
```
swryan: I think you should remove the `guess` function from this example..  
It's extraneous for this feature and just makes the example more complicated for the reader.

Kenneth-T-Moore: We iterated on this, but the guess_nonlinear is necessary so that the user knows that its signature includes the discrete_inputs and discrete_outputs arguments.

PR #1031, opened 2019-08-20
n2 viewer now shows Solvers in the solver hierarchy when attached to an implicit comp

tadkollar:
Do we make tests for this sort of thing for N2?


hschilling:
> Do we make tests for this sort of thing for N2?

Since the fix is on the Python side, I would say yes.

PR #1037, opened 2019-08-30
UnstructuredMetaModel Visualization

DKilkenny:


openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+
+        # Layout creation
+        self.layout = row(self.contour_data(), self.right_plot(), self.sliders_and_selects)
+        self.layout2 = row(self.bot_plot())
+        curdoc().add_root(self.layout)
+        curdoc().add_root(self.layout2)
+        curdoc().title = 'MultiView'
+
+    def make_predictions(self, data):
```
JustinSGray: is this supposed to be a method that a user would call? or an internal function?

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+
+        # Parse dict into shape [n**2, number of inputs] list
+        inputs = np.empty([self.resolution**2, self.nx])
+        for idx, values in enumerate(data.values()):
+            inputs[:, idx] = values.flatten()
+
+        # Pair data points with their respective prob name. Loop to make predictions
+        for idx, tup in enumerate(inputs):
+            for name, val in zip(data.keys(), tup):
+                self.prob[self.surrogate_comp.name + '.' + name] = val
```
JustinSGray: there are a few more efficient ways to handle this kind of string manipulation. for python 3, format-strings are the fastest. For backwards compatibility with python 2, the `.format()` function is the better option. 

this approach is the slowest, unfortunately

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+        ye = np.zeros((n, n, self.ny))
+
+        # Query the slider dictionary, append the name and current value to the ordered dictionary
+        self.slider_value_and_name = OrderedDict()
+        for title, slider_params in self.slider_dict.items():
+            self.slider_value_and_name[title] = slider_params.value
+        
+        # Cast the current values of the slider_value_and_name dictionary values to a list
+        self.input_point_list = list(self.slider_value_and_name.values())
+        for ix in range(self.nx):
```
JustinSGray: this looks like it could probably be done more quickly with a full broadcast, rather than a for loop

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+        self.input_list = [self.x_input.value, self.y_input.value]
+        for title in self.slider_value_and_name.keys():
+            if title == self.x_input.value or title == self.y_input.value:
+                pass
+            else:
+                self.input_list.append(title)
+        
+        # Append the key (input_list) and the values copied from xe to pred_dict 
+        # where it is then ordered in pred_dict_ordered.
+        for idx, title in enumerate(self.slider_value_and_name.keys()):
```
JustinSGray: I think you can just do this without the for loop by passing `self.slider_value_and_name.items()` directly to the update method on the dict

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+        # Set the first two columns of the points array to x/y inputs, respectively
+        points[:, self.x_index] = self.input_point_list[self.x_index]
+        points[:, self.y_index] = self.input_point_list[self.y_index]
+        points = np.divide(points, self.limit_range)
+        tree = cKDTree(points)
+        dist_limit = np.linalg.norm(self.dist_range * self.limit_range)
+        scaled_x0 = np.divide(self.input_point_list, self.limit_range)
+        # Query the nearest neighbors tree for the closest points to the scaled x0 array
+        dists, idx = tree.query(scaled_x0, k=len(xt), distance_upper_bound=dist_limit)
+        idx = idx[idx != len(xt)]
```
JustinSGray: this for loop looks wonky to me. Why do you assign idx like this on line 514, then use an if condition in the for loop right after.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,551 @@
...
+        if self.input_dropdown_checks(self.x_input.value, new) == False:
+            raise ValueError("Inputs should not equal each other")
+        else: 
+            self.update_all_plots()
+
+    def output_value_update(self, attr, old, new):
+        self.output_variable = self.output_list.index(new)
+        self.update_all_plots()
+
+    def training_points(self):
```
JustinSGray: I would suggest that you write a unit test for this function specifically, since it seems easily isolatable functionality

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,553 @@
...
+                info[0:2] = infos[i, :]
+                info[2] = dists[dist_index] / dist_limit
+                info[3] = y_training[i, output_variable]
+                info[4] = (1. - info[2] / self.dist_range) ** 0.5
+                data[dist_index] = info
+
+        return data
+
+    def stack_outputs(self, outputs_dict):
+        """
```
Kenneth-T-Moore: This can be a function (instead of a class method) since it doesn't use anything in `self`.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,553 @@
...
+        # Parse dict into shape [n**2, number of inputs] list
+        inputs = np.empty([self.resolution**2, self.num_of_inputs])
+        for idx, values in enumerate(data.values()):
+            inputs[:, idx] = values.flatten()
+
+        # Pair data points with their respective prob name. Loop to make predictions
+        for idx, tup in enumerate(inputs):
+            for name, val in zip(data.keys(), tup):
+                self.prob[self.surrogate_comp.name + '.' + name] = val
+            self.prob.run_model()
```
Kenneth-T-Moore: We might need to discuss some use-cases for this visualizer, but if the user gives you a prob object that has a ton of other calculations in it, then this will be needlessly slow. An alternative might be to create a new clean `Problem` instance, and add self.surrogate_comp to it, along with appropriate indepvarcomps for the inputs.  Then, when it runs, you are only running the surrogate model.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,553 @@
...
+        for idx, title in enumerate(self.slider_value_and_name.keys()):
+            pred_dict.update({title: x_data[:, :, idx]})
+        pred_dict_ordered = OrderedDict((k, pred_dict[k]) for k in self.input_list)
+
+        # Pass the dict to make predictions and then reshape the output to (n, n, number of outputs)
+        y_data[:, :, :] = self.make_predictions(pred_dict_ordered).reshape(
+            (resolution, resolution, self.num_of_outputs))
+        Z = y_data[:, :, self.output_variable]
+        Z = Z.reshape(resolution, resolution)
+        self.Z = Z
```
Kenneth-T-Moore: self.Z isn't listed in the class attributes docstring.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,553 @@
...
+        Returns
+        -------
+        array
+            np.stack of values
+        """
+        output_lists_to_stack = []
+        for values in outputs_dict.values():
+            output_lists_to_stack.append(np.asarray(values))
+
+        return np.stack(output_lists_to_stack, axis=-1)
```
Kenneth-T-Moore: Another useful tool you can use is `vulture`, which you can pip install.  You can run it on a file (or a group of files) and It can find dead code such as variables or methods that are assigned and not used.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,553 @@
+"""Define output of Meta Models and visualize the results."""
```
Kenneth-T-Moore: When you made your initial branch, you should have made it on your repository, not the OpenMDAO one.  (This PR comes from OpenMDAO/mybug1.)  I think for continuity, you should close the PR and resubmit it by first pushing the branch up to your repo, then performing a PR from there.  But first, address the changes here so we don't lose them.

Going to pull down and play with the viewer, so i may have further comments.

openmdao/visualization/meta_model_viewer/tests/test_unstruct.py
```diff
@@ -0,0 +1,68 @@
+import unittest
+
+import numpy as np
+import openmdao.api as om
+from openmdao.visualization.meta_model_viewer.meta_model_visualization import UnstructuredMetaModelVisualization
```
Kenneth-T-Moore: I think we can add UnstructuredMetaModelVisualization to openmdao.api.

Also, why not just call it MetaModelVisualization.  We can detect which kind of metamodel it is in init, and run it accordingly.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: I can't do that because I rely on the idx from enumerate to append the data from `x_data` to the dict values.

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: This is a great tool. There were few lines I was able to delete.

setup.py
```diff
@@ -141,6 +142,7 @@
         'pyparsing',
         'scipy',
         'six',
+        'bokeh'
```
swryan: I wonder if we should add a new `visualization` section under `optional_dependencies`... 
I don't think we want this as a core dependency? (would have to add appropriate handling of failed import)

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,554 @@
...
+        for name, slider_object in self.slider_dict.items():
+            if name == self.x_input.value:
+                self.x_input_slider = slider_object
+                self.x_input_slider.on_change('value', self._scatter_plots_update)
+            elif name == self.y_input.value:
+                self.y_input_slider = slider_object
+                self.y_input_slider.on_change('value', self._scatter_plots_update)
+            else:
+                setattr(self, name, slider_object)
+                obj = getattr(self, name)
```
swryan: It's not clear to me why you are creating these attributes, rather than just doing `slider_object.on_change(...)`
I don't see where they are used?

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,554 @@
...
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        Bokeh figure
+        """
+        for title in self.input_data_dict:
```
swryan: is the loop necessary here?  couldn't you just use `x_input.value` and  `y_input.value` directly? 
```
self.x_data = self.input_data_dict[self.x_input_value]
self.y_value = self.y_input_slider.value
alt_index = np.where(np.around(
                    self.input_data_dict[self.y_input_value], 5) == np.around(self.y_value, 5))[0]
```

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,554 @@
...
+    """
+    Top-level container for the Meta Model Visualization.
+
+    Attributes
+    ----------
+    prob : om.Problem
+        Name of variable corresponding to Problem Component
+    surrogate_comp : MetaModel
+        Name of variable corresponding to Meta Model Component
+    resolution : int
```
swryan: both `prob` and `surrogate_comp` attributes are just object references rather than "name of variable corresponding to..."
also Problem is not a Component...

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
```diff
@@ -0,0 +1,554 @@
...
+
+        # Color bar formatting
+        color_mapper = LinearColorMapper(palette="Viridis11", low=np.amin(Z), high=np.amax(Z))
+        color_bar = ColorBar(color_mapper=color_mapper, ticker=BasicTicker(), label_standoff=12,
+                             location=(0, 0))
+
+        # Contour Plot
+        self.contour_plot = figure(
+            tooltips=[(self.x_input.value, "$x"), (self.y_input.value, "$y"),
+                      (self.output_select.value, "@image")], tools="pan")
```
swryan: could make a local reference here also so you don't have to go through `self.` in the next 10 references

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: I create these attributes because I create a slider for every input so that a user can reduce an n dimensional input space to 3 dimensions. Consider a 4 dimensional space, the user will select 2 dimensions to be a single value input while the remaining 2 dimension input sliders allow the user to make slices in the contour at any given point which are then projected on to the subplots. 

I create them so that each slider has a unique name in the slider_dict which will later be looped through so bokeh can display them

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: This is a `self.` reference because I add to the contour_plot figure in the `bot_plot` and `right_plot` functions

swryan: sorry, I meant that you could make a temporary local reference, something like:
`self.contour_plot = contour_plot = figure(`
so you wouldn't need to go through `self.` for the remaining references in this function

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: Upon further clarification and discussion, I can just use `slider_object.on_change()`

openmdao/visualization/meta_model_viewer/meta_model_visualization.py
DKilkenny: Ah ok. I'll make that change

setup.py
DKilkenny: Yeah, I think that's reasonable to create a new section of optional dependencies

openmdao/utils/om.py
```diff
@@ -229,6 +232,88 @@ def _viewconns(prob):
...
+            script_path, file_name)
+    os.system(bash_command)
+
+
+def _hidden_func_parser(parser):
+
+    parser.add_argument('file', nargs=1, help='Python file containing the model.')
+    parser.add_argument('-s', '--surrogate_model_ref', action='store', dest='sur_ref',
+                        help='Variable name of the surrogate model reference.')
+
```
swryan: I think this hidden_func is not necessary... see slack and PR#1045

PR #1038, opened 2019-08-30
DirectSolver now works under MPI

naylor-b:
- also fixed a few other MPI-related issues
- modified BroydenSolver to raise an exception when used with comm.size > 1. Put a story in to fix BroydenSolver under MPI.
- added a tracer function that will dump openmdao, mpi4py, and petsc4py function calls in order to more quickly zero in on the problem when an MPI test is hanging.


openmdao/vectors/petsc_vector.py
```diff
@@ -123,41 +143,29 @@ def get_norm(self):
...
-                    owning_rank = system._owning_rank[name]
-                    if not abs2meta[name]['distributed'] and owning_rank != system.comm.rank:
-                        dup_slice.extend(range(idx_slice.start, idx_slice.stop))
-
-                self._dup_slice = dup_slice
+            dup_inds = self._get_dup_inds()
 
             if self._ncol == 1:
-                data_cache = self._data.copy()
+                data_cache = self._data.copy().real
```
Kenneth-T-Moore: Why are we only caching and restoring the real part? Doesn't this remove the imaginary part when we are under complex step?

openmdao/solvers/nonlinear/broyden.py
```diff
@@ -13,6 +13,7 @@
 from openmdao.solvers.solver import NonlinearSolver
 from openmdao.utils.class_util import overrides_method
 from openmdao.utils.general_utils import simple_warning
+from openmdao.utils.mpi import MPI
```
Kenneth-T-Moore: Is the MPI import used?

openmdao/vectors/petsc_vector.py
naylor-b: I don't think it actually matters what we assign the self._petsc.array to, because we always reassign it later before we do a scatter or whatever.  Under complex step we treat the real and imaginary parts separately as real, then combine them into a complex array when we assign them to the local data vector.  When I updated get_norm I only assigned the real part because it didn't seem right to assign a complex array to the petsc array attribute when we created it using a real array.  Technically though, I guess we don't really need to reset the petsc array at all in get_norm, because it will get set to the proper array later prior to any distriuted operations.

Kenneth-T-Moore: ok, i forgot that we did things differently in the petsc array.

openmdao/solvers/linear/direct.py
```diff
@@ -10,12 +10,55 @@
...
+        Either 'row' or 'col'.
+    loc : int
+        Index of row or column.
+
+    Returns
+    -------
+    str
+        New error string.
+    """
+    start = end = 0
```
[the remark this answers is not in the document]
naylor-b: Turns out I had not updated the singular matrix error code or the NaN error code to handle the MPI case.  Should now be fixed.

openmdao/solvers/nonlinear/broyden.py
naylor-b: Nope.  Removed.

PR #1042, opened 2019-09-10
 list_inputs/list_outputs on Case now prints in execution order

swryan:
- list_inputs/list_outputs on Case now prints in execution order
- list_inputs/list_outputs on Case now prints the appropriate system name
- list_inputs/list_outputs tag processing was very slightly refactored

### DIFF
--- a/openmdao/core/indepvarcomp.py
+++ b/openmdao/core/indepvarcomp.py
@@ -7,7 +7,7 @@ import collections
 from six import string_types
 
 from openmdao.core.explicitcomponent import ExplicitComponent
-from openmdao.utils.general_utils import warn_deprecation, convert_user_defined_tags_to_set
+from openmdao.utils.general_utils import warn_deprecation, make_set
 
 
 class IndepVarComp(ExplicitComponent):
@@ -49,13 +49,11 @@ class IndepVarComp(ExplicitComponent):
         self._indep_external_discrete = []
 
         if 'tags' not in kwargs:
-            kwargs['tags'] = set()
+            kwargs['tags'] = {'indep_var'}
         else:
             if not isinstance(kwargs['tags'], (str, set, list)):
                 raise TypeError('The tags argument should be a str, set, or list')
-
-        kwargs['tags'] = convert_user_defined_tags_to_set(kwargs['tags'])
-        kwargs['tags'].add('indep_var')  # Tag all indep var comps this way
+            kwargs['tags'] = make_set(kwargs['tags']).add('indep_var')
 
         # A single variable is declared during instantiation
         if isinstance(name, string_types):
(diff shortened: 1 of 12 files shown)


openmdao/recorders/sqlite_recorder.py
```diff
@@ -361,6 +363,7 @@ def startup(self, recording_requester):
             var_settings.update(objectives)
             var_settings.update(constraints)
             var_settings = self._cleanup_var_settings(var_settings)
+            var_settings['execution_order'] = var_order
```
Kenneth-T-Moore: Should we update the case version for this change, and add a backwards-compatibility test?

swryan: We have not been changing the version when the content of the metadata changes.. only when the data format changes.
The presence of the data field is not required, and it will fall back to a sorted order.  I'll add a test for that branch of the code.

swryan: added backwards-compatibility test

openmdao/core/indepvarcomp.py
```diff
@@ -158,11 +156,12 @@ def add_output(self, name, val=1.0, shape=None, units=None, res_units=None, desc
...
 
-        if tags is not None and not isinstance(tags, (str, list)):
-            raise TypeError('The tags argument should be a str or list')
-
-        tags = convert_user_defined_tags_to_set(tags)
-        tags.add('indep_var')
+        if tags is None:
+            tags = {'indep_var'}
+        else:
+            if not isinstance(tags, (str, set, list)):
```
naylor-b: Do we need this check since make_set will raise its own exception if tags is not of the proper type?

openmdao/core/indepvarcomp.py
```diff
@@ -186,10 +185,12 @@ def add_discrete_output(self, name, val, desc='', tags=None):
...
             list_outputs.
         """
-        if tags is not None and not isinstance(tags, (str, list)):
-            raise TypeError('The tags argument should be a str or list')
-
-        tags = convert_user_defined_tags_to_set(tags)
+        if tags is None:
+            tags = {'indep_var'}
+        else:
+            if not isinstance(tags, (str, set, list)):
```
naylor-b: Same comment as above, don't think we need this check.

openmdao/recorders/case.py
```diff
@@ -448,28 +448,26 @@ def list_inputs(self,
         inp_vars = {}
         inputs = []
 
-        tags = convert_user_defined_tags_to_set(tags)
-
         if self.inputs is not None:
             for abs_name in self.inputs.absolute_names():
                 if abs_name not in inp_vars:
                     inp_vars[abs_name] = {'value': self.inputs[abs_name]}
```
naylor-b: I know this isn't part of the PR, but it seems odd here that we are creating a dict with only one item in it.  Why don't we just store self.inputs[abs_name] in inp_vars[abs_name] directly?

swryan: cleaned this up in both `list_inputs` and `list_outputs`

openmdao/core/indepvarcomp.py
swryan: removed and added arg to `make-set` to enable user-friendlier error message

PR #1054, opened 2019-09-19
connection viewer updates

naylor-b:
- added toggle buttons for every column at the bottom of the table so that any column can now be shown/hidden.
- updated connection viewer docs
- updated table docs
- updated summary docs
- tweaked summary output


openmdao/docs/other/om_command.rst
```diff
@@ -89,15 +89,23 @@ openmdao view_connections
...
-output variables in the model.  Units can be compared for each connection, and unconnected inputs
-and outputs can be easily identified.  The displayed variables can be filtered by source system
-and/or target system.  They can also be filtered by NO CONNECTION, which will show all of the
-unconnected inputs or outputs, depending on whether the NO CONNECTION filter is active for the
-source or target side.  When units differ between a source and a target, they are highlighted in
-red, and when inputs are connected to outputs outside of the currently-selected, top-level system,
-they are highlighted in purple.  This highlighting can be used to easily identify variables that are connected
-across group boundaries.  Below is an example of a connection viewer for a pycycle propulsor
-model obtained using the command:
+output variables in the model.  It's primary purpose is to help debug a model by making the following
```
Kenneth-T-Moore: It's -> Its

openmdao/docs/other/om_command.rst
```diff
@@ -110,16 +118,38 @@ model obtained using the command:
 
    An example of a connection viewer.
 
+
+By default the promoted names columns of both inputs and outputs are shown, but in the example
+above, the absolute input names are shown and the promoted input names are hidden.
+
+Unconnected inputs can easily be identified by typing 'NO CONNECTION' or 'NO ', into
+the filter field of either the absolute or promoted output column.  Unconnected outputs can
```
Kenneth-T-Moore: While it is rather unlikely, I believe "NO CONNECTION" is a valid openmdao variable name.   Maybe something like an "!" (or anything restricted in names) can be the shorthand code for showing unconnected inputs.

naylor-b: I noticed when looking at the _valid_var_name function that we were not preventing names like ' ' and '\t', so I just went ahead and made both of those illegal characters.  So now 'NO CONNECTION' is no longer a valid variable name.

openmdao/docs/other/om_command.rst
naylor-b: fixed

openmdao/docs/other/om_command.rst
naylor-b: I removed the restriction on spaces within a variable name and changed 'NO CONNECTIONS' to '[NO CONNECTIONS]', which is not a legal name because of the '[' and ']'.  So once people know what they're looking for, the unconnecteds can be shown by filtering on just '[' or ']'.

PR #1057, opened 2019-09-26
P168431589 user can set specific slices of variables using the set_val method on problem

hschilling:
User wants a way to input slices using the same syntax as you would when accessing a numpy array, e.g.,  arr[2:10, :99]. Based on that, we could provide a helper class, maybe called Slicer or something, that would define a __getitem__ method that would simply return the arg it was passed, allowing things like prob.get_val('comp.x', 'm', indices=Slicer[2:10, :99])

### DIFF
--- a/openmdao/core/tests/test_problem.py
+++ b/openmdao/core/tests/test_problem.py
@@ -1012,19 +1012,6 @@ class TestProblem(unittest.TestCase):
         import numpy as np
         import openmdao.api as om
 
-        # 1D array
-        prob = om.Problem()
-        prob.model.add_subsystem('comp', om.ExecComp('y=x+1.',
-                                                     x={'value': np.array([100.0, 33.3]),},
-                                                     y={'shape': (2, ), }))
-
-        prob.setup()
-        prob.run_model()
-
-        assert_rel_error(self, prob.get_val('comp.x', indices=om.slicer[0]), 100.0, 1e-6)
-        assert_rel_error(self, prob.get_val('comp.x', indices=om.slicer[:1]), 100.0, 1e-6)
-
-        # 2D array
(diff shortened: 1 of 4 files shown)


openmdao/core/tests/test_problem.py
```diff
@@ -1008,6 +1008,25 @@ def test_feature_get_set_array_with_units(self):
...
+        prob = om.Problem()
+        prob.model.add_subsystem('comp', om.ExecComp('y=x+1.',
+                                                     x={'value': np.array([[1., 2.], [3., 4.]]),},
+                                                     y={'shape': (2, 2), }))
+
+        prob.setup()
+        prob.run_model()
+
+        assert_rel_error(self, prob.get_val('comp.x', indices=om.slicer[:,0]), [1., 3.], 1e-6)
+
```
Kenneth-T-Moore: PEP - space after comma.

openmdao/core/tests/test_problem.py
```diff
@@ -1008,6 +1008,25 @@ def test_feature_get_set_array_with_units(self):
...
+        prob.set_val('comp.x', [5., 6.], indices=om.slicer[:,0])
+        assert_rel_error(self, prob.get_val('comp.x', indices=om.slicer[:,0]), [5., 6.], 1e-6)
+        prob.run_model()
+        assert_rel_error(self, prob.get_val('comp.y', indices=om.slicer[:,0]), [6., 7.], 1e-6)
```
Kenneth-T-Moore: I wonder if we should show an example for a single index get too, like [0, 1]. Just to drive home the point that all of the normal slicing syntax is valid.  Could also show that it works with a -1.

PR #1070, opened 2019-10-03
Added ability user to set resolution of view_mm in command line

DKilkenny:
https://www.pivotaltracker.com/story/show/168761376


openmdao/utils/om.py
```diff
@@ -285,11 +287,16 @@ def _view_metamodel(prob):
 
         pathname = options.pathname
         port_number = options.port_number
+        try:
+            resolution = int(options.resolution)
```
naylor-b: when you do the `parser.add_argument` above for resolution, you can set the 'type' arg to int and argparse will automatically type check the value for you.

openmdao/visualization/meta_model_viewer/tests/test_view_mm_cmd_line.py
```diff
@@ -0,0 +1,38 @@
...
+
+@unittest.skipUnless(bokeh, "Bokeh is required")
+class ViewMMCommandLineTest(unittest.TestCase):
+
+    def test_unspecified_metamodel(self):
+        script = os.path.join(os.path.dirname(__file__), 'example.py')
+        cmd = 'openmdao view_mm {}'.format(script)
+        output = subprocess.check_output(cmd.split()).decode('utf-8', 'ignore')
+        expected_output = "Metamodel not specified. Try one of the following: ['interp1', 'interp2']."
+        self.assertTrue(expected_output in output)
```
naylor-b: follow the check in the assertTrue call with a string explaining the failure.  Otherwise, the person running the tests just sees 'False is not true', which tells them nothing about what went wrong.

PR #1074, opened 2019-10-07
added some re-initialization of certain data structures whenever setup is called.

naylor-b:


swryan:
I'm guessing this all has to do with subproblems...  would a test or two be appropriate?

naylor-b:
It's actually for reconfig.  Rob ran into an issue where we weren't re-initializing the connections dictionary and I thought I'd go through and find any other data structs that were also not being re-initialized.  You're right though, I should have added some tests.  I'll close this and reopen after the tests have been added.

PR #1083, opened 2019-10-10
commit to sqlite db after each execute cmd

fzahle:
minor fix to ensure that data is flushed to the sqlite file after each iteration


swryan:
I'm curious if you have run into an issue related to commits... 
My understanding is that the use of the context manager means the explicit commit is not necessary.
```
Connection objects can be used as context managers that automatically commit or rollback transactions. In the event of an exception, the transaction is rolled back; otherwise, the transaction is committed:
```
https://docs.python.org/3/library/sqlite3.html#using-the-connection-as-a-context-manager

PR #1089, opened 2019-10-21
command line updates

naylor-b:
- more flexible hook system
- fixed log level check_config bug
- removed openmdao total_sparsity sub-command and associated code
- added name to Problem so hooks can be associated with a specific Problem instance
- updated Problem error msgs to include the Problem name
- fixed openmdao error messages so now invalid arg errors will display the usage info for the correct sub-command
- changed -j and -jtext in various `openmdao *_coloring` sub-commands to --view and --textview
- made some features optional for certain sub-commands (like tree and view_connections) so they can be called earlier (post setup instead of post final_setup) if those features are not required.
- changed n2 hook to pre final_setup instead of post final_setup


openmdao/utils/om.py
```diff
@@ -563,87 +608,98 @@ def _post_setup_exec(options):
...
+    'view_coloring': (_view_coloring_setup_parser, _view_coloring_exec, None,
+                      'Colored jacobian viewer.'),
+    'view_connections': (_view_connections_setup_parser, _simple_exec, _view_connections_cmd,
+                         'Connection viewer showing values and source/target units.'),
+    'view_mm': (_meta_model_parser, _simple_exec, _meta_model_cmd, "Meta Model Viewer."),
+    'view_model': (_n2_setup_parser, _view_model_cmd, None,
+                   'Display an interactive N2 diagram of the problem. '
+                   '(Deprecated, please use n2 instead.)'),
+    'xdsm': (_xdsm_setup_parser, _xdsm_cmd, None, 'XDSM viewer.'),
 }
```
Kenneth-T-Moore: Asked Bret to make the grammatical structure more parallel in the command line help descriptions.

naylor-b: fixed

PR #1090, opened 2019-10-23
Added error message for empty group promotion

DKilkenny:
https://www.pivotaltracker.com/story/show/169262664

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2021,14 +2021,17 @@ class System(object):
 
             not_found = (set(names).union(renames).union(patterns)) - found
             if not_found:
-                print(not_found)
+                if not self._var_abs2meta and isinstance(self, openmdao.core.group.Group):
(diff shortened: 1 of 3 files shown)


openmdao/core/system.py
```diff
@@ -2020,6 +2020,10 @@ def resolve(to_match, io_types, matches, proms):
                             pmap[name] = gname + name if gname else name
 
             not_found = (set(names).union(renames).union(patterns)) - found
+            if self._var_promotes['any']:
+                if not self._var_abs2meta and isinstance(self, openmdao.core.group.Group):
+                    raise RuntimeError('No inputs or outputs found. Check %s '
```
naylor-b: this error message provides no context.  I would recommend, instead of creating a whole new error message, just use the existing one, and if the group happens to be empty, then add the 'check to make sure it's not empty' at the end of the existing message, since that message already provides all of the other relevant information.

PR #1097, opened 2019-10-30
add pip-tools generated requirements.txt

capsulecorplab:
`requirements.txt` is generated, using the `pip-tools` toolchain, by running `pip-compile` without any arguments. This works by reading `install_requires` in `setup.py`. See [requirements-from-setuppy](https://github.com/jazzband/pip-tools/#requirements-from-setuppy)

To update, run:

    pip-compile --output-file requirements.txt setup.py

Having a verbose `requirements.txt` alongside a minimal `install_requires` is also the recommended configuration in the PyPA user guide. See [install-requires-vs-requirements](https://packaging.python.org/discussions/install-requires-vs-requirements/)


Kenneth-T-Moore:
Based on the link you shared, I wonder if the requirements.txt should be more of an 'ideal' openmdao build, which would include everything you need to make use of all capabilities, so it would perhaps include `matplotlib`, `pyXDSM`, `bokeh`, etc.  I'd also say `mpi4py` and `petsc4py`, but sometimes those are a little trickier to install (or have been in the past).  Then there are packages like `pyoptsparse`, which I don't believe is on pypi, but you can get it from another github repo.  I don't know how to handle those in the requirements.txt format.

onodip:
> Based on the link you shared, I wonder if the requirements.txt should be more of an 'ideal' openmdao build, which would include everything you need to make use of all capabilities, so it would perhaps include `matplotlib`, `pyXDSM`, `bokeh`, etc. I'd also say `mpi4py` and `petsc4py`, but sometimes those are a little trickier to install (or have been in the past). Then there are packages like `pyoptsparse`, which I don't believe is on pypi, but you can get it from another github repo. I don't know how to handle those in the requirements.txt format.

In a "requirements.txt" file you can refer to the repo the same way, as you would with a `pip install` command. For example the line for _pyXDSM_ in the file would be:

`git+https://github.com/mdolab/pyXDSM.git
`

If installing one dependency fails with `pip install -r requirements.txt`, none of the packages listed in the file will be installed. So I am not sure if it wouldn't be better to include only the minimum for the sake of robustness. At the same time it would be nice to see all the optional packages at one place...

capsulecorplab:
> If installing one dependency fails with pip install -r requirements.txt, none of the packages listed in the file will be installed. So I am not sure if it wouldn't be better to include only the minimum for the sake of robustness. At the same time it would be nice to see all the optional packages at one place...

The point of the `requirements.txt` is to provide more deterministic builds by serving as a flattened _lockfile_. So if an installation of a package were to fail using `pip install -r requirements.txt`, it would serve its purpose as you would then know that particular package would need to be rolled back and pinned to a working version, or removed entirely, from `install_requires` - unless it's a broken 3rd party dependency, in which case you should probably not use its parent dependency and/or contact the package author - then use `pip-compile --output-file requirements.txt setup.py` to regenerate the `requirements.txt`.

onodip:
> > If installing one dependency fails with pip install -r requirements.txt, none of the packages listed in the file will be installed. So I am not sure if it wouldn't be better to include only the minimum for the sake of robustness. At the same time it would be nice to see all the optional packages at one place...
> 
> The point of the `requirements.txt` is to provide more deterministic builds by serving as a flattened _lockfile_. So if an installation of a package were to fail using `pip install -r requirements.txt`, it would serve its purpose as you would then know that particular package would need to be rolled back and pinned to a working version, or removed entirely, from `install_requires` - unless it's a broken 3rd party dependency, in which case you should probably not use its parent dependency and/or contact the package author - then use `pip-compile --output-file requirements.txt setup.py` to regenerate the `requirements.txt`.

Okay, so the purpose would be the same as doing a `pip freeze`. In this case only referring to a git repo is questionable. Maybe the best would be to include a tag in this case, like this:

`git+https://github.com/mdolab/pyXDSM.git@1.0
`

sanbales:
> Based on the link you shared, I wonder if the requirements.txt should be more of an 'ideal' openmdao build, which would include everything you need to make use of all capabilities, so it would perhaps include `matplotlib`, `pyXDSM`, `bokeh`, etc. I'd also say `mpi4py` and `petsc4py`, but sometimes those are a little trickier to install (or have been in the past). Then there are packages like `pyoptsparse`, which I don't believe is on pypi, but you can get it from another github repo. I don't know how to handle those in the requirements.txt format.

We were using the `extras_require` as the way to define these additional dependencies.
https://stackoverflow.com/questions/41268863/difference-between-extras-require-and-install-requires-in-setup-py

Not sure how to use pip-compile to look at `extras_require` instead of `install_requires`.

capsulecorplab:
> > Based on the link you shared, I wonder if the requirements.txt should be more of an 'ideal' openmdao build, which would include everything you need to make use of all capabilities, so it would perhaps include `matplotlib`, `pyXDSM`, `bokeh`, etc. I'd also say `mpi4py` and `petsc4py`, but sometimes those are a little trickier to install (or have been in the past). Then there are packages like `pyoptsparse`, which I don't believe is on pypi, but you can get it from another github repo. I don't know how to handle those in the requirements.txt format.
> 
> We were using the `extras_require` as the way to define these additional dependencies.
> https://stackoverflow.com/questions/41268863/difference-between-extras-require-and-install-requires-in-setup-py
> 
> Not sure how to use pip-compile to look at `extras_require` instead of `install_requires`.

At the time of writing, pip-tools ignores `extras_require`. Though, [625](https://github.com/jazzband/pip-tools/issues/625) may suggest alternative workflows.

sanbales:
Sorry for taking so long to get back on this.

Couple of questions for the group:
* Should we be concerned about the bigger CI/CD picture if we start pinning the dependencies?
* Travis and Appveyor pipelines check multiple versions of numpy and other packages, how do we keep those versions and these pinned versions consistent?  Should Travis and Appveyor be using `requirements.txt` for the packages they are not specifically testing against?
* Should the CI/CD pipeline make a new `requirements.txt` file and if it passes it saves it as an artifact or pushes it back to the repo?  Then it becomes a question or which version should be kept, latest?
* If it has to be done manually, who is going to be responsible for compiling a new `requirements.txt`, or when should that be done?
* Should we have something in the README (maybe under the `Test OpenMDAO 2` section?) explaining how to compile a new `requirements.txt`?

swryan:
Due to questions like those raised in @sanbales last post, we are closing this PR for now.

For development, we test multiple configurations and versions and address compatibility issues as they come up.  It seems that a `requirements.txt` would be specific to a particular production installation.

If this is something that the community thinks is needed, then it might be a good issue to discuss further in a POEM:  https://github.com/OpenMDAO/POEMs

PR #1099, opened 2019-10-31
Added Error Bars For KrigingSurrogates when eval_rmse is True

DKilkenny:


JustinSGray:
one potential problem with this PR is that, since Kriging is currently the only surrogate with error bars we haven't really thought through that part of the surrogate API. So that needs to be considered more carefully before we can accept the PR

Kenneth-T-Moore:
So, I recall that in the meeting last week, we discussed this story as if it were a done deal. @JustinSGray

DKilkenny:
No, that was about the dashed lines to show the reach of the scatter distance input.

PR #1101, opened 2019-11-04
units updates for POEM_001

robfalck:
Reference implementation for updates for some astrodynamics units as given by POEM 001


openmdao/utils/unit_library.ini
```diff
@@ -114,7 +114,7 @@ e: 1.60217653e-19*C, elementary charge
 eV: e*V, electron volt
 Da: 1.66053886e-27*kg, dalton
 u: Da, unified atomic mass unit
-ua: 1.49597870691e11*m, astronomical unit
```
swryan: I wonder if we need to make this deprecation

PR #1104, opened 2019-11-04
Newton deprecation warning to print on comm 0

DKilkenny:
https://www.pivotaltracker.com/story/show/169262784

### DIFF
--- a/openmdao/utils/assert_utils.py
+++ b/openmdao/utils/assert_utils.py
@@ -51,7 +51,6 @@ def assert_warning(category, msg):
             raise AssertionError("Did not see expected %s: %s" % (category.__name__, msg))
 
 
-
 def assert_check_partials(data, atol=1e-6, rtol=1e-6):
     """
     Raise assertion if any entry from the return from check_partials is above a tolerance.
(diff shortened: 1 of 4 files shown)


openmdao/utils/assert_utils.py
```diff
@@ -35,16 +36,19 @@ def assert_warning(category, msg):
...
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
             yield
 
-    for warn in w:
-        if (issubclass(warn.category, category) and str(warn.message) == msg):
-            break
-    else:
-        raise AssertionError("Did not see expected %s: %s" % (category.__name__, msg))
+    if rank == 0:
```
naylor-b: I think it would be better to keep assert_warning blissfully unaware of MPI, and handle any rank related behaviors in your unit test.  There could be instances where a warning might be generated on some other rank than rank 0 and a test might want to check that.

openmdao/utils/assert_utils.py
```diff
@@ -47,6 +47,35 @@ def assert_warning(category, msg):
...
+    with reset_warning_registry():
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            yield
+
+    for warn in w:
+        if (issubclass(warn.category, category) and str(warn.message) == msg):
+            raise AssertionError("Found warning: ", msg)
+        else:
+            break
```
naylor-b: By putting a break here, you're only checking the first warning.  The else block should be removed.

PR #1107, opened 2019-11-06
Driver debug print mode and indices

DKilkenny:
https://www.pivotaltracker.com/story/show/169513241

### DIFF
--- a/openmdao/core/total_jac.py
+++ b/openmdao/core/total_jac.py
@@ -1229,6 +1229,10 @@ class _TotalJacInfo(object):
                              sub_do_ln=model._linear_solver._linearize_children())
         model._linear_solver._linearize()
 
+        # forward is columns and reverse is rows
+        # If the mode is fwd or rev
+        # inds is the indices
+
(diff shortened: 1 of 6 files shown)


openmdao/core/total_jac.py
```diff
@@ -1242,7 +1242,8 @@ def compute_totals(self):
                             varlist = '(' + ', '.join([name for name in par_deriv[key]]) + ')'
                             print('Solving color:', key, varlist)
                         else:
-                            print('Solving variable:', key)
+                            print('Solving variable: {0} \nIn mode: {1} \n'
```
Kenneth-T-Moore: Did you run the tests? I suspect this should make tests fail in test_pyoptsparse_driver and test_scipy_optimizer.

If they don't fail, then you should add a test for this.

DKilkenny: Yes, I did run the tests but I'll write one for it

DKilkenny: added

PR #1114, opened 2019-11-11
N2 detecting offscreen connections

tadkollar:
N2 can now tell whether a connection exists offscreen and where it should be drawn. Does not actually draw the connection, yet (separate story). Drawing the arrow will be easy, but labeling, especially when there are a lot of persistent ones, could be a problem.

Originally I intended to precompute the matrix once and find connections from that, but that wasn't useful because collapsing and searching causes too many changes. Offscreen connections are now computed every time the matrix is rebuilt.


tadkollar:
I'm not sure why the appveyor Python 2.7 test is failing, there aren't any Python changes in this PR

swryan:
> I'm not sure why the appveyor Python 2.7 test is failing, there aren't any Python changes in this PR

most likely a change in underlying conda packages.. I'll see what I can figure out

PR #1116, opened 2019-11-13
Additional info for driver debug total options

DKilkenny:


openmdao/core/tests/test_coloring.py
```diff
@@ -908,6 +939,64 @@ def test_multipoint_with_coloring(self):
...
+        model.add_objective('obj.y')
+
+        try:
+            p.setup()
+
+            failed, output = run_driver(p)
+
+            self.assertFalse(failed, "Optimization failed.")
+
+            self.assertTrue('In mode: fwd, Solving variable(s):' in output)
```
naylor-b: should also check for specific variables and indices appearing in the output

openmdao/core/tests/test_coloring.py
```diff
@@ -908,6 +939,64 @@ def test_multipoint_with_coloring(self):
...
+
+        except Exception as err:
```
naylor-b: probably should get rid of this try block since it'll prevent the test from ever failing.

openmdao/core/total_jac.py
```diff
@@ -1242,8 +1242,16 @@ def compute_totals(self):
...
-                                  'Sub Indices: {2}'.format(key, mode, inds))
+                            if key == '@simul_coloring':
+                                local_inds = imeta['coloring']._local_indices(
+                                    inds=inds, mode=self.mode)
+                                print('In mode: {0}, Solving variable(s):'.format(
+                                      mode))
+                                print(*local_inds, sep='\n')
+                                pass
+                            else:
+                                print('In mode: {0}, Solving variable(s): {1}\n'
```
naylor-b: the output in the coloring case is now inconsistent with the output in other cases, i.e. in the non-colored case, it has a separate line for 'Sub Indices' which we got rid of in the colored case.   I would get rid of it here as well and make the non-colored case look like the colored case.

openmdao/core/tests/test_parallel_derivatives.py
```diff
@@ -168,10 +168,11 @@ def test_debug_print_option_totals_color(self):
             sys.stdout = stdout
 
         output = strout.getvalue()
-
+        print(output)
```
naylor-b: leftover debug print statement

openmdao/core/total_jac.py
```diff
@@ -1242,8 +1242,18 @@ def compute_totals(self):
...
+                                local_inds = imeta['coloring']._local_indices(
+                                    inds=inds, mode=self.mode)
+                                print('In mode: {0}, Solving variable(s):'.format(
+                                      mode))
+                                print(*local_inds, sep='\n')
+                            else:
+                                print('In mode: {0}, Solving variable(s):'.format(
+                                      mode))
+                                print("('{0}', [{1}])".format(key, inds))
+                                # print('In mode: {0}, Solving variable(s): {1}\n'
```
naylor-b: leftover commented out code

openmdao/core/total_jac.py
```diff
@@ -1242,8 +1242,18 @@ def compute_totals(self):
                             varlist = '(' + ', '.join([name for name in par_deriv[key]]) + ')'
                             print('Solving color:', key, varlist)
                         else:
-                            print('Solving variable: {0} \nIn mode: {1} \n'
-                                  'Sub Indices: {2}'.format(key, mode, inds))
+                            if key == '@simul_coloring':
+                                local_inds = imeta['coloring']._local_indices(
+                                    inds=inds, mode=self.mode)
+                                print('In mode: {0}, Solving variable(s):'.format(
```
naylor-b: This first print statement is now identical in both branches of the if block, so it could be pulled outside of the if block.

PR #1121, opened 2019-11-15
Added no_browser flag to prevent browser from opening 

DKilkenny:
https://www.pivotaltracker.com/story/show/169784084

### DIFF
--- a/openmdao/docs/features/model_visualization/meta_model_basics.rst
+++ b/openmdao/docs/features/model_visualization/meta_model_basics.rst
@@ -75,5 +75,6 @@ want to visualize if there is more than one, and optionally a port number:
     :cmd: openmdao view_mm -h
 
 .. note::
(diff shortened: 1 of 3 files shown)


openmdao/docs/features/model_visualization/meta_model_basics.rst
```diff
@@ -74,3 +74,6 @@ want to visualize if there is more than one, and optionally a port number:
 .. embed-shell-cmd::
     :cmd: openmdao view_mm -h
 
+.. note::
+    When using Bash on Windows set --no_browser option to start the server, then open an internet
+    browser and copy/paste the path to viewer.
```
swryan: I think this note needs a little more explanation.. I think you are saying that you _have_ to use the `--no_browser` option when running in a WSL shell because the default option does not work (I tried it and it seems to hang up when trying to open the browser).  Is that right?  If so maybe you could fill in some of that detail to make it more clear what you're saying here...

PR #1122, opened 2019-11-15
Arrows to offscreen connections

tadkollar:
Offscreen connection arrows are working with tooltips naming the offscreen nodes, but the tooltips need some work on improving placement. However, they're useful in their current state and I wanted to PR this before I go on travel next week.

Some extra fixes in this PR are better clipping on the matrix and tree groups, which is noticeable during transitions, a few fixes for bugs that Herb discovered, and as a result of that, testing on some callback functions to make sure the correct "this" is being used.


Kenneth-T-Moore:
Sent Tad a small test that shows the problem.

tadkollar:
Should be fixed now. Connection sources/parents were not being tracked by the target nodes.

PR #1123, opened 2019-11-18
fix issue causing test failures on cluster

swryan:
process items within tuples when assigning parameterized test names
remove attempt to pass invalid arg to itrace setup()

### DIFF
--- a/openmdao/core/tests/test_parallel_groups.py
+++ b/openmdao/core/tests/test_parallel_groups.py
@@ -4,6 +4,8 @@ from __future__ import division, print_function
 
 import unittest
 import itertools
+
+# note: this is a Python 3.3 change, clean this up for OpenMDAO 3.x
 try:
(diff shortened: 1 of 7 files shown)


openmdao/core/tests/test_parallel_groups.py
```diff
@@ -4,6 +4,10 @@
 
 import unittest
 import itertools
+try:
+    from collections.abc import Iterable
```
Kenneth-T-Moore: Perhaps a note that this is a Python 3.3 change, so that we clean this up when we do openmdao 3.

swryan: ok.. we have this in more than one place.. so will have to search the code base

Kenneth-T-Moore: Marking this as approved since it affects the cluster tests.

PR #1126, opened 2019-11-25
P169183805 n2 does not handle colon vars properly

hschilling:
N2 didn't handle the situation where component had inputs like f_approx:f, f_approx:r and an output like f_approx. This happens in dymos. 

Vars with double colons caused problems.

There were also some bugs in XDSM that showed up when running testflo locally but aren't showing up on CI for some reason!


Kenneth-T-Moore:
Is the `_addChildren` function only called to add the tree children that represent colon vars? I think it could use a rename to something more descriptive.

PR #1130, opened 2019-12-01
XDSM v2 component type update

onodip:
Added the new component type mapping based on pyXDSM 2.0. The current implementation sets as default the new version. This might break the code for users of 1.0. If someone really wants, they can change the "pyxdsm_version" of the XDSMWriter, and pass it as a custom writer. It can be made friendlier after https://github.com/mdolab/pyXDSM/pull/14.

Fixed a small bug, which caused to place the output blocks incorrectly in some cases, when mixed left & right outputs sides were set.

Currently ImplicitGroups are not considered.

image

### DIFF
--- a/openmdao/visualization/xdsm_viewer/xdsm_writer.py
+++ b/openmdao/visualization/xdsm_viewer/xdsm_writer.py
@@ -18,16 +18,16 @@ from __future__ import print_function
 
 import json
 import os
+from distutils.version import LooseVersion
 
+from numpy.distutils.exec_command import find_executable
 from six import iteritems, string_types
 
 from openmdao.core.problem import Problem
 from openmdao.utils.general_utils import simple_warning
-from openmdao.visualization.n2_viewer.n2_viewer import _get_viewer_data
 from openmdao.utils.webview import webview
+from openmdao.visualization.n2_viewer.n2_viewer import _get_viewer_data
 from openmdao.visualization.xdsm_viewer.html_writer import write_html
-
-from numpy.distutils.exec_command import find_executable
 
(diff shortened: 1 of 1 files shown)


onodip:
Anyone has an idea of the precise definition of an ImplicitGroup? I would say, a group is an implicit group, if it contains feedbacks between its subsystems, which are not included in any solver process of a subsystem of the group. The more intuitive versions would be, any group that "needs a solver".

Kenneth-T-Moore:
I would think that an ImplicitGroup is as you described, but in addition to feedback, you could also have implicit states on the subsystems.  `ImplicitGroup` isn't an official class in openmdao though.

openmdao/visualization/xdsm_viewer/xdsm_writer.py
```diff
@@ -687,13 +705,24 @@ def __init__(self, name='pyxdsm', box_stacking=_DEFAULT_BOX_STACKING,
                 Defaults to False.
             add_component_indices : bool
                 If true, display components with numbers.
-            equations : bool, optional
```
Kenneth-T-Moore: Did the "equations" feature go away, or does the new version of pyXDSM always render indepvarcomps with equations?  This option has further integration in "om.py" as a command-line argument to `openmdao xdsm`, so it might need to be removed there (and anywhere in between).  Alternatively, maybe the option needs to be included for those still running legacy.

onodip: I implemented it outside the class, this was by mistake left in the code and now cleaned up. In the functionality nothing changed.

PR #1132, opened 2019-12-02
Use running python executable

toddrme2178:
Currently it uses the system-default version of Python, which may not be the one running the code and may not have openMDAO installed.


toddrme2178:
Fixed the issue in more places

PR #1135, opened 2019-12-04
P169069250 solver recording options

hschilling:
User can specify includes/excludes for solver recording options using relative promoted names for outputs and relative absolute names for inputs.

Also fixed an issue with pydocstyle where it now tests for Google style which we don't conform to. 

Also added some docs to explain what kinds of wildcards we can use in the includes and excludes.


openmdao/docs/features/recording/saving_data.rst
```diff
@@ -114,11 +122,17 @@ differentiate the cases.  This prefix can be specified when calling :code:`run_m
...
+
+The :code:`includes` and :code:`excludes` recording options provide support for Unix shell-style wildcards,
+which are not the same as regular expressions. The documentation for the :code:`fnmatchcase` function from the Python
+standard library documents the wildcards: https://docs.python.org/2/library/fnmatch.html#fnmatch.fnmatchcase.
 
 Recording Options Precedence
 ----------------------------
 
-The recording options that determine what gets recorded can sometime be a little confusing. Here is
+The recording options precedence that determine what gets recorded can sometime be a little confusing. Here is
```
Kenneth-T-Moore: determine -> determines

PR #1142, opened 2019-12-30
Work for POEM_007 adding command string compatibility

DKilkenny:
https://www.pivotaltracker.com/story/show/170129605


DKilkenny:
This fails on one of Ken's new tests from POEM_002. Will talk to him about this on Thursday

openmdao/components/tests/test_external_code_comp.py
```diff
@@ -273,6 +273,9 @@ def setup(self):
         self.options['external_input_files'] = [self.input_file]
         self.options['external_output_files'] = [self.output_file]
 
+        # If you want to write your own string command, the code below will also work.
```
Kenneth-T-Moore: Can remove the commented-out lines.

openmdao/components/tests/test_external_code_comp.py
```diff
@@ -610,13 +613,96 @@ def setup(self):
...
+
+        area_ratio = 1.3
+        super_sonic = True
+        prob['area_ratio'] = area_ratio
+        mach_comp.options['super_sonic'] = super_sonic
+        prob.run_model()
+        assert_rel_error(self, prob['mach'], mach_solve(area_ratio, super_sonic=super_sonic), 1e-8)
+
+    def test_string_command(self):
+        import sys
```
Kenneth-T-Moore: This new test isn't a feature test, so it doesn't need these imports. However, maybe it should be in the docs. Maybe it makes sense to use the string version as the preferred way of demonstrating how to set the commands.

openmdao/components/tests/test_external_code_comp.py
DKilkenny: Leaving in lines for OpenMDAO docs

openmdao/components/tests/test_external_code_comp.py
naylor-b: I agree that passing a string seems like a simpler option for users and should probably be shown in the docs.

openmdao/components/tests/test_external_code_comp.py
```diff
@@ -273,9 +273,12 @@ def setup(self):
...
 
-        self.options['command'] = [
-            sys.executable, 'extcode_paraboloid.py', self.input_file, self.output_file
-        ]
+        # If you want to write your command as a list, the code below will also work.
+        # self.options['command'] = [
+        #     sys.executable, 'extcode_paraboloid.py', self.input_file, self.output_file
+        # ]
+
+        self.options['command'] = 'python extcode_paraboloid.py paraboloid_input.dat paraboloid_output.dat'
```
swryan: I don't think you should hard-code the file names here, since they are specified via attributes.
Should probably do something like:
_cmd_ = `python extcode_paraboloid.py {} {}'.format(self.input_file, self.output_file)`

openmdao/components/tests/test_external_code_comp.py
```diff
@@ -610,13 +613,17 @@ def setup(self):
...
                 self.options['command_apply'] = [
                     sys.executable, 'extcode_mach.py', self.input_file, self.output_file,
                 ]
                 self.options['command_solve'] = [
                     sys.executable, 'extcode_mach.py', self.input_file, self.output_file,
                 ]
 
+                # If you want to write your own string command, the code below will also work.
+                # self.options['command_apply'] = 'python extcode_mach.py mach_input.dat mach_output.dat'
+
```
swryan: I don't think you should hard-code the file names here, since they are specified via attributes.
Should probably do something like:
_cmd_ = `python extcode_mach.py {} {}'.format(self.input_file, self.output_file)`

openmdao/docs/features/building_blocks/components/external_code_comp.rst
```diff
@@ -61,6 +61,10 @@ does the same computation as the :ref:`Paraboloid Tutorial <tutorial_paraboloid_
 
 The following example demonstrates how to build an OpenMDAO component that makes use of this external code.
 
+.. note::
+    If you pass a string as a command, OpenMDAO sets `shell=True` which can add overhead leading to
+    a decrease in performance and a security loophole. Use list when possible.
```
swryan: I would suggest linking to https://docs.python.org/3/library/subprocess.html#frequently-used-arguments, which contains this description:

```If shell is True, the specified command will be executed through the shell. This can be useful if you are using Python primarily for the enhanced control flow it offers over most system shells and still want convenient access to other shell features such as shell pipes, filename wildcards, environment variable expansion, and expansion of ~ to a user’s home directory. However, note that Python itself offers implementations of many shell-like features (in particular, glob, fnmatch, os.walk(), os.path.expandvars(), os.path.expanduser(), and shutil).```

as well as to https://docs.python.org/3/library/subprocess.html#security-considerations

(`shell=True` is not necessarily meaningful to the user without context)

PR #1144, opened 2020-01-04
POEM005-Plugins

naylor-b:
Implementation PR for POEM005 - Proposed OpenMDAO Plugin System

The changes in the PR are primarily to allow for local (installed) discovery of plugins using entry points and to make it easier for outside users to create plugins.  There is nothing here to facilitate global discovery by accessing the PyPI web API.  That will come later.

- added entry point groups and entry points for various OpenMDAO types
- added 'openmdao list_installed' command line tool to list installed classes/sub-commands based on entry points
- updated 'openmdao scaffold' command line tool to generate the directory structure for a python package and generate proper entry points.
- added 'compute_entry_points' function to generate list of entry point strings based on classes found in the source code of a given python package. We'll use this for testing to make sure classes don't get added without registering the corresponding entry point. 
- added support for loading plugin command line tools.  This will allow, for example, the XDSM stuff to be moved into a plugin while still allowing there to be an 'openmdao xdsm' command line tool for users.
- simplified the API for adding command line tools.
- updated all internal openmdao command line tools to properly handle user args (args that follow '--' on the command line).
- fixed a bunch of unittest.skipIf and unittest.skipUnless decorators to properly skip tests if either MPI or petsc are not present.
- added code to clean up the signal handler cache in pyoptsparse_driver in order to prevent a memory leak test failure.
- added `if __name__ == '__main__'` block to a few modules to prevent having those modules run a bunch of code when imported while running compute_entry_points.
- moved `openmdao dump_dist_idxs` out into a plugin package (om_devtools).


openmdao/utils/scaffolding_templates/LinearSolver_template
```diff
@@ -0,0 +1,45 @@
...
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Options dictionary.
+        """
+        super({class_name}, self).__init__(**kwargs)
+
+        # initialize needed atributes here...
+
```
Kenneth-T-Moore: I would add `_declare_options` and `_setup_solvers` here just to be consistent with the nl solver template.

PR #1147, opened 2020-01-08
Added test to verify that design vars, constraints, and objectives

DKilkenny:
https://www.pivotaltracker.com/story/show/170517752

### DIFF
--- a/openmdao/drivers/tests/test_pyoptsparse_driver.py
+++ b/openmdao/drivers/tests/test_pyoptsparse_driver.py
@@ -106,7 +106,6 @@ class DataSave(om.ExplicitComponent):
 
         partials['y', 'x'] = 2.0*x - 6.0
 
-@contextmanager
 @unittest.skipIf(OPT is None or OPTIMIZER is None, "only run if pyoptsparse is installed.")
 @unittest.skipUnless(MPI, "MPI is required.")
 class TestMPIScatter(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -104,6 +106,38 @@ def compute_partials(self, inputs, partials):
...
+
+        model.add_design_var('x', lower=-50.0, upper=50.0)
+        model.add_design_var('y', lower=-50.0, upper=50.0)
+        model.add_objective('f_xy')
+        model.add_constraint('c', upper=-15.0)
+
+        prob.setup()
+        prob.run_driver()
+
+        design_vars = MPI.COMM_WORLD.allgather([prob['x'], prob['y'], prob['c'], prob['f_xy']], root=0)
```
Kenneth-T-Moore: Don't need to do an allgather in the test. Just assert that each variable(e.g., prob['x]) equals the static value (e.g., 7.1666667). The test runs on both procs, so such a test would verify that the correct value of the desvars, objs, and cons have been propagated to each proc.

DKilkenny: Updated

openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -3,6 +3,7 @@
 import copy
 import sys
 import unittest
+from contextlib import contextmanager
```
naylor-b: Don't need this any more.

DKilkenny: removed

PR #809, opened 2018-11-10
Reconfigurable Model Execution - connection tests

onodip:
Added tests with "expected failure" for connections of a reconfigurable problem.

Based on the following Stackoverflow question:
https://stackoverflow.com/questions/53231118/how-to-update-connection-sizes-in-a-reconfigurable-model-in-openmdao-2-5-0

PR #1150, opened 2020-01-12
Reconfigurable model execution bugfix

onodip:
This is a follow up to #809, and partially fixes the issues mentioned there (for now only with the default "RunOnce" solver, with Newton or NLBGS it still fails).

My proposal would be also to add a "reconfigurable" flag to systems that gets inherited from the parents. By default it should be False, and all reconfiguration checks performed in every iteration could be skipped, since in 99% of the problems this functionality is not used. I will make a POEM on this, if it makes sense.


onodip:
> The OpenMDAO team is planning to completely overhaul how reconfigurability works at some point in the future, but so far it hasn't reached the top of our priority list. Is reconfigurability something that you need in the near future? I have some ideas about reconfigurability-related API changes that I plan to lay out in a POEM at some point.

I don't need it urgently, but would be nice if this functionality would work again in the long term. I also don't mind if this PR does not get merged, if the reconfigurability will be reworked at some point anyways. I will run the tests again, but it seems, that the fix is bad for the performance.

swryan:
Closing this PR for now, but we have saved a reference to it in a reconfigurability epic in our backlog.

A POEM will be written for community feedback as part of the process of the overhaul.

Thanks!

PR #1151, opened 2020-01-13
fixes for  list_inputs/list_outputs

swryan:
- make `desc` arg work for  list_inputs/list_outputs on Case, add tests
- fix bug in list_inputs/list_outputs under MPI, add tests


onodip:
Thanks for the fix!

PR #1152, opened 2020-01-15
New tests for N2 GUI

tadkollar:
This update performs headless browser testing on N2 HTML files generated from the models stored in n2_viewer/tests/gui_test_models. It consists of a Node script, test_gui.js, a set of model-specific tests in gui_scripts.json, and several test models. There's a set of toolbar tests that can be performed on any model. Changes were made to .travis.yml to install Node and add another test matrix row to test the GUI. Some changes were also made to the N2 source code to make sure each element had an id attrib.

Here's what the output on Travis looks like:
[travis_output.txt](https://github.com/OpenMDAO/OpenMDAO/files/4065990/travis_output.txt)


.travis.yml
```diff
@@ -150,10 +166,16 @@ script:
 
 # run the tests from down here to see if it can work without being at top level
 # only do coverage on the upload machine.
-- if [ "$UPLOAD_DOCS" ]; then
-    testflo -n 1 openmdao --coverage  --coverpkg openmdao --cover-omit \*tests/\*  --cover-omit \*devtools/\* --cover-omit \*test_suite/\* --cover-omit \*docs/\* --cover-omit \*code_review/\*;
-  else
-    testflo -n 1 --timeout=120 openmdao;
+- if [ -z "$SKIP_TESTFLO" ]; then
```
swryan: could just use `-z "$GUI"` here and no need for SKIP_TESTFLO, right?

.travis.yml
```diff
@@ -11,16 +11,21 @@ filter_secrets: false
 os:
 - linux
 
+# language: node_js
+
+# node_js: node
```
swryan: these commented lines can be deleted?

.travis.yml
tadkollar: Having both allows for it to be tested in any other combo in the matrix by putting GUI=1 in that row - for instance in addition to/instead of testing standalone, we might want to test it with the Python 3.6 row because it usually completes faster.

.travis.yml
tadkollar: Yes, didn't need that anymore after I changed it so it only installs Node.js when GUI=1

tadkollar:
Since it's not a Python package, I'm having a lot of trouble getting Node.js into setup.py so it would install with [test] or [all]. I'm thinking that because the GUI tests run outside of testflo, they should have their own script to kick them off, and that could check for the presence of node and automatically install if necessary.

Kenneth-T-Moore:
@tadkollar   :  I wonder if there is something @naylor-b can do to testflo to give it the capability to run the GUI tests.

tadkollar:
AppVeyor keeps failing on the Python 2.7 build with:

> ImportError: 
> Importing the multiarray numpy extension module failed.  Most
> likely you are trying to import a failed build of numpy.
> If you're working with a numpy git repo, try `git clean -xdf` (removes all
> files not under version control).  Otherwise reinstall numpy.
> Original error was: DLL load failed: The specified module could not be found.

Not sure why that's happening, doesn't seem to be anything to do with my changes...

PR #1157, opened 2020-01-21
Code for POEM_008: Nonlinear Refactor

DKilkenny:
https://www.pivotaltracker.com/story/show/169906417


openmdao/solvers/nonlinear/tests/test_nonlinear_block_jac.py
```diff
@@ -158,7 +158,7 @@ class TestNonlinearBlockJacobiMPI(unittest.TestCase):
 
     N_PROCS = 2
 
-    @unittest.skipUnless(MPI, "MPI is not active.")
+    # @unittest.skipUnless(MPI, "MPI is not active.")
```
Kenneth-T-Moore: A  couple of things commented out on this test that should be restored.

openmdao/solvers/nonlinear/newton.py
```diff
@@ -68,13 +68,17 @@ def _declare_options(self):
         """
         super(NewtonSolver, self)._declare_options()
 
-        self.options.declare('solve_subsystems', types=bool, default=False,
```
Kenneth-T-Moore: I think there is behavior missing here. If the user does pass a value for "solve_subsystems" as an attribute into the constructor, then there should be a DeprecationWarning raised that tells the user that in openmdao 3.x and beyond, the "solve_subsystems" argument is going to be required.  Then, to conserve backwards compatibility, we should set it to False.

Also, we need a deprecation test that confirms this.

openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -2549,7 +2549,7 @@ def test_linesearch(self):
 
         model.nonlinear_solver = om.NewtonSolver()
         model.nonlinear_solver.options['maxiter'] = 3
-        # model.nonlinear_solver.options['solve_subsystems'] = True
```
Kenneth-T-Moore: Why are we changing solve_subsystems here from true to false?

openmdao/solvers/nonlinear/tests/test_nonlinear_block_jac.py
```diff
@@ -170,6 +170,7 @@ def test_reraise_analylsis_error(self):
         sub.add_subsystem('c1', AEComp())
         sub.add_subsystem('c2', AEComp())
         sub.nonlinear_solver = om.NonlinearBlockJac()
+        # sub.nonlinear_solver.options['reraise_child_analysiserror'] = True
```
Kenneth-T-Moore: can remove this line

openmdao/recorders/tests/test_sqlite_reader.py
DKilkenny: Found that it was commented out originally. Uncommented and changed

PR #1159, opened 2020-01-22
Fixed bug to prevent design variables being different on different MPI procs

DKilkenny:
Follow up to https://www.pivotaltracker.com/story/show/170517752


openmdao/drivers/scipy_optimizer.py
```diff
@@ -569,6 +570,8 @@ def _objfunc(self, x_new):
 
             # Pass in new parameters
             i = 0
+            if MPI:
+                x_new = model.comm.bcast(x_new, root=0)
```
naylor-b: Should probably replace bcast with Bcast.  bcast pickles the array before sending it, so it's slower.

DKilkenny: Changed

PR #1161, opened 2020-01-24
Error raise to prevent add_subsystem being called in the config method

DKilkenny:
https://www.pivotaltracker.com/story/show/170734950

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -323,6 +323,8 @@ class Group(System):
         self.options._parent_name = self.msginfo
         self.recording_options._parent_name = self.msginfo
 
+        self._setup_procs_finished = False
+
         if self._num_par_fd > 1:
             info = self._coloring_info
             if comm.size > 1:
@@ -726,7 +728,6 @@ class Group(System):
         super(Group, self)._setup_var_sizes()
 
         self._var_offsets = None
-        self._setup_procs_finished = False
 
         iproc = self.comm.rank
         nproc = self.comm.size
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -81,6 +81,8 @@ class Group(System):
         Mapping of local subsystem names to their corresponding System.
     _approx_subjac_keys : list
         List of subjacobian keys used for approximated derivatives.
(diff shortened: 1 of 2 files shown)


openmdao/core/group.py
```diff
@@ -1435,6 +1442,9 @@ def add_subsystem(self, name, subsys, promotes=None,
             enable users to instantiate and add a subsystem at the
             same time, and get the reference back.
         """
+        if self._setup_procs_finished:
+            raise RuntimeError('Cannot call add_subsystem in the configure method')
```
Kenneth-T-Moore: You should prepend the self.msginfo info so that the user knows which component raised the error. See the next error message below this one to see how it is done.

openmdao/core/tests/test_group.py
```diff
@@ -71,6 +71,28 @@ def setup(self):
     def compute(self, inputs, outputs):
         self._order_list.append(self.pathname)
 
+class TestSubsystemConfigError(unittest.TestCase):
+
+    def test_add_subsystem_error_on_config(self):
+        class SimpleGroup(om.Group):
+
+            def __init__(self):
+                super(SimpleGroup, self).__init__()
```
swryan: this should prob be `setup` rather than `__init__`

openmdao/core/group.py
DKilkenny: Changed

openmdao/core/tests/test_group.py
DKilkenny: Fixed

openmdao/core/tests/test_group.py
```diff
@@ -71,6 +71,28 @@ def setup(self):
     def compute(self, inputs, outputs):
         self._order_list.append(self.pathname)
 
+class TestSubsystemConfigError(unittest.TestCase):
+
+    def test_add_subsystem_error_on_config(self):
+        class SimpleGroup(om.Group):
+
+            def setup(self):
+                super(SimpleGroup, self).setup()
```
swryan: this super call should not be necessary..

PR #1162, opened 2020-01-26
Bugfix for DOEDriver full factorial array inputs

onodip:
This PR fixes a bug of the `FullFactorialGenerator` in the `DOEDriver`

**Bug:**
Wrong factoring, if there exists at least one array design variable with array size larger than one, and it is not the last one. The array test of this class was testing with only one design variable, therefore it was not caught before.

**Fix:**
Corrected a wrong indexing in `doe_generators.py`. Location in the comments
Added a new test.

I also made some other small changes for performance and readability

The only test so far, where it was used with multiple array inputs, was in `test_csv_array()`. Here the expected result was not really a full factorial, since it included non-unique cases (each repeated twice), as the figure below shows. 

image

### DIFF
--- a/openmdao/drivers/doe_generators.py
+++ b/openmdao/drivers/doe_generators.py
@@ -320,12 +320,6 @@ class _pyDOE_Generator(DOEGenerator):
         """
         sizes = [meta['size'] for name, meta in iteritems(design_vars)]
 
-        if isinstance(self._levels, int):
-            self._level_lst = [self._levels] * len(sizes)
-        else:
-            default = self._levels.get('default', _LEVELS)
-            self._level_lst = [self._levels.get(name, default) for name in design_vars.keys()]
-
         size = sum(sizes)
 
         doe = self._generate_design(size)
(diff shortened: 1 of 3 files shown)


openmdao/drivers/tests/test_doe_driver.py
```diff
@@ -337,36 +318,36 @@ def test_csv_array(self):
         prob.run_driver()
         prob.cleanup()
 
-        expected = {
```
onodip: This is the expected result shown in the figure. 

Which is equivalent to:
`pyDOE2.fullfact(levels=[2, 2, 2])
`

Which correctly results in:
```
array([[0., 0., 0., 0.],
       [1., 0., 0., 0.],
       [0., 1., 0., 0.],
       [1., 1., 0., 0.],
       [0., 0., 1., 0.],
       [1., 0., 1., 0.],
       [0., 1., 1., 0.],
       [1., 1., 1., 0.],
       [0., 0., 0., 1.],
       [1., 0., 0., 1.],
       [0., 1., 0., 1.],
       [1., 1., 0., 1.],
       [0., 0., 1., 1.],
       [1., 0., 1., 1.],
       [0., 1., 1., 1.],
       [1., 1., 1., 1.]])
```

openmdao/drivers/doe_generators.py
```diff
@@ -347,26 +349,24 @@ def __call__(self, design_vars, model=None):
...
         for idxs in doe.astype('int'):
             retval = []
-            var = row = 0
-            for name, meta in iteritems(design_vars):
+            row = 0
+            for var, (name, meta) in enumerate(iteritems(design_vars)):
                 size = meta['size']
                 val = np.empty(size)
                 for k in range(size):
-                    idx = idxs[var + k]
```
onodip: This is the actual bug. `var` was the index of the design variable, but array DVs take multiple rows in `idxs`

onodip:
No idea, how the error in the tests is related to the changes.

swryan:
> No idea, how the error in the tests is related to the changes.

see: https://github.com/onodip/OpenMDAO/pull/3

onodip:
Thanks for the help @swryan

PR #1165, opened 2020-01-29
list_inputs and list_outputs have the option to display global size

swryan:
- list_inputs and list_outputs should have the option to display local size as well as global size for distributed components


swryan:
closing this PR until I add docs

PR #1166, opened 2020-01-30
Fixed some UX bugs in the N2 library

Subraiz:
The first bug was not being able to uncollapse a collapsed view that was zoomed in on.

The second bug was not being able to press the back button to undo collapsing a view.

Tests have been added for both of these bugs as well and are tested using the Turbojet.py model 

Both of these bugs are being tracked on the Pivotal tracker.


openmdao/visualization/n2_viewer/tests/gui_test_models/turbojet.py
```diff
@@ -0,0 +1,30 @@
+import openmdao.api as om
```
naylor-b: This is a simple paraboloid model, not a turbojet model, so the filename should be changed.

openmdao/visualization/n2_viewer/turbojet.py
```diff
@@ -0,0 +1,30 @@
+import openmdao.api as om
```
naylor-b: Filename is incorrect as described before.  And why are we duplicating this file in 2 places?

tadkollar:
A few things -

- There are some huge diffs on multiple files only due to whitespace, so that makes it hard to tell exactly where the important updates were made
- The n2.html file shouldn't be included
- The CI tests failed because of Python formatting issues

Might be easiest to reapply only the fixes/tests to a fresh clone of master and push again? Make sure you have a local backup so you don't lose anything

swryan:
closing this PR, as I believe it is superseded by #1169

PR #1170, opened 2020-01-30
Promote variables in configure (POEM_003)

DKilkenny:
https://www.pivotaltracker.com/story/show/170749130

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1374,6 +1374,34 @@ class Group(System):
         if self._conn_discrete_in2out:
             self._vector_class.TRANSFER._setup_discrete_transfers(self, recurse=recurse)
 
+    def promotes(self, subsys_name, any=None, inputs=None, outputs=None):
+        """
+        Promote a variable in the model tree.
+
+        Parameters
+        ----------
+        subsys_name : str
+            The name of the child subsystem whose inputs/outputs are being promoted.
+        any : Sequence of str or tuple
+            A Sequence of variable names (or tuples) to be promoted, regardless
+            of if they are inputs or outputs. This is equivalent to the items
+            passed via the `promotes=` argument to add_subsystem.  If given as a
+            tuple, we use the "promote as" standard of ('real name', 'promoted name')*[]:
+        inputs : Sequence of str or tuple
+            A Sequence of input names (or tuples) to be promoted. Tuples are
+            used for the "promote as" capability.
+        outputs : Sequence of str or tuple
+            A Sequence of output names (or tuples) to be promoted. Tuples are
+            used for the "promote as" capability.
+        """
+        subsys = getattr(self, subsys_name)
+        if any:
+            subsys._var_promotes['any'] = any
+        if inputs:
+            subsys._var_promotes['input'] = inputs
+        if outputs:
+            subsys._var_promotes['output'] = outputs
+
     def add(self, name, subsys, promotes=None):
         """
         Add a subsystem (deprecated version of <Group.add_subsystem>).
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1396,11 +1396,11 @@ class Group(System):
         """
         subsys = getattr(self, subsys_name)
         if any:
-            subsys._var_promotes['any'] = any
+            subsys._var_promotes['any'].extend(any)
         if inputs:
-            subsys._var_promotes['input'] = inputs
+            subsys._var_promotes['input'].extend(inputs)
         if outputs:
-            subsys._var_promotes['output'] = outputs
+            subsys._var_promotes['output'].extend(outputs)
 
     def add(self, name, subsys, promotes=None):
         """
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1404,12 +1404,12 @@ class Group(System):
 
         list_comp = [i if isinstance(i, tuple) else (i, i) for i in subsys._var_promotes['input']]
 
-        for i in list_comp:
-            original, new = i
-            for j in list_comp:
-                original_inside, new_inside = j
+        # Fix this
+        for original, new in list_comp:
+            for original_inside, new_inside in list_comp:
                 if original == original_inside and new != new_inside:
-                    raise RuntimeError("%s: Trying to promote '%s' when it has been aliased to '%s'." % (self.msginfo, original_inside, new))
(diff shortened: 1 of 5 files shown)


openmdao/core/tests/test_group.py
```diff
@@ -226,6 +318,42 @@ def test_group_rename_connect(self):
...
+
+            def configure(self):
+                self.promotes('comp1', any=['*'])
+
+        top = om.Problem(model=SimpleGroup())
+        top.setup()
+
+        self.assertEqual(top['x'], 5)
+
+    def test_promotes_inputs_and_outputs(self):
```
Kenneth-T-Moore: Tests that are used in the feature docs need to include all imports.  If you forget to do this, it should fail when the docs build (tests are run independently in that process), so if you built these docs, something isn't working right there.

openmdao/core/group.py
```diff
@@ -1374,6 +1374,34 @@ def _setup_transfers(self, recurse=True):
...
+        inputs : Sequence of str or tuple
+            A Sequence of input names (or tuples) to be promoted. Tuples are
+            used for the "promote as" capability.
+        outputs : Sequence of str or tuple
+            A Sequence of output names (or tuples) to be promoted. Tuples are
+            used for the "promote as" capability.
+        """
+        subsys = getattr(self, subsys_name)
+        if any:
+            subsys._var_promotes['any'] = any
```
Kenneth-T-Moore: This looks like it would replace any existing values in the _var_promotes dictionary.  I think it is more correct to append them to the existing dictionary instead, in case the user want to call `promotes` twice on the same subsystem.

openmdao/core/tests/test_group.py
DKilkenny: Yeah I forgot. This was added in the latest commit

openmdao/core/group.py
DKilkenny: Fixed

openmdao/core/tests/test_group.py
```diff
@@ -96,6 +96,123 @@ def configure(self):
...
+        top.setup()
+
+        self.assertEqual(top['x'], 5)
+        with self.assertRaises(UnboundLocalError) as cm:
+            top['a']
+
+        self.assertEqual(str(cm.exception),
+                         "local variable 'val' referenced before assignment")
+
+    def test_multiple_promotes(self):
```
naylor-b: I don't see any tests in here that do promotes with aliasing.  We should probably have a couple of those, one with multiple nested groups that does promotion with aliasing successfully and another that maybe tries and fails to promote an aliased child name using the original name.

DKilkenny: Added

openmdao/core/tests/test_group.py
```diff
@@ -96,6 +96,146 @@ def configure(self):
...
+                self.promotes('comp2', outputs=['b'])
+
+        top = om.Problem(model=SimpleGroup())
+        top.setup()
+
+        self.assertEqual(top['b'], 1)
+        with self.assertRaises(UnboundLocalError) as cm:
+            top['a']
+
+        self.assertEqual(str(cm.exception),
```
naylor-b: This error message seems unrelated to the action being checked.  `top['a']` results in `local variable "val" referenced before assignment'?  Something doesn't seem right here.  This happens in several places.

openmdao/core/tests/test_group.py
```diff
@@ -96,6 +96,146 @@ def configure(self):
...
+                         "local variable 'val' referenced before assignment")
+
+    def test_promotes_inputs_in_config(self):
```
naylor-b: I think we still need a test where a subsystem promotes a variable to one alias, and then a parent tries to promote it to a different alias.  That should raise an exception.  Promoting it to the same alias would be ok.

openmdao/core/system.py
```diff
@@ -2038,9 +2038,27 @@ def resolve(to_match, io_types, matches, proms):
...
                 else:
                     call = 'promotes_%ss' % io_types[0]
-                raise RuntimeError("%s: '%s' failed to find any matches for the following "
-                                   "names or patterns: %s.%s" %
-                                   (self.msginfo, call, sorted(not_found), empty_group_msg))
+
+                for p in patterns:
+                    for name, alias in renames.items():
+                        if fnmatchcase(name, p):
+                            raise RuntimeError("%s: %s '%s' matched '%s'. You are attempting to "
```
naylor-b: Since we have the information available here, we should maybe say " ... matched 'bb' but 'bb' was aliased to 'xx'."

openmdao/core/tests/test_group.py
```diff
@@ -96,6 +96,223 @@ def configure(self):
...
+
+            def configure(self):
+                self.sub.promotes('comp1', inputs=['bb'])
+
+        top = om.Problem(model=TopGroup())
+
+        # with self.assertRaises(RuntimeError) as context:
+        top.setup()
+
+        # self.assertEqual(str(context.exception),
```
naylor-b: Get rid of commented out code.

openmdao/core/system.py
DKilkenny: fixed

openmdao/core/group.py
```diff
@@ -1374,6 +1374,43 @@ def _setup_transfers(self, recurse=True):
...
+        if any:
+            subsys._var_promotes['any'].extend(any)
+        if inputs:
+            subsys._var_promotes['input'].extend(inputs)
+        if outputs:
+            subsys._var_promotes['output'].extend(outputs)
+
+        list_comp = [i if isinstance(i, tuple) else (i, i) for i in subsys._var_promotes['input']]
+
+        # Fix this
```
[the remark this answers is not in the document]
DKilkenny: fixed

openmdao/core/tests/test_group.py
```diff
@@ -985,8 +980,7 @@ def test_promote_not_found1(self):
         with self.assertRaises(Exception) as context:
             p.setup()
         self.assertEqual(str(context.exception),
-                         "ExecComp (C2): 'promotes_outputs' failed to find any matches for "
-                         "the following name: ['x*'].")
+                         "ExecComp (C2): 'promotes_outputs' failed to find any matches for the following pattern: x*.")
```
naylor-b: should put pattern in quotes

DKilkenny: fixed

PR #1173, opened 2020-02-03
Fixed setup_descrete_transfers when using MPI and not using PETSc

DKilkenny:
https://www.pivotaltracker.com/story/show/171027633


openmdao/vectors/default_transfer.py
```diff
@@ -180,14 +181,73 @@ def _setup_discrete_transfers(group, recurse=True):
...
+        owns = group._owning_rank
 
         for tgt, src in iteritems(group._conn_discrete_in2out):
-            src_sys, src_var = src[offset:].split('.', 1)
-            tgt_sys, tgt_var = tgt[offset:].split('.', 1)
+            src_sys, src_var = src[name_offset:].split('.', 1)
+            tgt_sys, tgt_var = tgt[name_offset:].split('.', 1)
             xfer = (src_sys, src_var, tgt_sys, tgt_var)
             transfers[tgt_sys].append(xfer)
-            transfers[None].append(xfer)
```
naylor-b: If group.comm.size == 1 then it looks like transfers[None] doesn't get defined.  I think we might need a new test as well since our current tests don't catch this.

DKilkenny: Put in story to capture this issue in the future

PR #1176, opened 2020-02-05
a fix to some regex to get rid of warnings from new pycodestyle

naylor-b:
- also pinned to a newer version,  2.4.0, of pycodestyle
- added some docs describing how to fix a couple of MPI errors when using newer versions of OpenMPI (3+).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ optional_dependencies = {
     'test': [
         'parameterized',
         'numpydoc>=0.9.1',
-        'pycodestyle==2.3.1',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@
     'test': [
         'parameterized',
         'numpydoc>=0.9.1',
-        'pycodestyle==2.3.1',
+        'pycodestyle==2.4.0',
```
Kenneth-T-Moore: As we speak, 2.5 just came out, so Bret is testing against that.

PR #1178, opened 2020-02-07
list_inputs/list_outputs have the option to display global size

swryan:


openmdao/core/system.py
```diff
@@ -3491,22 +3526,26 @@ def _write_table(self, var_type, var_data, hierarchical, print_arrays, out_strea
         # If parallel, gather up the vars.
         if MPI and self.comm:
             # All procs must call this. Returns a list, one per proc.
-            all_var_dicts = self.comm.gather(var_dict, root=0)
+            all_var_dicts = self.comm.allgather(var_dict)
```
naylor-b: if all_procs is False you could just do a gather here since in that case you only use all_var_dicts on proc 0.

PR #1180, opened 2020-02-10
Code for POEM_010: Added recordable arg

DKilkenny:
https://www.pivotaltracker.com/story/show/171079395

### DIFF
--- a/openmdao/components/tests/test_balance_comp.py
+++ b/openmdao/components/tests/test_balance_comp.py
@@ -97,6 +97,30 @@ class TestBalanceComp(unittest.TestCase):
         for (of, wrt) in cpd['balance']:
             assert_almost_equal(cpd['balance'][of, wrt]['abs error'], 0.0, decimal=5)
 
+    def test_balance_comp_options_exclude(self):
+
+        prob = om.Problem()
+
+        bal = om.BalanceComp()
+        bal.add_balance('x', val=1.0)
+
+        prob.model.add_subsystem(name='balance', subsys=bal)
+
+        recorder = om.SqliteRecorder('cases.sql')
+
+        prob.model.add_recorder(recorder)
+
+        prob.model.recording_options['record_inputs'] = True
+        prob.model.recording_options['record_outputs'] = True
+        prob.model.recording_options['record_residuals'] = True
+        prob.model.recording_options['record_metadata'] = False
+        prob.model.recording_options['options_excludes'] = ['*']
+
+        prob.setup()
+
+        prob.run_model()
+
+
     def test_create_on_init_no_normalization(self):
 
         prob = om.Problem()
--- a/openmdao/components/tests/test_balance_comp.py
+++ b/openmdao/components/tests/test_balance_comp.py
@@ -11,6 +11,7 @@ import numpy as np
 from numpy.testing import assert_almost_equal
 
 import openmdao.api as om
+from openmdao.utils.assert_utils import assert_warning, assert_no_warning
 
 
 class TestBalanceComp(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


openmdao/utils/options_dictionary.py
```diff
@@ -44,6 +44,23 @@ def __init__(self, parent_name=None, read_only=False):
...
+        Return state as a dict.
+
+        Returns
+        -------
+        dict
+            State to get.
+        """
+        state = self.__dict__.copy()
+        user_options = state['_dict'].copy()
+        for i in user_options.copy():
```
naylor-b: `i` by convention is used for integer looping variables.  I would use `name` or `key` here.

openmdao/utils/options_dictionary.py
```diff
@@ -44,6 +44,23 @@ def __init__(self, parent_name=None, read_only=False):
...
+            if not user_options[i]['recordable']:
+                del user_options[i]
+
+        return user_options
```
naylor-b: I would add an attribute to `OptionsDictionary` called `_all_recordable` or something like that.  Default it to True and set it to False if `recordable` ever equals False in `declare`.  Then in your `__getstate__`, if `_all_recordable` is True you can just skip all of the looping and just return `self.__dict__`.

openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -1941,12 +1941,11 @@ def test_system_options_pickle_fail(self):
...
-        msg = "Trying to record options which cannot be pickled on system with name: subs. " \
-              "Use the 'options_excludes' recording option on system objects to avoid " \
-              "attempting to record options which cannot be pickled. Skipping recording " \
-              "options for this system."
-        with assert_warning(RuntimeWarning, msg):
-            prob.run_model()
+        msg = ("Trying to record option 'options value to fail' which cannot be pickled on system "
+               "IndepVarComp (subs). Set 'recordable' to False. Skipping recording options for "
+               "this system.")
+        # with assert_warning(UserWarning, msg):
```
naylor-b: any reason why this assert_warning is commented out?

openmdao/utils/options_dictionary.py
```diff
@@ -44,6 +46,25 @@ def __init__(self, parent_name=None, read_only=False):
...
+        Return state as a dict.
+
+        Returns
+        -------
+        dict
+            State to get.
+        """
+        if self._all_recordable:
+            state = self.__dict__
+            return state
```
naylor-b: No need for a temporary variable here, just do `return self.__dict__`

openmdao/recorders/tests/test_sqlite_reader.py
DKilkenny: I was debugging and forgot to change it. Fixed

openmdao/utils/options_dictionary.py
DKilkenny: Fixed

openmdao/recorders/sqlite_recorder.py
```diff
@@ -594,13 +594,14 @@ def record_metadata_system(self, recording_requester):
...
             except Exception:
-                pickled_metadata = pickle.dumps(OptionsDictionary(), self._pickle_version)
-                simple_warning("Trying to record options which cannot be pickled "
-                               "on system with name: %s. Use the 'options_excludes' "
-                               "recording option on system objects to avoid attempting "
-                               "to record options which cannot be pickled. Skipping "
-                               "recording options for this system." % recording_requester.name,
-                               RuntimeWarning)
+                try:
+                    for key, values in user_options._dict.items():
```
Kenneth-T-Moore: Wouldn't it be more efficient to do the full pickle first, then if there is an exception, in the except block, loop over each one to find the name of the one that fails.  Otherwise, we are pickling everything twice the majority of the time.

openmdao/components/tests/test_balance_comp.py
```diff
@@ -97,6 +98,51 @@ def test_create_on_init(self):
...
+        prob.model.add_recorder(recorder)
+
+        prob.model.recording_options['record_inputs'] = True
+        prob.model.recording_options['record_outputs'] = True
+        prob.model.recording_options['record_residuals'] = True
+        bal.recording_options['record_metadata'] = False
+
+        prob.setup()
+
+        msg = ("Trying to record options which cannot be pickled on system "
```
Kenneth-T-Moore: This warning text in msg doesn't match the warning text that would be printed, so I think this test would pass even without the fix. I recommend temporarily setting the `recordable` to True in that balance option, and fix this so that it fails, then restore it to False.

openmdao/recorders/sqlite_recorder.py
DKilkenny: That's what we're doing already

Kenneth-T-Moore: you are right, I didn't notice we were already inside an exception block.

openmdao/components/tests/test_balance_comp.py
DKilkenny: fixed

PR #1182, opened 2020-02-10
Fix for our petsc transfers

naylor-b:
- changed over to using partial seeds during reverse derivative computation
- we now do an Allreduce to compute our final total jacobian row values in rev mode and a scatter to compute our final total jacobian column values in fwd mode.
- transfer indices are the same for fwd and rev modes.
- DirectSolver and BroydenSolver no longer work if they contain distributed or non-local variables. This was intentional and was done in order to allow certain parallel deriv coloring cases to continue working.
- removed all distributed DirectSolver and BroydenSolver tests
- removed all code involved in computing the non-duplicated parts of distributed vectors because they're no longer needed.


swryan:
```DirectSolver and BroydenSolver no longer work if they contain distributed or non-local variables.```

Shouldn't this situation be caught and an error raised?

naylor-b:
Yeah, I'll add that.

PR #1190, opened 2020-02-18
fixed bug in naming of non-local systems

naylor-b:
### Summary

Fixed bug in naming of non-local systems that was causing hangs in a few mpi tests under petsc 3.12.

### Related Issues

- Resolves https://www.pivotaltracker.com/story/show/170736753

None

### New Dependencies

None


naylor-b:
Added pathname test.

PR #1199, opened 2020-02-20
added better err msg for connecting output to output

### LINKED ISSUE #1193: User should get better error when attempting to make an output the target of a connection
### Summary of Issue

Currently a user gets an error that says "input does not exist for 'foo'" when attempting to make an output the target of a connection.  While true, this can be frustrating for users with large models who might not immediately realize whether a given variable is an input or output.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

See example.

### Example

The following test should be added to `test_group.py:TestConnect`

```
    def test_connect_target_is_output(self):
        msg = "Group (sub): Cannot connect source 'src.x' to target 'tgt.y' - The target is an output."

        self.sub.connect('src.x', 'tgt.y')

        with assertRaisesRegex(self, RuntimeError, msg):
            self.prob.setup()
```

### Environment

N/A

naylor-b:
### Summary

Added a specific error message for the case of trying to connect to an output.

### Related Issues

- Resolves #1193

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -881,21 +881,29 @@ class Group(System):
             # (not traceable to a connect statement, so provide context)
             if (prom_out not in allprocs_prom2abs_list_out and
                     prom_out not in self._var_allprocs_discrete['output']):
-                raise NameError(
-                    "%s: Output '%s' does not exist for connection in '%s' from '%s' to '%s'." %
-                    (self.msginfo, prom_out, self.pathname, prom_out, prom_in))
+                if (prom_out in allprocs_prom2abs_list_in or
+                        prom_out in self._var_allprocs_discrete['input']):
+                    raise NameError(
+                        "%s: Attempted to connect from '%s' to '%s', but '%s' is an input. "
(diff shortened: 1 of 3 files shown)


openmdao/core/group.py
```diff
@@ -887,9 +887,15 @@ def _setup_global_connections(self, recurse=True, conns=None):
 
             if (prom_in not in allprocs_prom2abs_list_in and
                     prom_in not in self._var_allprocs_discrete['input']):
-                raise NameError(
-                    "%s: Input '%s' does not exist for connection from '%s' to '%s'." %
-                    (self.msginfo, prom_in, prom_out, prom_in))
+                if (prom_in in allprocs_prom2abs_list_out or
+                        prom_in in self._var_allprocs_discrete['output']):
+                    raise NameError(
+                        "%s: Attempted to connect to output '%s' from output '%s'." %
```
Kenneth-T-Moore: Maybe add another sentence, something like "the target in 'connect' must be an input."

PR #1204, opened 2020-02-24
reapplied unit changes from POEM 001

robfalck:
### Summary

Adds units for astronomical units (au) and parsec (pc) to the units table, and updates the definition of light year (ly) to the IAU definition.

### Related Issues

N/A

The definition of a light year and astronomical unit change slightly to their proper IAU definitions.

### New Dependencies

None


robfalck:
> Maybe i am crazy, but can you add a test for this for the new unit names and values?
> 
> Also, is this going in for 2.10? If so, we need deprecation warning for the old 'ua' unit.

We should probably test every possible unit we have, and in that case i would not hold up 2.10.0 for this.  That's going to be a good bit of tests.   And how do we test, against the unix units command?

I'll work on that and say don't worry about this one for 2.10.0

Kenneth-T-Moore:
I was mostly just thinking of tests for the new units, and for the conversion factors that changed.

Although, an exhaustive test could uncover any mistakes in the obscure units. I recall fixing a unit a long time ago, but we haven't carried over any tests about it from old OpenMDAO:

https://github.com/OpenMDAO/OpenMDAO-Framework/commit/9fa873379df189a78ccd7d9b8cbcdd29b02c3068#diff-f20122ff6276297277f3a57110eb860c

PR #1214, opened 2020-02-28
Removing Six imports

DKilkenny:
### Summary

Removed Six imports in preparation for dropping Python 2.x support

### Related Issues

- Resolves #

Python 2.x

### New Dependencies

Python 3.6+

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -280,7 +280,7 @@ class DirectSolver(LinearSolver):
                     if 'exactly singular' in str(err):
                         raise RuntimeError(format_singular_csc_error(system, matrix))
                     else:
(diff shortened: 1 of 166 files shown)


openmdao/solvers/linear/direct.py
```diff
@@ -285,7 +280,7 @@ def _linearize(self):
                     if 'exactly singular' in str(err):
                         raise RuntimeError(format_singular_csc_error(system, matrix))
                     else:
-                        reraise(*sys.exc_info())
+                        raise sys.exc_info()
```
naylor-b: `raise sys.exc_info()` isn't valid, at least not in python 3.8.

DKilkenny: Fixed

PR #1234, opened 2020-03-05
A few fixes for support of IPOPT

### LINKED ISSUE #1224: A couple of IPOPT issues.
### Summary of Issue

Now that we are using IPOPT, a couple of issues have turned up.

### Issue Type

- [x] Bug
- [ ] Enhancement  
      POEM #___ N/A

### Description

1. The pyoptsparse test `test_show_exception_bad_opt` uses IPOPT to test an optimizer the tester doesn't have. We should switch that over to one of the other commercial opts.  Also, could skip the test if the user has the optimizer, since it defeats the purpose. We are checking the error message text here.

2. Seems to be a small bug (TypeERROR raised) after IPOPT finishes with the signal handling code.  Probably an easy fix.  Add a test as well.

### Example

'test_show_exception_bad_opt' in test_pyoptsparse shows both problems.

### Environment

Operating System: All

Python environment: All

Packages: IPOPT must be installed

Kenneth-T-Moore:
### Summary

1. Added some default opt_settings for IPOPT suitable for use with Mumps as installed by our IPOPT installer.
2. Added a test for IPOPT applied to Sellar
3. Fixed a test that used IPOPT to test the error message for setting an optimizer that is not installed.

### Related Issues

- Resolves #1224

None

### New Dependencies

None


openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -1534,6 +1536,11 @@ def test_debug_print_option(self):
 
     def test_show_exception_bad_opt(self):
 
+        # First, check if we have the optimizer for this test. If the do, then just skip it.
```
naylor-b: typo: 'If the do'

Kenneth-T-Moore: Fixed

PR #1241, opened 2020-03-06
compute_total_coloring can now be called directly with 'of' and 'wrt' lists

### LINKED ISSUE #1235: Enable coloring from compute_totals

naylor-b:
### Summary

compute_total_coloring can now be called directly with 'of' and 'wrt' lists.  This allows the coloring to be determined without defining design variables, constraints, or objectives.

### Related Issues

- Resolves #1235

Some new default args were added to compute_total_coloring, but they won't break existing code.

### New Dependencies

None


openmdao/core/problem.py
```diff
@@ -1302,7 +1302,7 @@ def check_totals(self, of=None, wrt=None, out_stream=_DEFAULT_OUT_STREAM, compac
         return data['']
 
     def compute_totals(self, of=None, wrt=None, return_format='flat_dict', debug_print=False,
-                       driver_scaling=False):
+                       driver_scaling=False, global_names=False):
```
robfalck: Why not use abs_names here to be consistent with the notion of absolute names?

naylor-b: I agree that this is inconsistent with abs_names.  I was just copying the naming used in Driver._compute_totals.  We should probably change it across the board, but that would mean deprecating global_names in Driver._compute_totals.

PR #1244, opened 2020-03-06
add -dev suffix to version

swryan:
### Summary

add `-dev` suffix to version to reduce confusion for users who pull intermediate versions between releases

### Related Issues

- Resolves #

None

### New Dependencies

None


swryan:
I guess it's not going to be that easy...

PR #1253, opened 2020-03-11
Added ability to deprecate options in OptionsDictionary

### LINKED ISSUE #1252: OptionsDictionary options support deprecation
### Summary of Issue

OptionsDictionary options which are deprecated currently must be handled during setup.  This change will raise a deprecation warning any time a deprecated option's `__setitem__` or `__getitem__` method is called.

### Issue Type

- [ ] Bug
- [X] Enhancement  
POEM N/A

### Description

The internal options for an options dictionary are currently stored in `self.__dict__`.

```
        self._dict[name] = {
            'value': default,
            'values': values,
            'types': types,
            'desc': desc,
            'upper': upper,
            'lower': lower,
            'check_valid': check_valid,
            'has_been_set': default_provided,
            'allow_none': allow_none,
            'recordable': recordable,
        }
```
This issue would add `'deprecation'` to the list.  It's default value would be None.  If that given option is deprecated, `'deprecation'` would take a string value to be issued as a DeprecationWarning during `__setitem__` and `__getitem__`.

### Example

The following would declare a system option 'foo' as deprecated.

```
    self.options.declare(name='foo',
                         types=str,
                         deprecation='Option "foo" is deprecated.  Please change to use the new option "bar" instead.')
```

### Environment

N/A

hschilling:
### Summary

OptionsDictionary options which are deprecated currently must be handled during setup. This change will raise a deprecation warning any time a deprecated option's __setitem__ or __getitem__ method is called.

### Related Issues

- Resolves #1252

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/options_dictionary.py
+++ b/openmdao/utils/options_dictionary.py
@@ -1,5 +1,7 @@
 """Define the OptionsDictionary class."""
 
+from openmdao.utils.general_utils import warn_deprecation
+
 # unique object to check if default is given
 _undefined = object()
 
@@ -24,6 +26,10 @@ class OptionsDictionary(object):
         If True, no options can be set after declaration.
     _all_recordable : bool
         Flag to determine if all options in UserOptions are recordable.
+    _setitem_warning_issued : list
+        Option names that are deprecated and a warning has been issued for their use in _setitem.
+    _getitem_warning_issued : list
+        Option names that are deprecated and a warning has been issued for their use in _getitem.
     """
 
     def __init__(self, parent_name=None, read_only=False):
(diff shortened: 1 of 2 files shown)


openmdao/utils/options_dictionary.py
```diff
@@ -43,6 +49,9 @@ def __init__(self, parent_name=None, read_only=False):
 
         self._all_recordable = True
 
+        self._setitem_warning_issued = []
```
naylor-b: Do we really need to bookkeep setitem and getitem warnings separately?

hschilling: I kind of thought so but maybe not. I can't think of a scenario where it would be really critical to do both. I will make it one. Thanks

PR #1257, opened 2020-03-12
Improvement to reporting of ill-conditioned jacobians in DirectSolver

### LINKED ISSUE #1192: Improved singular matrix errors
### Summary of Issue

Currently OpenMDAO detects some forms of singular matrix problems, but not all.  As a result, a user sometimes gets an error message that a singular matrix has been detected with no other information.  The singularity detection should be improved to provide better feedback to the user.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Consider three scenarios where a singular matrix issue can arise:

- There is no way to impact a residual in the system (a zero row)

    OpenMDAO currently handles this.

- Two rows are linearly dependent.

    If a matrix singularity is encountered and there are no zero-rows, OpenMDAO should check to see if any rows are linear combinations of others.

- Multiple rows combine to form a singularity.

    If the previous two cases are not encountered, its possible that interactions between multiple components result in a system which is singular, but is not as simple as a zero-row or a linear combination of two rows.  In this case, OpenMDAO should assess the matrix at varying levels of depth (depth-first), until the singularity is encountered.  This would at least inform the user as to the group at which the singularity exists.

### Example

We need test cases for these three scenarios.

Kenneth-T-Moore:
### Summary

1. Previously, when the splu solver reported a singularity, and we did not find a zero row or column, then we raised a general exception.  This generally occurs because a row (equation) is a linear combination of other rows (equations).  Now, we return a set of likely residual equation names.  These are determined by looking at the non-zero elements of the singular vector associated with the lowest singular value determined by np.linalg.svd.

NOTE: The singular vector sometimes includes some related equations that aren't part of the immediate problem. I have not found a way to tighten this down yet, and would like to see feedback on this version applied to some real models.

2. Previously, when the lu solver (which is used for dense assembled jacobians) encountered a similar problem, the "index" returned from numpy is not correct.  This has been fixed by ignoring that index and recomputing it using the same function used for sparse matrices, so we can correctly identify where to look for zero rows/cols and for under-determined subsets.

### Related Issues

- Resolves #1192 

None

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -116,6 +116,11 @@ def format_singular_csc_error(system, matrix):
     if zero_cols.size <= zero_rows.size:
 
         if zero_rows.size == 0:
+            u, s, v = np.linalg.svd(dense)
+            idx1 = np.where(np.abs(u[:, -1]) > 1e-15)[0]
+            idx2 = np.where(np.abs(v[-1, :]) > 1e-15)[0]
+            idx = set(idx1).intersection(idx2)
+
             # Underdetermined: duplicate columns or rows.
             msg = "Identical rows or columns found in jacobian in '{}'. Problem is " + \
                   "underdetermined."
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -67,45 +67,9 @@ def loc_to_error_msg(system, loc_txt, loc):
     return msg.format(system.msginfo, loc_txt)
 
 
-def format_singular_error(err, system, mtx):
+def format_singular_error(system, matrix):
     """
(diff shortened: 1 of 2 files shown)


openmdao/solvers/linear/direct.py
```diff
@@ -1,40 +1,31 @@
...
-    loc_txt : str
-        Either 'row' or 'col'.
     loc : int
         Index of row or column.
 
     Returns
     -------
     str
-        New error string.
+        Sting containing variable absolute name (and promoted name if there is one) and index.
```
naylor-b: typo 'Sting'

Kenneth-T-Moore: Fixed

PR #1264, opened 2020-03-18
Deprecated assert_rel_error(), replaced with assert_near_equal()

### LINKED ISSUE #1260: clean up `assert_rel_error`
### Summary of Issue

added new assert method `assert_almost_equal`

### Issue Type

- [ ] Bug
- [x] Enhancement  
      N/A -- Not needed for this. the assert methods are not part of the public API, though I know some external packages do use them. 

### Description

Current code in asser_utils has `assert_rel_error` taking the TestCase instance as its first argument. This is awkward and caused some difficulty with the command_line tool (thought a workaround was found). 

It would be simpler to just follow the numpy style and not use the TestCase instance at all, and just raise a ValueError instead. 

Unfortunately, we can't adjust the `asser_rel_error` function without breaking a lot of older code. So we can deprecate that functionality, but we need to make a new method that doesn't require that argument. 

so let's deprecate the old method, and add a new one `assert_almost_equal` that doesn't require the TestCase argument. 

### Environment

N/A -- all environments

tadkollar:
### Summary

- New function called `assert_near_equal()` because `assert_almost_equal()` conflicts with numpy function of the same name
- `assert_near_equal()` has the same behavior as `assert_rel_error()`, but drops testcase as the first argument, raises `ValueError` if the value is not within tolerance, or `KeyError` if there's a mismatch in the number of dictionary keys
- Added deprecation warning to `assert_rel_error()`
- Replaced all instances of `assert_rel_error()` with `assert_near_equal()`

### Related Issues

- Resolves #1260

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/assert_utils.py
+++ b/openmdao/utils/assert_utils.py
@@ -14,6 +14,7 @@ from openmdao.core.component import Component
 from openmdao.core.group import Group
 from openmdao.jacobians.dictionary_jacobian import DictionaryJacobian
 from openmdao.utils.general_utils import pad_name, reset_warning_registry
+from openmdao.utils.general_utils import warn_deprecation
 
 
 @contextmanager
@@ -42,7 +43,8 @@ def assert_warning(category, msg):
         if (issubclass(warn.category, category) and str(warn.message) == msg):
             break
     else:
-        raise AssertionError("Did not see expected %s: %s" % (category.__name__, msg))
+        raise AssertionError("Did not see expected %s: %s" %
+                             (category.__name__, msg))
 
 
 @contextmanager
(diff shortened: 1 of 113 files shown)


openmdao/utils/assert_utils.py
```diff
@@ -151,7 +155,8 @@ def assert_check_partials(data, atol=1e-6, rtol=1e-6):
                             norm_string = '{}'.format(norm)
                             err_msg = '{0} | {1} | {2} | {3}'.format(
                                 pad_name(wrt_string, len_wrt_width),
-                                pad_name(error_type.split()[0], len_absrel_width),
+                                pad_name(error_type.split()[
+                                         0], len_absrel_width),
```
Kenneth-T-Moore: Not sure why the split is so weird here, but we have agreed for this project to allow 100-character lines, and have adjusted our linting as such.

PR #1265, opened 2020-03-19
POEM0017 - Add a units argument for all design variables, constraints, and objectives.

### LINKED ISSUE #1189: unit conversion for design variables, constraints, and objectives
It would be nice if constraints and objectives had the option to specify units.  Many times I have to reach down into sub-models to grab variables for constraints and objectives and I am not totally sure what their units are (especially if i come back to a model after a few months of not working on it, or if the sub-model was created by someone else). This would also be useful if you wanted to change all the top level units for consistency purposes. For completeness design variables would be nice too,  but that is probably less useful b/c they are typically at the top level of my.

-Jeff

Kenneth-T-Moore:
### Summary

Added the 'units' argument to `add_design_var`, `add-constraint`, and `add_objective`, giving the ability to convert quantities in the model to a new unit before the driver sees them.

### Related Issues

- Resolves #1189

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -1658,8 +1657,77 @@ def _setup_relevance(self, mode, relevant=None):
         relevant : dict or None
             Dictionary mapping VOI name to all variables necessary for computing
             derivatives between the VOI and all other VOIs.
-
         """
+        # Update driver quantities to apply unit conversions.
```
naylor-b: This setup of unit conversions doesn't really belong in the `setup_relevance` method.  Could this stuff be done in the driver after it collects the desvars/constraints/objective?

Kenneth-T-Moore: Moved that stuff into its own function and called it from the first method in driver that is called under final_setup.

PR #1267, opened 2020-03-20
POEM013 implementation - add convert_units and unit_conversion to openmdao.api

### LINKED ISSUE #1242: POEM013 implementation
### Summary of Issue

This enhancement involves two changes:
1. `openmdao.utils.units.convert_units` will be added to `openmdao.api`
2. `openmdao.utils.units.get_conversion(from, to)` will be added to `openmdao.api` and renamed to `unit_conversion(from, to)`.  The existing `get_conversion` will be deprecated.

### Issue Type

- [ ] Bug
- [x] Enhancement  

POEM: OpenMDAO/POEMs#27

### Description

POEM013 exists to give users easier access to OpenMDAO's unit conversion capabilities so they can perform unit conversions in their models and tools that are consistent with the conversions in OpenMDAO.

### Example

```
from openmdao.units.units import convert_units
...
distance_ft = convert_units(inputs['distance'], 'm', 'ft')
```

becomes

```
import openmdao.api as om
...
distance_ft = om.convert_units(inputs['distance'], 'm', 'ft')
```

### Environment

N/A

tadkollar:
### Summary

- Renamed existing get_conversion to unit_conversion
- get_conversion now consists of a deprecation warning and a call to unit_coversion
- convert_units and unit_conversion added to openmdao.api
- All calls to get_conversion replaced with unit_conversion

### Related Issues

- Resolves #1242

None

### New Dependencies

None


openmdao/utils/tests/test_units.py
```diff
@@ -245,11 +245,11 @@ def test_name(self):
         y = x2 / (x1**2)
         self.assertEqual(y.name(), 'kg/m**2')
 
-    def test_get_conversion(self):
```
swryan: I think typically we would add a test for the deprecation as well

tadkollar: I can add it, but get_conversion now only does the deprecation warning and returns a call to unit_coversion with the same args, so there's not much being tested

swryan: yeah, the test would really just be that you get the deprecation warning

PR #1277, opened 2020-03-24
abs2prom not recorded to driver_metadata table since it is already recorded to metadata table

### LINKED ISSUE #1276: abs2prom is recorded twice in the case recorder file
### Summary of Issue

abs2prom is recorded twice in the case recorder file: the metadata table and the driver_metadata table.

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #011

### Description

abs2prom is currently recorded into the driver_metadata table. But it is already recorded into the metadata table which is part of every case recorder file. Update the code to skip recording abs2prom into the driver_metadata table and then in the n2_viewer.py function, _collect_metadata, get the abs2prom info from what is already there from the metadata table. 

### Example

N/A

### Environment

N/A

hschilling:
### Summary

abs2prom not recorded to driver_metadata table since it is already recorded to metadata table

### Related Issues

- Resolves #1276 

None

### New Dependencies

None


swryan:
It seems like the  n2 viewer does not need `abs2prom` at all since removing it from  `viewer_data` does not break `test_viewmodeldata.py:TestViewModelData.test_n2_from_sqlite`. Is that right?

If so then it seems like we should just not add it to `data_dict` in `_get_viewer_data` at all rather than pop it later...

swryan:
from a quick look at `n2.html`, it looks like `abs2prom` is optional:
```
        this.abs2prom = modelJSON.abs2prom; // May be undefined.
```
```
            .on("mouseover", function(d) {
                if (self.model.abs2prom != undefined) {
                    if (d.isParam()) {
                        return self.dom.toolTip.text(
                                self.model.abs2prom.input[d.absPathName])
                            .style("visibility", "visible");
                    }
                    if (d.isUnknown()) {
                        return self.dom.toolTip.text(
                                self.model.abs2prom.output[d.absPathName])
                            .style("visibility", "visible");
                    }
                }
            })
```
so I am wondering what the effect is of not having `abs2prom` in the dict when viewing the n2 from a recording...

PR #1295, opened 2020-04-01
Add recording of derivatives to Problem

### LINKED ISSUE #1281: Add recording of derivatives to Problem
### Summary of Issue

When calling `record_state` on Problem, there should also be an option to record derivatives. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #011

### Description

Problem should be able to record the same data as Driver, and that includes recording of derivatives. 

Need to add an option to Problem to allow for recording derivatives and then provide a way to access this information via a case reader.

### Example

N/A

### Environment

N/A

hschilling:
### Summary

Added an option to Problem to allow for recording derivatives and then provided a way to access this information via a case reader.

### Related Issues

- Resolves #1281

None

### New Dependencies

None


openmdao/core/driver.py
```diff
@@ -1053,10 +1053,18 @@ def record_iteration(requester, prob, case_name):
     outs = model._retrieve_data_of_kind(filt, 'output', 'nonlinear', parallel)
     ins = model._retrieve_data_of_kind(filt, 'input', 'nonlinear', parallel)
 
+    from openmdao.core.problem import Problem
+    if not isinstance(requester, (Problem, Driver)):
+        raise ValueError('Unexpected object requesting recording: {0}'.format(requester))
```
Kenneth-T-Moore, line 1058: Does this happen anywhere?  What could the user do to trigger this?

hschilling: Not that I know of. Just being extra cautious.

Kenneth-T-Moore: Since it's not public API, we probably don't need to do this.  The line has no test coverage anyway.

hschilling: ok

PR #1310, opened 2020-04-08
handle skip decorators when running doc tests

### LINKED ISSUE #1216: Current doc building code does not make use of skip decorators
### Summary of Issue

The doc building code should honor skip decorators on embedded tests. 

### Issue Type

- [X] Bug
- [ ] Enhancement  
      POEM #N/A

### Description

Currently, the embed code is supposed to run the code with the decorators so that skip decorators can skip tests that can't run on a given environment. There's a lot of code to handle skipped tests but it's never used because of this. Skipped tests are still supposed to result in the code being put into the docs but for the output section, we state the code was skipped and why.

Skip should raise a warning in order to fail on CI if a skip is encountered.

### Example

For example, this test decorator is ignored when doing the embed code when doing the doc build.

```
   @unittest.skipUnless(LooseVersion(scipy_version) >= LooseVersion("1.2"),
                         "scipy >= 1.2 is required.")
    def test_feature_shgo_rastrigin(self):

```

### Environment

Operating System: <All>

Python environment: <All>

Packages: N/A

swryan:
### Summary

handle skip decorators when running doc tests

### Related Issues

- Resolves #1216

None

### New Dependencies

None


openmdao/docs/_exts/embed_code.py
```diff
@@ -169,9 +176,23 @@ def run(self):
                 if 'plot' in layout:
                     code_to_run = code_to_run + ('\nmatplotlib.pyplot.savefig("%s")' % plot_file_abs)
 
-            skipped, failed, run_outputs = run_code(code_to_run, path, module=module, cls=class_,
-                                                    imports_not_required=imports_not_required,
-                                                    shows_plot=shows_plot)
+            if is_test and decorators is not None:
+                # check whether test is skipped due to decorator
+                test_suite = unittest.TestLoader().loadTestsFromName(path)
+                test_result = unittest.TextTestRunner(stream=StringIO()).run(test_suite)
```
Kenneth-T-Moore, line 182: What is the performance penalty for running each test twice like this whenever we build the docs?

swryan: there is a hit no doubt, but the test has to run in the test context for the decorator and again in the isolated context for the docs..   I don't see a way around it if we want to build docs with skipped tests as described in the issue...  open to suggestions

swryan: this is only done for decorated tests of course

naylor-b: It turns out that if a method is going to be skipped, the method instance will have two attributes called '__unittest_skip__' and '__unittest_skip_why__', so you can simply check for the existence of '__unittest_skip__' for example and know that it'll be skipped without actually running the test.  Our current 'get_source_code' function already looks up the method object although it currently doesn't return it, but we could easily return it and look for the `__unittest_skip__` attribute to see if that test is going to be skipped.

swryan: ok will look at this

PR #1311, opened 2020-04-08
P1296 problem in out resids record

### LINKED ISSUE #1296: When recording Problem states, user should be able to record inputs, outputs, and residuals
### Summary of Issue

Add recording of inputs, outputs, and residuals to Problem state recording.  

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #011

### Description

Add options for recording inputs, outputs, and residuals to Problem.

Store those values in the problem recording table in the case recording file.

Provide a way to access those values from Problem cases.

### Example

N/A

### Environment

N/A

hschilling:
### Summary

Add recording of inputs, outputs, and residuals to Problem. Add recording of outputs and residuals to Driver

### Related Issues

- Resolves #1296

None

### New Dependencies

None


openmdao/recorders/sqlite_recorder.py
```diff
@@ -687,3 +701,19 @@ def shutdown(self):
         # close database connection
         if self.connection:
             self.connection.close()
+
+    def delete_recordings(self):
+        """
```
Kenneth-T-Moore, line 706: Is this function part of another story? it never gets used and isn't tested (and the comment is cut-and-paste.)

hschilling: I use it four times in test_sqlite_recorder.py. I use it to clear out the case recorder file so I can do a lot of runs with different options and not have to make a whole new test. 

Good catch on the comment! Deleted.

Kenneth-T-Moore: I see, i must have had that file closed when I searched.  It makes sense now.

openmdao/core/driver.py
```diff
@@ -1061,20 +1073,39 @@ def record_iteration(requester, prob, case_name):
...
-            prob.driver._designvars and prob.driver._responses:
-        totals = requester.compute_totals(return_format='flat_dict_structured_key')
-        data['totals'] = totals
+    if isinstance(requester, Problem):
+        inputs, outputs, residuals = model.get_nonlinear_vectors()
+        vec_name = 'nonlinear'
+
+        discrete_inputs = model._discrete_inputs
+        discrete_outputs = model._discrete_outputs
+        filt = requester._filtered_vars_to_record
```
swryan, line 1097: The above lines appear to be unnecessary? 
- 1092-1093 repeat 1076-1077
- 1095-1096 are unused?
- 1097 repeats 1073

Also, if requester is Problem lines 1081-1088 are executed unnecessarily?

hschilling: let me look! I wouldn't doubt this needs work

PR #1323, opened 2020-04-14
Adds option to KSComp to automatically add a corresponding constraint

### LINKED ISSUE #1315: KSComp should add a constraint to its output.
### Summary of Issue

KSComp currently forms an aggregated constraint but doesn't actually add the constraint to the optimization.  It should optionally add this constraint (and in the future, this constraint should be added by default)

### Issue Type

- [x] Bug
- [ ] Enhancement  
      POEM #N/A

### Description

The output of KSComp `KS` is a function that should be constrained to enforce an aggregated constraint.

If option `add_constraint=False`, the default, the current behavior will result.
If `add_constraint=True`, then KSComp will add the appropriate constraint in its setup method.

### Example

N/A

### Environment

N/A

robfalck:
### Summary

KSComp will now accept `add_constraint` as an option.  If True, it will automatically add a corresponding constraint on its output during setup.

### Related Issues

- Resolves #1315

None

### New Dependencies

None


openmdao/components/tests/test_ks_comp.py
```diff
@@ -121,6 +121,42 @@ def test_beam_stress(self):
         self.assertTrue(np.all(stress0 < 100.0))
```
Kenneth-T-Moore, line 121: I think you need to add an example to the feature doc for KSComp.  This is mostly to address the weakness in the current documentation that led to this story, namely that none of the examples clue the user in to the fact that they need to add a constraint (or now, set the new option to True) in order for their KSComp to be active in the optimization problem.

Some of the existing docs and writing should also be updated because of this.

robfalck: Example added for `add_constraint=True`

PR #1325, opened 2020-04-14
set_solver_print is now a public method on System

### LINKED ISSUE #1301: set_solver_print needs to be made callable from a group level
### Summary of Issue

Need finer grained control of how this macro is applied

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM not needed

### Description

It is useful to be able to call this on one group specifically, rather than just on the problem as a whole.  

Leave the problem level method, but also make a version of it on system that works recursively from whatever level it's called down.

robfalck:
### Summary

This PR makes set_solver_print available on Systems (in addition to still being available on Problem).

### Related Issues

- Resolves #1301 

Previous `_set_solver_print` method on System has been renamed to `set_solver_print`.

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2460,6 +2460,11 @@ class System(object):
     def _setup_solver_print(self, recurse=True):
         """
         Apply the cached solver print settings during setup.
+
+        Parameters
+        ----------
+        recurse : bool
+            Whether to call this method in subsystems.
         """
         for level, depth, type_ in self._solver_print_cache:
             self._set_solver_print(level, depth, type_)
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -34,9 +34,9 @@ from openmdao.utils.name_maps import name2abs_name
 from openmdao.utils.coloring import _compute_coloring, Coloring, \
     _STD_COLORING_FNAME, _DEF_COMP_SPARSITY_ARGS
 import openmdao.utils.coloring as coloring_mod
-from openmdao.utils.general_utils import determine_adder_scaler, find_matches, \
+from openmdao.utils.general_utils import determine_adder_scaler, \
     format_as_float_or_array, ContainsAll, all_ancestors, \
-    simple_warning, make_set, match_includes_excludes
+    simple_warning, make_set, match_includes_excludes, warn_deprecation
 from openmdao.approximation_schemes.complex_step import ComplexStep
 from openmdao.approximation_schemes.finite_difference import FiniteDifference
 from openmdao.utils.units import unit_conversion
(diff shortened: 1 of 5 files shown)


openmdao/core/system.py
```diff
@@ -2445,13 +2445,36 @@ def _set_solver_print(self, level=2, depth=1e99, type_='all'):
...
 
-            subsys._set_solver_print(level=level, depth=depth - current_depth, type_=type_)
+            subsys.set_solver_print(level=level, depth=depth - current_depth, type_=type_)
 
             if subsys._linear_solver is not None and type_ != 'NL':
                 subsys._linear_solver._set_solver_print(level=level, type_=type_)
             if subsys.nonlinear_solver is not None and type_ != 'LN':
                 subsys.nonlinear_solver._set_solver_print(level=level, type_=type_)
 
+    def _set_solver_print(self, level=2, depth=1e99, type_='all'):
```
swryan, line 2455: We shouldn't have to deprecate an underscored function, no?
Is this just because it has become  a _de facto_ API function?

Kenneth-T-Moore: I agree, though if you feel you do have to deprecate it, then add a test too.

robfalck: I'm OK with the standard of not deprecating non-API functions, I'll remove the deprecation.

swryan:
I think maybe this doc also needs updated, since the method is now available on any System and not just on Problem:
```
Controlling Solver Output in Large Models
When you have a large model with multiple solvers, it is easier to use a shortcut method that recurses over the entire model.
The set_solver_print method on problem can be used to set the iprint to one of four specific values for all solvers in the model
```
http://openmdao.org/twodocs/versions/latest/features/core_features/controlling_solvers/solver_options.html#solver-options-set-solver-print

robfalck:
Making a test now, please don't merge yet.

robfalck:
Due to some issues I found, this is now a more substantial rewrite.

The `_solver_print_cache` has been moved into System, and the cached print settings are applied during final setup.

openmdao/solvers/tests/test_solver_iprint.py
```diff
@@ -316,6 +316,45 @@ def test_feature_set_solver_print3(self):
...
+
+        g2.nonlinear_solver = om.NewtonSolver(solve_subsystems=False)
+        g2.linear_solver = om.ScipyKrylov()
+        g2.linear_solver.precon = om.LinearBlockGS()
+        g2.linear_solver.precon.options['maxiter'] = 2
+
+        prob.set_solver_print(level=-1)
+        g2.set_solver_print(level=2, type_='NL')
+
+        prob.setup()
```
Kenneth-T-Moore, line 355: Need to capture the stdout and test the contents to enforce that this works as expected.  Do this in a second test so that the extra code doesn't show up in the feature example.

robfalck: Added output checks to all non-feature tests in test_solver_iprint.py

PR #1329, opened 2020-04-15
Fix for MPI hang when using DirectSolver on a serial model run on multiple procs.

### LINKED ISSUE #1322: Unexpected behavior in degenerate MPI use case
### Summary of Issue

A degenerate Dymos case that is entirely serial yet run with `mpirun -n 2` experiences singularities in the nonlinear solves and then locks up (fails to complete)

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

In developing the capability to specify whether phases are run serially or in parallel in Dymos, we encountered what seems to be a bug in OpenMDAO.

The case is setup entirely as a serial case.  Running with `mpirun -n 2` causes singularities in the nonlinear solve.  

- If the phases are placed in a parallel group (parallel_phases=True in the Trajectory setup), the problem behaves as expected
- If the problem as-is is run with `mpirun -n 1` or without MPI, it behaves as expected

### Example

The example in question is at: https://github.com/robfalck/dymos/tree/ted_parallelPhases
in directory: `dymos/examples/battery_multibranch/test`

Run with
`mpirun -n 2 python test_multibranch_trajectory_rk_ivp.py TestBatteryRKIVP.test_static_input_params` experiences the failure but running with
`mpirun -n 1 python test_multibranch_trajectory_rk_ivp.py TestBatteryRKIVP.test_static_input_params` will complete successfully.

### Environment

openmdao/master as of 04/13/2020

Kenneth-T-Moore:
### Summary

Fix for MPI hang when using DirectSolver on a serial model run on multiple procs.

### Related Issues

- Resolves #1322 

None

### New Dependencies

None

### DIFF
--- a/openmdao/jacobians/assembled_jacobian.py
+++ b/openmdao/jacobians/assembled_jacobian.py
@@ -11,8 +11,6 @@ from openmdao.matrices.csr_matrix import CSRMatrix
 from openmdao.matrices.csc_matrix import CSCMatrix
 from openmdao.utils.units import unit_conversion
 from openmdao.utils.array_utils import _flatten_src_indices
-from openmdao.utils.mpi import MPI
-from openmdao.vectors.vector import INT_DTYPE
 
 _empty_dict = {}
 
@@ -114,13 +112,11 @@ class AssembledJacobian(Jacobian):
         ext_mtx = self._matrix_class(system.comm, False)
 
         iproc = system.comm.rank
-        abs2idx = system._var_allprocs_abs2idx['nonlinear']
         in_sizes = system._var_sizes['nonlinear']['input']
         out_ranges = self._out_ranges
         in_ranges = self._in_ranges
 
         abs2prom_out = system._var_abs2prom['output']
-        owns = system._owning_rank
         conns = {} if isinstance(system, Component) else system._conn_global_abs_in2out
         abs_key2shape = self._abs_key2shape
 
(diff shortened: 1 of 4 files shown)


openmdao/jacobians/assembled_jacobian.py
```diff
@@ -280,23 +276,16 @@ def _get_subjac_iters(self, system):
...
-                iscomp = True
             else:
                 global_conns = system._conn_global_abs_in2out
-                iscomp = False
 
-            output_names = set(n for n in system._var_abs_names['output']
-                               if iscomp or owned[n] == irank or meta[n]['distributed'])
-            input_names = set(n for n in system._var_abs_names['input']
-                              if iscomp or owned[n] == irank or meta[n]['distributed'])
+            output_names = set(n for n in system._var_abs_names['output'])
```
naylor-b, line 287: this can be reduced to `set(system._var_abs_names['output']) and the next line similarly.

openmdao/matrices/coo_matrix.py
```diff
@@ -75,17 +63,13 @@ def _build_coo(self, system):
...
-                        delta = info['shape'][0] * len(src_indices)
-                    else:
-                        delta = shape[0] * len(src_indices)
+                    delta = shape[0] * len(src_indices)
+
             elif rows is None:  # sparse matrix
                 delta = val.data.size
+
             else:  # list sparse format
                 delta = len(rows)
```
naylor-b, line 75: There's a method called '_get_assembled_matrix' here and in dense_matrix.py that is no longer called and can be removed.

naylor-b, line 75: There's a method here called _get_assembed_matrix that is no longer called and can be removed.  The same method can be removed in dense_matrix.py.  Also, there's an attribute here called _first_gather that can be removed.

openmdao/jacobians/assembled_jacobian.py
Kenneth-T-Moore: Fixed

openmdao/matrices/coo_matrix.py
Kenneth-T-Moore: Got rid of all the extra stuff.

Issue #1331, opened 2020-04-16
Cannot use outputs of distributed components as pyoptsparse objectives or constraints


bbrelje:
### Summary of Issue

While developing on a distributed component in another project, I found a couple of undocumented (I think) things about the distributed component implementation.

1) Setting the output of a distributed component as either a constraint or objective of an optimization problem does not work correctly (at least with pyoptsparsedriver) when running in parallel

2) Check_partials and check_totals do not appear to work correctly when running parallel

### Issue Type

- [ X] Bug
- [ X] Enhancement  

### Description

Adding an output of a distributed component as an optimization constraint does not work (produces a pyoptsparse error message) although list_outputs and other openmdao-based interactions display correctly concatenated distributed arrays.

See the gist below.

Also, check_totals and check_partials appear to be broken right now when run in parallel on a distributed component which makes developing parallel components much trickier. If only non-distributed comps are include= in the check_partials, then it works, so the problem is not about parallel in general, but about the distributed outputs.

### Example

See the gist here: [link](https://gist.github.com/bbrelje/b6ac496b96c1ce1d393226078314d25d)

When run under mpi with n=2 or more, the GroupThatDoesNotWork() the following error is raised:
```
`+------------------------------------------------------------------------------+
| pyOptSparse Error: 1 constraint values were returned in distcomp.y, but      |
| expected 2.                                                                  |
+------------------------------------------------------------------------------+

Traceback (most recent call last):
  File "toy.py", line 88, in <module>
    prob.run_driver()
  File "/usr/local/lib/python2.7/dist-packages/openmdao/core/problem.py", line 592, in run_driver
    return self.driver.run()
  File "/usr/local/lib/python2.7/dist-packages/openmdao/drivers/pyoptsparse_driver.py", line 410, in run
    storeHistory=self.hist_file, hotStart=self.hotstart_file)
  File "/usr/local/lib/python2.7/dist-packages/pyoptsparse/pySLSQP/pySLSQP.py", line 281, in __call__
    self._waitLoop()
  File "/usr/local/lib/python2.7/dist-packages/pyoptsparse/pyOpt_optimizer.py", line 634, in _waitLoop
    self._masterFunc2(*info)
  File "/usr/local/lib/python2.7/dist-packages/pyoptsparse/pyOpt_optimizer.py", line 352, in _masterFunc2
    fcon = self.optProb.processConstraints(funcs)
[...]

PR #1333, opened 2020-04-18
Prevent users from adding obj or con from distributed components

bbrelje:
### Summary

This is a tempfix to add a NotImplementedError in the situation where a user tries to use a distributed output as an objective or constraint

### Related Issues

- Tempfix for  #1331 

- It is theoretically possible that someone has a runscript that only runs the model (no driver) but also adds a distributed output as a response variable. This change will cause that runscript to fail with an insightful error message

### New Dependencies
None


swryan:
We should probably have a test for this, even though it's just a temporary fix.

PR #1337, opened 2020-04-19
Fix FD derivatives on distributed components closes #1279

### LINKED ISSUE #1279: Approximating derivatives with FD on a distributed component does not appear to work.
### Summary of Issue

Approximating derivatives with FD on a distributed component does not appear to work.

### Issue Type

- [ x] Bug
- [ ] Enhancement  

### Description

This may be unsupported at present. If that is the case, we should raise an exception if someone tries it.

### Environment

Operating System: All

Python environment: All

Packages: All

bbrelje:
### Summary

Finite difference approximations were broken for distributed components.
The reason was that the FD routine had separate paths for parallel groups but not distributed components and the way proc ownership was handled wasn't quite right, causing an array mismatch error.

### Related Issues

- Resolves #1279

If test coverage on parallel groups is poor then I might have messed something up?

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -277,6 +277,7 @@ class ApproximationScheme(object):
             self._approx_groups.append((wrt, data, in_idx, tmpJ, [(arr, in_idx)], None))
 
     def _compute_approximations(self, system, jac, total, under_cs):
+        from openmdao.core.component import Component
         # Clean vector for results
         results_array = system._outputs._data.copy() if total else system._residuals._data.copy()
 
@@ -290,6 +291,7 @@ class ApproximationScheme(object):
         par_fd_w_serial_model = use_parallel_fd and system._num_par_fd == system._full_comm.size
         num_par_fd = system._num_par_fd if use_parallel_fd else 1
         is_parallel = use_parallel_fd or system.comm.size > 1
+        is_distributed = isinstance(system, Component) and system.options['distributed']
 
         results = defaultdict(list)
         iproc = system.comm.rank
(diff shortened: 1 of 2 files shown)


openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -290,6 +291,7 @@ def _compute_approximations(self, system, jac, total, under_cs):
         par_fd_w_serial_model = use_parallel_fd and system._num_par_fd == system._full_comm.size
         num_par_fd = system._num_par_fd if use_parallel_fd else 1
         is_parallel = use_parallel_fd or system.comm.size > 1
+        is_distributed = isinstance(system, Component) and system.options['distributed']
```
naylor-b, line 294: What happens is we're doing group FD instead of component FD and the group contains distributed components?

Kenneth-T-Moore: Group FD is still broken, but this story fixes component FD and check_partials.

openmdao/core/tests/test_distribcomp.py
```diff
@@ -327,6 +365,32 @@ def test_distrib_full_in_out(self):
...
+        p.final_setup()
+
+        # run the model and check partials
+        p.run_model
+
+        # this used to fail (bug #1279)
+        cpd = p.check_partials(out_stream=None)
+        for (of, wrt) in cpd['C2']:
+            print(of)
+            print(wrt)
```
Kenneth-T-Moore, line 390: I would remove these prints.

bbrelje: Good catch, thanks

openmdao/approximation_schemes/approximation_scheme.py
naylor-b: OK I'll put in a story to fix group FD as well.

PR #1338, opened 2020-04-19
Fixed bug in options passing for bspline comp

johnjasa:
### Summary

The kwargs for the `bsplines` method of `SplineComp` were not initialized correctly.
This means that when a user sets the order of the bspline in the `interp_options` dictionary, that order was not actually used, and instead `order=4` was always used.
In the case where we only have three control points, for example, this causes a halting error as the bspline cannot be formed.

### Related Issues

N/A

None

### New Dependencies

None


openmdao/components/interp_util/interp_bsplines.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, grid, values, interp=None, **kwargs):
         **kwargs : dict
             Interpolator-specific options to pass onward.
         """
-        super(InterpBSplines, self).__init__(grid, values, interp)
+        super(InterpBSplines, self).__init__(grid, values, interp, **kwargs)
 
         self._vectorized = True
         self.k = self.options['order'] + 1
```
Kenneth-T-Moore, line 51: Can you add a simple test that fails because the order was not passed correctly?

johnjasa: Just pushed one now, thanks!

PR #1343, opened 2020-04-20
Converted output lists to np.arrays and added test

### LINKED ISSUE #1292: When discrete variables are recorded, integer arrays are turned into integer lists.
### Summary of Issue

When discrete variables are recorded, integer arrays are turned into integer lists.

### Issue Type

- [ x] Bug
- [ ] Enhancement  

### Description

This happens in the SQLite recorder during serialization in the `make_serializable` call.

### Example

Inspect the outputs in the case object in TestDOEDriver.test_discrete_arraydesvar_list.

### Environment

Operating System: 

Python environment: 

Packages:

DKilkenny:
### Summary

When reading the contents of a case from `get_case` in `sqlite_reader.py` it will now convert the outputs from a list to a np.ndarray

### Related Issues

- Resolves #1292 

None

### New Dependencies

None


openmdao/recorders/sqlite_reader.py
```diff
@@ -648,7 +648,12 @@ def get_case(self, case_id, recurse=False):
 
         for table in tables:
             case = table.get_case(case_id)
-            if case:
+            if case and case.outputs is not None:
+                for key, value in case.outputs.items():
+                    if isinstance(case[key], list):
+                        case.outputs[key] = np.asarray(value)
```
swryan, line 654: This is the wrong place to fix this... 

The current logic is here:
https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/utils/record_util.py#L192
It only converts to an array if there is no shape associated with the value...  allowing for an actual list to be serialized and deserialized.

I think a more correct fix is to add the shape to the metadata in `add_discrete_input` and `add_discrete_output`  (in both component.py and indepvarcomp.py) if the type of the discrete is an array.

PR #1349, opened 2020-04-22
Both promoted and absolute names now in node info panel

### LINKED ISSUE #1309: N2 Info panel for inputs should contain the promoted and absolute names
### Summary of Issue

When you hover over any variable, you get a tooltip with its promoted name. The info panel only shows the absolute name. Show them both, and label them both in the info panel. 

Remove tool tip once this is complete. 

### Issue Type

- [ ] Bug
- [x] Enhancement

tadkollar:
### Summary

The node info panel used to show just the absolute name, and now it shows both the absolute and promoted names with proper labels. The additional tooltip was removed since it was redundant. The cursor changes to a crosshair when the info panel mode is enabled.  Also, "param" is now renamed to "Input" and "unknown" renamed to "Output" in this view.

A test was added for the previous problem of offscreen arrows being drawn for collapsed nodes.

### Related Issues

- Resolves #1309

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/src/N2Diagram.js
+++ b/openmdao/visualization/n2_viewer/src/N2Diagram.js
@@ -349,7 +349,7 @@ class N2Diagram {
             });
 
         // Create a new SVG group for each node in zoomedNodes
-        let nodeEnter = selection.enter().append("svg:g")
+        let nodeEnter = selection.enter().append("g")
             .attr("class", function (d) {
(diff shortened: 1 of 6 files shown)


hschilling:
Except for using a different cursor, it looks great!

tadkollar:
Travis builds passed but still showing checks pending, not sure why

PR #1355, opened 2020-04-23
Rename Problem.record_state to Problem.record

### LINKED ISSUE #1354: Rename Problem.record_case to just Problem.record
### Summary of Issue

record_state is not ideal because "state" is a loaded term.

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

record_state is not ideal because "state" is a loaded term. The previous name, record_iteration, was not correct either since it wasn't really recording an iteration. Just taking a snapshot of Problem. 

### Example

N/A

### Environment

N/A

hschilling:
### Summary

Rename Problem.record_state to Problem.record since users do not think only "state variables" are being recorded. Actually, all variables are being recorded.

### Related Issues

- Resolves #1354

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -645,7 +645,7 @@ class Problem(object):
         for system in self.model.system_iter(include_self=True, recurse=True):
             system.cleanup()
 
-    def record_state(self, case_name):
+    def record(self, case_name):
         """
         Record the variables at the Problem level.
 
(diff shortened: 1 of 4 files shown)


openmdao/core/problem.py
```diff
@@ -666,7 +666,7 @@ def record_iteration(self, case_name):
             Name used to identify this Problem case.
         """
         warn_deprecation("'Problem.record_iteration' has been deprecated. "
-                         "Use 'Problem.record_state' instead.")
+                         "Use 'Problem.record' instead.")
```
Kenneth-T-Moore, line 669: Should we deprecate record_state (and add a test for that?)

hschilling: I thought about that but record_state has only been out there for a while and was never released. I can easily add it, obviously!

Kenneth-T-Moore: Oh, if it's not in a release then maybe we don't need to.

PR #1358, opened 2020-04-23
Prevent users from setting driver.supports options

### LINKED ISSUE #1294: User should not be able to change options in driver.supports.
### Summary of Issue

User should not be able to change options in driver.supports.

### Issue Type

- [x ] Bug
- [ ] Enhancement  

### Description

The support options dictionary defines what a driver can do. Changing one of these options from False to True doesn't add support for unsupported features.  I think all we have to do is set the `_read_only` attribute on the optionsdictionary to True.

### Environment

Operating System:

Python environment

Packages:

DKilkenny:
### Summary

User is no longer able to change the options in `driver.supports` dictionary

### Related Issues

- Resolves #1294 

Will now throw error if user tries to set an option on `driver.supports` OptionsDictionary.

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/scipy_optimizer.py
+++ b/openmdao/drivers/scipy_optimizer.py
@@ -132,6 +132,7 @@ class ScipyOptimizeDriver(Driver):
         self.supports['multiple_objectives'] = False
         self.supports['active_set'] = False
         self.supports['integer_design_vars'] = False
+        self.supports._read_only = True
 
         # The user places optimizer-specific settings in here.
         self.opt_settings = OrderedDict()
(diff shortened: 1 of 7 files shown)


openmdao/drivers/scipy_optimizer.py
```diff
@@ -188,6 +189,7 @@ def _setup_driver(self, problem):
         super(ScipyOptimizeDriver, self)._setup_driver(problem)
         opt = self.options['optimizer']
 
+        self.supports._read_only = False
         self.supports['gradients'] = opt in _gradient_optimizers
         self.supports['inequality_constraints'] = opt in _constraint_optimizers
         self.supports['two_sided_constraints'] = opt in _constraint_optimizers
```
Kenneth-T-Moore, line 195: Let's set these back to `_read__only =True` when done setting.

DKilkenny: Fixed

PR #1385, opened 2020-05-04
Configure timing

JustinSGray:
Initial step in restructuring the setup/configure stack slightly to make it simpler and enable the availability of list_inputs/list_outputs from within the configure of any group. 

This first step removed the _post_configure methods from all the components. It was being used to add additional I/O after the normal configure stack was called, but upon inspection that delayed I/O addition wasn't necessary. Instead we just added the I/O when the methods were called directly. 

There was some remaining error checking that was being done in `component.py` and `indepvarcomp.py` which verified the i/o configuration. Since that did not change it, I left that timing alone, but renamed the method to `_configure_check`


openmdao/core/indepvarcomp.py
```diff
@@ -69,31 +70,18 @@ def __init__(self, name=None, val=1.0, **kwargs):
...
-        self._var_rel2meta = {}
-
-        for (name, val, kwargs) in self._indep + self._indep_external:
-            super(IndepVarComp, self).add_output(name, val, **kwargs)
-
-        for (name, val, kwargs) in self._indep_external_discrete:
-            super(IndepVarComp, self).add_discrete_output(name, val, **kwargs)
-
+       
         if len(self._indep) + len(self._indep_external) + len(self._indep_external_discrete) == 0:
```
Kenneth-T-Moore, line 78: I don't think we need all these _indep, _indep_external, or _indep_external_discrete anymore.  If there isn't already a way in the parent class to keep track of the number of outputs (discrete and nondiscrete), then replace all three of these with a single counter or even a boolean that is set to True if `add_output` or `_add_discrete_output` is called.

openmdao/components/add_subtract_comp.py
```diff
@@ -150,58 +158,67 @@ def add_equation(self, output_name, input_names, vec_size=1, length=1, val=1.0,
         res_ref : float or ndarray
             Scaling parameter. The value in the user-defined res_units of this output's residual
             when the scaled value is 1. Default is 1.
+        append : bool
+            If True, add the given equation to the list.  If False, add the appropriate I/O but
```
Kenneth-T-Moore, line 162: probably should add a test for `append=False`

openmdao/components/add_subtract_comp.py
```diff
@@ -150,58 +158,67 @@ def add_equation(self, output_name, input_names, vec_size=1, length=1, val=1.0,
...
-                                  scaling_factors, kwargs))
 
-    def add_output(self):
-        """
-        Use add_equation instead of add_output to define equation systems.
-        """
-        raise NotImplementedError(self.msginfo + ': Use add_equation method, not add_output '
-                                  'method to create an addition/subtraction relation')
+        if isinstance(input_names, str):
+            input_names = input_names
```
naylor-b, line 170: did you mean `input_names = [input_names]` here?

openmdao/core/indepvarcomp.py
naylor-b: Agreed.  Looks like those just aren't needed anymore at all.  It could just check the length of self._static_var_rel_names['output'] and self._var_rel_names['output'] or the length of self._static_var_rel2meta and self._var_rel2meta.

openmdao/components/add_subtract_comp.py
JustinSGray: actually, a length 1 input_names set makes no sense... so I changes it to an error condition

PR #1386, opened 2020-05-04
Improved several N2 resizing issues

### LINKED ISSUE #1332: N2 diagram sizing/resizing needs improvement
### Summary of Issue

The N2 diagram initially maximizes its height based only on the height of the window, ignoring the width, which can lead to a very wide diagram going offscreen. The size selection slider doesn't always behave as intended and can sometimes be very touchy. Also, when embedded in a document, the width of the embedded `<object>` should respect the width of the rest of the content in the document.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

- Check both the window height and width when selecting a size for drawing the diagram on the initial load. This can probably be done in N2Layout since the proportions won't be known until the numbers of columns and their widths are computed.
- Augment or replace the toolbar size slider control by making the bottom-right corner of the diagram draggable (e.g. using style `resize: both` on a container `<div>`)
- The `embed-n2` code may need some work to properly select the `<object>` size when the diagram is embedded and not in control of the entire browser window

tadkollar:
### Summary

Updates occurred in several areas:
- Very large models will initially be scaled to fit in the window. The layout is computed with 95% of the window height, and if it overflows the side of the window, the entire diagram is scaled down.
- Resizing the window will automatically resize the diagram (when dragging stops) **if** the user hasn't manually resized it yet.
- The diagram itself can be resized within the window by dragging the lower right corner. This creates a box outline with the current dimensions displayed until the user releases the mouse button.
- The toolbar resizing slider now displays the scale as a percent of window height. There's also a `Fit` button that will resize the diagram to fill the window (this causes the diagram to automatically resize with the window again, until the user manually resizes).
- Embedded N2s are now in an iframe container with a light gray border, which is resizable and has the same effect as resizing the window in a non-embedded N2.
- Added a `toolbar` optional arg to embed-n2 to allow the toolbar to be displayed by default when specified in docs.
- A small fix to the docs Makefile to create the _n2html directory first.

### Related Issues

- Resolves #1332

None

### New Dependencies

None


tadkollar:
> Only saw one bug. When you resize the diagram, you lose the drawing of the pinned connection arrows.

It's always been that way - it's really out of scope to fix it for this issue as it's not completely trivial. There is an old related Pivotal story that we should create a new issue for: https://www.pivotaltracker.com/story/show/167200513

PR #1391, opened 2020-05-04
Raise error if component has 0 length array

### LINKED ISSUE #1360: Partial coloring can't handle 0 length arrays
### Summary of Issue

When attempting to generate a partial coloring for a component containing a length 0 array, the run died in _setup_partials.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Example

This happened in a complicated model, so no test case was provided, but it shouldn't be too hard to reproduce the problem by just creating a simple component containing a size 0 array.

DKilkenny:
### Summary

Raise error if component has 0 length array

### Related Issues

- Resolves #1360 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1259,9 +1259,9 @@ class Component(System):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
-                if abs2meta[abs_key[0]]['size'] == 0:
-                    msg = '{}: Is an array of size 0'
-                    raise ValueError(msg.format(self.msginfo))
+                if shape[0] == 0 or shape[1] == 0:
+                    msg = "{}: '{}' is an array of size 0"
+                    raise ValueError(msg.format(self.msginfo, abs_key[0]))
 
                 if val is None:
                     # we can only get here if rows is None  (we're not sparse list format)
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1259,6 +1259,10 @@ class Component(System):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
+                if abs2meta[abs_key[0]]['size'] == 0:
+                    msg = '{}: Is an array of size 0'
+                    raise ValueError(msg.format(self.msginfo))
+
                 if val is None:
                     # we can only get here if rows is None  (we're not sparse list format)
                     meta['value'] = np.zeros(shape)
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1261,7 +1261,10 @@ class Component(System):
 
                 if shape[0] == 0 or shape[1] == 0:
                     msg = "{}: '{}' is an array of size 0"
-                    raise ValueError(msg.format(self.msginfo, abs_key[0]))
+                    if shape[0] == 0:
+                        raise ValueError(msg.format(self.msginfo, abs_key[0]))
+                    elif shape[1] == 0:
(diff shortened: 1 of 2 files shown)


openmdao/core/component.py
```diff
@@ -1259,6 +1259,10 @@ def _declare_partials(self, of, wrt, dct):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
+                if abs2meta[abs_key[0]]['size'] == 0:
+                    msg = '{}: Is an array of size 0'
+                    raise ValueError(msg.format(self.msginfo))
+
```
Kenneth-T-Moore, line 1265: To clarify, are we allowed to have zero-length arrays as component inputs and outputs as long as we don't delcare partial derivatives for them explicitly?

openmdao/core/tests/test_coloring.py
```diff
@@ -319,6 +319,55 @@ def test_dynamic_fwd_simul_coloring_snopt_approx_fd(self):
...
+
+        p.driver.declare_coloring(show_summary=True, show_sparsity=True)
+
+        p.setup(mode='fwd')
+
+        with self.assertRaises(Exception) as context:
+            p.run_driver()
+        self.assertEqual(str(context.exception),
+                         "DynamicPartialsComp (arctan_yox): Is an array of size 0")
+
```
Kenneth-T-Moore, line 369: It would be helpful to include the variable name in the message string too.

openmdao/core/component.py
```diff
@@ -1259,6 +1259,10 @@ def _declare_partials(self, of, wrt, dct):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
+                if abs2meta[abs_key[0]]['size'] == 0:
+                    msg = '{}: Is an array of size 0'
```
naylor-b, line 1263: msg should include the name of the variable that is size 0  (abs_key[0] in this case). Also, is there a reason we're not checking if the variable corresponding to abs_key[1] is size 0 as well?

openmdao/core/component.py
DKilkenny: That's correct

openmdao/core/component.py
```diff
@@ -1259,6 +1259,10 @@ def _declare_partials(self, of, wrt, dct):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
+                if shape[0] == 0 or shape[1] == 0:
+                    msg = "{}: '{}' is an array of size 0"
+                    raise ValueError(msg.format(self.msginfo, abs_key[0]))
+
```
Kenneth-T-Moore, line 1265: This error message only prints out the variable name `abs_key[0]`, even if `abs_key[1]` is the one that is zero-shape.  You should fix the 2nd test so that it fails, and then fix this bug.

DKilkenny: Good catch, fixed

openmdao/core/component.py
```diff
@@ -1259,6 +1259,13 @@ def _declare_partials(self, of, wrt, dct):
                 meta['cols'] = cols
                 meta['shape'] = shape = (abs2meta[abs_key[0]]['size'], abs2meta[abs_key[1]]['size'])
 
+                if shape[0] == 0 or shape[1] == 0:
+                    msg = "{}: '{}' is an array of size 0"
+                    if shape[0] == 0:
+                        raise ValueError(msg.format(self.msginfo, abs_key[0]))
+                    elif shape[1] == 0:
```
Kenneth-T-Moore, line 1266: can just be 'else'

PR #1393, opened 2020-05-05
BoundsEnforceLS linesearch UserWarning Remove

### LINKED ISSUE #1370: Remove warning about BoundsEnforceLS linesearch is active but no bounds have been set.
### Summary of Issue

The BoundsEnforceLS code warns about linesearch being active but no bounds set. The warning is no longer needed.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

A bi-product of making the line search the default is that now we get warnings like 

 `UserWarning:BoundsEnforceLS in Circuit (circuit): linesearch is active but no bounds have been set.`

There isn't a performance hit because of this anymore so the warning can be removed. 

The warning is in LinesearchSolver._setup_solvers.

### Example

Run the `test_feature_circuit_with_recorder` test in test_sqlite_recorder.py to see the warning.

### Environment

All

DKilkenny:
### Summary

Removed UserWarning because it's no longer needed. There isn't a performance hit because of this anymore so the warning can be removed.

### Related Issues

- Resolves #1370

None

### New Dependencies

None

### DIFF
--- a/openmdao/recorders/tests/test_sqlite_recorder.py
+++ b/openmdao/recorders/tests/test_sqlite_recorder.py
@@ -11,6 +11,7 @@ from tempfile import mkdtemp
 
 import openmdao.api as om
 from openmdao.utils.general_utils import set_pyoptsparse_opt
+from openmdao.utils.assert_utils import assert_no_warning
 
 from openmdao.test_suite.components.ae_tests import AEComp
 from openmdao.test_suite.components.sellar import SellarDerivatives, SellarDerivativesGrouped, \
@@ -2588,7 +2589,12 @@ class TestFeatureSqliteRecorder(unittest.TestCase):
         prob['circuit.n1.V'] = 10.
         prob['circuit.n2.V'] = 1.
 
-        prob.run_driver()
+        msg = ("BoundsEnforceLS in Circuit (circuit): linesearch is active but no bounds have been "
+               "set.")
+
+        with assert_no_warning(UserWarning, msg):
+            prob.run_driver()
+
         prob.cleanup()
 
         # create the case reader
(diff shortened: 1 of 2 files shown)


openmdao/recorders/tests/test_sqlite_recorder.py
```diff
@@ -2588,7 +2589,12 @@ def setup(self):
         prob['circuit.n1.V'] = 10.
         prob['circuit.n2.V'] = 1.
 
-        prob.run_driver()
+        msg = ("BoundsEnforceLS in Circuit (circuit): linesearch is active but no bounds have been "
+               "set.")
+
+        with assert_no_warning(UserWarning, msg):
+            prob.run_driver()
```
swryan, line 2596: It seems odd to check for a very specific warning that doesn't exist anywhere in the code base...
Maybe we just want a version of this assert that checks for no warnings at all?

Kenneth-T-Moore: Wasn't there a problem where other warnings sometimes show up depending on changes we make in future stories (or sometimes temporarily due to updates in other packages like numpy) which would cause this to fail because it expects no warning?

Kenneth-T-Moore: You make a good point though, that maybe we should just remove the test.

swryan: Agree... seems like just removing the assert would be appropriate

naylor-b: Agree.

DKilkenny: Removed

PR #1400, opened 2020-05-07
Allow partials on distributed comps with zero-length inputs/outputs on some procs

### LINKED ISSUE #1350: declare_partials with empty row / col indices fails
### Summary of Issue

A user may wish to distribute an output value across processors such that one processor has all the values and all the other processors have no values. Alternatively, the output may be of size m < n_procs resulting in some processors with no values.

If a user declares rows=[] or cols=[] in the declare_partials method of a component, this edge case is not handled correctly and raises a value error.

Per communication with @JustinSGray the intent is for zero length outputs to be allowed in OpenMDAO so I think this is a bug?

### Issue Type

- [X ] Bug

### Description

A common design pattern in distributed components is:
`        self.declare_partials('y', 'x', rows=np.arange(0, sizes[rank]), cols=np.zeros((sizes[rank],)))
`
If sizes[rank] is zero, then a value error is raised [here](https://github.com/OpenMDAO/OpenMDAO/blob/a938f8ad5e20f61134398fef2663a6015af9d756/openmdao/core/component.py#L1272). 

### Example

See [gist](https://gist.github.com/bbrelje/402910db2b7f3fc7076763aadbda3c99)

### Environment

Python environment: Python 3.7

Packages: OpenMDAO 9931ddd4b246485d651930a91ffde3dcec1bb07c, numpy 1.16.4

bbrelje:
### Summary

This is a fix for issue #1350 and enables partial derivative computation on distributed components with zero-length outputs or inputs on some processors

### Related Issues

- Resolves #1350

- I removed some code that Bret had added in jacobians.py which didn't break any tests, but I'm not sure what the purpose of that logic was in the first place?

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_distrib_derivs.py
+++ b/openmdao/core/tests/test_distrib_derivs.py
@@ -588,6 +588,49 @@ class MPIFeatureTests(unittest.TestCase):
         assert_near_equal(J[('C2.outvec', 'indep.x')], 
                           np.eye(15)*np.append(2*np.ones(8), -3*np.ones(7)))
 
+@unittest.skipUnless(MPI and PETScVector, "MPI and PETSc are required.")
+class ZeroLengthInputsOutputs(unittest.TestCase):
+
+    N_PROCS = 4
+    # this test case targets situations when zero-length inputs
+    # or outputs are located on some processors
+    # issue 1350
+
+    def test_distribcomp_zerolengthinputsoutputs(self):
+        import numpy as np
+        import openmdao.api as om
+        from openmdao.test_suite.components.distributed_components import DistribCompDerivs, SummerDerivs
+        from openmdao.utils.assert_utils import assert_check_partials
+
+        size = 3  # set to one less than number of procs, leave zero inputs/outputs on proc 3
+
+        model = om.Group()
+        model.add_subsystem("indep", om.IndepVarComp('x', np.zeros(size)))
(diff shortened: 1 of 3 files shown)


openmdao/jacobians/jacobian.py
```diff
@@ -90,13 +90,6 @@ def _abs_key2shape(self, abs_key):
         system = self._system()
         abs2meta = system._var_allprocs_abs2meta
         of, wrt = abs_key
-        if system.comm.size > 1:
```
Kenneth-T-Moore, line 93: Interesting. Just tested this change on my branch with distributed constraints, and tests all pass so guess it was not needed.

openmdao/core/tests/test_distrib_derivs.py
```diff
@@ -588,6 +588,49 @@ def test_distribcomp_derivs_feature(self):
...
+@unittest.skipUnless(MPI and PETScVector, "MPI and PETSc are required.")
+class ZeroLengthInputsOutputs(unittest.TestCase):
+
+    N_PROCS = 4
+    # this test case targets situations when zero-length inputs
+    # or outputs are located on some processors
+    # issue 1350
+
+    def test_distribcomp_zerolengthinputsoutputs(self):
+        import numpy as np
```
Kenneth-T-Moore, line 600: Don't need to put imports here unless this test is going to show up in the docs.

PR #1409, opened 2020-05-08
Fixed N2 home button bug; improvements to search, forward, and back

### LINKED ISSUE #1361: Home button on N2 diagram does not work after doing a search
### Summary of Issue

Home button on N2 diagram does not work after doing a search.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When you do a search and then click the home button, you get a grey rectangle instead of the N2 diagram. There are lots of errors in the console log.

### Example

Go to the page http://openmdao.org/twodocs/versions/latest/features/model_visualization/n2_details.html and 

and in the N2 diagram embedded on the page, do a search on R2. Then click the home button. Nothing appears in the n2 diagram. Just a grey rectangle.

### Environment

All

tadkollar:
### Summary

The primary issue was that following a search, the behavior of the home button was unpredictable (prior to 3.0, the home button was disabled following a search). The new behavior is that when the home button is pressed, the diagram is reset to its original state.

There were also many cleanups in the behavior of the search bar. Previously, the search was triggered by hovering over the magnifying glass, but now a click is required to open it and a click outside the box will close it. With the former hover behavior, it was too difficult to stay in focus and the box would close unintentionally. There were also several visual changes to the search box.

The forward and back buttons will now restore the state of collapsed/hidden nodes as well as previous searches, rather than just previously zoomed elements.

### Related Issues

- Resolves #1361

None

### New Dependencies

None


tadkollar:
That's a good idea, but I'll add it as a separate issue.

PR #1412, opened 2020-05-08
Out of Order ParallelGroups Warning

### LINKED ISSUE #1334: Connections within parallel groups should raise warning on check
### Summary of Issue

Parallel groups currently will not warn the user if the output of one parallel component is used as input to another peer parallel component. 

### Issue Type

 [X ] Bug

### Description

OpenMDAO currently warns users when components are connected out of order in serial.
If the the output system of an input/output pair is executed second, then openmdao will return a warning if check=True. The reason for this is that users expect output variables to be freshly computed at the time that they are used by downstream components.

In parallel groups, similar behavior does not occur. If two components in a parrallel group are somehow connected to each other, then an effect similar to the out-of-order serial condition will occur, except it will be harder to detect because it will depend on the number of processors.

### Example

This [gist](https://gist.github.com/bbrelje/f1d8af3b3665f3c7fa04ade4ce084193) illustrates the problem. The system output value changes depending on whether 1 or more processes is used. And no warning is raised.

### Desired Behavior

Groups that may execute out of order in a parallel configuration should raise a warning when check is run. 

### Environment

Packages: OpenMDAO  commit 9931ddd4b246485d651930a91ffde3dcec1bb07c

DKilkenny:
### Summary

OpenMDAO now warns the user about out of order parallel components and recommends the use of a `NonlinearBlockJac` to converge the components in a parallel group.

### Related Issues

- Resolves #1334

None

### New Dependencies

None


openmdao/error_checking/check_config.py
```diff
@@ -165,6 +172,11 @@ def _get_used_before_calc_subs(group, input_srcs):
             oparts = src_abs.split('.')
             src_sys = oparts[glen]
             tgt_sys = iparts[glen]
+            if (src_sys in parallel_connections and tgt_sys in parallel_connections and
+                    (parallel_connections[src_sys] != "NL: NLBJ")):
+                simple_warning("Need to attach NonlinearBlockJac to '%s' when connecting "
```
bbrelje, line 177: Strictly speaking, couldn't it be another nonlinear solver? Would Newton not work in this case?

JustinSGray:
hrm... thats a good point. newton and broyden would both work in this case too. We need to allow them.

openmdao/error_checking/check_config.py
DKilkenny: Added Newton and Broyden

Kenneth-T-Moore:
So, i made a simple model with two parallel systems in series. (One follows the other with a connection between.)

https://gist.github.com/Kenneth-T-Moore/23ec1b571c314fdb42a9b87d1d0c9c19

I get a warning with this system

   WARNING: The following systems are executed out-of-order:
   System 'parallel_copy.comp1' executes out-of-order with respect to its source systems ['parallel.c1']

I don't think we should warn in this case because you definitely don't need a solver for this to work. I think the check should confine itself solely to systems that are contained in a single parallel group, not interconnections between separate parallel groups.

openmdao/error_checking/check_config.py
```diff
@@ -155,7 +155,14 @@ def _get_used_before_calc_subs(group, input_srcs):
...
         source Systems that execute after them.
     """
-    sub2i = {sub.name: i for i, sub in enumerate(group._subsystems_allprocs)}
+    sub2i = {}
+    parallel_connections = {}
+    for i, sub in enumerate(group._subsystems_allprocs):
+        if hasattr(sub, '_mpi_proc_allocator') and sub._mpi_proc_allocator.parallel:
+            parallel_connections.update({sub.name: sub.nonlinear_solver.SOLVER})
+
+        sub2i.update({sub.name: i})
```
naylor-b, line 164: `sub2i[sub.name] = i` is clearer and more efficient.

openmdao/error_checking/check_config.py
```diff
@@ -155,7 +155,14 @@ def _get_used_before_calc_subs(group, input_srcs):
         A dict mapping names of target Systems to a set of names of their
         source Systems that execute after them.
     """
-    sub2i = {sub.name: i for i, sub in enumerate(group._subsystems_allprocs)}
+    sub2i = {}
+    parallel_connections = {}
+    for i, sub in enumerate(group._subsystems_allprocs):
+        if hasattr(sub, '_mpi_proc_allocator') and sub._mpi_proc_allocator.parallel:
+            parallel_connections.update({sub.name: sub.nonlinear_solver.SOLVER})
```
naylor-b, line 162: It's not clear why this is called `parallel_connections` when it's a dict that maps a parallel group's name to its solver name.

openmdao/error_checking/tests/test_check_config.py
```diff
@@ -92,6 +93,105 @@ def test_dataflow_1_level(self):
...
+        with assert_warning(UserWarning, msg):
+            prob.run_model()
+
+        expected_warning = ("The following systems are executed out-of-order:\n"
+                            "   System 'parallel.c2' executes out-of-order with respect to its source systems ['parallel.c1']\n"
+                            "   System 'parallel_copy.comp2' executes out-of-order with respect to its source systems ['parallel_copy.comp1']\n")
+
+        testlogger.find_in('warning', expected_warning)
+
+    def test_serial_in_parallel(self):
```
Kenneth-T-Moore, line 139: If you add the `assert_no_warning` check here, this will be ready to go.

DKilkenny: Added

PR #1413, opened 2020-05-09
Fix redundant calls to matrix-free in parallel coloring algorithm clo…

### LINKED ISSUE #1405: jacvec products called twice per apply_linear when parallel coloring turned on
### Summary of Issue
Matrix-free components take a 2x performance hit when parallel coloring is turned on

### Issue Type

- [ X] Bug

### Description

The [parallel coloring](http://openmdao.org/twodocs/versions/latest/features/core_features/working_with_derivatives/parallel_derivs.html) feature example adds a delay to the apply_linear method and illustrates the theoretically correct speedup when coloring is turned on.

However, in reverse mode (at least, maybe fwd too) when the matrix-free API is being used, the compute_jacvec_product method gets called 2x per apply linear when coloring is on, and 1x per apply linear when coloring is off, independent of the number of colors or number of MPI processes.

### Example

See [gist](https://gist.github.com/bbrelje/bf0d003cb96310ab0682007625395541)

### Environment

Ubuntu 16.04, Python 3.7, numpy 1.16.4, Intel MPI

bbrelje:
### Summary

Redundant calls to the matrix-free API were being done when running under parallel deriv coloring (for fan-out parallel problems). This adds a fix and tests.

### Related Issues

- Resolves #1405 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/total_jac.py
+++ b/openmdao/core/total_jac.py
@@ -460,7 +460,7 @@ class _TotalJacInfo(object):
                 if matmat or parallel_deriv_color:
                     rhsname = name
 
-                    if parallel_deriv_color and self.debug_print:
+                    if parallel_deriv_color:
                         if parallel_deriv_color not in self.par_deriv:
                             self.par_deriv[parallel_deriv_color] = []
                         self.par_deriv[parallel_deriv_color].append(name)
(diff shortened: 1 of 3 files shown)


openmdao/core/total_jac.py
```diff
@@ -1336,7 +1336,13 @@ def compute_totals(self):
                             model._solve_linear(model._lin_vec_names, self.mode, rel_systems)
                             self._save_linear_solution(vec_names, cache_key, self.mode)
                         else:
-                            model._solve_linear(model._lin_vec_names, mode, rel_systems)
+                            if par_deriv and key in par_deriv:
+                                # parallel colored derivatives only need to solve
+                                # the vectors relevant to this color, not all of them
+                                vecnames_par_deriv = par_deriv[key].copy()
```
naylor-b, line 1342: Do we need a copy here?

PR #1416, opened 2020-05-12
Fix for ParallelComps over allocating processors

### LINKED ISSUE #1339: list_outputs and list_inputs fail when overallocating processors to parallel components
### Summary of Issue

List inputs and list outputs will fail when the second (or higher) component in a parallel group is allocated more than one process.

### Issue Type

- [X ] Bug

### Description

A model with a parallel group breaks the list_outputs and list_inputs methods when the second (or higher) component gets allocated more than one process.
For example, a two-component group with 4 or more processes, or a three-component group with 5 or more processes.

The problem is most likely with the way the metadata dict is gathered on processes that don't own their components.

### Example

Run this [gist](https://gist.github.com/bbrelje/ebeb687d49b19d0131fe52434edb8e4c) with 3 or fewer procs and it works fine - breaks with the following traceback for 4 or more.

```
Traceback (most recent call last):
  File "parallel_toy.py", line 60, in <module>
    prob.model.list_outputs()
  File "/home/ben/Documents/OpenMDAO/openmdao/core/system.py", line 3548, in list_outputs
    all_procs, out_stream)
  File "/home/ben/Documents/OpenMDAO/openmdao/core/system.py", line 3633, in _write_table
    is_distributed = meta[name]['distributed']
KeyError: 'parallel.c2.y'
```

### Environment

Packages: OpenMDAO commit 9931ddd4b246485d651930a91ffde3dcec1bb07c

DKilkenny:
### Summary

Added an MPI gather to pull in the meta dicts from all procs

### Related Issues

- Resolves #1339

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -3493,11 +3493,10 @@ class System(object):
         else:
             raise RuntimeError(self.msginfo +
                                ': You have excluded both Explicit and Implicit components.')
-
+    import wingdbstub
     def _write_table(self, var_type, var_data, hierarchical, print_arrays, all_procs, out_stream):
         """
         Write table of variable names, values, residuals, and metadata to out_stream.
-
         Parameters
         ----------
         var_type : 'input', 'explicit' or 'implicit'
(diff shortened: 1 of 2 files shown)


openmdao/core/tests/test_expl_comp.py
```diff
@@ -166,6 +166,118 @@ def test_compute_and_list(self):
         with self.assertRaisesRegex(RuntimeError, msg):
             prob.model.list_outputs(explicit=False, implicit=False)
 
+    def test_list_inputs_outputs_with_parallel_comps(self):
+        class TestComp(om.ExplicitComponent):
+
+            def initialize(self):
+                self.options['distributed'] = False
+
```
Kenneth-T-Moore, line 174: To reproduce (and test) reported issue,, you need an MPI test with 3 procs.

DKilkenny: I actually need 4 to make reproduce the the issue. Fixed

openmdao/core/tests/test_expl_comp.py
```diff
@@ -980,6 +980,121 @@ def compute_jacvec_product(self, inputs, d_inputs, d_outputs, mode):
...
+            def setup(self):
+                self.add_input('x', shape=1)
+                self.add_output('y', shape=1)
+                self.declare_partials('y', 'x')
+
+            def compute(self, inputs, outputs):
+                print(self.comm.rank)
```
swryan, line 1016: print left over from debugging?
... but isn't this class identical to TestComp anyway?

DKilkenny: Yep, fixed

openmdao/core/system.py
```diff
@@ -3559,20 +3559,18 @@ def _write_table(self, var_type, var_data, hierarchical, print_arrays, all_procs
                     if name not in var_dict:     # If not in the merged dict, add it
                         var_dict[name] = proc_vars[name]
                     else:
-                        # In there already, only need to deal with it if it is a distributed array
-                        # Checking to see if distributed depends on if it is an input or output
-                        if var_type == 'input':
-                            is_distributed = meta[name]['src_indices'] is not None
-                        else:
+                        try:
```
naylor-b, line 3562: This try block is unnecessary.  Just use `allprocs_meta[name]['distributed'] `for all cases.  The 'distributed' entry exists in both `meta` and `allprocs_meta`, but only local variables will be found in `meta`.  `allprocs_meta` will contain all variables from all procs.

PR #1426, opened 2020-05-19
Add support for response indices on distributed variables.

### LINKED ISSUE #1406: Support `indices` for distributed driver responses.
### Summary of Issue

Support `indices` for distributed driver responses.

### Issue Type

- [x] Bug
- [] Enhancement  

### Description

Bret: "I think you'll need to treat indices as global in the same way as we treat src_indices, which means you'll have to check what part of the indices refer to a given proc and only gather those from each proc."

### Environment

<For bugs, otherwise N/A>

Kenneth-T-Moore:
### Summary

Response indices are now supported on distributed variables.
Specifying an objective that points to a location in a distributed variable is also now supported.
The check_totals output now displays when a variable is distributed.

### Related Issues

- Resolves #1406

None

### New Dependencies

None


openmdao/core/driver.py
```diff
@@ -310,16 +314,66 @@ def _setup_driver(self, problem):
...
+                                if rank == irank:
+                                    new_index = index - offsets[irank] + dist_sizes[irank]
+                                    local_indices.append(new_index)
+
+                            indices = local_indices
+                            dist_sizes = true_sizes
+
+                        dist_con_dict[vname] = (indices, dist_sizes)
+
+                    if vname in self._objs:
```
naylor-b, line 351: Seems like this whole block could go away since the code is identical to the block above, and you could just check inside of the `if vname in self._responses` block to see if vname is in self._objs and if so set the dist_obj_dict entry there (since self._responses contains both the constraints and the objectives).

openmdao/core/driver.py
```diff
@@ -310,16 +314,66 @@ def _setup_driver(self, problem):
...
+                                irank = np.argwhere(offsets > index)[0][0]
+                                true_sizes[irank] += 1
+                                if rank == irank:
+                                    new_index = index - offsets[irank] + dist_sizes[irank]
+                                    local_indices.append(new_index)
+
+                            indices = local_indices
+                            dist_sizes = true_sizes
+
+                        dist_con_dict[vname] = (indices, dist_sizes)
```
naylor-b, line 349: Should there be a check around this to make sure this is actually an constraint and not an objective?

openmdao/core/driver.py
Kenneth-T-Moore: I was thinking we allowed a con and obj to be declared on the same var, but apparently we don't. I should just combine these into one dict for responses then.

naylor-b: I thought we could to that too.

openmdao/core/driver.py
Kenneth-T-Moore: Refactored so that we only have responses.

PR #1430, opened 2020-05-20
Fixed collapsing the zoomed element's parent, as well as other collapse/search bugs

### LINKED ISSUE #1429: N2: Collapsing the parent of the currently focused group disables the parent button
### Summary of Issue

Collapsing the parent of the currently focused group disables the parent button, and as a result,  users cannot navigate back up the hierarchy by clicking the parent button anymore.
### Issue Type

- [x] Bug
- [ ] Enhancement  
      POEM #N/A

### Description

In the N2 diagrams, users can focus on groups with a left click, and collapse and expand groups with a right click. When focused to a group, when the users right-click to the parent (the area on the left), the parent is collapsed and this leaves an empty space, and the user cannot click on the parent to navigate back up.

[mphys_aero.html.zip](https://github.com/OpenMDAO/OpenMDAO/files/4654217/mphys_aero.html.zip)

This effect only happens when focused on a group that is at least 2 levels deep, i.e. when the parent is the model itself, the top level model does not collapse; however, when the parent is a group that is under the top level model, the issue is there.

### Example

Open the attached N2. Focus on the group `s0` (which is under the `mp_group` group). Then, right click to the parent. 

### Environment

Operating System: N2 is opened with safari on macos 10.15.4. N2 is created on Ubuntu 16 

Python version: Python 3.6.9 

Packages: OpenMDAO version: 3.1.1-dev (commit 280b581c17a31aca6c3c10c9f14c74b0f688fa7c)

tadkollar:
### Summary

If a group/component was zoomed and the user right-clicked on its parent, the parent would be collapsed, preventing navigation it. This fixes that behavior.

Also, uncollapse wasn't working, and there were still some issues with hitting the home button after a search that have been fixed.

Fixed the test code for performing a search and added tests that use it and check element after the search and then again after pressing home.

### Related Issues

- Resolves #1429

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/src/N2UserInterface.js
+++ b/openmdao/visualization/n2_viewer/src/N2UserInterface.js
@@ -413,7 +413,6 @@ class N2UserInterface {
      * @param {N2TreeNode} node The node that was right-clicked.
      */
     rightClick(node) {
-        console.log("rightClick() called")
         testThis(this, 'N2UserInterface', 'rightClick');
 
         d3.event.preventDefault();
--- a/openmdao/visualization/n2_viewer/src/N2UserInterface.js
+++ b/openmdao/visualization/n2_viewer/src/N2UserInterface.js
@@ -373,19 +373,25 @@ class N2UserInterface {
     }
 
     /**
+     * Make sure the clicked node is deeper than the zoomed node, that
+     * it's not the root node, and that it actually has children.
+     * @param {N2TreeNode} node The right-clicked node to check.
+     */
+    isCollapsible(node) {
+        return (node.depth > this.n2Diag.zoomedElement.depth &&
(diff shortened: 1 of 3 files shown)


openmdao/visualization/n2_viewer/src/N2UserInterface.js
```diff
@@ -403,20 +408,30 @@ class N2UserInterface {
...
     }
 
-    /* When a node is right-clicked, collapse it. */
-    rightClick(node1, node2) {
+    /**
+     * When a node is right-clicked, collapse it if it's allowed.
+     * @param {N2TreeNode} node The node that was right-clicked.
+     */
+    rightClick(node) {
+        console.log("rightClick() called")
```
swryan, line 416: temporary debugging?

PR #1438, opened 2020-05-27
Precursor to the PR that adds AutoIndepVarComps to eliminate unconnected inputs.

naylor-b:
Adds warnings for connected inputs without a source that have different units and/or values, and adds the Group.add_input method, which is mostly non-functional until AutoIndepVarComps are added.

### DIFF
--- a/openmdao/utils/units.py
+++ b/openmdao/utils/units.py
@@ -1055,6 +1055,20 @@ def convert_units(val, old_units, new_units=None):
 
 
 def _has_val_mismatch(units1, val1, units2, val2):
+    """
+    Return True if values differ after unit conversion or if values differ when units are None.
+
+    Parameters
+    ----------
+    units1 : str or None
+        Units for first value.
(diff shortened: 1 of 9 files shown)


openmdao/utils/units.py
```diff
@@ -1052,6 +1054,31 @@ def convert_units(val, old_units, new_units=None):
...
+        Second value.
+    """
+    if units1 != units2:
+        if units1 is None or units2 is None:
+            return True
+
+        # convert units
+        val1 = convert_units(val1, units1, new_units=units2)
+
+    return np.linalg.norm(val1 - val2) > 1e-10
```
Kenneth-T-Moore, line 1079: Wonder if this comparison should be a relative difference instead of absolute -- for cases where we have 2 very large numbers (like maybe modulus of elasticity.)

naylor-b: done

PR #1441, opened 2020-05-28
Several N2 appearance changes

### LINKED ISSUE #1362: Coloring and highlighting of items in the N2 hierarchy can be confusing
### Summary of Issue

To highlight the items in the hierarchy when hovering over connections, some method is being used that leads to some strange and possibly confusing displays in the N2.

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

The highlight uses some coloring of the text in the hierarchy cells that blends with the colors of the items in those cells that leads to strange and confusing coloring of the cells.

### Example

One of the best ways to see this is to go to the page http://openmdao.org/twodocs/versions/latest/features/model_visualization/n2_details.html and in the live N2 diagram on that page, left click on the n2 component. Then hover over the connections. Here is an example of what it looks like.

image

### Environment

All

### LINKED ISSUE #1410: N2 info panel should be pinnable
### Summary of Issue

With the current hover behavior of the N2 info panel, it's not possible to interact with the text at all, for example to select and copy it.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

Change the behavior of the info panel mode:
- Left-clicking while in info panel mode will no longer zoom the current node.
- If a hovered node is left-clicked, the info panel becomes pinned and doesn't change when other nodes are hovered.
- An alternate way of pinning with a keyboard command (e.g. space) should be implemented for situations when it's difficult to click without moving the pointer. The alternative method should be displayed on the bottom of the info panel (e.g. "Press Space to Pin").
- When the info panel is pinned, a close button should appear on its upper left. Clicking will unpin the panel. A tooltip should be added to the close button.

tadkollar:
### Summary

There are several N2 updates:
- The issue with the weird variable name highlighting is addressed by drawing a vertical line next to the variable name with the same color as the node, similar to the method used in Excel.
- The info panel is now pinnable. When using info-panel mode and a node is left-clicked, the panel is pinned and won't change when the mouse moves off the node. To unpin the panel, either left-click on another node, click the pushpin on the panel, or exit info-panel mode.
- Connection nodes in the matrix are now represented by a gray, curved arrow instead of a black square. The arrow makes it obvious that the connections and variables are different entities, and the gray color doesn't contrast as harshly with the rest of the diagram colors.
- Connection arrows now curve through the connection nodes instead of forming a right angle with a dot at the vertex.

- Resolves #1362 
- Resolves #1410

n2_new_matrix

n2_info_panel

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/index.html
+++ b/openmdao/visualization/n2_viewer/index.html
@@ -122,8 +122,9 @@
                     <g id='n2outer'>
                         <g id='n2inner'>
                             <rect id='backgroundRect' class='background'></rect>
-                            <g id='n2elements' clip-path='url(#n2MatrixClip)'></g>
+                            <g id='n2background' clip-path='url(#n2MatrixClip)'></g>
                             <g id='n2gridlines' clip-path='url(#n2MatrixClip)'></g>
+                            <g id='n2elements' clip-path='url(#n2MatrixClip)'></g>
                             <g id='n2componentBoxes' clip-path='url(#n2MatrixClip)'></g>
                             <g id='n2dots' clip-path='url(#n2MatrixClip)'></g>
                             <g id='n2arrows'></g>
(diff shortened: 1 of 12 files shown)


tadkollar:
image

PR #1448, opened 2020-06-02
Fix for issue #1447

### LINKED ISSUE #1447: Bug during setup.
### Summary of Issue

During (second) setup, a user got `TypeError: unsupported operand type(s) for *: 'Undefined' and 'Undefined'`. 

### Issue Type

- [x ] Bug
- [ ] Enhancement

naylor-b:
### Summary

User got the error message `TypeError: unsupported operand type(s) for *: 'Undefined' and 'Undefined'` after calling setup on a problem where the top level system was a group that was already setup earlier in a different problem.  The problem was due to the fact that the internal vector dictionaries were not being reinitialized early enough.

### Related Issues

- Resolves #1447

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/meta_model_unstructured_comp.py
+++ b/openmdao/components/meta_model_unstructured_comp.py
@@ -70,7 +70,7 @@ class MetaModelUnStructuredComp(ExplicitComponent):
         self._static_surrogate_output_names = []
         self._static_input_size = 0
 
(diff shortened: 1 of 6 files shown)


swryan:
maybe add a test that calls "setup on a problem where the top level system was a group that was already setup earlier in a different problem"?

naylor-b:
Yeah, hold off on the merge and I'll add that test.

PR #1452, opened 2020-06-05
Add guess_nonlinear to NonlinearBlockGS.

### LINKED ISSUE #1365: NonlinearBlockGS should call guess_nonlinear
### Summary of Issue

NonlinearBlockGS should call guess_nonlinear at the start of its solve.

### Issue Type

- [x] Bug
- [ ] Enhancement 

### Description

For iterative solutions, an explicit output (lets call it `t` in this case) is often fed back to a previous component in the hierarchy.  WIth a NonlinearBlockGS solver attached, this sequence is repeated until convergence is achieved.

Often times, iterative algorithms have a suggested starting value for the feedback variable `t`.  The canonical OpenMDAO way to set a guess value of a variable based on inputs is to use `guess_nonlinear` per the documentation here: http://openmdao.org/twodocs/versions/latest/features/core_features/grouping_components/guess_method.html

However, the NLBGS solver doesn't call `guess_nonlinear`, so there's no way to initialize `t` to the appropriate value outside of `set_val`, which doesn't account for changes in the inputs during the run of the model.

There is some question as to whether `guess_nonlinear` should be called by default, or whether this should be an option on the solver.

### Example

Pending...

### Environment

N/A

Kenneth-T-Moore:
### Summary

Add support for guess_nonlinear to the NonlinearBlockGS solver.

### Related Issues

- Resolves #1365

None

### New Dependencies

None


openmdao/solvers/nonlinear/nonlinear_block_gs.py
```diff
@@ -101,6 +101,9 @@ def _iter_initialize(self):
         if system.under_complex_step and self.options['cs_reconverge']:
             system._outputs._data += np.linalg.norm(system._outputs._data) * 1e-10
 
+        # Execute guess_nonlinear if specified.
+        system._guess_nonlinear()
```
naylor-b, line 105: Should we check system._has_guess here before calling _guess_nonlinear?  I noticed that ImplicitComponent does some stuff in its _guess_nonlinear that isn't needed if no guesses are provided.  Broyden does the same thing.  Pretty minor, but still...

Kenneth-T-Moore: Well, we don't do it in Newton. The only place we check it is in Group._guess_nonlinear.  I have added it into ImplicitComponent._guess_nonlinear to be consistent.

Kenneth-T-Moore: Note though, the only time where this will come into play is the case where a nonlinear solver is plugged into a component.  When the solver is in a group, we never recurse down to the component level if there isn't a guess.

PR #1460, opened 2020-06-11
Adding apply_nonlinear to _guess_nonlinear

### LINKED ISSUE #1397: Update docs for guess_nonlinear
### Summary of Issue

All of the guess_nonlinear docs show a direct setting of the guess, with not checking of values or residuals. This is pretty inefficient, and probably not how we should show things. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM N/A

### Description

The original intent of this Issue was to update the docs to encourage users to check the magnitude of their residuals before just changing the implicit variable value in the guess function.  It turns out that the updated residual wasn't being calculated before guess_nonlinear, so this didn't provide useful data.

Now the decided-upon path is to call _apply_nonlinear from _guess_nonlinear if the user has overridden `guess_nonlinear` for their system.  Once that's done, then the following documents should be updated.

http://openmdao.org/twodocs/versions/3.1.0/features/core_features/defining_components/implicitcomp.html?highlight=guess_nonlinear

http://openmdao.org/twodocs/versions/3.1.0/features/building_blocks/components/balance_comp.html?highlight=guess_nonlinear

http://openmdao.org/twodocs/versions/3.1.0/examples/keplers_equation/keplers_equation.html?highlight=guess_nonlinear

http://openmdao.org/twodocs/versions/3.1.0/features/core_features/grouping_components/guess_method.html?highlight=guess_nonlinear

DKilkenny:
### Summary

Added `_apply_nonlinear` to `_guess_nonlinear` in `Group` and `ImplicitComponent`.

### Related Issues

- Resolves #1397 

None

### New Dependencies

None


openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -254,12 +254,60 @@ def linearize(self, inputs, outputs, jac):
...
+        y = outputs['y']
+        z = outputs['z']
+
+        residuals['y'] = y - x - 2.0*z
+        residuals['z'] = x*z + z - 4.0
+
+        self.counter += 1
+        if self.counter > 20 and self.counter < 25:
+            raise om.AnalysisError('catch me')
+
```
Kenneth-T-Moore, line 286: Do you get any failures if you change the lower and upper checks here in the original `ImplCompTwoStatesAE` class? I think it is only used in 2 tests, and the other test deals with the iprint output.

If you do, then it's okay, but you can simplify this by just deriving a new class from `ImplCompTwoStatesGuess` with a replacement of the apply_nonlinear method.  A lot less code and one less unused class.

openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -320,7 +368,7 @@ def test_read_only_bug(self):
         top.model.add_subsystem('px', om.IndepVarComp('x', 7.0))
```
Kenneth-T-Moore, line 369: I think we are missing a test where you query the residuals in a group's guess_nonlinear to determine if you should provide a new guess.

openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -254,12 +254,60 @@ def linearize(self, inputs, outputs, jac):
         jac[('z', 'z')] = -inputs['x'] + 1.0
         jac[('z', 'x')] = -outputs['z']
 
+class ImplCompTwoStatesAEBug(om.ImplicitComponent):
```
Kenneth-T-Moore, line 257: If you just set new counter limits in `ImplCompTwoStatesAE`, do you get a test failure? If you don't, then you don't need the extra class.  If you do, then you can still make this simpler by inheriting from `ImplCompTwoStatesGuess` and just replacing the apply_nonlinear function with the new one that has new limits.  (You could also do it in one class if you let the upper and lower check numbers be set as an attribute.

openmdao/core/tests/test_impl_comp.py
```diff
@@ -500,6 +500,120 @@ def guess_nonlinear(self, inputs, outputs, resids):
...
+
+                partials['x', 'a'] = x ** 2
+                partials['x', 'b'] = x
+                partials['x', 'c'] = 1.0
+                partials['x', 'x'] = 2 * a * x + b
+
+            def guess_nonlinear(self, inputs, outputs, resids):
+                # Default initial state of zero for x takes us to x=1 solution.
+                # Here we set it to a value that will take us to the x=3 solution.
+                outputs['x'] = 5.0
```
Kenneth-T-Moore, line 540: The story was to make sure we had valid residuals when we reached this point inside of `guess_nonlinear`.  You should add an assert in each of these tests right here in the guess_nonlinear function that asserts that the residual 'x' is non-zero.

swryan:
The original issue had to do with examples in the docs.. 

I think this change makes it possible to do what was requested (per the comments on the issue), but does not actually update the docs per the issue summary?

swryan:
I think we just need to remove the "Resolves #1397" from this PR, since it doesn't actuallyu update the docs...  maybe replace with another issue that describes the code change...

bbrelje:
The description says that apply_nonlinear was added to guess_nonlinear for Group and ImplicitComponent but it looks like it's only happening in ImplicitComponent. Is that per design?

PR #1467, opened 2020-06-16
Edited error message for incompatible src_indices

### LINKED ISSUE #1340: src_indices notation allows non-local transfers to be defined on local groups
### Summary of Issue

When running in parallel, a non distributed component is allowed to specify src_indices in excess of the size of the incoming vector. For a distributed component, this is supposed to specify a transfer from another processor. For a non-distributed component, this undefined behavior can create erroneous results.

### Issue Type

- [ X] Bug

### Description

A group with all non-distributed components and no parallel groups will use the default_vector class for variable transfers. However, when run in parallel, the non-distributed input variables are not prevented from specifying src_indices in excess of the size of the incoming variable connection. 

This results in the default value of the variable being read for all time even if the variable value is updated on the other processor rank (since there is no Petsctransfer to overwrite that portion of the local vector with the correct value)

### Example

See [gist](https://gist.github.com/bbrelje/01ed972b8ac0b4870a9983dc068c4fa8)

Note that if SRC_INDICES is set to 0 the value on each proc is returned (default)
Alternatively, if distributed=True, src_indices works as intended for distributed systems and returns the result on proc=SRC_INDICES for all procs.

### Proposed Fix

Do not allow src_indices specification out of bounds of the variable input shape, unless the component is distributed and the PETSc transfer exists for the owning group. Raise an error on setup?

### Environment

Packages: OpenMDAO 9931ddd4b246485d651930a91ffde3dcec1bb07c

DKilkenny:
### Summary

Error is already raised when src_indices does not match input vector when running in parallel on a non distributed component.

### Related Issues

- Resolves #1340

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1364,8 +1364,8 @@ class Group(System):
                                         msg = f"{self.msginfo}: The source indices " + \
                                               f"do not specify a valid index for the " + \
                                               f"connection '{abs_out}' to '{abs_in}'. " + \
-                                              f"Index '{i}' is out of range for source " + \
-                                              f"dimension of size {d_size}."
+                                              f"src_indices index: {i} must out of range for " + \
+                                              f"the input vector: {d_size}."
                                         if self._raise_connection_errors:
                                             raise ValueError(msg)
                                         else:
--- a/openmdao/core/tests/test_connections.py
+++ b/openmdao/core/tests/test_connections.py
@@ -339,9 +339,9 @@ class TestConnectionsIndices(unittest.TestCase):
         # the valid range for the source
         self.prob.model.connect('idvp.arrout', 'arraycomp.inp1', src_indices=[100000])
 
-        expected = "Group (<model>): The source indices do not specify a valid index " + \
-                   "for the connection 'idvp.arrout' to 'arraycomp.inp1'. " + \
-                   "Index '100000' is out of range for source dimension of size 5."
+        expected = "Group (<model>): The source indices do not specify a valid index for the connection "+\
+                   "'idvp.arrout' to 'arraycomp.inp1'. src_indices index: 100000 must out of range for the " +\
+                   "input vector: 5."
 
         try:
             self.prob.setup()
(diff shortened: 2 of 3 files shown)


openmdao/core/group.py
```diff
@@ -1364,8 +1364,8 @@ def _setup_connections(self, recurse=True):
                                         msg = f"{self.msginfo}: The source indices " + \
                                               f"do not specify a valid index for the " + \
                                               f"connection '{abs_out}' to '{abs_in}'. " + \
-                                              f"Index '{i}' is out of range for source " + \
-                                              f"dimension of size {d_size}."
+                                              f"src_indices index: {i} must out of range for " + \
```
naylor-b, line 1367: typo:  'must'.  Did you intend 'must be'?  I think the new form of the error message is confusing.  It's confusing at the end where it says that the index {i} is out of range for the input vector.  `src_indices` reference into the source vector, not the input vector, and `d_size` is the size of the current source dimension.  I don't think the original issue mentioned any problems with the form of the error message, did it?

DKilkenny: Yeah that was a typo. 
I'm open to leaving the error message alone. To me the original error message didn't seem very clear. The original story was asking to raise an error if `src_indices` was out of bounds of the input vector, but this error already catches that.

naylor-b: The `src_indices` reference into the source vector, not the input vector, so, for example, you could have an input vector of size 1 with a `src_indices` of [99], and as long as the source vector is of size 100 or greater, it's fine.  It is true though that the *number* of indices contained in `src_indices` must match the input size.  I added Ben as a requested reviewer for this one because I'm not sure we're addressing his actual issue.  I think he was talking about a situation where DefaultVectors are used under MPI.

Kenneth-T-Moore: @DKilkenny : I agree with Bret that the old messages are more clear.

Per Bret's suggestion, you could try running the test you showed me under MPI (without adding any distributed comps or parallelgroups to the model) and see if a correct error is raised.

DKilkenny: I confirmed that the error is still raised under those conditions

robfalck:
This PR will be dropped but a test aded with @bbrelje's gist above.

PR #1478, opened 2020-06-23
Made N2 --use_declare_partial_info the default and corrected behavior

### LINKED ISSUE #1356: use_declare_partial_info option does not work with the N2
### Summary of Issue

The use_declare_partial_info option on the creating of the N2 diagram is completely broken.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The use_declare_partial_info option is not  behaving correctly. It is removing information from the N2 diagram. 

Also, it was recommended that use_declare_partial_info=True be the default behavior. 

A GUI test should be added to make sure this feature works in the future. 

### Example

Consider this model

`./openmdao/test_suite/scripts/circuit_analysis.py`

without the option, the N2 looks like this.

image

with the option it looks like this

image

### Environment

All

tadkollar:
### Summary

The bug originally described is that connections were hidden if they weren't in the declare_partial list when `--use_declare_partial_info` was specified. It _should_ have only affected the drawing internal to each component.

That behavior is fixed, and now by default, only declared partials are shown internal to components if _anything_ is in declare_partial_list.

The `--use_declare_partial_info` switch is now ignored from the command line and the option is ignored in the `n2()` call.

Declared partials are now drawn as black boxes, and the legend was updated to show the representation. Tests were updated to reflect the new node counts.

### Related Issues

- Resolves #1356

image

For reference, here's the Python that assembles `declare_partials_list` (unchanged in this update):

```
def _get_declare_partials(system):
    declare_partials_list = []

    def recurse_get_partials(system, dpl):
        if isinstance(system, Component):
            subjacs = system._subjacs_info
            for abs_key, meta in subjacs.items():
                dpl.append("{} > {}".format(abs_key[0], abs_key[1]))
        elif isinstance(system, Group):
            for s in system._subsystems_myproc:
                recurse_get_partials(s, dpl)
        return

    recurse_get_partials(system, declare_partials_list)
    return declare_partials_list
```


tadkollar:
Did AppVeyor fail because of the new version of numpy?

swryan:
new version of SciPy (1.5.0), yup..  per #1480

PR #1484, opened 2020-06-26
Added test from Ben Brelje

### LINKED ISSUE #1340: src_indices notation allows non-local transfers to be defined on local groups

DKilkenny:
### Summary

Added test from Ben Brelje to capture error message when running in parallel with non distributed components

### Related Issues

- Resolves #1340

None

### New Dependencies

None


openmdao/core/tests/test_connections.py
```diff
@@ -613,5 +613,66 @@ def compute_partials(self, inputs, J):
...
+        if MPI.COMM_WORLD.rank == 0:
+            setval = np.array([2.0,3.0])
+        else:
+            setval = np.array([10.0,20.0])
+
+        # no parallel or distributed comps, so default_vector is used (local xfer only)
+        model.add_subsystem('p1', om.IndepVarComp('x', setval))
+        model.add_subsystem('c3', TestComp())
+        if FORCE_PETSCTRANSFER:
+            model.add_subsystem('c4', TestCompDist())
```
swryan, line 666: I think the intent was to add a test that runs 
```under MPI (without adding any distributed comps or parallelgroups to the model) and see if a correct error is raised.```
(quoting @Kenneth-T-Moore from #1467 discussion)
So the `FORCE_PETSCTRANSFER=False` version of @bbrelje 's gist?

swryan: hmm I see the `COMP_DISTRIBUTED` flag does this.. so not clear on the function of the FORCE_PETSCTRANSFER flag

DKilkenny: Yeah I could clean that up a bit. I'll fix it

bbrelje:
@swryan @DKilkenny This looks correct to me, sorry for the delay in the review.

PR #1486, opened 2020-06-26
All case recorder files now always contain system options information about all systems in the model

### LINKED ISSUE #1482: All case recorder files should contain system options information about all systems in the model
### Summary of Issue

System options is not something that is tied to a specific sub-set of cases. It's just the information about the model that we need to know to ensure that any data saved can be put into the right context.

### Issue Type

- [ ] Bug
- [X] Enhancement  
      POEM #011

### Description

For all case recorders, record the system options for all systems in the model.

### Example

N/A

### Environment

N/A

hschilling:
### Summary

Currently, what systems have their system options recorded in the case recording file depends on what the case recorder is attached to. All case recorder files should always record all system options for all systems in the model so the complete context is available.

### Related Issues

- Resolves #1482 

None

### New Dependencies

None


openmdao/recorders/sqlite_recorder.py
```diff
@@ -650,10 +650,6 @@ def record_metadata_system(self, recording_requester):
             pickled_metadata = sqlite3.Binary(pickled_metadata)
 
             with self.connection as c:
-                # Because we can have a recorder attached to multiple Systems,
-                #   and because we are now recording System metadata recursively,
-                #   we can store System metadata multiple times. Need to ignore when that happens
-                #   so we don't get database errors. So use OR IGNORE
                 c.execute("INSERT OR IGNORE INTO system_metadata"
```
swryan, line 653: should we remove the IGNORE now, since we shouldn't be trying to record a system multiple times?

robfalck:
@hschilling I'm moving the base of this PR to auto-ivc, and it looks like theres a conflict to fix now.  Could you please address it?

PR #1491, opened 2020-07-02
Added om.slicer support for all src_indices/indices arguments 

### LINKED ISSUE #1471: Allow slice objects where indices/src_indices are specified.
### Summary of Issue

This issue would allow slice objects to be provided wherever indices or src_indices are specified in OpenMDAO. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #18  https://github.com/OpenMDAO/POEMs/pull/36

### Description

set_val and get_val support a seldom-used and seldom-documented slicer object for setting/getting slices of variables.

```
prob['A'][:, 0] = 5
```
```
prob.set_val('A', 5.0, indices=om.slicer[:, 0])
```

There are a lot of other places in OpenMDAO that make use of indices/src_indices where it would be nice to utilize slicer objects, notably in connect and promote statements.

For the implementation, it might be easiest resolve the slice objects as indices at some point during setup.

### Example

Connect the left column of a 3x3 matrix to a vector input:

```
self.connect('M', 'x', src_indices=om.slicer[:, 0])
```

### Environment

N/A

DKilkenny:
### Summary

Added support for `om.slicer` as a way to pass in slices to src_indices/indices arguments in `add_design_var`, `add_response`, `add_constraint`, `promotes`, and `connect`

### Related Issues

- Resolves #1471 

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -1775,6 +1775,11 @@ def update_src_indices(name, key):
                 abs_name = self._var_allprocs_prom2abs_list['input'][name][0]
                 meta = self._var_abs2meta[abs_name]
 
+                if isinstance(src_indices, tuple):
```
naylor-b, line 1778: I can't see where 'indices' is used anywhere in this function.  I think this whole conditional block may be unnecessary.

openmdao/core/system.py
```diff
@@ -1792,7 +1797,13 @@ def update_src_indices(name, key):
                                            (self.msginfo, name, str(flat_src_indices),
                                             str(meta['flat_src_indices'])))
 
-                meta['src_indices'] = np.asarray(src_indices, dtype=INT_DTYPE)
+                if isinstance(src_indices, np.ndarray):
```
naylor-b, line 1800: could add `and src_indices.dtype==object` to the 'if' check since that's the only situation where an ndarray could contain a slice.

openmdao/core/system.py
```diff
@@ -1792,7 +1797,13 @@ def update_src_indices(name, key):
...
+                    contains_slice = [True if isinstance(i, slice) else False for i in src_indices]
```
naylor-b, line 1801: Since you do this check in multiple places, I'd recommend creating a function to do it.  Also, `contains_slice = any(isinstance(i, slice) for i in src_indices)` is cleaner.  It also gives you a simple bool instead of a list of bools that you have to iterate through again later.

openmdao/core/group.py
```diff
@@ -1636,7 +1643,12 @@ def connect(self, src_name, tgt_name, src_indices=None, flat_src_indices=False):
             raise TypeError("%s: src_indices must be an index array, did you mean"
                             " connect('%s', %s)?" % (self.msginfo, src_name, tgt_name))
 
-        if isinstance(src_indices, Iterable):
+        if isinstance(src_indices, tuple):
+            contains_slice = [True if isinstance(i, slice) else False for i in src_indices]
+            if True not in contains_slice and isinstance(src_indices, Iterable):
```
naylor-b, line 1648: The `isinstance(src_indices, Iterable)` is unnecessary since you're inside of a conditional block where src_indices is a tuple.

openmdao/core/component.py
```diff
@@ -484,7 +484,12 @@ def add_input(self, name, val=1.0, shape=None, src_indices=None, flat_src_indice
         }
 
         if src_indices is not None:
-            metadata['src_indices'] = np.asarray(src_indices, dtype=INT_DTYPE)
+
+            contains_slice = _is_slice(src_indices)
+            if contains_slice:
```
naylor-b, line 489: No need to create a temporary variable here.  Just do  `if _is_slice(src_indices):`

openmdao/core/group.py
```diff
@@ -1216,7 +1217,13 @@ def _setup_connections(self):
                             simple_warning(msg)
 
                 elif src_indices is not None:
-                    src_indices = np.atleast_1d(src_indices)
+
+                    contains_slice = _is_slice(src_indices)
+                    if contains_slice:
```
naylor-b, line 1222: No temporary variable needed here either.

openmdao/utils/general_utils.py
```diff
@@ -222,7 +222,10 @@ def ensure_compatible(name, value, shape=None, indices=None):
                                  (name, shape, value.shape))
 
     # finally make sure shape of indices is compatible
-    if indices is not None and shape != ind_shape[:len(shape)]:
+    if isinstance(indices, np.ndarray):
+        contains_slice = _is_slice(indices)
+
+    if indices is not None and shape != ind_shape[:len(shape)] and not contains_slice:
```
naylor-b, line 228: If indices is not None but also not an ndarray, it looks like contains_slice will be undefined.  If the test suite isn't catching this, we probably need another test.

openmdao/vectors/default_transfer.py
```diff
@@ -94,7 +94,15 @@ def _setup_transfers(group):
                     if src_indices is None:
                         pass
                     elif src_indices.ndim == 1:
-                        src_indices = convert_neg(src_indices, meta_out['global_size'])
+                        if isinstance(src_indices, tuple) or isinstance(src_indices, np.ndarray):
+                            contains_slice = [True if isinstance(i, slice) else
```
naylor-b, line 98: this should just call the _is_slice function.  Also, it should only call _is_slice if src_indices is a tuple or (src_indices is an ndarray AND dtype==object).

openmdao/core/group.py
```diff
@@ -1216,7 +1217,11 @@ def _setup_connections(self):
                             simple_warning(msg)
 
                 elif src_indices is not None:
-                    src_indices = np.atleast_1d(src_indices)
+                    if _is_slice(src_indices):
+                        val = self._abs_get_val(abs_out)
+                        src_indices = np.array([i for i in range(len(val[tuple(src_indices)]))])
```
naylor-b, line 1222: There are some problems here.  What if abs_out is a distributed variable?  In that case, src_indices will reference into the full distributed variable.  `val = self._abs_get_val(abs_out)` will only return the local part of the distributed variable.  Also, the value could have greater than a single dimension.  You really need a function that takes the global shape of the output and the 'fancy' indexing tuple and computes the index values along each dimension.  The current implementation will always give indices of [0, 1, 2, .....] which may not be correct.  This also means that we need a new test that has src_indices with distributed variables.

openmdao/vectors/default_transfer.py
```diff
@@ -94,7 +95,15 @@ def _setup_transfers(group):
                     if src_indices is None:
                         pass
                     elif src_indices.ndim == 1:
-                        src_indices = convert_neg(src_indices, meta_out['global_size'])
+                        if isinstance(src_indices, tuple) or \
+                                     (isinstance(src_indices, np.ndarray) and
+                                      src_indices.dtype == object):
+                            if _is_slice(src_indices):
+                                indices = _slice_indices(src_indices, meta_out['size'],
```
naylor-b, line 102: Use meta_out['global_size'] and meta_out['global_shape'] here.  Note that when a variable is NOT distributed, global_size == size and global_shape == shape.

openmdao/core/group.py
```diff
@@ -1216,7 +1217,16 @@ def _setup_connections(self):
                             simple_warning(msg)
 
                 elif src_indices is not None:
-                    src_indices = np.atleast_1d(src_indices)
+                    inds_size = None
+                    if _is_slice(src_indices):
+                        if MPI:
```
naylor-b, line 1222: No need to retrieve the value here. You need the global size and shape which you can find in `self._var_allprocs_abs2meta`.

naylor-b:
This PR still says it's merging into master.  Does it need to be closed and reopened to point to auto_ivc?

openmdao/utils/general_utils.py
```diff
@@ -975,6 +977,48 @@ def common_subpath(pathnames):
...
+    ----------
+    indices : ndarray
+        Dotted pathnames of systems.
+
+    Returns
+    -------
+    bool
+        Returns True if indices contains a slice.
+    """
+    contains_slice = any(isinstance(i, slice) for i in indices)
```
naylor-b, line 994: no need for temporary here.  Just return any(...

openmdao/utils/general_utils.py
```diff
@@ -975,6 +977,48 @@ def common_subpath(pathnames):
...
+        Size of output array
+    out_shape : tuple
+        Tuple of output array shape
+
+    Returns
+    -------
+    array
+        Returns the sliced indices.
+    """
+    sliced_inds = np.arange(out_size, dtype=int).reshape(out_shape)[tuple(slicer)]
```
naylor-b, line 1017: no need for temporary here

openmdao/core/tests/test_group.py
```diff
@@ -514,6 +514,191 @@ def compute(self, inputs, outputs):
                          "Group (<model>): src_indices has been defined in both "
                          "connect('indep.x', 'C1.x') and add_input('C1.x', ...).")
 
+    def test_om_slice_in_connect(self):
```
naylor-b, line 517: I haven't seen any tests that use a slicer with a negative stop.  Could we add that for both serial and MPI?

PR #1492, opened 2020-07-06
Added flag for `under_approx` and a new counter `iter_count_without_approx`

johnjasa:
### Summary

Just as a user may want to know when a system is [under complex step](http://openmdao.org/twodocs/versions/latest/features/core_features/working_with_derivatives/approximating_partials.html?highlight=under_complex#complex-step), it's helpful to know when the system is under approximation of any kind. This PR adds a flag for all systems that is True when the system is under approximation, and False when it is not.

Additionally, this PR adds `iter_count_without_approx` to systems, which counts the number of times `compute()` has been called while not counting any `compute()` calls that have been called as part of an approximation. This allows the user to add logic to a system based on the number of analysis calls to it regardless of the approximation scheme used.

I've added a test for the counting and the relevant docstrings. Feel free to suggest any syntax or docs as needed.

### Related Issues

No known issues

None, only adds attributes to systems

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -281,10 +281,10 @@ class ApproximationScheme(object):
 
     def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
-        
+
         # Set system flag that we're under approximation to true
         system._set_approx_mode(True)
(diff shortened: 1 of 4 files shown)


swryan:
Since this is a new feature, we would like to have a POEM associated with it.

PR #1493, opened 2020-07-06
Added a new NLopt driver

johnjasa:
### Summary

[NLopt](https://nlopt.readthedocs.io/en/latest/) is an open-source library for nonlinear optimization, similar to pyOptSparse. It grants access to a large number of global and local, derivative-free and gradient-based methods. It's a well-documented package with support for multiple programming languages, though this driver uses the Python version.

This PR adds an NLopt driver, associated tests, and updates the driver documentation. Not all of the methods included in NLopt have been tested through this PR. Instead, we focus on relevant and useful methods for our specific application purposes at NREL, specifically methods that can handle nonlinear constraints.

I'd be comfortable listing NLopt as an experimental feature because of that, and since the core OpenMDAO team didn't implement it, but that might also be worth a discussion. It doesn't currently have any logic for MPI, in general it's not quite as full-featured as pyOptSparse.

Please let me know what you want to see added or changed! We can talk it through on a call too if it's easier.

### Related Issues

None

None

### New Dependencies

Optional NLopt dependency if using the driver


johnjasa:
Closing to implement as a plugin instead

Issue #1498, opened 2020-07-10
Case recording on a DOEDriver with a parallel model does not record cases for all remote inputs.


Kenneth-T-Moore:
### Summary of Issue

Case recording on a DOEDriver with a parallel model does not record all remote variables.

This bug is present in OpenMDAO 3.1.1 and also in 3.2 (auto_ivc).
### Issue Type

- [x] Bug
- [ ] Enhancement  
      POEM #___ <N/A or POEM number at https://github.com/OpenMDAO/POEMs>

### Description

The problem stems from how the case recording works for "run_parallel" = True, wherein we record a separate file for each set of DOE points that we do in parallel.

1. When we have a parallel model, and we try to additionally run a parallel DOE, the driver splits the comm up to give each parallel model `procs_per_node` number of processors.  The bug here is that it only records one sql file per model, so you aren't recording any values that are remote to the proc that does the recording.  [For example: i have 4 procs and a model that needs 2 procs, so i can run two DOE points in parallel.] The solution that makes sense is for every rank to always record a file.

2. The following two tests are partially disabled and should be restored:
TestParallelDOE:test_fan_in_grouped
TestParallelDOEFeature2:test_fan_in_grouped

3. While working on this, take a loo at (add a test for) `run_parallel`=False on a parallel model where you have a recorder. This case terminates with an error during setup..

### Example

TestParallelDOE:test_fan_in_grouped
TestParallelDOEFeature2:test_fan_in_grouped

For enhancements this can be left as N/A assuming the enhancement is captured in a POEM

### Environment

<For bugs, otherwise N/A>

Operating System: <i.e. OS X 10.14.6, Windows 10, Ubuntu 16.04>

Python environment: <i.e. Anaconda Python 3.7.1>

Packages: <versions of Dymos, OpenMDAO, and other pertinent Python packages. You can just paste the output of `pip freeze` or `conda list` here.>

PR #1499, opened 2020-07-13
Auto ivc2 bug fixes - set 1.

Kenneth-T-Moore:
### Summary

1.  Relevancy `vec_names` and `lin_vec_names` converted to point to the auto_ivc sources during setup.
2.  Certain debug printing under parallel derivatives now prints the promoted input name instead of the auto_ivc source name.
3. Fixed a bug where global sizes were incorrect in auto_ivc source vars connected to inputs inside of parallel groups. (thanks @naylor-b )
4. Disabled a check in a test for parallel models under a parallel DOE. Apparently there has always been a problem, but the auto_ivc change brought it to the forefront. It will be addressed in #1498 .
5. Fixed the new data structures in case recording to correctly reference the all_procs versions.
6. Feature doc updated to showcase auto_ivc in the rest of the components library (thanks @DKilkenny )

### Related Issues

- Resolves #

None

### New Dependencies

None


openmdao/components/tests/test_external_code_comp.py
```diff
@@ -612,6 +608,7 @@ def solve_nonlinear(self, inputs, outputs):
 
         prob.setup()
 
+        prob.set_val('area_ratio', 0.5)
```
naylor-b, line 611: Is this set intentional since you override it 3 lines later?

openmdao/core/system.py
```diff
@@ -4245,18 +4269,28 @@ def _retrieve_data_of_kind(self, filtered_vars, kind, vec_name, parallel=False):
                         else:
                             ivc_path = conns[prom2abs_in[name][0]]
                             vdict[ivc_path] = views[ivc_path]
+                            print(name, views_flat[ivc_path].size)
```
naylor-b, line 4272: is this left over from debugging?

openmdao/core/system.py
```diff
@@ -4245,18 +4269,28 @@ def _retrieve_data_of_kind(self, filtered_vars, kind, vec_name, parallel=False):
...
-                    vdict = {n: views[n] for n in variables if views_flat[n].size > 0}
+                    for name in variables:
+                        if name in views:
+                            if views_flat[name].size > 0:
+                                vdict[name] = views[name]
+                        else:
+                            ivc_path = conns[prom2abs_in[name][0]]
+                            if views_flat[ivc_path].size > 0:
+                                vdict[ivc_path] = views[ivc_path]
+                            print(name, views_flat[ivc_path].size)
```
naylor-b, line 4292: is this left over from debugging?

openmdao/core/system.py
Kenneth-T-Moore: it must be

openmdao/core/system.py
Kenneth-T-Moore: it must be

openmdao/components/tests/test_external_code_comp.py
Kenneth-T-Moore: That was the old default value set in the IVC, but it is redundant and this is an example, so removed.

PR #1500, opened 2020-07-13
Auto IVC Doc Changes: Solvers and Surrogates

DKilkenny:
### Summary

This PR covers all solver tests under Building Blocks.

### Related Issues

- Resolves #NA

None

### New Dependencies

None


openmdao/solvers/linesearch/tests/test_backtracking.py
```diff
@@ -858,9 +856,8 @@ def test_feature_armijogoldsteinls_basic(self):
         from openmdao.test_suite.components.implicit_newton_linesearch import ImplCompTwoStatesArrays
 
         top = om.Problem()
-        top.model.add_subsystem('px', om.IndepVarComp('x', np.ones((3, 1))))
-        top.model.add_subsystem('comp', ImplCompTwoStatesArrays())
-        top.model.connect('px.x', 'comp.x')
+        top.model.set_input_defaults('x')
```
naylor-b, line 859: Why are we calling `set_input_defaults` without actually setting any defaults?

DKilkenny: It gets set with `top['x'] = 0.5` on line 952

naylor-b: Calling `set_input_defaults` with just a variable name, without setting 'units' or 'value', is literally doing nothing, so why are we calling it at all?  And if you do want to disambiguate the value of any inputs that map to 'x', it's better to do it by passing the value into `set_input_defaults` instead of setting it via the problem interface because it lets the framework do the ambiguity checking more efficiently, i.e., if you don't specify what the input default is, we have to compare all of the input values and make sure they're identical.

DKilkenny: Fixed

PR #1534, opened 2020-07-16
SimpleGA fixes

wright:
### Summary

This is a few minor bug fixes for the SImpleGA driver.

1 - The results of the test_distribute_obj test were not close the actual optimal values. I increased the number of bits and generations to make it close.

2 - The option to set an environment variable to fix the random number generator seed was not working (tests were not repeatable),

3 - As a result of changing the RNG, the results of the pareto test changed. I put in new values for the unit test, and Ken (author of the parato code) looked them over and says they look OK. 

### Related Issues

- Resolves #

None

### New Dependencies

None


wright:
I reduced the bits and generations even further and made the assert_near_equal much more lenient, and this caused an _additional_ Travis test to time out on test_distributed_obj.

Also, after this change, two of the Travis runs are now also failing test_option_parallel. I don't understand what is going on with the Travis MPI tests.

wright:
The RNG changes use the os.environ seed to initialize a new random number generator used within the GA, so only setting the environment variable is necessary.

The changes are pretty minor: https://github.com/OpenMDAO/OpenMDAO/pull/1534/files
Does this behave badly in MPI?

Kenneth-T-Moore:
@wright  I suspect there is some problem under MPI.  Don't know why we don't experience it locally.

Try removing this line in the MPI tests and pushing up to see if those tests pass:

   np.random.seed(1)

Here is a story that might be related:
https://github.com/numpy/numpy/issues/9248

wright:
I rebased onto master to help debug the merged code.

Kenneth-T-Moore:
II *think* removing the extra seed in the tests actually worked.

wright:
I also think removing the seed call got rid of the Travis timeouts, but I have been having a lot of Travis issues, so I am not 100% sure.

PR #1536, opened 2020-07-17
support options on IndepVarComp

### LINKED ISSUE #1481: options don't work when using IndepVarComp
### Summary of Issue

A user tired to inherit from IndepVarComp and add some custom options that would determine what outputs got added. however, the kwargs are not passed down into the super command, so the options silently kept their default values. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

See test case below, but the result should be that the options affect the definition of the outputs, but that doesn't happen. The cause is that **kwargs are not passed down into the super call. 

### Example

```
import numpy as np

import openmdao.api as om 

# definition
class parameters(om.IndepVarComp):
    def initialize(self):
        self.options.declare('num_x', default=0)
        self.options.declare('val_y', default=0.)
        
    def setup(self):

        self.add_discrete_output('num_x', val = np.zeros(self.options['num_x']))
        self.add_output('val_y',val = self.options['val_y'])
#run script
prob = om.Problem()
model = prob.model
fileName = 'tmp.sql'
model.add_recorder(om.SqliteRecorder(fileName))

parameters = parameters(num_x=4, val_y=2.5)

model.add_subsystem('params',parameters)

prob.setup()
prob.run_model()

# Inputs with includes
inputs = case.list_inputs(print_arrays=True)
outputs = case.list_outputs(print_arrays=True) 

```

swryan:
### Summary

IndepVarComp did not use the `options` mechanism like other Components, which was an issue for a user that wanted to subclass it with additional options. 

### Related Issues

- Resolves #1481

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/indepvarcomp.py
+++ b/openmdao/core/indepvarcomp.py
@@ -1,5 +1,7 @@
 """Define the IndepVarComp class."""
 
+import numpy as np
+
 from openmdao.core.explicitcomponent import ExplicitComponent
 from openmdao.utils.general_utils import make_set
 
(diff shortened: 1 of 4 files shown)


openmdao/core/indepvarcomp.py
```diff
@@ -48,6 +50,49 @@ def __init__(self, name=None, val=1.0, **kwargs):
...
+                    desc="Units in which the output variables will be provided to the "
+                         "component during execution.")
+        opt.declare('res_units', types=str, default=None,
+                    desc="Units in which the residuals of this output will be given to "
+                         "the user when requested.")
+        opt.declare('desc', types=str,
+                    desc="Description of the variable")
+        opt.declare('lower', types=(int, float, list, tuple, np.ndarray), default=None,
+                    desc="Lower bound(s) in user-defined units. It can be (1) a float, "
+                         "(2) an array_like consistent with the shape arg (if given), or "
```
Kenneth-T-Moore, line 74: Many of these options are going to go away based on channel discussion (in particular, solver-related options like the bounds and scaling.)  Guess that just means you did a little extra work....

swryan: yeah.. this was easy enough without resolving those issues right now

PR #1541, opened 2020-07-18
WIP: Cache Kriging trained surrogates to expedite execution

bbrelje:
### Summary

The Kriging surrogate can take a long time to train, especially if multiple instances are created in a model.
This PR implements caching of the trained kriging model for repeat runs using pickle

None

### New Dependencies

None

### Status

I developed this for a different project. Looking for feedback on whether it's something the team finds desirable in the main project or whether I should keep it as a subclass in my other project. If thumbs up then I'll go ahead and write the OpenMDAO tests or make mods as needed.

Another open question is whether we want to validate that the model training inputs match what was used to train the cached model which potentially increases the size of the cache file but avoids the opportunity for weird behavior (surrogate model training data changes but behaviour doesn't because it's using a stale cache)


openmdao/surrogate_models/kriging.py
```diff
@@ -91,6 +92,12 @@ def _declare_options(self):
                                   "or 'gesvd' which is slower but more reliable."
                                   "'gesvd' is the default.")
 
+        self.options.declare('cache_trained_model', default=False, types=bool,
+                             desc='Cache the trained model to avoid repeat training')
```
Kenneth-T-Moore, line 96: You could probably get a way with just having one option 'cached_model_filename', that can be set to None when you don't want to load a cache. (I guess what you lose then is the standardized default filename.)

Maybe you could also have a similar option to save the surrogate model to a a pickle.

bbrelje:
@Kenneth-T-Moore or anybody on team really: I'm looking for some more generic feedback here - is this a POEM-level change or appropriate to just do as a PR? If the former, I'll close this PR and write a POEM. If the latter, I'll go ahead and write tests and address your comment above.

Kenneth-T-Moore:
I think maybe @robfalck is the best to decide. (He is back next week.)  This does feel POEM level to me because we still have some open API questions (how do we record the cache and what should the API be for that? do users want to select different formats? should this data be recorded in the case recorders somehow?)

JustinSGray:
technically, any change to the user interface (which includes options) should get a POEM. When we started the POEM process, I thought "We can probably only do this for larger changes", but as we've used it I've come to appreciate the value it brings more, not just for the community but for the devs too. I like having a POEM that explains the motivation and context for a feature, because I think that forcing ourselves to write that down often leads to discovering cleaner APIs. 

In this case, if it was just an single option, id probably skip the POEM. However, a short one detailing the caching logic is probably warranted. Here is some more food for thought: 

1) Are there other surrogates that would benefit from the caching? Is this generic enough? 

2)You mentioned the need to potentially check the training data to see if it matches the cache. From my read of the code, this seems pretty important. Right now if you had multiple instances of kriging with different training data, but they all found the same cache file then they would all use it anyway. So I agree with your concern about "weird behavior" 

Note: instead of storing the training data as a floating point, you can do some kind of checksum on the training data array and key to that instead. That will save some memory and be plenty safe. I admit I don't have a checksum algorithm ready at hand. Probably some combination of [md5 or a crc](https://docs.python.org/3/library/hashlib.html) and the [tobytes method](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.ndarray.tobytes.html)

Dakror:
I've taken inspiration from this PR to suggest a new POEM 036 with a (hopefully) cleaner slate: https://github.com/OpenMDAO/POEMs/pull/79

JustinSGray:
superseded by #1753

PR #1543, opened 2020-07-20
Document prerequisite calls to Problem.record()

payoto:
### Summary

Document in the `Problem.record` doc-string that it must be called
after `Problem.final_setup`. Otherwise the following error gets thrown:

    AttributeError: 'Problem' object has no attribute 
    '_filtered_vars_to_record'

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -741,6 +741,10 @@ class Problem(object):
         """
         Record the variables at the Problem level.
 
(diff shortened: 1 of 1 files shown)


robfalck:
Thank you for the pull request.  I'm going to add a related issue and suggest hat the AttributeError is changed to a RuntimeError with an error message that tells the user to invoke `final_setup`, `run_model`, or `run_driver`

PR #1568, opened 2020-07-24
Reworked error/warning behavior for set_input_defaults.

### LINKED ISSUE #1565: Change error in the event of conflicting metadata entries in multiple `set_input_defaults` calls to a warning.
### Summary of Issue

Currently if there are multiple set_input_defaults calls at different levels in the system hierarchy that have conflicting metadata values, we raise an exception.  The desired behavior is that higher level calls should override lower level calls and the exception should just be a warning.  The final metadata items will be a union of all metadata items set in all the calls pertaining to the same set of inputs, with higher level values overriding those from lower levels.  The warning message will include information about any metadata items brought up from lower level calls.

### Issue Type

- [x] Bug
- [ ] Enhancement

naylor-b:
### Summary

Ambiguity resolution now happens at the top level only.  Ambiguities that are resolved at a higher level are now warnings instead of errors.  A warning is issued if a higher level call to set_input_defaults has a missing metadata entry that is supplied from a lower level.

### Related Issues

- Resolves #1565

None

### New Dependencies

None


openmdao/core/tests/test_group.py
```diff
@@ -2198,14 +2199,14 @@ def test_conflicting_units(self):
         with self.assertRaises(Exception) as cm:
             p.setup()
 
-        self.assertEqual(cm.exception.args[0], "Groups 'par.G4' and 'par.G5' added the input 'x' with conflicting 'units'.")
+        self.assertEqual(cm.exception.args[0], "Group (<model>): The subsystems G1.G2 and par.G4 called set_input_defaults for promoted input 'x' with conflicting values for 'units'. Call <group>.set_input_defaults('x', units=?), where <group> is the Group named '' to remove the ambiguity.")
```
swryan, line 2202: I wonder if we should have special handling for the top level group...  
It seems like this message would make more sense to the user if it referred to "the 'model'" instead of "the group named ''", since that's how the user would reference it.

naylor-b: changed "Group named ''" to "model"

openmdao/core/group.py
```diff
@@ -53,6 +53,10 @@ class Group(System):
         Dictionary of input_name: (output_name, src_indices) connections.
     _group_inputs : dict
         Mapping of promoted names to certain metadata (src_indices, units).
+    _static_group_inputs : dict
+        Group inputs added outside of setup/configure.
+    _post_setup_group_inputs : dict
```
Kenneth-T-Moore, line 58: This name confuses me a little. At first, I thought it meant after the user calls prob.setup.  At second, I thought that configure is technically "post setup" too.  Maybe pre_configure is better?

openmdao/core/group.py
```diff
@@ -53,6 +53,10 @@ class Group(System):
         Dictionary of input_name: (output_name, src_indices) connections.
     _group_inputs : dict
         Mapping of promoted names to certain metadata (src_indices, units).
+    _static_group_inputs : dict
+        Group inputs added outside of setup/configure.
```
Kenneth-T-Moore, line 57: Where else can we add inputs? Does this just mean after prob.setup?

openmdao/core/group.py
```diff
@@ -762,14 +789,16 @@ def _setup_var_data(self):
...
+            for rank, (myproc_abs_names, myproc_abs_names_discrete, myproc_discrete,
+                       myproc_prom2abs_list, all_abs2prom, myproc_abs2meta, oscale,
+                       rscale, ginputs) in enumerate(gathered):
                 self._has_output_scaling |= oscale
                 self._has_resid_scaling |= rscale
 
-                group_inputs.extend(ginputs)
+                if rank != myrank:
+                    for p, mlist in ginputs.items():
+                        self._group_inputs[p].extend(mlist)
```
Kenneth-T-Moore, line 801: Is this a bug fix?

openmdao/core/group.py
```diff
@@ -796,42 +825,87 @@ def _setup_var_data(self):
...
+                    continue
+                for i, submeta in enumerate(metalist):
+                    if key in submeta:
+                        if fullmeta[key] is _undefined:
+                            val = fullmeta[key] = submeta[key]
+                            origin = submeta['path']
+                            origin_prom = submeta['prom']
+                            if origin != top_origin:
+                                simple_warning(f"Group '{top_origin}' did not set a default "
+                                               f"'{key}' for input '{top_prom}', so the value of "
```
Kenneth-T-Moore, line 868: is there a test for this warning?

openmdao/core/group.py
naylor-b: Static means everything outside of system setup and configure.  If things are added in setup/configure they're considered 'dynamic' and they get blown away if problem setup is called again.  Static stuff is never removed.

openmdao/core/group.py

openmdao/core/group.py
naylor-b: There is now.  :)

openmdao/core/group.py
naylor-b: The data structure was changed to make error checking easier later.

openmdao/core/group.py
swryan: this would be a nice comment to put in the code somewhere

naylor-b: updated comment

PR #1574, opened 2020-07-29
Added Stall Detection to the nonlinear solvers.

### LINKED ISSUE #1523: Adding stall detection to nonlinear solvers
### Summary of Issue

Add some new convergence options to solvers to handle common cases we see and improve efficiency

### Issue Type

- [ ] Bug
- [x] Enhancement  
      

### Description

In some cases, nonlinear solvers can stall out where the norm of the residual stops changing at all. This can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in a circle, or you can get stuck on a bound and the line search just keeps running into the same spot no matter what. Either way, if you have say 100 max iterations and you stall at 15 ... you waste a lot of compute time. 

This happens particularly often in pyCycle where the chem_eq can stall out, especially during intermediate steps. 

We need to add a new pair of options `stall_limit` and `stall_tol`. 

`stall_limit` will cause the solver to stop iterating if the same residual norm is detected for the set number of time. Default 3

`stall_tol` is the measure used to detect a change in the residual. You shouldn't just use an equality check because we're talking about very small floating point numbers. So we should look to see if the norm of the residual is changing by less than some amount. It shouldn't matter if we use the absolute or relative norm, since the delta should be the same. I think this should be a relative change. In well behaved solvers, you should see at around 0.5 orders of magnitude between iterations (at least). So if you see it barely changing (say less than .001 %) then we can stop. Default `1e-6`. 

Notes: 

- Termination due to stall should raise a non-convergence error (only if the options is set to raise non conevergence errors) 
- The non-convergence error (or warning) should be modified to say that it died to to the stall limit in this case. 

### Example
In this example, I intentionally set the lower limit on the state variable to -0.1. The correct answer is -0.333333, but now the solver can't reach it because of my limit. This causes it to stall out and do a bunch of iterations that don't make progress: 

```
import openmdao.api as om


p = om.Problem()

p.model.add_subsystem('comp', om.ExecComp('y=3*x+1'), promotes=['*'])
balance = p.model.add_subsystem('balance', om.BalanceComp(), promotes=['*'])

balance.add_balance('x', lower=-.1, upper=10, rhs_val=0, lhs_name='y')

newton = p.model.nonlinear_solver = om.NewtonSolver()
newton.options['iprint'] = 2
newton.options['solve_subsystems'] = True
p.model.linear_solver = om.DirectSolver()

p.setup()
p.run_model()

p.model.list_outputs()
```

**gives**: 

NL: Newton 0 ; 4 1
NL: Newton 1 ; 0.7 0.175
NL: Newton 2 ; 0.7 0.175
NL: Newton 3 ; 0.7 0.175
NL: Newton 4 ; 0.7 0.175
NL: Newton 5 ; 0.7 0.175
NL: Newton 6 ; 0.7 0.175
NL: Newton 7 ; 0.7 0.175
NL: Newton 8 ; 0.7 0.175
NL: Newton 9 ; 0.7 0.175
NL: Newton 10 ; 0.7 0.175

with default settings, this should terminate after NL: Newton 3 ; 0.7 0.175)

Kenneth-T-Moore:
### Summary

Added Stall Detection capability to the nonlinear solvers via new 'stall_limit' and 'stall_tol' options.

### Related Issues

- Resolves #1523

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/features/building_blocks/solvers/nonlinear/broyden.rst
+++ b/openmdao/docs/features/building_blocks/solvers/nonlinear/broyden.rst
@@ -119,7 +119,7 @@ number of iterations, though keep in mind that solving for the derivatives adds 
   In this example, we set stall_limit to 3. While the solver iterates, it will compare the value of the
   residual norm to the value computed in the previous iteration.  If the value matches for three iterations
   in a row, then iteration will terminate due to detection of a stall. If "err_on_non_converge" is set
-  to True, then an ``AnalysisError`` just as if we had reached the iteration count limit.
+  to True, then an ``AnalysisError`` will be raised just as if we had reached the iteration count limit.
 
   We also set the `stall_tol` to 1e-6, which is the threshold below which a change in the relative residual
   norm is considered to be unchanged.
--- a/openmdao/docs/features/building_blocks/solvers/nonlinear/broyden.rst
+++ b/openmdao/docs/features/building_blocks/solvers/nonlinear/broyden.rst
@@ -106,3 +106,24 @@ number of iterations, though keep in mind that solving for the derivatives adds 
   .. embed-code::
       openmdao.solvers.nonlinear.tests.test_broyden.TestBryodenFeature.test_circuit_options
       :layout: code, output
+
+**stall_limit and stall_tol**
+
+  In some cases, nonlinear solvers can stall out where the norm of the residual stops changing at all. This
+  can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in
+  a circle, or you can get stuck on a bound and the line search just keeps running into the same spot no
+  matter what. Either way, if you have say 100 max iterations and you stall at 15 ... you waste a lot of
+  compute time. To remedy this, you can turn on stall detection in all nonlinear solvers by setting the
+  "stall_limit" option to a number greater than zero.
+
+  In this example, we set stall_limit to 3. While the solver iterates, it will compare the value of the
+  residual norm to the value computed in the previous iteration.  If the value matches for three iterations
+  in a row, then iteration will terminate due to detection of a stall. If "err_on_non_converge" is set
+  to True, then an ``AnalysisError`` just as if we had reached the iteration count limit.
+
+  We also set the `stall_tol` to 1e-6, which is the threshold below which a change in the relative residual
+  norm is considered to be unchanged.
+
+  .. embed-code::
+      openmdao.solvers.tests.test_solver_features.TestSolverFeatures.test_feature_stall_detection_broyden
+      :layout: interleave
--- a/openmdao/docs/features/building_blocks/solvers/nonlinear/newton.rst
+++ b/openmdao/docs/features/building_blocks/solvers/nonlinear/newton.rst
@@ -113,6 +113,28 @@ NewtonSolver Option Examples
 
   This feature can be set on any iterative nonlinear or linear solver.
 
+**stall_limit and stall_tol**
+
+  In some cases, nonlinear solvers can stall out where the norm of the residual stops changing at all. This
+  can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in
+  a circle, or you can get stuck on a bound and the line search just keeps running into the same spot no
+  matter what. Either way, if you have say 100 max iterations and you stall at 15 ... you waste a lot of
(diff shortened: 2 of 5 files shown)


openmdao/docs/features/building_blocks/solvers/nonlinear/broyden.rst
```diff
@@ -106,3 +106,24 @@ number of iterations, though keep in mind that solving for the derivatives adds
...
+  can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in
+  a circle, or you can get stuck on a bound and the line search just keeps running into the same spot no
+  matter what. Either way, if you have say 100 max iterations and you stall at 15 ... you waste a lot of
+  compute time. To remedy this, you can turn on stall detection in all nonlinear solvers by setting the
+  "stall_limit" option to a number greater than zero.
+
+  In this example, we set stall_limit to 3. While the solver iterates, it will compare the value of the
+  residual norm to the value computed in the previous iteration.  If the value matches for three iterations
+  in a row, then iteration will terminate due to detection of a stall. If "err_on_non_converge" is set
+  to True, then an ``AnalysisError`` just as if we had reached the iteration count limit.
```
naylor-b, line 122: missing some words after `AnalysisError`, maybe 'will be raised`?

openmdao/docs/features/building_blocks/solvers/nonlinear/newton.rst
```diff
@@ -113,6 +113,28 @@ NewtonSolver Option Examples
...
+  can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in
+  a circle, or you can get stuck on a bound and the line search just keeps running into the same spot no
+  matter what. Either way, if you have say 100 max iterations and you stall at 15 ... you waste a lot of
+  compute time. To remedy this, you can turn on stall detection in all nonlinear solvers by setting the
+  "stall_limit" option to a number greater than zero.
+
+  In this example, we set stall_limit to 3. While the solver iterates, it will compare the value of the
+  residual norm to the value computed in the previous iteration.  If the value matches for three iterations
+  in a row, then iteration will terminate due to detection of a stall. If "err_on_non_converge" is set
+  to True, then an ``AnalysisError`` just as if we had reached the iteration count limit.
```
naylor-b, line 128: As before, missing words after `AnalysisError`.

openmdao/solvers/solver.py
```diff
@@ -633,6 +570,101 @@ def _iter_initialize(self):
...
+            stall_norm = norm0
+
+        while self._iter_count < maxiter and norm > atol and norm / norm0 > rtol and not stalled:
+            with Recording(type(self).__name__, self._iter_count, self) as rec:
+                self._single_iteration()
+                self._iter_count += 1
+                self._run_apply()
+                norm = self._iter_get_norm()
+                # With solvers, we want to record the norm AFTER the call, but the call needs to
+                # be wrapped in the with for stack purposes, so we locally assign  norm & norm0
```
naylor-b, line 605: typo:  'in the with'

openmdao/solvers/solver.py
```diff
@@ -784,6 +816,77 @@ def solve(self, vec_names, mode, rel_systems=None):
...
+        self._mpi_print(self._iter_count, norm, norm / norm0)
+
+        while self._iter_count < maxiter and norm > atol and norm / norm0 > rtol:
+            with Recording(type(self).__name__, self._iter_count, self) as rec:
+                self._single_iteration()
+                self._iter_count += 1
+                self._run_apply()
+                norm = self._iter_get_norm()
+                # With solvers, we want to record the norm AFTER the call, but the call needs to
+                # be wrapped in the with for stack purposes, so we locally assign  norm & norm0
```
naylor-b, line 844: typo: 'in the with'

openmdao/docs/features/building_blocks/solvers/nonlinear/newton.rst
Kenneth-T-Moore: Fixed

openmdao/solvers/solver.py
Kenneth-T-Moore: I didn't write this comment, but it is very confusing and roundabout. Rewrote to simplify.

PR #1575, opened 2020-07-29
Modified ambiguity message

johnjasa:
### Summary

When using auto IVCs, a descriptive error message pops up if you need to call `set_input_defaults()` to fix ambiguous metadata entries. However, if the values differed between inputs, the error message suggested calling `<group>.set_input_defaults(<var>, value=?)` when in reality the correct kwarg is  `<group>.set_input_defaults(<var>, val=?)`. This PR changes that error message to show `val` instead of `value` and also updates the relevant tests.

### Related Issues

N/A

N/A

### New Dependencies

N/A

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -2881,7 +2881,7 @@ class Group(System):
 
                 if tgt in all_discrete_ins:
                     if 'value' not in gmeta and sval != tval:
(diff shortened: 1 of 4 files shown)


johnjasa:
Tabling this for now since it requires a larger discussion of standardizing val/value usage.

PR #1578, opened 2020-07-30
Raised an error if set_val is called before setup

### LINKED ISSUE #1566: Create Error Message for Incorrect Call Order
### Summary of Issue

Currently if prob.set_val is called before prob.setup the error message that is thrown does not properly explain what happened. The desired behavior is that if set_val is called before setup, an error message that explains what happened is thrown.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Example

The following provides the requested short sample script that throws the current bad behavior.

import openmdao.api as om

class Paraboloid(om.ExplicitComponent):

    def setup(self):
        self.add_input('x', val=0.0)
        self.add_input('y', val=0.0)

        self.add_output('f_xy', val=0.0)

        self.declare_partials('*', '*', method='fd')

        x = inputs['x']
        y = inputs['y']

        outputs['f_xy'] = (x - 3.0)**2 + x * y + (y + 4.0)**2 - 3.0

if __name__ == "__main__":

    model = om.Group()
    model.add_subsystem('parab_comp', Paraboloid())

    prob = om.Problem(model)

    prob.set_val('parab_comp.x', 3.0)
    prob.set_val('parab_comp.y', -4.0)

    prob.setup()

    prob.run_model()
    print(prob['parab_comp.f_xy'])

DKilkenny:
### Summary

Raised an error if set_val is called before setup

### Related Issues

- Resolves #1566

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -431,7 +431,7 @@ class Problem(object):
         try:
             conns = self._metadata['connections']
         except AttributeError:
-            raise RuntimeError(f"{self.msginfo} '{name}': Cannot call set_val before setup.")
+            raise RuntimeError(f"{self.msginfo}: '{name}' Cannot call set_val before setup.")
 
         all_meta = model._var_allprocs_abs2meta
         n_proms = 0  # if nonzero, name given was promoted input name w/o a matching prom output
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -431,7 +431,7 @@ class Problem(object):
         try:
             conns = self._metadata['connections']
         except AttributeError:
-            raise RuntimeError("Cannot call set_val before setup.")
+            raise RuntimeError(f"{self.msginfo} '{name}': Cannot call set_val before setup.")
 
         all_meta = model._var_allprocs_abs2meta
         n_proms = 0  # if nonzero, name given was promoted input name w/o a matching prom output
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -428,7 +428,11 @@ class Problem(object):
             Indices or slice to set to specified value.
         """
         model = self.model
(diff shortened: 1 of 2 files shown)


openmdao/core/problem.py
```diff
@@ -428,7 +428,11 @@ def set_val(self, name, value, units=None, indices=None):
             Indices or slice to set to specified value.
         """
         model = self.model
-        conns = self._metadata['connections']
+        try:
+            conns = self._metadata['connections']
+        except AttributeError:
+            raise RuntimeError("Cannot call set_val before setup.")
```
naylor-b, line 434: Error message should contain the name of the variable they're trying to set and should start with f"{self.msginfo}:".

openmdao/core/problem.py
```diff
@@ -428,7 +428,11 @@ def set_val(self, name, value, units=None, indices=None):
             Indices or slice to set to specified value.
         """
         model = self.model
-        conns = self._metadata['connections']
+        try:
+            conns = self._metadata['connections']
+        except AttributeError:
+            raise RuntimeError(f"{self.msginfo} '{name}': Cannot call set_val before setup.")
```
Kenneth-T-Moore, line 434: looks like for uniformity, we always put the colon after the self.msginfo.  Look at the other error messages for example.

openmdao/core/tests/test_problem.py
```diff
@@ -2044,6 +2044,14 @@ def test_feature_list_problem_vars(self):
                                           'vectorize_derivs',
                                           'cache_linear_solution'])
 
+    def test_out_of_order_set_val(self):
```
Kenneth-T-Moore, line 2047: I think a more explicit test name would be better.  Something like "test_error_msg_set_val_before_setup"

PR #1580, opened 2020-07-31
Enable tagging of add_subtract outputs

bbrelje:
### Summary

As the API has changed over time, the feature comps have not always kept up. This became apparent recently when I needed to tag an output of an AddSubtractComp, but the add_equation method didn't support the tag kwarg. 

There are a number of components (e.g. CrossProduct, MuxDemux) that should be completely generic and support most or all of the kwargs that ExplicitComponent.add_output() or add_input() takes. This PR just fixes the AddSubtractComp, but if it's appropriate to fix most or all of the components at the same time, or if this is a POEM-level change, then we can close this without a merge for now.

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/add_subtract_comp.py
+++ b/openmdao/components/add_subtract_comp.py
@@ -106,7 +106,7 @@ class AddSubtractComp(ExplicitComponent):
 
     def add_equation(self, output_name, input_names, vec_size=1, length=1, val=1.0,
                      units=None, res_units=None, desc='', lower=None, upper=None, ref=1.0,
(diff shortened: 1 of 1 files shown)


Kenneth-T-Moore:
This feels more like a 'bug' fix to me than something that would require a tag, and I think we could accept this so that this bug is no longer present (probably could use a test, though I am pretty sure we don't test the others like 'res_ref'.

One thing to ponder though, is how you would tag the inputs.  Maybe we need to move towards an interface like we have on ExecComp where we pass in a dictionary of arguments for each input.  That would be POEM level stuff right there.

PR #1584, opened 2020-07-31
Updated deprecation warning to only show if user uses ref and ref0 args

DKilkenny:
### Summary

Previous fix made it so that every time `add_outputs` was called, a deprecation warning would be raised

### Related Issues

- Resolves #N/A

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/indepvarcomp.py
+++ b/openmdao/core/indepvarcomp.py
@@ -106,7 +106,7 @@ class IndepVarComp(ExplicitComponent):
         super(IndepVarComp, self)._configure_check()
 
     def add_output(self, name, val=1.0, shape=None, units=None, res_units=None, desc='',
-                   lower=None, upper=None, ref=1.0, ref0=0.0, res_ref=None, tags=None):
+                   lower=None, upper=None, ref=None, ref0=None, res_ref=None, tags=None):
         """
         Add an independent variable to this component.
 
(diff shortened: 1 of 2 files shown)


openmdao/core/indepvarcomp.py
```diff
@@ -152,19 +152,23 @@ def add_output(self, name, val=1.0, shape=None, units=None, res_units=None, desc
             list_outputs.
         """
         if res_units is not None:
-            warn_deprecation(
-                "'res_units' has been deprecated and will be removed in a future version")
+            warn_deprecation("'res_units' has been deprecated and unused. It will be removed in "
+                             "a future version")
```
Kenneth-T-Moore, line 156: Original wording was better on all of these warnings.

openmdao/core/indepvarcomp.py
```diff
@@ -139,10 +139,10 @@ def add_output(self, name, val=1.0, shape=None, units=None, res_units=None, desc
             Default is None.
         ref : float
             Scaling parameter. The value in the user-defined units of this output variable when
-            the scaled value is 1. Default is 1.
```
Kenneth-T-Moore, line 142: What I meant was these lines here. We don't need to describe in detail the default value and the definition of the ref/ref0/etc arguments when we don't even use them.  Just say "This argument is deprecated.  Default can be None. 

So, this:

    ref : float
        Scaling parameter. The value in the user-defined units of this output variable when
        the scaled value is 1. Default is 1.

Should become

    ref : None
        This argument is deprecated.

DKilkenny: Fix

PR #1587, opened 2020-08-03
Fix bogus units error

naylor-b:
### Summary

When multiple inputs promoted  to the same name were explicitly connected to an output, a units ambiguity error was generated even if all of the inputs had the same units.


openmdao/core/system.py
```diff
@@ -4215,7 +4215,12 @@ def _get_input_from_src(self, name, abs_names, conns, units=None, indices=None,
                 try:
```
Kenneth-T-Moore, line 4215: Can you add a docstring to `_get_input_from_src` ?

naylor-b: done

PR #1588, opened 2020-08-03
raise error when using Problem.__getitem__ on a distributed variable

### LINKED ISSUE #1352: get_val behavior not consistent between distributed components and parallel groups
### Summary of Issue

problem['val'] and problem.get_val() do not behave the same way for distributed components and parallel components.

### Issue Type

- [X ] Bug

### Description

Currently, using the __getitem__notation (problem['var']) or problem.get_val('var') to read out the value of a parallel component output (the output of a component in a parallel group) will raise a RuntimeError because get_remote=False is the default. But if get_remote=True, it will gather the entire vector, not just a piece.

Conversely, using problem['var'] or get_val() with a _distributed_ component output will successfully return only the portion of the variable that lives on the current processor. 

It should either
- Return the entire value after a gather operation (useful default), or
- Raise a RuntimeError to be consistent with parallel group behavior

### Example

See the [gist](https://gist.github.com/bbrelje/f0d57f6311655e46d90c0055ac3dee7f)

### Environment

<For bugs, otherwise N/A>

Packages: OpenMDAO a938f8ad5e20f61134398fef2663a6015af9d756

swryan:
### Summary

Raise error when using `Problem.__getitem__` on a distributed variable, 
so that problem['val'] and problem.get_val() behave consistently for 
distributed components and parallel components.

### Related Issues

- Resolves #1352

Some users may have been using `__getitem__` access for distributed variables, they will have to switch to using get_val()

### New Dependencies

None


openmdao/core/problem.py
```diff
@@ -364,10 +364,11 @@ def get_val(self, name, units=None, indices=None, get_remote=False):
             Units to convert to before return.
         indices : int or list of ints or tuple of ints or int ndarray or Iterable or None, optional
             Indices or slice to return.
-        get_remote : bool
+        get_remote : bool or None
             If True, retrieve the value even if it is on a remote process.  Note that if the
```
Kenneth-T-Moore, line 368: Since this argument can be True, False, or None, the docstring should clearly describe what happens if you set it to each of these three cases. (Seems to be missing False right now.).

PR #1590, opened 2020-08-05
Added an option to control NLBGS initial atkien relaxation factor 

joanibal:
### Summary

There is no reason the the relaxation factor used for the AitKen's acceleration must start at one. In fact, the reference for the implementation, "Scalable Parallel Approach for High-Fidelity Steady-State Aero-elastic Analysis and Adjoint Derivative Computations (ln 22 of Algo 1)", states that the initial relaxation factor should just be between 0 and 1. It has been observed that in practice for aero-structural and conjugate heat transfer problems an initial relaxation factor closer to 0.4 is often required for convergence. 

Therefore, I added the option `aitken_initial_factor` to NLBGS solver so users can tweak the initial relaxation factor themselves. 

None

### New Dependencies

None


Kenneth-T-Moore:
Can you add a test to `test_nonlinear_block_gs.py` where you set the new option and make sure it is working correctly? After that, it will be good to go, though we may also need a POEM since this is an enhancement to the API.

JustinSGray:
@shamsheersc19 in the original Aitken code you wrote, you set the aitken stuff not to kick in before _iter_count >=2. What was the motivation for that delay? 

@joanibal  any idea why that is there?

joanibal:
As far as I can tell the algorithm for relaxation is taken directly from the Scalable Parallel paper as stated in the comments. 
[```
                # Compute relaxation factor. This method is used by Kenway et al. in
                # "Scalable Parallel Approach for High-Fidelity Steady-State Aero-
                # elastic Analysis and Adjoint Derivative Computations" (ln 22 of Algo 1)
```]

In the algorithm from that paper you need the change in "displacements" and "displacement increment" which requires 1 prior iteration (see line 19 through 21 in the listing). 

![image](https://user-images.githubusercontent.com/20216619/89910120-d78f0e00-dbbd-11ea-88bb-1599b1f1553b.PNG)

**Comment by JustinSGray** [comment]
[the iter_count starts at 0](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/solver.py#L159). So you need one backplane of data. That means you can start when the iter_count >= 1. 

If you're willing, Id appreciate you adding that change to your PR and making sure the tests still pass. If not, we'll sort it out in a different PR

**Comment by joanibal** [comment]
hmmm, that is odd. 

It is because in Algorithm 1 you are given u^(0) w^(0) and start at iteration 1, but in the implementation u^(0) and w^(0) are calculated on iteration 0?

**Comment by JustinSGray** [comment]
if I understand correctly, u^(0) w^(0) are just the initial state variable vectors? At the start of 0th iteration those values are just the initial guesses. At the end of the 0th iteration you have the computed values from each disciplines solve. 

So you could compute a delta based on the change from the initial guesses to the first computed values. But perhaps thats not ideal. Waiting till the next pass through might make sense then ... It would depend on the initial state of the output vectors.

**Comment by joanibal** [comment]
I think there is also a bug in [`_run_apply` of NLBGS](https://github.com/OpenMDAO/OpenMDAO/blob/2969e3dbdaf9bef486fa9d562a38a62648962070/openmdao/solvers/nonlinear/nonlinear_block_gs.py#L204)

I'm not certain, but it looks like someone was trying to alias `self._itercount` on line 210 
`        itercount = self._iter_count`
and then increment it on line 226
`            itercount += 1`

but this won't effect self._iter_count because it's type (int) isn't mutable, so it [can't be aliased like that](http://pythontutor.com/visualize.html#code=class%20solver%28%29%3A%0A%20%20%20%20def%20__init__%28self%29%3A%0A%20%20%20%20%20%20%20%20self._iter_count%20%3D%200%0A%0A%20%20%20%20def%20_run_apply%28self%29%3A%0A%20%20%20%20%20%20%20%20itercount%20%3D%20self._iter_count%0A%20%20%20%20%20%20%20%20itercount%20%2B%3D%201%20%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%20%20%20%20print%28self._iter_count,%20itercount%29%0A%20%20%20%20%20%20%20%20print%28itercount%20is%20self._iter_count%29%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%20%20%20%20%0Ags%20%3D%20solver%28%29%0Ags._run_apply%28%29%0A%0Aclass%20solver%28%29%3A%0A%20%20%20%20def%20__init__%28self%29%3A%0A%20%20%20%20%20%20%20%20self._iter_count%20%3D%20%5B0%5D%0A%0A%20%20%20%20def%20_run_apply%28self%29%3A%0A%20%20%20%20%20%20%20%20itercount%20%3D%20self._iter_count%0A%20%20%20%20%20%20%20%20itercount%20%2B%3D%20%5B1%5D%20%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%20%20%20%20print%28self._iter_count,%20itercount%29%0A%20%20%20%20%20%20%20%20print%28itercount%20is%20self._iter_count%29%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%20%20%20%20%0Ags%20%3D%20solver%28%29%0Ags._run_apply%28%29%0A%0A&cumulative=false&curInstr=0&heapPrimitives=nevernest&mode=display&origin=opt-frontend.js&py=3&rawInputLstJSON=%5B%5D&textReferences=false).

this would mean that the first time _single_iteration is called the iteration count should start at 1, since iteration 0 should(?) occur during set `_iter_initialize()`

**Comment by JustinSGray** [comment]
Im not sure I follow. The following works: 

```
>>> foo = 1 
>>> bar = foo 
>>> bar += 1
>>> bar 
2
```

which I believe is similar to what you're concerned about?

**Comment by joanibal** [comment]
yes but at the end foo = 1.

```

    def _run_apply(self):
        """
        Run the apply_nonlinear method on the system.
        """
        system = self._system()
        maxiter = self.options['maxiter']
        itercount = self._iter_count

        if self.options['use_apply_nonlinear'] or (itercount < 1 and maxiter < 2):

            # This option runs apply_nonlinear to calculate the residuals, and thus ends up
            # executing ExplicitComponents twice per iteration.

            self._recording_iter.push(('_run_apply', 0))
            try:
                system._apply_nonlinear()
            finally:
                self._recording_iter.pop()

        elif itercount < 1:
            # Run instead of calling apply, so that we don't "waste" the extra run. This also
            # further increments the iteration counter.
            itercount += 1
            outputs = system._outputs
            residuals = system._residuals

            with system._unscaled_context(outputs=[outputs]):
                outputs_n = outputs.asarray(copy=True)

            self._solver_info.append_subsolver()
            for isub, subsys in enumerate(system._subsystems_allprocs):
                system._transfer('nonlinear', 'fwd', isub)
                if subsys._is_local:
                    subsys._solve_nonlinear()

            self._solver_info.pop()
            with system._unscaled_context(residuals=[residuals]):
                residuals.set_val(outputs._data - outputs_n)

```

so in the above code the line itercount += 1 doesn't do anything, because itercount isn't used after the incrementation and it doesn't effect self._iter_count (which i think is the bug)

JustinSGray:
ah, I see. @Kenneth-T-Moore thoughts? sounds like a real bug to me... 

@joanibal you're welcome to fix this in your PR as well ... some kind of test would be nice, but this is tricky to check. Maybe  up a test with max_iter of 1 and the specifically look at teh value of the attribute in question.

joanibal:
So do you still think `_iter_count >=2` should be changed?

JustinSGray:
actually no. I think its safer as is. The addition of the option to change the starting relaxation should make this work correctly now.

joanibal:
I fixed the issue with the `itercount` and a similar issue in the Aitken's algorithm ( theta_n_1 wasn't getting updated properly) 

I also added two tests. One to check that the initial Aitken's factor was set correctly and one to check that the factor min and max options work correctly.
I also modified `test_feature_maxiter` to check that `_iter_count` matches `maxiter` when exiting after hitting the `maxiter`. 
Finally, in the tests that compared the value `_iter_count` to an expected value, I increased the expected value by 1.

PR #1592, opened 2020-08-05
Raise error if set_order in configure or after setup

### LINKED ISSUE #1572: Calling set_order within configure should raise an exception.
### Summary of Issue

Calling set_order from inside of configure currently results in a shape mismatch error later during transfers.

### Issue Type

- [x ] Bug
- [ ] Enhancement  

### Description

If a user calls set_order from inside of configure, or at any point after setup for a given system, an exception should be raised saying that's not allowed.

DKilkenny:
### Summary

We now raise an error if a user calls `set_order` in configure or after setup 

### Related Issues

- Resolves #1572

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/constants.py
+++ b/openmdao/core/constants.py
@@ -1,0 +1,11 @@
+"""File to define different setup status constants."""
+
+# PRE_SETUP: Newly initialized problem or newly added model.
+# POST_CONFIGURE: Configure has been called.
+# POST_SETUP: The `setup` method has been called, but vectors not initialized.
+# POST_FINAL_SETUP: The `final_setup` has been run, everything ready to run.
+
+PRE_SETUP = 0
+POST_CONFIGURE = 1
+POST_SETUP = 2
+POST_FINAL_SETUP = 3
--- a/openmdao/core/constants.py
+++ b/openmdao/core/constants.py
@@ -1,11 +1,25 @@
 """File to define different setup status constants."""
 
-# PRE_SETUP: Newly initialized problem or newly added model.
-# POST_CONFIGURE: Configure has been called.
-# POST_SETUP: The `setup` method has been called, but vectors not initialized.
-# POST_FINAL_SETUP: The `final_setup` has been run, everything ready to run.
+from enum import IntEnum
 
-PRE_SETUP = 0
-POST_CONFIGURE = 1
-POST_SETUP = 2
-POST_FINAL_SETUP = 3
+
+class _SetupStatus(IntEnum):
+    """
+    Class used to define different states of the setup status.
+
+    Attributes
+    ----------
+    PRE_SETUP : int
+        Newly initialized problem or newly added model.
+    POST_CONFIGURE : int
+        Configure has been called.
+    POST_SETUP : int
+        The `setup` method has been called, but vectors not initialized.
+    POST_FINAL_SETUP : int
+        The `final_setup` has been run, everything ready to run.
+    """
+
+    PRE_SETUP = 0
+    POST_CONFIGURE = 1
+    POST_SETUP = 2
+    POST_FINAL_SETUP = 3
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1896,6 +1896,10 @@ class Group(System):
         """
         # Make sure the new_order is valid. It must contain all subsystems
         # in this model.
+
+        if self._setup_procs_finished:
+            raise RuntimeError("%s: Cannot call set_order after setup" % (self.msginfo))
+
         newset = set(new_order)
         if self._static_mode:
             subsystems = self._static_subsystems_allprocs
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -122,7 +122,6 @@ class Group(System):
         self._raise_connection_errors = True
         self._config_finished = False
         self._order_set = False
-        self._ready_for_setup = True
 
         # TODO: we cannot set the solvers with property setters at the moment
         # because our lint check thinks that we are defining new attributes
@@ -347,6 +346,7 @@ class Group(System):
                 self.matrix_free = True
 
         self._config_finished = True
+        self._problem_meta['_setup_status'] = 1
         self.configure()
 
     def _setup_procs(self, pathname, comm, mode, prob_meta):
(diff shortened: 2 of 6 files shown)


openmdao/core/group.py
```diff
@@ -341,6 +344,7 @@ def _configure(self):
             if subsys.matrix_free:
                 self.matrix_free = True
 
+        self._problem_meta['_setup_status'] = 1
```
naylor-b, line 347: For consistency, problem metadata variables shouldn't begin with underscores, and use a constant instead of 1.

openmdao/core/group.py
```diff
@@ -1897,6 +1901,9 @@ def set_order(self, new_order):
         new_order : list of str
             List of system names in desired new execution order.
         """
+        if self._problem_meta is not None and self._problem_meta['_setup_status'] == 1:
```
naylor-b, line 1904: use constant instead of 1

openmdao/core/problem.py
```diff
@@ -172,11 +175,10 @@ def __init__(self, model=None, driver=None, comm=None, name=None, **options):
 
         self._initial_condition_cache = {}
 
-        # Status of the setup of _model.
-        # 0 -- Newly initialized problem or newly added model.
-        # 1 -- The `setup` method has been called, but vectors not initialized.
-        # 2 -- The `final_setup` has been run, everything ready to run.
-        self._setup_status = 0
+        self.PRE_SETUP = 0
```
naylor-b, line 178: I would not make these constants attributes of Problem because it'll cause circular imports if you try to use them in system.py for example.  Maybe we need a constants.py file that contains nothing but constants and then anybody could include that without any issues.

openmdao/core/problem.py
```diff
@@ -277,7 +279,7 @@ def is_local(self, name):
         bool
             True if the named system or variable is local to this process.
         """
-        if self._setup_status < 1:
+        if not hasattr(self, '_metadata'):
```
naylor-b, line 282: Would probably be better to set self._metadata = None in the `Problem.__init__` method and just check if it's None here.  I should have done that when I added it but I forgot.

openmdao/core/problem.py
```diff
@@ -902,7 +906,7 @@ def final_setup(self):
         else:
             mode = self._orig_mode
 
-        if self._setup_status < 2:
+        if self._metadata['_setup_status'] < self.POST_FINAL_SETUP:
```
naylor-b, line 909: Should we just go ahead and check here if setup status is 0, and if so call setup as well as final_setup instead of raising an exception if someone set the order after setup_procs (or changed any number of things that would also require a full setup)?

openmdao/devtools/debug.py
```diff
@@ -234,7 +234,7 @@ def config_summary(problem, stream=sys.stdout):
                          if s.nonlinear_solver is not None]
 
     max_depth = max([len(name.split('.')) for name in sysnames])
-    setup_done = problem._setup_status == 2
+    setup_done = model._problem_meta['_setup_status'] == 3
```
naylor-b, line 237: use constant instead of 3

openmdao/core/group.py
```diff
@@ -1930,6 +1937,10 @@ def set_order(self, new_order):
 
         subsystems[:] = [olddict[name] for name in new_order]
 
+        self._order_set = True
+        if self._problem_meta is not None:
+            self._problem_meta['_setup_status'] = 0
```
naylor-b, line 1942: use constant instead of 0

openmdao/core/problem.py
```diff
@@ -929,15 +933,23 @@ def final_setup(self):
...
-        if self._setup_status == 1:
+        if self._metadata['_setup_status'] == self.POST_SETUP:
             driver._setup_recording()
             self._setup_recording()
             record_viewer_data(self)
             record_system_options(self)
 
-        if self._setup_status < 2:
-            self._setup_status = 2
+        if hasattr(self.model, '_in_good_standing') and not self.model._in_good_standing:
```
naylor-b, line 948: I didn't see any mention anywhere else of '_in_good_standing'

openmdao/core/problem.py
```diff
@@ -96,11 +96,14 @@ class Problem(object):
...
         until they can be processed.
-    _setup_status : int
-        Current status of the setup in _model.
-        0 -- Newly initialized problem or newly added model.
-        1 -- The `setup` method has been called, but vectors not initialized.
-        2 -- The `final_setup` has been run, everything ready to run.
+    PRE_SETUP : int
+        Newly initialized problem or newly added model.
+    POST_CONFIGURE : int
+        Configure has been called.
```
Kenneth-T-Moore, line 102: This status is never used, and I am confused about just what it means from a problem perspective. @naylor-b

naylor-b: Yeah I agree there's a little confusion here, and I commented earlier that these constants should not be attributes of Problem.  I think for clarity we may want to differentiate between Problem setup and System setup, or possibly instead of system setup we should have a constant for POST_SETUP_PROCS instead.  That would be enough info to tell us when we're done with all of the system setup stuff (which all happens in setup_procs).

DKilkenny: So I've put the setup statuses in a constants file. Also, `POST_CONFIGURE` is used in group, I just wasn't referencing it as such because I didn't have `constants.py` at the time

openmdao/core/problem.py
DKilkenny: Leaving this as is per our conversation with Justin and keeping the raise error

openmdao/core/constants.py
```diff
@@ -0,0 +1,11 @@
...
+
+# PRE_SETUP: Newly initialized problem or newly added model.
+# POST_CONFIGURE: Configure has been called.
+# POST_SETUP: The `setup` method has been called, but vectors not initialized.
+# POST_FINAL_SETUP: The `final_setup` has been run, everything ready to run.
+
+PRE_SETUP = 0
+POST_CONFIGURE = 1
+POST_SETUP = 2
+POST_FINAL_SETUP = 3
```
swryan, line 11: did you consider using an `enum` (`IntEnum`)? 
https://docs.python.org/3/library/enum.html

seems like a better/cleaner solution?

openmdao/core/problem.py
```diff
@@ -869,15 +863,16 @@ def setup(self, check=False, logger=None, mode='auto', force_alloc_complex=False
             'connections': {},
             'remote_systems': {},
             'remote_vars': {},  # does not include distrib vars
-            'prom2abs': {'input': {}, 'output': {}}  # includes ALL promotes including buried ones
+            'prom2abs': {'input': {}, 'output': {}},  # includes ALL promotes including buried ones
+            '_setup_status': _SetupStatus.PRE_SETUP
```
naylor-b, line 867: can you get rid of the underscore on `_setup_status` to make it consistent with the other members of `_metadata`?

DKilkenny: Fixed

naylor-b, line 867: missed one

DKilkenny: Also fixed

PR #1598, opened 2020-08-07
Added check to allow connections to flat arrays

### LINKED ISSUE #1593: Allow sliced-indices in connections to flat arrays
### Summary of Issue

When connecting arrays via slice, it should be possible to connect some arbitrary slice to a flat array-valued target as long as the shapes match (C-ordering will be preserved).

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When connecting with slices, the requirement that the src_indices be the shape of the target cannot be satisfied if the target is a flat array.  In this case, as long as the target has the same size as the source, the transfer should be allowed, preserved C-ordering of the elements as given by the slicer.

### Example

Below are two examples, the later works, but the first one (using slices) does not.

```
    def test_connect_to_flat_array_with_slice(self):
        class SlicerComp(om.ExplicitComponent):
            def setup(self):
                self.add_input('x', np.ones((12,)))
                self.add_output('y', 1.0)

                outputs['y'] = np.sum(inputs['x']) ** 2.0

        p = om.Problem()

        p.model.add_subsystem('indep', om.IndepVarComp('x', arr_large_4x4))
        p.model.add_subsystem('row123_comp', SlicerComp())

        idxs = np.array([0, 2, 3], dtype=int)

        p.model.connect('indep.x', 'row123_comp.x', src_indices=om.slicer[idxs, ...])

        p.setup()
        p.run_model()

        assert_near_equal(p['row123_comp.x'], arr_large_4x4[(0, 2, 3), ...].ravel())
        assert_near_equal(p['row123_comp.y'], np.sum(arr_large_4x4[(0, 2, 3), ...]))

    def test_connect_to_flat_array(self):
        class SlicerComp(om.ExplicitComponent):
            def setup(self):
                self.add_input('x', np.ones((4,)))
                self.add_output('y', 1.0)

                outputs['y'] = np.sum(inputs['x'])

        p = om.Problem()

        p.model.add_subsystem('indep', om.IndepVarComp('x', val=arr_large_4x4))
        p.model.add_subsystem('trace_comp', SlicerComp())

        idxs = np.array([0, 5, 10, 15], dtype=int)

        p.model.connect('indep.x', 'trace_comp.x', src_indices=idxs, flat_src_indices=True)

        p.setup()
        p.run_model()

        assert_near_equal(p['trace_comp.x'], np.diag(arr_large_4x4))
        assert_near_equal(p['trace_comp.y'], np.sum(np.diag(arr_large_4x4)))
```

### Environment

OpenMDAO 3.2.0

DKilkenny:
### Summary

Added check to allow connections to flat arrays

### Related Issues

- Resolves #1593

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1420,10 +1420,7 @@ class Group(System):
                     if np.prod(src_indices.shape) == 0:
                         continue
 
-                    if is_slice and src_indices.size == np.prod(in_shape):
-                        flat_array_slice_check = True
-                    else:
-                        flat_array_slice_check = False
+                    flat_array_slice_check = is_slice and src_indices.size == np.prod(in_shape)
 
                     if any('flat_src_indices' in subsys._var_abs2meta[name]
                            for name in subsys._var_abs2meta):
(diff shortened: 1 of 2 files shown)


openmdao/core/group.py
```diff
@@ -1419,9 +1420,20 @@ def _setup_connections(self):
...
+                    if any('flat_src_indices' in subsys._var_abs2meta[name]
+                           for name in subsys._var_abs2meta):
+                        msg = ("%s: flat_src_indices has no effect when using om_slicer to "
+                               "slice array." % (self.msginfo))
+                        simple_warning(msg)
+
                     # initial dimensions of indices shape must be same shape as target
                     for idx_d, inp_d in zip(src_indices.shape, in_shape):
-                        if idx_d != inp_d:
+                        if idx_d != inp_d and not flat_array_slice_check:
```
Kenneth-T-Moore, line 1436: If `flat_array_slice_check` never changes in the loop, can't you pull the check outside the loop (all of these loops)? Since you aren't checking, you could just do a `continue` before that whole block of checks to move to the next index in the outer loop.

openmdao/core/group.py
```diff
@@ -1419,9 +1420,20 @@ def _setup_connections(self):
                     if np.prod(src_indices.shape) == 0:
                         continue
 
+                    if is_slice and src_indices.size == np.prod(in_shape):
```
Kenneth-T-Moore, line 1423: could do this in one line: `flat_array_slice_check = is_slice and src_indices.size == np.prod(in_shape)`

openmdao/core/group.py
Kenneth-T-Moore: Also, is the flag definition backwards? We check when `and *not* flat_array_slice_check` is True.  Sounds like when check is True, we don't check, unless i am missing something.

DKilkenny: We're checking to see if it's a slice and if the sizes match and if it's False, we skip the check

PR #1610, opened 2020-08-12
I1411 added more information to the node info window in the n2

### LINKED ISSUE #1411: Export more info to N2 to display in the node info panel
### Summary of Issue

Additional properties can be exported by n2_viewer.py so that they can be viewed in the node info panel.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

Potential new properties to export:
- Variable shapes
- Units
- Solver options
- Distributed
- Discrete

hschilling:
### Summary

Main goal was to add additional information to the node info window in the N2

- Variable shapes
- Units
- Solver options
- Distributed
- Discrete

In addition, the initial values for the variables is shown in the node info window. Depending on the size and number of dimensions of the variable, there can also be a button to let the user see the full value in a separate window. As opposed to the node info window, multiple value info windows can be displayed and they can be dragged around and re-sized. There is also a button to put the value of the variable onto the clipboard. This is formatted so that it can be pasted into Python code. 

The values of system options was added to the node info windows for systems.

Another big change was that now the node info window for the Solvers on the right-hand side display solver options, rather than just repeat what is shown by the node info window in the system hierarchy on the left. 

Some other minor changes:

- A small bug fix with the N2 UI toolbar tooltips
- Removed a small amount of code that supported Python 2
- Re-did a few of the view model data tests so that now the expected values are stored in JSON files rather than as strings in the test code. With the addition of the extra information in the view model data, the strings were getting too long to have in the code. 

### Related Issues

- Resolves #1411 

None

### New Dependencies

None


openmdao/visualization/n2_viewer/n2_viewer.py
```diff
@@ -116,33 +148,59 @@ def _get_tree_dict(system, component_execution_orders, component_execution_index
...
             if system.nonlinear_solver.SOLVER == NewtonSolver.SOLVER:
                 tree_dict['solve_subsystems'] = system._nonlinear_solver.options['solve_subsystems']
         else:
             tree_dict['nonlinear_solver'] = ""
+            tree_dict['nonlinear_solver_options'] = None
 
     tree_dict['children'] = children
 
+    options = {}
+    from openmdao.solvers.solver import Solver
```
Kenneth-T-Moore, line 195: Is this import used?

hschilling: no. left over from some code that used to be there. Will remove. Thanks

PR #1635, opened 2020-08-19
Fix for CaseRecorder overwriting model options if run driver called twice

### LINKED ISSUE #1488: Case recording of system options does not handle multiple calls to Problem.run_driver correctly
### Summary of Issue

If a case recorder is being used, each time run_driver is called, the system options for all the systems in the Problem are saved to the case recorder file. If the user changes the system options between the first and second run, for example, the context for the cases recorded on the first run will have been lost since the system options for the second run will have overwritten the system options of the first run.

### Issue Type

- [X] Bug
- [ ] Enhancement  

### Description

Each call to run_driver calls final_setup which ultimately calls `record_system_metadata` (aka system options) on all the recorders in the problem. This works fine if only one run_driver is done and also if no changes to the system options is made between runs. To get around the problem now of writing the system metadata twice, we use the SQL phrase "OR IGNORE" but this kind of masks the fact that for each run, OpenMDAO should really be keeping track of the fact that another run is starting and a snapshot of the current context of all the system options should be saved. 

Some ideas on how to fix this include:

Checking to see if system options are different from the previous run. If not, fine. 

If it did change, write a new set of system options records and associate it with the global_iteration counter of the case recorder file so we know when that set of system options came in to effect.

If you wanted to get the system options associated with a case, use the system options record with the highest global iteration counter less than the case counter.

### Example

Any code example where there are two runs done by calling run_driver with a change to one of the system options between the runs will show the problem.

### Environment

All OSes

DKilkenny:
### Summary

- Fixed the CaseRecorder overwriting model options if run driver called twice
- Added function `list_model_options` for user to see model options 
- Changed system_options to a private dictionary 

### Related Issues

- Resolves #1488 

None

### New Dependencies

None


openmdao/recorders/sqlite_recorder.py
```diff
@@ -622,16 +622,21 @@ def record_viewer_data(self, model_viewer_data, key='Driver'):
...
         Parameters
         ----------
         recording_requester : System
             The System that would like to record its metadata.
+        run_counter : int or None
+            The number of times run_driver or run_model has been called.
         """
         if self.connection:
+            # if run_counter is None:
+            #     run_counter = self._counter
```
swryan, line 638: delete comment?

openmdao/recorders/base_case_reader.py
```diff
@@ -45,11 +59,11 @@ def system_metadata(self):
         Returns
         -------
         dict
-            reference to the 'system_options' attribute.
+            reference to the '_system_options' attribute.
         """
         warn_deprecation("The BaseCaseReader.system_metadata attribute is deprecated. "
                          "Use the BaseCaseReader.system_option attribute instead.")
```
swryan, line 65: this should say to use `list_model_options` instead of the _other_ deprecated attribute ;)

openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -2857,6 +2857,19 @@ def test_system_metadata_attribute_deprecated(self):
...
+        model.add_recorder(self.recorder)
+        prob = om.Problem(model)
+        prob.setup()
+        prob.run_model()
+        prob.cleanup()
+
+        cr = om.CaseReader(self.filename)
+        msg = "The system_options attribute is deprecated. Use `list_model_options` instead."
+        with assert_warning(DeprecationWarning, msg):
+            options = cr.system_options
```
swryan, line 2871: also need to document the new function, and remove the section for the deprecated attribute here:
http://openmdao.org/twodocs/versions/3.2.1/features/recording/reading_metadata.html?highlight=system_options

swryan (approved):

PR #1636, opened 2020-08-19
Added logic to stop error being raised if promotes * is used and no matches are found

### LINKED ISSUE #1602: `promotes=['*']` should not raise if no matches are found
### Summary of Issue

If the user specifies ['*'] as the glob pattern for promotes_inputs, and there no inputs to the model (e.g. it's an IndepVarComp), an error should not be raised.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Asterisk as a qualifier generally means "match zero or more", but as currently implemented, `promotes_inputs=['*']` fails if the system has zero inputs.  

While generally a user should know if their subsystem has inputs or not, there are use-cases where a user's system might incorporate submodels in a modular fashion, and might want to promote all inputs/outputs from those models.  The '*' glob pattern provides a covevnent way of doing that, but the current behavior means that the ['*'] pattern raises an error if the submodel is just an IndepVarComp.

### Example

The following test case should pass when the behavior is fixed:

```
    def test_promote_glob_no_inputs(self):
        p = om.Problem()
        p.model.add_subsystem('indep',
                              om.IndepVarComp('x', 2.0),
                              promotes_inputs=['*'],
                              promotes_outputs=['x'])
        p.model.add_subsystem('C1', om.ExecComp('y=x'), promotes_inputs=['x'], promotes_outputs=['y'])
        p.setup()
        p.run_model()
        self.assertEqual(p.get_val('x'), p.get_val('y'))
```

### Environment

OpenMDAO 3.2.1

DKilkenny:
### Summary

Added logic to stop error being raised if promotes * is used and no matches are found

### Related Issues

- Resolves #1602 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1954,9 +1954,10 @@ class System(object):
                         if fnmatchcase(i, p):
                             break
                     else:
-                        raise RuntimeError("%s: '%s' failed to find any matches for the following "
-                                           "pattern: '%s'.%s" %
-                                           (self.msginfo, call, p, empty_group_msg))
+                        if names and p != '*':
+                            raise RuntimeError("%s: '%s' failed to find any matches for the following "
+                                            "pattern: '%s'.%s" %
+                                            (self.msginfo, call, p, empty_group_msg))
                     if p == patterns[-1]:
                         break
                 else:
--- a/openmdao/core/tests/test_indep_var_comp.py
+++ b/openmdao/core/tests/test_indep_var_comp.py
@@ -67,6 +67,17 @@ class TestIndepVarComp(unittest.TestCase):
         assert_near_equal(prob.get_val('indep_var_1'), 1.0)
         assert_near_equal(prob.get_val('indep_var_2'), 2.0)
 
+    def test_promote_glob_no_inputs(self):
+        p = om.Problem()
+        p.model.add_subsystem('indep',
+                              om.IndepVarComp('x', 2.0),
+                              promotes_inputs=['*'],
+                              promotes_outputs=['x'])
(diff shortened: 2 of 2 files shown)


openmdao/core/system.py
```diff
@@ -1954,9 +1954,10 @@ def resolve(to_match, io_types, matches, proms):
                         if fnmatchcase(i, p):
                             break
                     else:
-                        raise RuntimeError("%s: '%s' failed to find any matches for the following "
-                                           "pattern: '%s'.%s" %
-                                           (self.msginfo, call, p, empty_group_msg))
+                        if names and p != '*':
```
naylor-b, line 1957: This logic will skip the error message for any pattern if names is empty.  I think the issue stated that it should only skip the error message if the pattern is '*'.

openmdao/core/tests/test_connections_in_configure.py
```diff
@@ -244,13 +244,7 @@ def test_queue_subcomponent_io_in_configure(self):
...
         p.model.add_subsystem('g', subsys=g)
 
-        with self.assertRaises(RuntimeError) as e:
-            p.setup()
-
-        err_msg = str(e.exception)
-        expected = "Cuber (g.cuber): 'promotes_inputs' failed to find any matches for the " \
-                   "following pattern: '*'."
-        self.assertEqual(err_msg, expected)
+        p.setup()
```
naylor-b, line 247: The comment for this test says it will fail, so either update the comment or maybe even get rid of this test.

DKilkenny: Good catch. Fixed

PR #1650, opened 2020-08-25
Raise UserWarning with location of AnalysisError

### LINKED ISSUE #1634: Drivers should not silently swallow AnalysisError.
### Summary of Issue

A driver should be verbose when it detects an AnalysisError.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Encountering an AnalysisError can degrade performance, but since these errors are often automatically handled by OpenMDAO, it can be difficult to tell that they are occuring.  Using SNOPT, for instance, just prints out a `D` status flag indicating that the line search size is being changed.  But in complex models, the user needs a better way to find the location of the AnalysisError.

For instance, while MetaModelStructured captures this AnalysisError and prints out a description of the problem, this information is absorbed by some drivers (pyOptSparseDriver with SNOPT is one) so the problem is hidden from the user.

When raised, AnalysisError should be provided with a message detailing why it is raised (like the one output in MetaModelStructured when it is asked to extrapolate).

The message should include information about where in the model the error was raised.

### Example

The following code users SNOPT which handles the AnalysisError silently:

```
import numpy as np
import openmdao.api as om

x_tr = np.linspace(0, 2*np.pi, 100)
y_tr = np.sin(x_tr)

p = om.Problem(model=om.Group())

p.driver = om.pyOptSparseDriver(optimizer='SNOPT')
p.driver.opt_settings['iSumm'] = 6

mm = om.MetaModelStructuredComp(extrapolate=False)
mm.add_input('x', val=1.0, training_data=x_tr)
mm.add_output('y', val=1.0, training_data=y_tr)
p.model.add_subsystem('interp', mm, promotes_inputs=['x'], promotes_outputs=['y'])

p.model.add_objective('y', scaler=-1)
p.model.add_design_var('x', lower=6, upper=10)

p.setup()

p.set_val('x', 0.75)
p.run_driver()
```

Setting `extrapolate=True` will cause the code to run successfully, there's nothing in that output stream to indicate what the problem is.

### Environment

N/A

DKilkenny:
### Summary

Raise UserWarning with location of AnalysisError 

### Related Issues

- Resolves #1634

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/tests/test_meta_model_structured_comp.py
+++ b/openmdao/components/tests/test_meta_model_structured_comp.py
@@ -1100,7 +1100,7 @@ class TestMetaModelStructuredPython(unittest.TestCase):
 
       p = om.Problem(model=om.Group())
 
-      p.driver = om.pyOptSparseDriver(optimizer=OPT)
+      p.driver = om.pyOptSparseDriver(optimizer=OPTIMIZER)
 
       mm = om.MetaModelStructuredComp(extrapolate=False)
       mm.add_input('x', val=1.0, training_data=x_tr)
--- a/openmdao/components/tests/test_meta_model_structured_comp.py
+++ b/openmdao/components/tests/test_meta_model_structured_comp.py
@@ -17,7 +17,7 @@ try:
 except ImportError:
     scipy_gte_019 = False
 
-OPT, OPTIMIZER = set_pyoptsparse_opt('SLSQP')
+OPT, OPTIMIZER = set_pyoptsparse_opt('SNOPT')
 if OPTIMIZER:
     from openmdao.drivers.pyoptsparse_driver import pyOptSparseDriver
 
--- a/openmdao/components/tests/test_meta_model_structured_comp.py
+++ b/openmdao/components/tests/test_meta_model_structured_comp.py
@@ -17,6 +17,8 @@ try:
 except ImportError:
     scipy_gte_019 = False
 
+# check that pyoptsparse is installed
+# if it is, try to use SNOPT but fall back to SLSQP
 OPT, OPTIMIZER = set_pyoptsparse_opt('SNOPT')
 if OPTIMIZER:
     from openmdao.drivers.pyoptsparse_driver import pyOptSparseDriver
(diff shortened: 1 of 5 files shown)


openmdao/core/analysis_error.py
```diff
@@ -11,4 +13,17 @@ class AnalysisError(Exception):
     code or a subsolver.
     """
 
-    pass
+    def __init__(self, error, location=None):
+        """
+        Instantiate DemuxComp and populate private members.
```
Kenneth-T-Moore, line 18: I think this is a cut and paste error.

openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -2047,6 +2049,32 @@ def test_ParOpt_basic(self):
...
+
+        p.model.add_objective('y', scaler=-1)
+        p.model.add_design_var('x', lower=6, upper=10)
+
+        p.set_solver_print(level=0)
+        p.setup()
+
+        p.set_val('x', 0.75)
+
+        msg = "Analysis Error: Line 205 of file {}".format(inspect.getsourcefile(MetaModelStructuredComp))
```
Kenneth-T-Moore, line 2074: This test should be in the metamodel tests, rather than in pyoptsparse/snopt.

DKilkenny: fixed

openmdao/components/tests/test_meta_model_structured_comp.py
```diff
@@ -1086,6 +1087,32 @@ def test_training_gradient_akima_basic(self):
 
         self.run_and_check_derivs(prob)
 
+    def test_snopt_analysis_error(self):
```
Kenneth-T-Moore, line 1090: remove "snopt_" from title.

DKilkenny: Changed

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -439,7 +439,12 @@ def run(self):
             self.set_design_var(name, dv_dict[name])
 
         with RecordingDebugging(self._get_name(), self.iter_count, self) as rec:
-            model.run_solve_nonlinear()
+            try:
+                model.run_solve_nonlinear()
+            except AnalysisError:
+                model._clear_iprint()
+                fail = 1
```
Kenneth-T-Moore, line 446: We don't need the `fail = 1` here.

PR #1659, opened 2020-08-28
Added support for distributed design variables in get_design_vars

### LINKED ISSUE #1631: Driver has access to metadata of distributed I/O
### Summary of Issue

Add support for distributed dvs to group `get_design_vars` method. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Current behavior reports the size of the DV on the root proc to ALL procs. This is correct for non-distributed dvs, but incorrect for distributed one. 

For distributed dvs, the local dv size should be given. 

### Example

```
from pprint import pprint
import numpy as np
import openmdao.api as om
from mpi4py import MPI

 

class DistribParaboloid(om.ExplicitComponent):

    def setup(self):
        self.options['distributed'] = True

        if self.comm.rank == 0:
            ndvs = 3
        else:
            ndvs = 2

        self.add_input('w', val=1.) # this will connect to a non-distributed IVC
        self.add_input('x', shape=ndvs) # this will connect to a distributed IVC

        self.add_output('y', shape=1) # all-gathered output, duplicated on all procs
        self.add_output('z', shape=ndvs) # distributed output
        self.declare_partials('y', 'x')
        self.declare_partials('y', 'w')
        self.declare_partials('z', 'x')

        x = inputs['x']
        local_y = np.sum((x-5)**2)
        y_g = np.zeros(self.comm.size)
        self.comm.Allgather(local_y, y_g)
        outputs['y'] = np.sum(y_g) + (inputs['w']-10)**2
        outputs['z'] = x**2

    def compute_partials(self, inputs, J):
        x = inputs['x']
        J['y', 'x'] = 2*(x-5)
        J['y', 'w'] = 2*(inputs['w']-10)
        J['z', 'x'] = np.diag(2*x)
 

if __name__ == "__main__":
    comm = MPI.COMM_WORLD

    p = om.Problem()
    d_ivc = p.model.add_subsystem('distrib_ivc',
                                   om.IndepVarComp(distributed=True),
                                   promotes=['*'])
    if comm.rank == 0:
        ndvs = 3
    else:
        ndvs = 2
    d_ivc.add_output('x', 2*np.ones(ndvs))

    ivc = p.model.add_subsystem('ivc',
                                om.IndepVarComp(distributed=False),
                                promotes=['*'])
    ivc.add_output('w', 2.0)
    p.model.add_subsystem('dp', DistribParaboloid(), promotes=['*'])

 

    p.model.add_design_var('x', lower=-100, upper=100)
    p.model.add_objective('y')
    p.setup()
    p.run_model()
    # p.model.list_outputs(print_arrays=True)
    # p.check_totals(of=['y'], wrt=['x'])
    # J = p.compute_totals(of=['y'], wrt=['x'])
    # pprint(J)

 

    # Check the local size of the design variables on each proc
    dvs = p.model.get_design_vars()
    for name, meta in dvs.items():
        model_size = dvs[name]['size']
        # print("Local model dv size = {0}".format(model_size))
        
        assert(model_size == ndvs)
        
```

DKilkenny:
### Summary

Added support for distributed dvs in get_design_vars

### Related Issues

- Resolves #1631

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -2940,7 +2940,10 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
 
                 if 'size' not in meta:
                     if src_name in abs2idx:
-                        meta['size'] = sizes[owning_rank[src_name], abs2idx[src_name]]
+                        if out[abs_name]['distributed']:
+                            meta['size'] = sizes[self.comm.rank, abs2idx[src_name]]
```
naylor-b, line 2944: For a distributed variable, you should just be able to set `meta['size'] = abs2meta[name]['global_size']`.

openmdao/core/system.py
```diff
@@ -2967,10 +2970,12 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
                     out.update(dvs)
 
             if self.comm.size > 1 and self._subsystems_allprocs:
-                allouts = self.comm.allgather(out)
-                out = OrderedDict()
-                for all_out in allouts:
-                    out.update(all_out)
+                for name in out:
```
naylor-b, line 2973: I think with the change above to the 'size' for distributed vars, the old loop here should work fine.  The new loop has a number of problems.  Try reverting to the old one now that the distrib size will be fixed and see if it works.

openmdao/core/system.py
DKilkenny: Global size doesn't own the size at that rank. Eg. three procs might have values of `3,2,2` respectively and global size holds `7` in that case. Also, not all abs2meta has `'size'`.

naylor-b: I was thinking that because drivers (at least all of our current ones) use the full distributed variable instead of just the local part of it, that the 'size' of a design var/response would just be set to the global size, but it looks like we label size and global_size separately for desvars/responses just like we do with input/output vector metadata.  I guess that makes sense because maybe there could be some sort of distributed driver someday that only used the local part.  I'm curious where you saw an abs2meta entry that didn't have 'size'.  They should all have that entry since abs2meta doesn't contain any discrete variables.

DKilkenny: You were right, I was using `name` when I should've been using `src_name`. Changed

openmdao/core/system.py
naylor-b, line 2973: This loop is still a problem.  'out' is the dict that contains dv metadata for ALL dvs, but you're allgathering it for every non-distributed variable.  Instead, you should only do it once like the original code did.  You're basically doing the exact same thing that the old code did except you're doing it multiple times.

DKilkenny: Yeah I was just about to change that back. I forgot to fix that in the new commit

openmdao/core/system.py
```diff
@@ -2940,7 +2940,10 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
 
                 if 'size' not in meta:
                     if src_name in abs2idx:
-                        meta['size'] = sizes[owning_rank[src_name], abs2idx[src_name]]
+                        if meta['distributed']:
+                            meta['size'] = abs2meta[src_name]['global_size']
```
naylor-b, line 2944: Ken pointed me to justin's issue that is the basis for this PR, and based on that, this line should probably be `meta['size'] = sizes[model.comm.rank, abs2idx[src_name]]`

naylor-b (changes requested):

openmdao/core/tests/test_distribcomp.py
```diff
@@ -342,6 +342,77 @@ def test_distrib_idx_in_full_out(self):
...
+        p.model.add_design_var('x', lower=-100, upper=100)
+        p.model.add_objective('y')
+        p.setup()
+        p.run_model()
+
+
+        # Check the local size of the design variables on each proc
+        dvs = p.model.get_design_vars()
+        for name, meta in dvs.items():
+            model_size = dvs[name]['size']
```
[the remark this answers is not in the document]
DKilkenny: Bret and I discussed and remembered we currently don't support distributed inputs connected to auto_ivcs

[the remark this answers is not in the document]
DKilkenny: Bret and I discussed and remembered we currently don't support distributed inputs connected to auto_ivcs

swryan (approved):

naylor-b (approved):

PR #1662, opened 2020-09-02
Differential Evolution driver

wright:
### Summary

This is a Differential Evolution variant of the Simple Genetic Algorithm driver. The DifferentialEvolutionDriver has advantages and disadvantages when compared to the SimpleGADriver:

Pros

- DifferentialEvolutionDriver is typically about 3 times faster than SimpleGADriver

- DifferentialEvolutionDriver is usually more accurate SimpleGADriver because it does not limit the number of bits available to represent inputs

- DifferentialEvolutionDriver does not require the user to manually specify a number of representation bits

Cons

- DifferentialEvolutionDriver only supports continuous input variables; SimpleGADriver also supports discrete

- DifferentialEvolutionDriver does not support SimpleGADriver’s “compute_pareto” option for multi-objective optimization

### Related Issues

None

None

### New Dependencies

None


openmdao/drivers/tests/test_differential_evolution_driver.py
```diff
@@ -0,0 +1,1052 @@
...
+try:
+    from openmdao.vectors.petsc_vector import PETScVector
+except ImportError:
+    PETScVector = None
+
+extra_prints = False  # enable printing results
+
+class TestDifferentialEvolution(unittest.TestCase):
+    def setUp(self):
+        import openmdao.api as om
```
Kenneth-T-Moore, line 28: You shouldn't need any imports in the setUp function here or in the other TestCase classes.

openmdao/drivers/tests/test_differential_evolution_driver.py
```diff
@@ -0,0 +1,1052 @@
...
+        import numpy as np
+        import os
+        os.environ['DifferentialEvolutionDriver_seed'] = '11'  # make RNG repeatable
+
+    def test_rastrigin(self):
+        ORDER = 6  # dimension of problem
```
Kenneth-T-Moore, line 34: This test is in the feature docs, so you do need to add the imports here.

openmdao/drivers/tests/test_differential_evolution_driver.py
```diff
@@ -0,0 +1,1052 @@
...
+        span = 5   # upper and lower limits
+        extra_prints = False
```
Kenneth-T-Moore, line 36: Get rid of all the "extra prints" stuff in the examples that are in the docs.  It will just confuse people.  (I think it is just this example.  The other one doesn't have them.)

openmdao/drivers/differential_evolution_driver.py
```diff
@@ -0,0 +1,593 @@
...
+                    c = rng.integers(0, self.npop)
+
+                # randomly select chromosome index for crossover
+                r = rng.integers(0, self.lchrom)
+
+                # crossover and mutation
+                population[ii] = parentPop[ii]  # start the same as parent
+                # clip mutant so that it cannot be outside the bounds
+                mutant = np.clip(parentPop[a] + F * (parentPop[b] - parentPop[c]), vlb, vub)
+                for j in range(self.lchrom):
```
Kenneth-T-Moore, line 588: Nested for-loops are slow in python.  I think you could speed this up with some numpy vectorized calls.  Off the top of my head, something like this:

rr = rng.random(self.lchrom)
idx = np.where(rr < Pc)
population[ii][idx] = mutant[idx]
population[ii][r] = mutant[r]

vectorizing the 'ii' loop might also be possible, but it looks like a difficult task, and the resulting code might not be very readable.

Kenneth-T-Moore (changes requested):

openmdao/drivers/tests/test_differential_evolution_driver.py
wright: Imports seem to be required in setup() for the tests that are included in the docs using embed-code. I will delete all the others.

Kenneth-T-Moore: You can put them in the actual test.  That way the imports also show up in the docs.

Kenneth-T-Moore (approved):

openmdao/docs/features/building_blocks/drivers/differential_evolution.rst
```diff
@@ -0,0 +1,76 @@
...
+optimizer supports only continuous variables.
+The DifferentialEvolutionDriver supports both constrained and unconstrained optimization.
+
+The DifferentialEvolutionDriver has advantages and disadvantages when compared to the SimpleGADriver:
+
+* Pros
+
+  * DifferentialEvolutionDriver is typically about 3 times faster than SimpleGADriver
+
+  * DifferentialEvolutionDriver is usually more accurate SimpleGADriver because it does not limit the number of bits available to represent inputs
```
swryan, line 19: more accurate *than* SimpleGaDriver

swryan (approved):

PR #1671, opened 2020-09-08
Implementation for POEM_022 - Determining variable shape at runtime based on connections

naylor-b:
### Summary

Implementation PR for POEM_022.  Adds the ability to determine variable shapes at runtime based on what those variables are connected to.  Also adds a new function hook, `setup_partials`, that is used to allow a component to compute its partials after all shape information is known.

None

### New Dependencies

None


openmdao/core/tests/test_dyn_sizing.py
```diff
@@ -0,0 +1,717 @@
...
+                               promotes_inputs=['*'])
+
+
+class TestDynShapes(unittest.TestCase):
+    def test_baseline_series(self):
+        # this is just a sized source and unsized sink, and we put a DynShapeGroupSeries in between them
+        p = om.Problem()
+        indep = p.model.add_subsystem('indep', om.IndepVarComp('x1', val=np.ones((2,3))))
+        indep.add_output('x2', val=np.ones((4,2)))
+        p.model.add_subsystem('Gdyn', DynShapeGroupSeries(3, 2, DynShapeComp))
```
Kenneth-T-Moore, line 439: Have you tested what happens if you accidentally set `shape_by_conn` to your own name?  I imagine you will just get a 'cannot resolve shape' exception, so probably fine.

Kenneth-T-Moore (approved):

openmdao/core/tests/test_dyn_sizing.py
naylor-b: Yeah, doing a `copy_shape` using your own name will either result in a failure to resolve, or if that variable happens to be connected to other variables by some other means (shape_by_conn or a resolvable variable connects to it by its own shape_by_conn or copy_shape) then it'll just work anyway.  Your comment also reminded me that I didn't have a test for when copy_shape is set to a non-existent variable, so I added that.

joanibal:
Wow! This is a really nice implementation. Thank you for all the work that went into this. 

I have just one suggestion. I think it would make it much easier for users to debug their models if they had some graphical representation of the size dependency graph. If there are unresolved variables the error message lists the variables that are an issue in alphabetical order, but doesn't say how they are connected.  

With a picture to go along with the error message it is much easier to see the connections and understand why the test `test_cycle_unresolved` doesn't work.
shape_dependency_graph



 This is something that could be done only if the model fails like this...
```python 
        if unresolved:
            unresolved = sorted(unresolved)
            #try to save a plot of the dependency graph for debuging
            msg = f"{self.msginfo}: Failed to resolve shapes for {unresolved}."
            try:
                import matplotlib.pyplot as plt
                nx.draw(graph, with_labels=True)
                plt.savefig('shape_dependency_graph.png')
                msg += " See shape_dependency_graph.png for more details"
            except ImportError:
                pass
            except err as e:
                raise err

            raise RuntimeError(msg)
```
However it is probably better to a have a utility function for this so users could verify that the dependencies are as expected even without an error. I'm not sure how to best implement that, but it sounds possible.

naylor-b:
@joanibal I added a 'view_dyn_shapes' command to the openmdao command line tool that will display the shape dependency graph using matplotlib, basically like you showed in your example.  At some point when I have more time I'd like to replace the matplotlib stuff with a better graph viewing library, but for now, it's at least better than nothing.

robfalck (approved):

openmdao/core/system.py
```diff
@@ -1346,6 +1364,7 @@ def _setup_global_shapes(self):
         """
         meta = self._var_allprocs_abs2meta
         loc_meta = self._var_abs2meta
+        myrank = self.comm.rank
```
swryan, line 1367: not used?

openmdao/core/tests/test_dyn_sizing.py
```diff
@@ -0,0 +1,730 @@
...
+            self.add_input('in', shape=1, src_indices=np.arange(0,1, dtype=int))
+        elif self.comm.rank == 1:
+            self.add_input('in', shape=2, src_indices=np.arange(1,3, dtype=int))
+        else:
+            self.add_input('in', shape=0, src_indices=np.arange(3,3, dtype=int))
+
+        self.add_output('out', shape=3)
+
+    def compute(self, inputs, outputs):
+        outputs['out'] *= self.comm.rank
```
swryan, line 144: this is strange, could you explain what's going on here (computing output based on itself)?

openmdao/docs/advanced_guide/derivs/partial_derivs_explicit.rst
```diff
@@ -22,7 +22,7 @@ Here is an example, based on the :ref:`Betz Limit Example <betz_limit_tutorial>`
 
 
 The calls to :code:`declare_partials` tell OpenMDAO which partial derivatives to expect.
-This is always done inside the :code:`setup` method.
+This should always be done inside the :code:`setup_partials` method.
```
swryan, line 25: `should` or `must`?
if `should` then more explanation is probably required...
if `must` then we do we need a deprecation or exception when it's done the old way?

openmdao/core/component.py
```diff
@@ -385,16 +405,16 @@ def _update_subjac_sparsity(self, sparsity):
                     # add sparsity info to existing partial info
                     self._subjacs_info[abs_key]['sparsity'] = tup
 
-    def add_input(self, name, val=1.0, shape=None, src_indices=None, flat_src_indices=None,
-                  units=None, desc='', tags=None):
+    def add_input(self, name, val=None, shape=None, src_indices=None, flat_src_indices=None,
+                  units=None, desc='', tags=None, shape_by_conn=False, copy_shape=None):
```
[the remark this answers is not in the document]
naylor-b: I reverted this back to 1.0.  So now, val is allowed to be set but it must be a scalar, and once the shape is known, that scalar is used to fill the array.

openmdao/core/component.py
```diff
@@ -333,10 +331,32 @@ def _setup_partials(self):
...
 
+    def setup_partials(self):
+        """
+        Declare partials.
+
+        This is meant to be overridden by component classes.  All partials should be
+        declared here since this is called after all size/shape information is known for
+        all variables.
+        """
+        pass
```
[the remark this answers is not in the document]
naylor-b: It's not backwardly incompatible.  Doing it in the old way still works, although we recommend that people do it now in setup_partials for consistency sake.

openmdao/core/system.py

openmdao/core/tests/test_dyn_sizing.py
naylor-b: I updated this so the output now depends upon the input.

openmdao/docs/advanced_guide/derivs/partial_derivs_explicit.rst
naylor-b: I added more explanation about why setup_partials is the recommended way.

swryan (approved):

PR #1677, opened 2020-09-10
I1673 add support for Nan values in the N2 diagram

### LINKED ISSUE #1673: Problem with N2 generation
### Summary of Issue

N2 generation fails if one input has been declared with `numpy.nan` as default value

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

We have use cases where we want to avoid that the user of an OpenMDAO problem uses default values without knowing it. In such cases, we use `numpy.nan` as default value in the input declaration. If the user forgets to provide a value, he is aware of that because the process crashes instantly.
(BTW, if there is a better way to achieve this, please let me know)

It works fine, except that since OpenMDAO 3.3, the generation of N2-diagram produces a blank page if there at least one `numpy.nan` as default value.

### Example
```python

import numpy as np
import openmdao.api as om


class SomeComponent(om.ExplicitComponent):
    def setup(self):
        self.add_input("foo", val=np.nan).  # replacing np.nan by 1.0 makes the N2 generation work again
        self.add_output("bar")

    def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
        outputs["bar"] = inputs["foo"]


problem = om.Problem()

ivc = om.IndepVarComp()
ivc.add_output("foo", 1.0)

problem.model.add_subsystem("ivc", ivc, promotes=["*"])
problem.model.add_subsystem("comp", SomeComponent(), promotes=["*"])

problem.setup()
om.n2(problem)
```

### Environment

Operating System: macOS 10.15.6

Python environment: Python 3.7.7

Packages: OpenMDAO 3.3.0

hschilling:
### Summary

The N2 diagram did not handle the case where initial values had Nan values in them.

Also did some re-factoring of related JavaScript code to reduce code duplication.

### Related Issues

- Resolves #1673

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/n2_viewer.py
+++ b/openmdao/visualization/n2_viewer/n2_viewer.py
@@ -34,6 +34,20 @@ _IND = 4  # HTML indentation (spaces)
 
 _MAX_ARRAY_SIZE_FOR_REPR_VAL = 1000  # If var has more elements than this do not pass to N2
 
+def _convert_nans_in_nested_list(l):
+    for i, val in enumerate(l):
+        if isinstance(val, list):
+            _convert_nans_in_nested_list(val)
+        else:
+            if np.isnan(val):
+                l[i] = "nan"
+            else:
+                l[i] = val
+
+def _handle_nans(a):
+    l = a.tolist()
+    converted = _convert_nans_in_nested_list(l)
+    return(l)
 
 def _get_var_dict(system, typ, name):
     if name in system._var_discrete[typ]:
@@ -77,7 +91,7 @@ def _get_var_dict(system, typ, name):
             var_dict['value'] = type(meta['value']).__name__
     else:
         if meta['value'].size < _MAX_ARRAY_SIZE_FOR_REPR_VAL:
-            var_dict['value'] = meta['value']
+            var_dict['value'] = _handle_nans(meta['value'])
         else:
             var_dict['value'] = None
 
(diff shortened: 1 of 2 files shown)


openmdao/visualization/n2_viewer/n2_viewer.py
```diff
@@ -77,7 +94,7 @@ def _get_var_dict(system, typ, name):
             var_dict['value'] = type(meta['value']).__name__
     else:
         if meta['value'].size < _MAX_ARRAY_SIZE_FOR_REPR_VAL:
-            var_dict['value'] = meta['value']
+            var_dict['value'] = _convert_to_support_nans_in_json(meta['value'])
         else:
```
Kenneth-T-Moore, line 98: I think you can probably add a test for this if calling `n2` on a prob with NaN formerly raised an exception.  Just do `n2(p, show_browser=False)` so that it doesn't open the browser or write a file.

hschilling: Great point. I had that in my head as a to do yesterday and then forgot today. Will add

Kenneth-T-Moore (approved):

openmdao/visualization/n2_viewer/src/N2UserInterface.js
```diff
@@ -35,7 +35,37 @@ class InfoPropYesNo extends InfoPropDefault {
...
+function element_to_string(element){
+    let val_string;
+    if (typeof element === 'number'){
+       if (Number.isInteger(element)) {
+          val_string = element.toString();
+       } else { /* float */
+          val_string = val_float_formatter(element);
+       }
+    } else {
+       val_string = stringify_with_support_for_nan(element);
```
swryan, line 65: this function is only used in this one place
since it's a simple if/else, I think you could just inline it and avoid a function call

naylor-b (approved):

openmdao/visualization/n2_viewer/n2_viewer.py
swryan: similar to other comment..
I think you could replace this single use function with:
`var_dict['value'] = _convert_nans_in_nested_list(val.tolist())`

swryan (changes requested):

swryan (approved):

PR #1678, opened 2020-09-11
Added declaration for distributed design variables to the `supports` dictionary. 

### LINKED ISSUE #1630: Explicit Driver interface to declare support for distributed I/O
### Summary of Issue

Allow drivers to declare if they support distributed I/O or not

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

No current drivers support this. So we want to be able to throw an error if distributed variables are added as either design vars, or quantities of interest. 

Add a `supports_distributed` descriptor to the list of things drivers can support and then mark all existing drivers as `False`

Kenneth-T-Moore:
### Summary

Added declaration for distributed design variables to the `supports` dictionary. No current drivers support this feature, so an error is raised if any distributed design variables are added.

### Related Issues

- Resolves #1630

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -176,6 +176,7 @@ class Driver(object):
         self.supports.declare('active_set', types=bool, default=False)
         self.supports.declare('simultaneous_derivatives', types=bool, default=False)
         self.supports.declare('total_jac_sparsity', types=bool, default=False)
+        self.supports.declare('distributed_design_vars', types=bool, default=False)
 
         self.iter_count = 0
         self.cite = ""
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -176,7 +176,7 @@ class Driver(object):
         self.supports.declare('active_set', types=bool, default=False)
         self.supports.declare('simultaneous_derivatives', types=bool, default=False)
         self.supports.declare('total_jac_sparsity', types=bool, default=False)
-        self.supports.declare('distributed_design_vars', types=bool, default=False)
+        self.supports.declare('distributed_design_vars', types=bool, default=True)
 
         self.iter_count = 0
         self.cite = ""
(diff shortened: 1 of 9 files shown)


openmdao/core/driver.py
```diff
@@ -176,6 +176,7 @@ def __init__(self, **kwargs):
         self.supports.declare('active_set', types=bool, default=False)
         self.supports.declare('simultaneous_derivatives', types=bool, default=False)
         self.supports.declare('total_jac_sparsity', types=bool, default=False)
+        self.supports.declare('distributed_design_vars', types=bool, default=True)
```
swryan, line 179: why default=True?  seems like default=False is safer and would not require all the existing drivers to set it

Kenneth-T-Moore: I set that to True when I thought some of our distributed tests were testing derivs wrt distributed variables, but it turns out those were all remote. I will set this back to False (but keeping it in all the sub drivers for completeness.

swryan (approved):

openmdao/core/driver.py
```diff
@@ -288,6 +289,31 @@ def _setup_driver(self, problem):
...
         responses = prom2ivc_src_dict(self._responses)
 
+        # Only allow distributed design variables on drivers that support it.
+        if self.supports['distributed_design_vars'] is False:
+            dist_vars = []
+            for dv, meta in self._designvars.items():
+
+                # For Auto-ivcs, we need to check the distributed metadata on the target instead.
+                if meta['ivc_source'].startswith('_auto_ivc.'):
+                    abs_name = model._var_allprocs_prom2abs_list['input'][dv][0]
```
naylor-b, line 299: Should we loop over these and set dist = True if any of them are distributed?  It seems unlikely but possible  that you could have an auto_ivc connected to multiple inputs where only some subset of them are distributed.

naylor-b (changes requested):

openmdao/core/driver.py
Kenneth-T-Moore: I guess we allow a connection like that, so yeah: I need to loop over each one in the list and check.

Kenneth-T-Moore: done

naylor-b (approved):

PR #1680, opened 2020-09-15
Added check to show or hide auto_ivc

### LINKED ISSUE #1666: Make auto_ivc-related warning optional.
### Summary of Issue

Warnings like the example below should not be displayed as part of a normal run but instead should be moved to an optional configuration check accessible via `openmdao check` or `problem.setup(check=???)`.  The error message should also be reworded to be more clear, e.g.,  "Group '' call to set_input_defaults( 'XXY', val=1.0) overrides lower level call from 'static_analysis.RotorGroup'" .

An example of the current warning is:
"UserWarning:Groups '' and 'static_analysis.RotorGroup' called set_input_defaults for the input 'diam_{hub}' with conflicting 'value'. The value (1.0) from '' will be used."

### Issue Type

- [x] Bug
- [ ] Enhancement

DKilkenny:
### Summary

Added check to show or hide auto_ivc

### Related Issues

- Resolves #1666

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_group.py
+++ b/openmdao/core/tests/test_group.py
@@ -2640,6 +2640,19 @@ class TestGroupAddInput(unittest.TestCase):
 
         self.assertEqual(cm.exception.args[0], "Group (<model>): The subsystems G1 and par.G4 called set_input_defaults for promoted input 'x' with conflicting values for 'units'. Call <group>.set_input_defaults('x', units=?), where <group> is the model to remove the ambiguity.")
 
+    def test_auto_ivc_warning(self):
+        p = self._make_tree_model()
+        model = p.model
+        g2 = model._get_subsystem('G1.G2')
+        g2.set_input_defaults('x', units='km')
+
+        g1 = model._get_subsystem('G1')
+        g1.set_input_defaults('x', units='inch', val=2.)
+
+        msg = "Groups 'G1' and 'G1.G2' called set_input_defaults for the input 'x' with conflicting 'units'. The value (inch) from 'G1' will be used."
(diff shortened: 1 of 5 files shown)


openmdao/error_checking/check_config.py
```diff
@@ -361,6 +361,20 @@ def _check_comp_has_no_outputs(problem, logger):
...
+    Issue a logger warning if any components have conflicting attributes.
+
+    Parameters
+    ----------
+    problem : <Problem>
+        The problem being checked.
+    """
+    if hasattr(problem.model, "_auto_ivc_warnings"):
+        for i in problem.model._auto_ivc_warnings:
+            simple_warning(i)
```
naylor-b, line 375: instead of using a simple_warning here, you should follow what the other check functions do, which is logger.warning(...)

openmdao/core/tests/test_group.py
```diff
@@ -2651,8 +2651,9 @@ def test_override_units(self):
         g1.set_input_defaults('x', units='inch', val=2.)
 
         msg = "Groups 'G1' and 'G1.G2' called set_input_defaults for the input 'x' with conflicting 'units'. The value (inch) from 'G1' will be used."
+        p.setup(check=True)
         with assert_warning(UserWarning, msg):
```
naylor-b, line 2655: After changing the check to use the logger (see other comment), you'll have to change this test.  To see other examples, search the source for 'TestLogger'.

naylor-b (changes requested):

naylor-b (approved):

Kenneth-T-Moore (approved):

PR #1689, opened 2020-09-17
Fix issue #1498 with case recording on a Driver with parallel models

### LINKED ISSUE #1498: Case recording on a DOEDriver with a parallel model does not record cases for all remote inputs.

swryan:
### Summary

Fixes issue with case recording on a DOEDriver with a parallel model where it did not record all remote variables.

NOTE: this was a bug for driver recording in general, not specific to DOEDriver

Also fixes an issue when setting `run_parallel=False` and fills out the test suite with combinations of the `run_parallel`  and `models_per_proc` options for a parallel model.

Comments in the tests were updated to help clarify the options being tested.

### Related Issues

- Resolves #1498

None

### New Dependencies

None


Kenneth-T-Moore (approved):
Discussed with Steve. Looks good.

naylor-b (approved):

PR #1690, opened 2020-09-17
add equivalent unit conversion check and tests

### LINKED ISSUE #1687: Promoting variables with different but equivalent units errors due to ambiguity
### Summary of Issue

Variables promoted with equivalent but different units (such as 'm/s/s' vs. 'm/s**2' will raise an exception without set_input_defaults.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

In `Group._resolve_ambiguous_input_meta`, the units check will flag different but equivalent units (such as 'W' vs. 'J/s') as being different, even though they are identical.

Instead of checking unit equivalence based on strings, it should probably be done by using the conversion factor.
This can be done as an O(n) operation by comparing the units of tgts[0] to tgts[1:], rather than comparing every possible combination among the targets.

### Example

The following test triggers a RuntimeError because the units on 'x' ('m/s/s' vs 'm/s**2') and 'z' ('W' vs 'J/s') are not the same, even though these units are equivalent.

```
    def test_promotes_equivalent_units(self):
        # multiple Group.set_input_defaults calls at same tree level with conflicting units args

        p = om.Problem()

        g1 = p.model.add_subsystem("G1", om.Group(), promotes_inputs=['x'])
        g1.add_subsystem("C1", om.ExecComp("y = 2. * x * z",
                                            x={'value': 5.0, 'units': 'm/s/s'},
                                            y={'value': 1.0, 'units': None},
                                            z={'value': 1.0, 'units': 'W'}),
                                            promotes_inputs=['x', 'z'])
        g1.add_subsystem("C2", om.ExecComp("y = 3. * x * z",
                                            x={'value': 5.0, 'units': 'm/s**2'},
                                            y={'value': 1.0, 'units': None},
                                            z={'value': 1.0, 'units': 'J/s'}),
                                            promotes_inputs=['x', 'z'])
        p.setup()
```

### Environment

OpenMDAO `ffc8a16f505267bb55ccb66d2511c320364b3bcb` (Wed Sep 16 14:46:15 2020 -0400)

wright:
### Summary

Variables promoted with equivalent but different units (such as 'm/s/s' vs. 'm/s**2' will raise an exception without set_input_defaults. Before raising the exception, make sure the units not equivalent.

### Related Issues

- Resolves #1687

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -25,7 +25,7 @@ from openmdao.utils.array_utils import convert_neg, array_connection_compatible,
     _flatten_src_indices
 from openmdao.utils.general_utils import ContainsAll, all_ancestors, simple_warning, \
     common_subpath, conditional_error, _is_slicer_op, _slice_indices
-from openmdao.utils.units import is_compatible, unit_conversion, _has_val_mismatch
+from openmdao.utils.units import is_compatible, unit_conversion, _has_val_mismatch, _find_unit
 from openmdao.utils.mpi import MPI, check_mpi_exceptions, multi_proc_exception_check
 from openmdao.utils.coloring import Coloring, _STD_COLORING_FNAME
 import openmdao.utils.coloring as coloring_mod
@@ -3196,12 +3196,7 @@ class Group(System):
                     tmeta = abs2meta[tgt] if tgt in abs2meta else all_abs2meta[tgt]
                     tunits = tmeta['units'] if 'units' in tmeta else None
                     if 'units' not in gmeta and sunits != tunits:
-                        try:
-                            test_conv = unit_conversion(sunits, tunits)
-                            if abs(test_conv[0] - 1.0) > 1e-9 or abs(test_conv[1] - 0.0) > 1e-9:
-                                errs.add('units')
-                                metadata.add('units')
-                        except TypeError:
+                        if _find_unit(sunits) != _find_unit(tunits):
                             errs.add('units')
                             metadata.add('units')
                     if 'value' not in gmeta:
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore (approved):

openmdao/core/group.py
```diff
@@ -3196,8 +3196,14 @@ def _resolve_ambiguous_input_meta(self):
                     tmeta = abs2meta[tgt] if tgt in abs2meta else all_abs2meta[tgt]
                     tunits = tmeta['units'] if 'units' in tmeta else None
                     if 'units' not in gmeta and sunits != tunits:
-                        errs.add('units')
-                        metadata.add('units')
+                        try:
+                            test_conv = unit_conversion(sunits, tunits)
```
naylor-b, line 3200: Since PhysicalUnit objects understand if they're equivalent or not, we could do this a little more cleanly:

```
if _find_unit(sunits) != _find_unit(tunits):
   errs.add('units')
   metadata.add('units')
```

_find_unit is found in openmdao.utils.units

openmdao/core/tests/test_units.py
```diff
@@ -880,6 +880,43 @@ def setup(self):
...
+                                            y={'value': 1.0, 'units': None},
+                                            z={'value': 1.0, 'units': 'W'}),
+                                            promotes_inputs=['x', 'z'])
+        g1.add_subsystem("C2", om.ExecComp("y = 3. * x * z",
+                                            x={'value': 5.0, 'units': 'm/s**2'},
+                                            y={'value': 1.0, 'units': None},
+                                            z={'value': 1.0, 'units': 'J/s'}),
+                                            promotes_inputs=['x', 'z'])
+        # trying to convert J/s/s to m/s**2 should cause Incompatible units TypeError exception
+        with self.assertRaises(TypeError):
```
naylor-b, line 917: We should check the contents of the error message here as well.

naylor-b (changes requested):

wright:
The requested changes have been made.

naylor-b (approved):

PR #1699, opened 2020-09-25
I1629 be smarter about sparse derivatives for ExecComps with multiple expressions

### LINKED ISSUE #1629: Improve ExecComp derivative sparsity for multiple expressions
ExecComp declares all partials as dense if has_diag_partials is False, but in cases where the ExecComp is created using multiple expressions, it's possible that some outputs may not depend on all of the inputs.  It's easy enough to determine if a given output does not depend on a particular input, and in that case we shouldn't declare those partials.

hschilling:
### Summary

Only declare and compute partials when an output really depends on the input. So, for example,

`ExecComp(['y1=2.0*x1+1.', 'y2=3.0*x2-1.'],x1=1.0, x2=2.0)`

only declare and compute `('y1','x1')` and `('y2','x2')`.

Also, removed some unneeded IndepVarComps in the ExecComp tests.

### Related Issues

- Resolves #1629 

None

### New Dependencies

None


openmdao/components/exec_comp.py
```diff
@@ -461,8 +465,9 @@ def compute_partials(self, inputs, partials):
                 self._residuals.set_val(0.0)
                 self.compute(pwrap, uwrap)
 
-                for u in out_names:
-                    partials[(u, input)] = imag(uwrap[u] * inv_stepsize).flat
+                for out, inp in self._declared_partials.keys():
```
naylor-b, line 468: It seems more efficient to keep the `for u in out_names` loop and just add a check for `if (u, input) in self._declared_partials: ...` rather than looping over all declared partials for each input and updating partials when the inputs match.

hschilling: Good idea. Changed.

openmdao/components/exec_comp.py
```diff
@@ -477,9 +482,10 @@ def compute_partials(self, inputs, partials):
                     self._residuals.set_val(0.0)
                     self.compute(pwrap, uwrap)
 
-                    for u in out_names:
-                        # set the column in the Jacobian entry
-                        partials[(u, input)][:, i] = imag(uwrap[u] * inv_stepsize).flat
+                    for out, inp in self._declared_partials.keys():
```
naylor-b, line 485: I forgot to mention this loop, which has the same issue as the one I mentioned earlier.

hschilling: Done

naylor-b (approved):

Kenneth-T-Moore (approved):

PR #1704, opened 2020-09-28
Added check for unitless connections

### LINKED ISSUE #1696: Units of `None` should be treated as truly unitless and not a special case.
### Summary of Issue

Units of `None` are treated somewhat differently than dimensionless units when doing comparisons for connection or promotion.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When programmatically creating units, a user might set units using something like:

```
units = f'1/s*{time_units}'
```

While this units string (`'1/s*s'`) reduces to `1` and is unitless, it induces a warning when connecting to an input with units of None.  None is handled through a separate logic branch, but it should be treated as just another set of units (where the exponents of all base units are 0).

### Example

The following test raises because two UserWarnings are generated:

```
    def test_connect_unitless_to_none(self):
        import warnings
        p = om.Problem()
        ivc = p.model.add_subsystem('indeps', om.IndepVarComp())
        ivc.add_output('x', val=5.0, units='1/s*s')
        ivc.add_output('y', val=10.0, units='Hz*s')
        p.model.add_subsystem('exec_comp', om.ExecComp('z = x + y', z={'units': None},
                                                       x={'units': None}, y={'units': None}))
        p.model.connect('indeps.x', 'exec_comp.x')
        p.model.connect('indeps.y', 'exec_comp.y')

        with warnings.catch_warnings():
            warnings.simplefilter("error")
            p.setup()

        p.run_model()
        assert_near_equal(p.get_val('exec_comp.z'), 15.0)
```

and for promotion:

```
    def test_promote_unitless_and_none(self):
        import warnings
        p = om.Problem()
        ivc = p.model.add_subsystem('indeps', om.IndepVarComp(), promotes_outputs=['x', 'y'])
        ivc.add_output('x', val=5.0, units='1/s*s')
        ivc.add_output('y', val=10.0, units='Hz*s')
        p.model.add_subsystem('exec_comp', om.ExecComp('z = x + y', z={'units': None},
                                                       x={'units': None}, y={'units': None}),
                              promotes_inputs=['x', 'y'])

        with warnings.catch_warnings():
            warnings.simplefilter("error")
            p.setup()

        p.run_model()
        assert_near_equal(p.get_val('exec_comp.z'), 15.0)
```

### Environment

OpenMDAO 3.3.1-dev

DKilkenny:
### Summary

Added a check which of the base units of the input/output units for "unitless" units when connecting to None

### Related Issues

- Resolves #1696

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1661,13 +1661,14 @@ class Group(System):
             in_units = allprocs_abs2meta_in[abs_in]['units']
 
             if out_units:
-                unit = _find_unit(out_units)
-                base_powers = any(i >= 1 for i in unit._powers)
-                if not in_units and base_powers:
+                out_unit_meta = _find_unit(out_units) if out_units else None
+                out_unit_base_powers = any(i >= 1 for i in out_unit_meta._powers)
+
+                if not in_units and out_unit_base_powers:
                     msg = f"{self.msginfo}: Output '{abs_out}' with units of '{out_units}' " + \
                           f"is connected to input '{abs_in}' which has no units."
                     simple_warning(msg)
-                elif base_powers and not is_compatible(in_units, out_units):
+                elif out_unit_base_powers and not is_compatible(in_units, out_units):
                     msg = f"{self.msginfo}: Output units of '{out_units}' for '{abs_out}' " + \
                           f"are incompatible with input units of '{in_units}' for '{abs_in}'."
                     if self._raise_connection_errors:
@@ -1675,9 +1676,12 @@ class Group(System):
                     else:
                         simple_warning(msg)
             elif in_units is not None:
-                msg = f"{self.msginfo}: Input '{abs_in}' with units of '{in_units}' is " + \
-                      f"connected to output '{abs_out}' which has no units."
-                simple_warning(msg)
+                in_unit_meta = _find_unit(in_units) if in_units else None
+                in_unit_base_powers = any(i >= 1 for i in in_unit_meta._powers)
+                if in_unit_base_powers:
+                    msg = f"{self.msginfo}: Input '{abs_in}' with units of '{in_units}' is " + \
+                        f"connected to output '{abs_out}' which has no units."
+                    simple_warning(msg)
 
             fail = False
 
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1674,7 +1674,7 @@ class Group(System):
                         raise RuntimeError(msg)
                     else:
                         simple_warning(msg)
-            elif in_units is not None:
+            elif not in_units:
                 if not _is_unitless(in_units):
                     msg = f"{self.msginfo}: Input '{abs_in}' with units of '{in_units}' is " + \
                         f"connected to output '{abs_out}' which has no units."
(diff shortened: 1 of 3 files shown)


openmdao/core/group.py
```diff
@@ -1661,21 +1661,27 @@ def _setup_connections(self):
             in_units = allprocs_abs2meta_in[abs_in]['units']
 
             if out_units:
-                if not in_units:
+                out_unit_meta = _find_unit(out_units) if out_units else None
+                out_unit_base_powers = any(i >= 1 for i in out_unit_meta._powers)
```
swryan, line 1665: since this logic relies on underlying implementation of units, I think it might be better to implement a `is_unitless(units)` function in the units module that does this.. it would also eliminate the code duplication

DKilkenny: Fixed

openmdao/core/group.py
```diff
@@ -1661,21 +1662,22 @@ def _setup_connections(self):
             in_units = allprocs_abs2meta_in[abs_in]['units']
 
             if out_units:
-                if not in_units:
+                if not in_units and _is_unitless(out_units):
                     msg = f"{self.msginfo}: Output '{abs_out}' with units of '{out_units}' " + \
                           f"is connected to input '{abs_in}' which has no units."
                     simple_warning(msg)
-                elif not is_compatible(in_units, out_units):
+                elif _is_unitless(out_units) and not is_compatible(in_units, out_units):
```
naylor-b, line 1669: Don't we still need a check for the case where out_units is NOT unitless but is not compatible with in_units?  I'm not sure why the _is_unitless check is even necessary here.  Shouldn't is_compatible handle this?

openmdao/utils/units.py
```diff
@@ -845,6 +845,11 @@ def _update_library(cfg):
 _UNIT_CACHE = {}
 
 
+def _is_unitless(units):
```
naylor-b, line 848: I don't think this gives the right answer.  For me it returns True for 'ft/s'.   Also, PhysicalUnit already has an is_dimensionless() method, so you should just be able to use that,  `return unit_meta.is_dimensionless()`

naylor-b (changes requested):

openmdao/utils/units.py
DKilkenny: Fixed

openmdao/utils/units.py
```diff
@@ -845,6 +845,11 @@ def _update_library(cfg):
...
+    unit_meta = _find_unit(units) if units else None
+    return not unit_meta.is_dimensionless()
```
naylor-b, line 850: Shouldn't this be `return unit_meta.is_dimensionless()` instead of `return not unit_meta.is_dimensionless()`?  Also,  you're not handling the case where unit_meta ends up being None.

naylor-b (changes requested):

openmdao/utils/units.py
DKilkenny: I removed the chance of it being None since it's private and protected by a none check in each instance. Fixed the `not` issue.

openmdao/utils/units.py
```diff
@@ -845,6 +845,11 @@ def _update_library(cfg):
 _UNIT_CACHE = {}
 
 
+def _is_unitless(units):
+    unit_meta = _find_unit(units)
```
naylor-b, line 849: _find_unit can return None if units are not valid, so maybe just to be safe, `return unit_meta is not None and unit_meta.is_dimensionless()`

naylor-b (changes requested):

naylor-b (approved):

swryan (approved):

PR #1709, opened 2020-09-30
Permit src_indices for an input in both group.promotes and group.connect provided that both are flat.

### LINKED ISSUE #1708: OpenMDAO should allow src_indices to be used on both a promote and a connection to the same input provided that the are both flat src_indices.
### Summary of Issue

OpenMDAO should allow src_indices to be used on both a promote and a connection to the same input provided that they are both flat src_indices.

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Motivation for this is a context where the `promotes` uses src_indices as a broadcast and the `connect` uses src_indices as a reduction, but it should work in all cases. It is possible for this to work for non-flat indices, but it may prove more difficult. Flat to flat is a simple indexing of one src_indices into the other.

Raise an error if either end is not flat.

### Example

see dymos: test_trajectory_connection.py

### Environment

<For bugs, otherwise N/A>

Kenneth-T-Moore:
### Summary

Permit src_indices for an input in both group.promotes and group.connect provided that both are flat.  A new cascaded src_indices is computed for the connection.

### Related Issues

- Resolves #1708

None

### New Dependencies

None


swryan (approved):

naylor-b (commented):
Are we really sure that we want to allow users to do this?  If we happen to see src_indices specified in multiple places, how can we be sure of the intent?  Did they intend to do this sort of nested src_indices thing, or did they just make a mistake?

Kenneth-T-Moore:
I don't know. There is some risk, but the risk stems from the user not knowing what they are connecting to, and if you are using src_indices, you should know the shape of your target.

Alternative is to make the user create an adapter component.

@robfalck @JustinSGray

naylor-b (commented):
Or maybe add another metadata entry like 'allow_nested_src_indices' or something so we know the intent.

Kenneth-T-Moore:
After discussion, we determined a better way to do this inside a dymos model.

PR #1711, opened 2020-10-01
Changed localized error msg

### LINKED ISSUE #1703: Rework of localized error msg string
### Summary of Issue

Current msg looks like: `SetWar (war)`

Reformat to be more clear. 

`'war' <type SetWar>`

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM # N/A 

### Description

The localization is there to help users figure out where their error is coming from. Including both class and instance name is helpful, but new users may not recognize the ordering of the current message. 
 
It would be more clear to specify which part is a class name, and which is an instance.

DKilkenny:
### Summary

Changed from `IndepVarComp (comp)` to `'comp' <class IndepVarComp>`

### Related Issues

- Resolves #1703

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -473,9 +473,9 @@ def msginfo(self):
         if self.pathname == '':
             return '{} (<model>)'.format(type(self).__name__)
```
naylor-b, line 474: Shouldn't this one also change to match the format of the others?  so  `<model> <class {}>`

swryan (approved):

naylor-b (approved):

PR #1715, opened 2020-10-01
Tweak to balance comp for complex safeness

JustinSGray:
### Summary

I noticed that the balance comp was barking in a check_partials call on a user model. This was scary, since we rely heavily on the balance. The derivs were fine, but the complex-check was bad because we used numpy absolute value, instead of our complex safe one. 

I borrowed the cs safe complex definition from the exec-comp to fix the problem


robfalck (changes requested):
This should be applied to EQConstraintComp as well.

JustinSGray:
@robfalck I added the change to EqConstraintComp too, but that forced me to make a separate module to hold the helper function... cause copying the helper function into three separate files seemed dumb.

Kenneth-T-Moore (approved):

robfalck (approved):

PR #1718, opened 2020-10-02
Added linear_constraint arg to check_totals and print formatting

### LINKED ISSUE #1700: check_totals should show linear constraints
### Summary of Issue

Add linear constraints to check_totals output

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM https://github.com/OpenMDAO/POEMs/blob/master/POEM_033.md

### Description

Add linear constraints to the check_totals output 

### Example

See POEM 033; 

When debugging with outside user, I was very confused by the output of `check_totals`. The constraints were reported by the `list_problem_vars` method, but didn't show up in `check_totals`. It took me a minute to figure out why. 

I can't think of a good reason to skip them in the check. They could be wrong, even if they are linear. The only potential problem might be the number of them. Hence it might be costly. So we can make it optiona, but default to True.

DKilkenny:
### Summary

Added linear_constraint arg to check_totals and print formatting

### Related Issues

- Resolves #1700

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -1528,7 +1528,8 @@ class Problem(object):
             out_stream = sys.stdout
 
         _assemble_derivative_data(data, rel_err_tol, abs_err_tol, out_stream, compact_print,
-                                  [model], {'': fd_args}, totals=True)
+                                  [model], {'': fd_args}, totals=True,
+                                  linear_constraints=self.driver._cons)
         return data['']
 
     def compute_totals(self, of=None, wrt=None, return_format='flat_dict', debug_print=False,
@@ -1824,7 +1825,7 @@ class Problem(object):
 
 def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out_stream,
                               compact_print, system_list, global_options, totals=False,
-                              indep_key=None, print_reverse=False,
+                              linear_constraints=None, indep_key=None, print_reverse=False,
                               show_only_incorrect=False):
     """
     Compute the relative and absolute errors in the given derivatives and print to the out_stream.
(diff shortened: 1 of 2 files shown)


openmdao/core/problem.py
```diff
@@ -1824,7 +1825,7 @@ def check_config(self, logger=None, checks=None, out_file='openmdao_checks.out')
 
 def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out_stream,
                               compact_print, system_list, global_options, totals=False,
-                              indep_key=None, print_reverse=False,
+                              linear_constraints=None, indep_key=None, print_reverse=False,
```
robfalck, line 1828: linear_constraints isn't in the doc string here.  It's a bit confusing.  Above you pass it self.driver._cons, does that include the nonlinears as well?

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2237,6 +2237,40 @@ def test_set_method_and_step_bug(self):
...
 
+    def test_check_totals_linear_constraint(self):
+        prob = om.Problem()
+        prob.model = SellarDerivatives()
+        prob.model.nonlinear_solver = om.NonlinearBlockGS()
+
+        prob.model.add_design_var('x', lower=-100, upper=100)
+        prob.model.add_design_var('z', lower=-100, upper=100)
+        prob.model.add_objective('obj')
+        prob.model.add_constraint('con1', upper=0.0)
```
robfalck, line 2248: should we be testing a linear constraint here?  Perhaps change the upper bound of z to be a linear constraint?

robfalck (changes requested):

Kenneth-T-Moore:
This PR wasn't meant for review.

PR #1725, opened 2020-10-05
added support for colons in variable names in ExecComp

fzahle:
### Summary

ExecComp does not support colons in variable names, but the rest of the framework does. This commit ports this functionality from OpenMDAO1.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -44,6 +44,7 @@ def _valid_name(s, exprs):
             return n
         i += 1
 
+
 def check_option(option, value):
     """
     Check option for validity.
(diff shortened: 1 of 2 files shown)


JustinSGray:
can you add a test that has an exec-comp with 2 separate equations, and the same input variable in both of them? What about an expression like `foo:bar = tree:leaf_count + tree:leaf_count`? Will the duplication of the input be handled correctly in either case?

fzahle:
colon variables of course have the drawback that array sizes can't be initialised in the instantiation since they aren't valid python names:

    om.ExecComp('foo:bar=x:bar+1.', x:bar=2.0)

but can of course be set through a dictionary. I just converted some code from OpenMDAO1 and came across this difference. Also there's a few numpy functions that aren't supported that I've added. I'll push that to another PR.

PR #1728, opened 2020-10-07
Added progress outputs for check totals

### LINKED ISSUE #1719: User will get some feedback on the progress of check_totals
### Summary of Issue

check_totals can be slow, especially for large models. We should give some indication of progress so they don't give up! 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM # N/A 

### Description

We should print to screen a status update on the check_totals progress as it goes. 

Before starting a check, print `1/45; Checking derivatives with respect to: <var name>[index]` ... 
Log the start time. Then after that check has finished print the time 

so final print should look like 

`1/45; Checking derivatives with respect to: <var name>[index]  ... 30 seconds`

The 1/45 part should represent the number relative to the total number of finite difference calls that need to be made.

DKilkenny:
### Summary

Added progress outputs for check totals

### Related Issues

- Resolves #1719

None

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -34,6 +34,8 @@ class ApproximationScheme(object):
         Array of sizes of data chunks that make up _j_colored. (Used for MPI Allgatherv)
     _j_data_offsets : ndarray of int
         Array of offsets of each data chunk that makes up _j_colored. (Used for MPI Allgatherv)
+    _progress_out : None
+        Attribute to output the progress of check_totals
     """
 
     def __init__(self):
@@ -47,6 +49,7 @@ class ApproximationScheme(object):
         self._j_data_offsets = None
         self._approx_groups_cached_under_cs = False
         self._exec_dict = defaultdict(list)
+        self._progress_out = None
 
     def __repr__(self):
         """
@@ -297,15 +300,7 @@ class ApproximationScheme(object):
     def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
 
-        if hasattr(system._approx_schemes, "out_stream"):
-            out_stream = system._approx_schemes.out_stream
-        else:
-            out_stream = None
-
-        if hasattr(system._approx_schemes, "_show_progress"):
-            show_progress = system._approx_schemes._show_progress
-        else:
-            show_progress = None
+        self._progress_out = self.out_stream if hasattr(self, "out_stream") else None
 
         # Set system flag that we're under approximation to true
         system._set_approx_mode(True)
@@ -379,7 +374,8 @@ class ApproximationScheme(object):
 
         # now do uncolored solves
         for wrt, data, col_idxs, tmpJ, idx_info, nz_rows in approx_groups:
-            start_time = time.time()
+            if self._progress_out:
+                start_time = time.time()
 
             J = tmpJ[wrt]
             full_idxs = J['loc_outvec_idxs']
@@ -406,11 +402,11 @@ class ApproximationScheme(object):
                     else:
                         J['data'][:, i_count] = self._transform_result(result[full_idxs])
 
-                end_time = time.time()
-                if show_progress is not None and out_stream is not None:
-                    out_stream.write(f"{fd_count+1}/{len(full_idxs)}: Checking "
-                                     f"derivatives with respect to: '{wrt} [{idxs}]' ... "
-                                     f"{round(end_time-start_time, 4)} seconds\n")
+                if self._progress_out:
+                    end_time = time.time()
+                    self.out_stream.write(f"{fd_count+1}/{len(full_idxs)}: Checking "
+                                          f"derivatives with respect to: '{wrt} [{idxs}]' ... "
+                                          f"{round(end_time-start_time, 4)} seconds\n")
 
                 fd_count += 1
 
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -35,7 +35,7 @@ class ApproximationScheme(object):
         Array of sizes of data chunks that make up _j_colored. (Used for MPI Allgatherv)
     _j_data_offsets : ndarray of int
         Array of offsets of each data chunk that makes up _j_colored. (Used for MPI Allgatherv)
-    _progress_out : None
+    _progress_out : None or file-like object
         Attribute to output the progress of check_totals
     """
 
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -296,11 +296,10 @@ class ApproximationScheme(object):
 
             self._approx_groups.append((wrt, data, in_idx, tmpJ, [(arr, in_idx)], None))
 
-    def _compute_approximations(self, system, jac, total, under_cs, out_stream=_DEFAULT_OUT_STREAM):
+    def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
 
-        if out_stream == _DEFAULT_OUT_STREAM:
-            out_stream = sys.stdout
+        out_stream = system._approx_schemes.out_stream
 
         # Set system flag that we're under approximation to true
         system._set_approx_mode(True)
@@ -402,7 +401,7 @@ class ApproximationScheme(object):
                         J['data'][:, i_count] = self._transform_result(result[full_idxs])
 
                 end_time = time.time()
-                if system._show_progress:
+                if system._approx_schemes._show_progress and out_stream is not None:
                     out_stream.write(f"{fd_count+1}/{len(full_idxs)}: Checking "
                                      f"derivatives with respect to: '{wrt} [{idxs}]' ... "
                                      f"{round(end_time-start_time, 4)} seconds\n")
--- a/openmdao/approximation_schemes/approximation_scheme.py
+++ b/openmdao/approximation_schemes/approximation_scheme.py
@@ -7,6 +7,7 @@ import numpy as np
 
 from openmdao.utils.array_utils import sub2full_indices, get_input_idx_split
 import openmdao.utils.coloring as coloring_mod
+from openmdao.utils.general_utils import _convert_auto_ivc_to_conn_name
 from openmdao.utils.mpi import MPI
 from openmdao.jacobians.jacobian import Jacobian
 from openmdao.vectors.vector import _full_slice
@@ -409,9 +410,11 @@ class ApproximationScheme(object):
 
                 if self._progress_out:
                     end_time = time.time()
+                    prom_name = _convert_auto_ivc_to_conn_name(system._conn_global_abs_in2out, wrt)
                     self._progress_out.write(f"{fd_count+1}/{len(full_idxs)}: Checking "
-                                          f"derivatives with respect to: '{wrt} [{idxs}]' ... "
-                                          f"{round(end_time-start_time, 4)} seconds\n")
+                                             f"derivatives with respect to: "
+                                             f"'{prom_name} [{idxs}]' ... "
+                                             f"{round(end_time-start_time, 4)} seconds\n")
 
                 fd_count += 1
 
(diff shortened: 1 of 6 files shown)


openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -296,6 +297,16 @@ def _init_approximations(self, system):
     def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
 
+        if hasattr(system._approx_schemes, "out_stream"):
```
Kenneth-T-Moore, line 300: I think we can do this with 1 attribute instead of 2. Something like `self._progress_out`, which is either None (do nothing), or a stream (print the progress to the stream.)

Kenneth-T-Moore: Note, this also needs to be declared in the __init__ , with default value None.  Also, needs to be in the class docstring.

openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -368,6 +379,8 @@ def _compute_approximations(self, system, jac, total, under_cs):
 
         # now do uncolored solves
         for wrt, data, col_idxs, tmpJ, idx_info, nz_rows in approx_groups:
+            start_time = time.time()
```
Kenneth-T-Moore, line 382: Let's not compute the time unless there is a progress_stream.

openmdao/core/total_jac.py
```diff
@@ -1494,9 +1499,16 @@ def compute_totals_approx(self, initialize=False):
             if model._coloring_info['coloring'] is not None:
                 model._update_wrt_matches(model._coloring_info)
 
+        if show_progress:
+            setattr(model._approx_schemes, "out_stream", out_stream)
```
Kenneth-T-Moore, line 1503: You can set directly without setattr.  e.g., `model._approx_schemes[method] = out_stream`  That way you don't need to call del either. Just set it back to None at the end.

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2287,6 +2287,38 @@ def test_cs(self):
         self.assertTrue('9.7743e+00' in compact_lines[11],
             "'9.7743e+00' not found in '%s'" % compact_lines[11])
```
Kenneth-T-Moore, line 2289: PEP extra carriage return.

openmdao/approximation_schemes/approximation_scheme.py
Kenneth-T-Moore: I think I see what you did. system._approximation_schemes is a dictionary that contains this approximation scheme. That is a roundabout way to do this, and I think the class attribute is a better way. You just need to know that `system._approximation_schemes` will be a dictionary that contains a single approximation scheme with the key either "fd" or "cs" depending on what was selected.

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2287,6 +2287,38 @@ def test_cs(self):
...
+        # We don't call run_driver() here because we don't
+        # actually want the optimizer to run
+        prob.run_model()
+
+        # check derivatives with complex step and a larger step size.
+        stream = StringIO()
+        totals = prob.check_totals(method='fd', show_progress=True, out_stream=stream)
+
+        lines = stream.getvalue().splitlines()
+        self.assertTrue(lines[0], "1/3: Checking derivatives with respect to: '_auto_ivc.v0 [0]' ... 0.0088 seconds")
```
Kenneth-T-Moore, line 2318: I think we also want a counterpart test that proves that we don't get the message if we don't specify `show_progress = True`

Kenneth-T-Moore (changes requested):

openmdao/core/tests/test_check_derivs.py
DKilkenny: fixed

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2287,6 +2287,68 @@ def test_cs(self):
...
+        prob.model.nonlinear_solver.options['atol'] = 1e-15
+        prob.model.nonlinear_solver.options['rtol'] = 1e-15
+
+        # We don't call run_driver() here because we don't
+        # actually want the optimizer to run
+        prob.run_model()
+
+        # check derivatives with complex step and a larger step size.
+        stream = StringIO()
+        totals = prob.check_totals(method='fd', show_progress=False, out_stream=stream)
```
Kenneth-T-Moore, line 2345: Just call `check_totals` a second time in the same test, but with show_progress=False.

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2287,6 +2287,68 @@ def test_cs(self):
...
+        # actually want the optimizer to run
+        prob.run_model()
+
+        # check derivatives with complex step and a larger step size.
+        stream = StringIO()
+        totals = prob.check_totals(method='fd', show_progress=True, out_stream=stream)
+
+        lines = stream.getvalue().splitlines()
+        self.assertTrue(lines[0], "1/3: Checking derivatives with respect to: '_auto_ivc.v0 [0]' ... 0.0088 seconds")
+        self.assertTrue(lines[1], "2/3: Checking derivatives with respect to: '_auto_ivc.v0 [1]' ... 9.2891 seconds")
```
Kenneth-T-Moore, line 2318: A few things:

1. Don't check the time -- that can change.  To make a check like that work, you can do `assertTrue(small_string in large_string)`
2. You should convert the '_auto_ivc.v0' name into its corresponding connected promoted input name. Otherwise this has no meaning.
3. For the "false" check, just assert that the text "Checking derivatives with respect to" isn't found anywhere in the output.  You can glue the stream output together into one string with ` "\n".join(lines)` and do a single check

openmdao/core/total_jac.py
```diff
@@ -1454,6 +1454,11 @@ def compute_totals_approx(self, initialize=False):
         initialize : bool
             Set to True to re-initialize the FD in model. This is only needed when manually
             calling compute_totals on the problem.
+        show_progress : bool
```
Kenneth-T-Moore, line 1457: I think you should combine these two arguments into a single one.

The question to ask is, will these two arguments ever be used separately? If not, then there should be a way to combine. LIke before, something like "progress_stream", which is None by default.

openmdao/core/total_jac.py
```diff
@@ -1494,6 +1499,9 @@ def compute_totals_approx(self, initialize=False):
             if model._coloring_info['coloring'] is not None:
                 model._update_wrt_matches(model._coloring_info)
 
+        if show_progress:
+            model._approx_schemes[method].out_stream = out_stream
```
Kenneth-T-Moore, line 1503: Take a look at the if branch a few lines above this, where `method` is defined.  There is a branch of the 'if' where it is not defined, so adjust accordingly.

openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -296,6 +300,8 @@ def _init_approximations(self, system):
     def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
 
+        self._progress_out = self.out_stream if hasattr(self, "out_stream") else None
```
Kenneth-T-Moore, line 303: Again cut out the middleman.  Put the out_stream in self._progress_out, as that is what it is for.

openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -33,6 +34,8 @@ class ApproximationScheme(object):
         Array of sizes of data chunks that make up _j_colored. (Used for MPI Allgatherv)
     _j_data_offsets : ndarray of int
         Array of offsets of each data chunk that makes up _j_colored. (Used for MPI Allgatherv)
+    _progress_out : None
```
Kenneth-T-Moore, line 37: technically, this is None or a file_like_object. (see comments below)

Kenneth-T-Moore (changes requested):

openmdao/core/problem.py
```diff
@@ -1443,6 +1444,8 @@ def check_totals(self, of=None, wrt=None, out_stream=_DEFAULT_OUT_STREAM, compac
         step_calc : string
             Step type for finite difference, can be 'abs' for absolute', or 'rel' for relative.
             Default is 'abs'.
+        show_progress : bool
+            Bool to show progress of check_totals
```
Kenneth-T-Moore, line 1448: I find it better to use less computer-science-like language:  "Set to True to show...

openmdao/core/total_jac.py
```diff
@@ -1486,14 +1489,21 @@ def compute_totals_approx(self, initialize=False):
                 method = list(model._approx_schemes)[0]
                 kwargs = model._owns_approx_jac_meta
                 model.approx_totals(method=method, **kwargs)
+                # if progress_out_stream is not None:
```
Kenneth-T-Moore, line 1492: This is correct, and should be uncommented.

openmdao/core/total_jac.py
```diff
@@ -1486,14 +1489,21 @@ def compute_totals_approx(self, initialize=False):
...
+                #     model._approx_schemes[method].progress_out_stream = progress_out_stream
             else:
                 model.approx_totals(method='fd')
+                # if progress_out_stream is not None:
+                #     model._approx_schemes.progress_out_stream = progress_out_stream
```
Kenneth-T-Moore, line 1497: This is almost right. should be
model._approx_schemes['fd'].progress_out_stream = progress_out_stream

openmdao/core/total_jac.py
```diff
@@ -1486,14 +1489,21 @@ def compute_totals_approx(self, initialize=False):
...
 
             model._setup_jacobians(recurse=False)
             model._setup_approx_partials()
             if model._coloring_info['coloring'] is not None:
                 model._update_wrt_matches(model._coloring_info)
 
+        if progress_out_stream is not None:
+            model._approx_schemes.progress_out_stream = progress_out_stream
```
Kenneth-T-Moore, line 1505: Don't need this now, with the other lines uncommented.

openmdao/approximation_schemes/approximation_scheme.py
```diff
@@ -296,6 +301,13 @@ def _init_approximations(self, system):
     def _compute_approximations(self, system, jac, total, under_cs):
         from openmdao.core.component import Component
 
+        if hasattr(self, "progress_out_stream"):
```
Kenneth-T-Moore, line 304: You don't need this entire if block, if the value of _progress_out is being set in the approximation scheme instead of the dictionary it lives in, as per the other changes I marked.

Kenneth-T-Moore (changes requested):

openmdao/core/total_jac.py
DKilkenny: Whoops, forgot to uncomment these

Kenneth-T-Moore (approved):

openmdao/core/total_jac.py
```diff
@@ -1454,6 +1454,9 @@ def compute_totals_approx(self, initialize=False):
         initialize : bool
             Set to True to re-initialize the FD in model. This is only needed when manually
             calling compute_totals on the problem.
+        progress_out_stream : None or file-like object
+            Where to send human readable output. By default it goes to stdout.
```
swryan, line 1458: Looks like the default is actually `None`

openmdao/core/tests/test_check_derivs.py
```diff
@@ -2287,6 +2288,46 @@ def test_cs(self):
...
+        self.assertTrue("3/3: Checking derivatives with respect to: 'd1.x [2]' ..." in lines[2])
+
+        prob.run_model()
+
+        # Check to make sure nothing is going to output
+        stream = StringIO()
+        totals = prob.check_totals(method='fd', show_progress=False, out_stream=stream)
+
+        lines = stream.getvalue().splitlines()
+        self.assertFalse("Checking derivatives with respect to" in "\n".join(lines))
```
swryan, line 2329: No need to split lines, since you're just joining them again

openmdao/core/tests/test_check_derivs.py
```diff
@@ -6,6 +6,7 @@
 import unittest
 
 import numpy as np
+import sys
```
swryan, line 9: has the code that needed this import been deleted?

DKilkenny: fixed

swryan (approved):

PR #1729, opened 2020-10-07
Check_totals overhaul 1.

### LINKED ISSUE #1700: check_totals should show linear constraints

### LINKED ISSUE #1701: cleanup the output on check totals
### Summary of Issue

Confusing to users who force reverse mode, or who use coloring to see the analytic compute mode always reported as `fwd`. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

There are three updates that are needed here: 

1) Update the compute_totals report so that the correct analytic derivative direction is listed. `Forward Magnitude` should be either `Forward Magnitude` or `Reverse Magnitude` depending on which direction was used to compute that derivative. Same with the `Raw Forward Derivative`. It should be either `Raw Forward Derivative` or `Raw Reverse Derivative`. 

------------

2) The `Jfor` label should be replaced with `Jan` for all cases. 

We need to careful check which direction was used for each specific tuple. If the user hard coded it, then its always forward or reverse. If they were using total derivative coloring, there there could be a mix of forward and reverse. We need to accurately report each specific case. 

------------

3) The label for relative error is inaccurate: (Jfor-Jfd) should be (Jfor-Jfd)/Jfd [assuming we normalize by the finite difference result]. We should accurately label it here and in `check_partials` as well. 

Current Output 
----------------
```
  Full Model: 'obj' wrt 'x'
         Forward Magnitude : 2.980614e+00
         Fd Magnitude : 2.980617e+00 (fd:None)
    Absolute Error (Jfor  - Jfd) : 3.404818e-06 *

    Relative Error (Jfor  - Jfd) : 1.142320e-06 *

    Raw Forward Derivative (Jfor)
[[2.98061391]]

    Raw FD Derivative (Jfd)
[[2.98061732]]
```

Desired Output 
----------------
```
  Full Model: 'obj' wrt 'x'
         Reverse Magnitude : 2.980614e+00
         Fd Magnitude : 2.980617e+00 (fd:None)
    Absolute Error (Jad  - Jfd) : 3.404818e-06 *

    Relative Error (Jad  - Jfd)/Jfd : 1.142320e-06 *

    Raw Reverse Derivative (Jad)
[[2.98061391]]

    Raw FD Derivative (Jfd)
[[2.98061732]]
```

Kenneth-T-Moore:
### Summary

1. Fixed bug where linear constraints where omitted from check_totals.
2. In the check_totals full output:
  a. mark the derivative of a linear constraint with "(Linear constraint)".
  b. change "Jfwd" to "Jan" (analytic).
  c. change "forward" to "analytic".
  d. correct printed formula for relative difference to include the denominator "Jan" or "Jfd".
3. 2.b, 2.c, and 2.d apply to check_partials as well, where applicable. Forward (fwd) and Reverse (rev) are now only mentioned in check_partials for of:wrt pairs that declare matrix-free partials.

### Related Issues

- Resolves #1700 
- Resolves #1701 

None

### New Dependencies

None


openmdao/core/problem.py
```diff
@@ -1470,6 +1471,21 @@ def check_totals(self, of=None, wrt=None, out_stream=_DEFAULT_OUT_STREAM, compac
 
         # TODO: Once we're tracking iteration counts, run the model if it has not been run before.
 
+        if wrt is None:
+            wrt = list(self.driver._designvars)
+            if wrt is None:
```
naylor-b, line 1476: Did you mean to put `if not wrt:` here instead? 'wrt' can never be None here.  Same goes for `if of is None:` check below.

openmdao/core/problem.py
```diff
@@ -1470,6 +1471,21 @@ def check_totals(self, of=None, wrt=None, out_stream=_DEFAULT_OUT_STREAM, compac
...
+                raise RuntimeError("Driver is not providing any design variables "
+                                   "for compute_totals.")
+
+        if of is None:
+            of = list(self.driver._objs)
+            of.extend(self.driver._cons)
+            lcons = [n for n, meta in self.driver._cons.items()
+                     if ('linear' in meta and meta['linear'])]
+            if of is None:
```
naylor-b, line 1485: Very minor, but I would put the if check before the `lcons = ...` line above.

openmdao/core/total_jac.py
```diff
@@ -155,17 +155,15 @@ def __init__(self, problem, of, wrt, use_abs_names, return_format, approx=False,
...
         driver_of = driver._get_ordered_nl_responses()
 
-        # Convert of and wrt names from promoted to absolute
+        # In normal use, of and wrt always contain variable names. However, there are unit tests
+        # that don't specify them, so we need these here.
         if wrt is None:
-            if driver_wrt:
-                prom_wrt = driver_wrt
-            else:
-                raise RuntimeError("Driver is not providing any design variables "
```
naylor-b, line 163: I'm not following why you removed the empty dv and empty response checks here and replaced them with checks in two different places (in compute_totals and check_totals).

naylor-b (changes requested):

openmdao/core/problem.py
Kenneth-T-Moore: oops, I missed these here. Fixed

openmdao/core/total_jac.py
Kenneth-T-Moore: Fixed

openmdao/core/problem.py
Kenneth-T-Moore: Fixed

openmdao/core/total_jac.py
Kenneth-T-Moore: This is what we talked about. Now, the of and wrt are always passed in to this init.  No reason to check them here anymore, since it should never happen.

naylor-b: It would happen if you weren't doing the upper level checks, wouldn't it?

naylor-b: I guess the checks were checking for of/wrt == None, so you're right, that wouldn't happen.

naylor-b (approved):

swryan (approved):

PR #1730, opened 2020-10-07
add support for numpy min, max and diff in ExecComp

fzahle:
### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -627,7 +627,7 @@ _import_functs(np, _expr_dict,
                       'e', 'pi',  # Constants
                       'isinf', 'isnan',  # Logic
                       'log', 'log10', 'log1p', 'power',  # Math operations
-                      'exp', 'expm1', 'fmax',
+                      'exp', 'expm1', 'fmax', 'min', 'max', 'diff',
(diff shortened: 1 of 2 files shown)


naylor-b (approved):

fzahle:
we also use np.interp here and there, but adding this is more tricky since it doesn't pass the complex step test. Looking at using the `SplineComp` instead.

JustinSGray:
wait ... min and max are not differentiable .... hrm....

fzahle:
if you use complex step for derivatives, it should work since it will find the min/max along the real axis and return that along with the corresponding complex part.

JustinSGray:
That is only mathematically valid if the location of the max-value in the array isn't changing. Will it work? probably for fairly well behaved functions. Anyway, its there now... so I guess we're stuck with it. So I can't object to adding the similar methods either. 

```
import numpy as np 

a = np.ones(4)*10

b = np.ones(4, dtype='complex')*10
b[0] += complex(0, 1e-50)

print(np.max(a))

print(np.max(b))

b[1] += 1e-10
print(np.max(b))
```

robfalck (approved):

PR #1744, opened 2020-10-16
Implemented non-modal, multi-capable node info panels

### LINKED ISSUE #1556: N2: Make info panels non-modal
### Summary of Issue

Info panels should remain unless explicitly closed by the user.

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #N/A

### Description

The info panels that pop-up are closed when clicking on the background.  They should remain open and relocatable by dragging until the user closes them.

### Example

N/A

### Environment

N/A

tadkollar:
### Summary
Previously, only one info panel was able to be displayed at once. This change allows for as many as the user wants, in conjunction with value windows.

Changes include:
- A PersistentNodeInfo class was created to manage each pinned info panel. A clone of the original panel HTML is generated and event handlers are added to its `Copy` and `Show more` buttons if it has them.
- z-index management for both info panels and value windows was necessary so older windows wouldn't be irretrievably buried under newer ones.
- `NodeInfo._addPropertyRow()` was removed, and each InfoProp class now either makes use of its `output()` and/or `addRow()` methods to add data to the panel
- Some updates to the styling of the value info windows:
  - Row and column indices are displayed
  - Row colors alternate to make it easier to follow long rows
  - The window template is defined in index.html now and new windows are clones of it

image

### Related Issues

- Resolves #1556 

None

### New Dependencies

None


hschilling (commented):
Something strange about when you click on an element multiple times. For example, using the circuit example `openmdao n2 openmdao/test_suite/scripts/circuit.py`, click on the circuit Group multiple times. You can end up with multiple push pins with not window that goes with it

hschilling:
Also, if you click on the circuit Group exactly where the word circuit is, the info window is created such that the title bar is off-screen so you can't click on it to drag it around or close it.

hschilling:
This is maybe a kind of nice to have but if a node info window is behind another node info window, you have to click on the title bar to bring it to the front. Could you make it so you could click anywhere in the back window it comes to the front? This would be more like how windows work in Windows or Mac OS X.

hschilling:
Should the push pins on the info windows be X's like in the value window? Do the push pins let you do anything other than the closing the info window?

hschilling:
Adding the indices to the value box was a great idea!

robfalck:
I do have one issue where, in the values box, a pathname is so long that I can't get to the button to close the window.  Maybe take the tail of the variable name as the box/window title when the name exceeds the width of the window?

hschilling (approved):

swryan (approved):

PR #1753, opened 2020-11-01
POEM 036 Implementation - Serialization of Kriging training weights

Dakror:
### Summary

This is an implementation of the POEM 036 - Serialization of Kriging training weights: https://github.com/OpenMDAO/POEMs/pull/79

### Related Issues

- Inspired by: https://github.com/OpenMDAO/OpenMDAO/pull/1541

None

### New Dependencies

None


bbrelje:
Bravo, @Dakror! Thank you for following up on the earlier issue with a good fix.

openmdao/surrogate_models/kriging.py
```diff
@@ -2,8 +2,11 @@
 import numpy as np
 import scipy.linalg as linalg
 from scipy.optimize import minimize
+import json
```
naylor-b, line 5: This is unused.

openmdao/surrogate_models/kriging.py
```diff
@@ -91,6 +92,14 @@ def _declare_options(self):
                                   "or 'gesvd' which is slower but more reliable."
                                   "'gesvd' is the default.")
 
+        self.options.declare('training_cache_output', types=str, default=None,
```
JustinSGray, line 95: I don't see the need for two separate options. It seems like one option should work fine, if it accepted `None` or a string. If it's a string, you check if the file exists first, then check for cache validity and apply it if valid. 

If not valid, cache is ignored, surrogate is retrained, then cache file is overwritten with the new data. 

You can *check cache validity* by using md5 hash function on the training data.

JustinSGray (changes requested):
please collapse dual options into a single one for cache, and add a hash-based validity check for the cache.

JustinSGray (changes requested):
The way you have set up the training cache needs a little work. Please add some cache validity checking based on a hash of the training data, and reduce to one input option

openmdao/surrogate_models/kriging.py
Dakror: Changed it to be a single option. I have some doubts about the hashing functionality though. Especially in my use case the numpy arrays are quite big and i dont think hashing these floats is very fast nor stable. From what i found online the common approach to hash numpy arrays is to hash the string representation which i dont expect to be very consistent with floats

Dakror: Also, should i keep updating the POEM with these changes or what is the best approach there?

JustinSGray: You can wait to update the poem till you're done with the implementation. Kind of a chicken and egg problem, but since we've indicated we'd accept the POEM you don't need to keep it tightly up to date with the code. Just match them up in the end. 

Ok, how about a simpler check-sum instead of a hash. Take the training data, flatten it. Then element wise sum it with an integer array like `arange(len(training_data))` then sum the resulting final array to a scalar. 

Thats a bit of a poorman's checksum, but it simple. The right away to do this is with a cyclic redundancy check: 
https://docs.python.org/3/library/binascii.html#binascii.crc32

We have to have some kind of cache validity check here... otherwise users are going to get weird results and not understand why.

Dakror: good call, I'll try that.

Dakror: I actually misunderstood and though i should hash all the weights not just the training data. So now i added hashing of training data only via simple md5 and i think that should suffice

Dakror:
I think i messed up with the rebase. will open a new PR

Dakror:
New PR: https://github.com/OpenMDAO/OpenMDAO/pull/1802

Sorry about the inconvenience

PR #1755, opened 2020-11-03
Current value appears in N2 node info panel instead of initial value

### LINKED ISSUE #1749: N2 diagram value seem to be wrong
### Summary of Issue

I think the N2 is reporting the default value, regardless of what value is in the actual vectors when the n2 was generated

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

```
import openmdao.api as om

p = om.Problem()

p.model.add_subsystem('foo', om.ExecComp('y=2*x'))

p.setup()
p.final_setup()


p['foo.x'] = 10 
om.n2(p)

p.run_model()

print(p['foo.y'])
````

In this example, the generated n2 reports the value of `x` to be `1` when it should be `10`
Screen Shot 2020-10-25 at 12 11 37 PM

tadkollar:
### Summary

The method to get the value of a variable was changed from `meta['value']` to `get_val()` in `n2_viewer.py`. This gets the most recent setting instead of the initial one. The `sellar_tree.json` file used by`test_viewmodeldata.py` was also updated to reflect the new expected model tree.

### Related Issues

- Resolves #1749 

None

### New Dependencies

None


Kenneth-T-Moore (approved):

robfalck (approved):

tadkollar:
The tests were failing when using MPI with this error: `RuntimeError: <model> <class Group>: Non-distributed variable 'summer.y' has a distributed source, 'plus.y', so you must retrieve its value using 'get_remote=True'.`

Only one rank generates the N2 file, and `get_val()` with `get_remote=True` needs to run across all ranks because it makes a collective call. So when running under MPI I set the value retrieval method back to `meta['value']`.

PR #1758, opened 2020-11-05
Fixed residuals_tol arg

DKilkenny:
### Summary

Added residuals_tol logic to filter out by residual tolerance specified by the user in list_outputs

### Related Issues

- Resolves # 1756

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -3631,7 +3631,19 @@ def list_outputs(self,
...
+            if residuals_tol:
+                for val in outputs.values():
+                    if not hasattr(val, "resids"):
+                        simple_warning("No residuals found, skipping residuals_tol")
+                        impl_outputs = {n: m for n, m in outputs.items() if n in states}
+                        break
+                    else:
+                        impl_outputs = {n: m for n, m in outputs.items()
+                                        if n in states and
+                                        (n in states and m['resids'] != residuals_tol)}
```
swryan, line 3643: per:
```
        residuals_tol : float, optional
            If set, limits the output of list_outputs to only variables where
            the norm of the resids array is greater than the given 'residuals_tol'.
```
I think this test should be ` np.linalg.norm(resids) > residuals_tol` 

See comparable logic here:
https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/recorders/case.py#L587

swryan (changes requested):

openmdao/core/system.py
DKilkenny: Whoops, fixed that now

openmdao/core/tests/test_impl_comp.py
```diff
@@ -363,6 +364,32 @@ def test_list_residuals(self):
...
+        prob.set_solver_print(level=-1)
+
+        prob.run_model()
+
+        stdout = StringIO()
+        outputs = model.list_outputs(residuals_tol=0.01, residuals=True, out_stream=stdout)
+        text = stdout.getvalue().split('\n')
+        self.assertEqual(text[14], 'd1')
+        self.assertEqual(text[16], 'd2')
+
```
JustinSGray, line 392: can you explain how this test works? 
it looks like you end up with d1 and d2 in the result. How do you make sure that anything actually got filtered?

JustinSGray (changes requested):

openmdao/core/tests/test_impl_comp.py
DKilkenny: `p1` has a resid of 0 and that gets filtered out. `d1` and `d2` have a resid higher than 0.01 so they don't get filtered out.

JustinSGray: ah, k. Thanks!

JustinSGray: good use of IVC there :)

JustinSGray (approved):

openmdao/core/tests/test_impl_comp.py
```diff
@@ -363,6 +365,35 @@ def test_list_residuals(self):
...
+        prob.set_solver_print(level=-1)
+
+        prob.run_model()
+
+        stdout = StringIO()
+        outputs = model.list_outputs(residuals_tol=0.01, residuals=True, out_stream=stdout)
+        text = stdout.getvalue().split('\n')
+        # P1 and D1 should not appear in the outputs section. This is being checked below
+        self.assertEqual(text[14], 'd2')
+        self.assertEqual(text[15], '  y2     [0.2323928]  [0.01679515]')
```
swryan, line 393: This test fails for me with numerical differences:
```AssertionError: '  y2     [0.23238004]  [0.0167883]' != '  y2     [0.2323928]  [0.01679515]'
-   y2     [0.23238004]  [0.0167883]
?                  ---          ^^^
+   y2     [0.2323928]  [0.01679515]
?                 ++           ^^^^
```
I think you don't really care about the y values for the purposes of this test... 
I think rather you would just want to check that `d1` is present and `p1` and `d2` are not (due to their outputs being filtered)

swryan (commented):
was `distributed_gradient_test.py` deleted intentionally?  doesn't seem to be related to the issue..

DKilkenny:
> was `distributed_gradient_test.py` deleted intentionally? doesn't seem to be related to the issue..

Yes that was intentional. I must have snuck it in a PR by accident

swryan (approved):

PR #1778, opened 2020-11-19
Internal refactor of vector to have real and complex step modes share memory.

### LINKED ISSUE #1767: Refactor of complex-step vectors to save memory usage
### Summary of Issue

Currently we allocate a complete separate array for complex values and swap it in when needed. This is wasteful though, and we can do a minor refactor to save the extra memory  

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #30 (related to this POEM, but not the API changes for the POEM. Just an internal refactor as part of it)

### Description

Numpy allows you to create separate views into the real and complex parts of a complex array. So we don't need to allocate a separate real array when `force_alloc_complex` is on. Instead, we can just assign the view to the real part of that array. 

This will save roughly 33% of memory usage in complex step mode. It should be measurable since we have to allocate memory for both the output and the residual vectors. 

### Example

Different options for making the views: 
`myview = arr.real[start:stop]`

```
>>> a1 = np.ones(3, dtype=complex)
>>> r1 = a1.real 
>>> c1 = a1.imag
>>> r1[0] = 2
>>> a1 
array([2.+0.j, 1.+0.j, 1.+0.j])
```

naylor-b:
### Summary

We used to allocate separate real and complex data arrays within a Vector.  Now, if complex step is being done, we allocate a complex data array and return/set the real part of it if the vector is not currently under a complex step operation.

### Related Issues

- Resolves #1767

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -2631,9 +2631,9 @@ class Group(System):
         complex_step = self._inputs._under_complex_step
 
         if complex_step:
-            self._inputs.set_complex_step_mode(False, keep_real=True)
-            self._residuals.set_complex_step_mode(False, keep_real=True)
-            self._outputs.set_complex_step_mode(False, keep_real=True)
+            self._inputs.set_complex_step_mode(False)
+            self._residuals.set_complex_step_mode(False)
+            self._outputs.set_complex_step_mode(False)
 
         if self._discrete_inputs or self._discrete_outputs:
             self.guess_nonlinear(self._inputs, self._outputs, self._residuals,
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -2636,8 +2636,7 @@ class Group(System):
 
             # The Group outputs vector contains imaginary numbers from other components, so we need
             # to save a cache and restore it later.
-            imag_cache = np.empty(len(self._outputs._data))
-            imag_cache[:] = self._outputs._data.imag
+            imag_cache = self._outputs._data.imag.copy()
             self._outputs.set_complex_step_mode(False, keep_real=True)
 
         if self._discrete_inputs or self._discrete_outputs:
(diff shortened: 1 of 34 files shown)


Kenneth-T-Moore (approved):

openmdao/core/group.py
```diff
@@ -2631,14 +2631,9 @@ def _guess_nonlinear(self):
         complex_step = self._inputs._under_complex_step
 
         if complex_step:
-            self._inputs.set_complex_step_mode(False, keep_real=True)
-            self._residuals.set_complex_step_mode(False, keep_real=True)
-
-            # The Group outputs vector contains imaginary numbers from other components, so we need
-            # to save a cache and restore it later.
-            imag_cache = np.empty(len(self._outputs._data))
-            imag_cache[:] = self._outputs._data.imag
```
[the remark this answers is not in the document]
naylor-b: And the other things we don't need any more are 2 dicts (_views, _views_flat) for every Vector in the entire model, so that's significant as well.  It's not ideal that I had to cache the old values in compute_approximations, and it may be that we could be smarter about that and maybe only do it in certain circumstances, but I tried turning it off after everything was passing and it caused a few failures.

openmdao/vectors/vector.py
```diff
@@ -589,7 +594,7 @@ def _in_matvec_context(self):
         raise NotImplementedError('_in_matvec_context not defined for vector type %s' %
                                   type(self).__name__)
 
-    def set_complex_step_mode(self, active, keep_real=False):
+    def set_complex_step_mode(self, active):
```
swryan, line 597: the extended description in this docstring is no longer accurate

naylor-b: fixed

swryan (approved):

PR #1796, opened 2020-12-09
New function to simplify units.

### LINKED ISSUE #1776: Add functionality to allow OpenMDAO to simplify unit strings
### Summary of Issue

Add the capability to allow unit strings to automatically be simplified when they are being parsed programmatically.

### Issue Type

- [ ] Bug
- [x] Enhancement  
   

### Description

In Dymos, sometimes we take an existing openmdao variable and create a new one that contains the time
derivative or integral of that variable. To do so, we need to also differentiate or integrate the units by multiplying
or dividing by the time unit. This can create unit strings that are not simplified (e.g., "m/s*s"). This
POEM proposes a units library function that we can call to clean those up.

This POEM proposes a new method:

```
new_unit_str = simplify_unit(old_unit_str)
```

### Example

'ft/s&ast;s' returns 'ft'
'm/s&ast;s' returns 'm'
's/s' returns None

If possible:
'J/s&ast;s' returns 'W'

Care should be taken to make sure that units don't get converted to their base units (i.e., 'ft' -> 'm').

### Environment

N/A

Kenneth-T-Moore:
### Summary

Added new function `simplify_unit` that simplified unit strings to remove units that are present in both the numerator and denominator.

### Related Issues

- Resolves #1776

None

### New Dependencies

None


swryan (approved):

naylor-b (approved):

JustinSGray (approved):
Should this be implemented as a stand alone function? or should it be integrated into the unit system directly (i guess it could be stand alone function still, just always used)? 

I see that you added the function, but didn't make it automatically used by the units system.

robfalck:
> Should this be implemented as a stand alone function? or should it be integrated into the unit system directly (i guess it could be stand alone function still, just always used)?
> 
> I see that you added the function, but didn't make it automatically used by the units system.

I'm trying to think if there's ever a case where we wouldn't want this to be applied, and I'm not able to think of one at the moment.

robfalck (changes requested):
Well this turned out to be simpler than I expected it to be.  I can't see of a good reason to ever allow something like `'m/m/s'` be a unit, though.  I agree with Justin that we should just always apply this logic to units.

Also, since we explicitly didn't take the POEM route on this, having a separate function for it would seem to require a POEM, as opposed to just making it the default behavior.

This gives me the behavior I expect in Dymos tests.

Kenneth-T-Moore:
Updated: Now we simplify the units automatically.

robfalck (approved):

PR #1804, opened 2020-12-15
Fixed issue where resid_tol failed with vectorized resids

### LINKED ISSUE #1801: Residual tolerance option with vectorized residual throwing a ValueError
### Summary of Issue

If a residual is vectorized and you try to sort residual values above a certain tolerance using list_outputs() with any value for residuals_tol option you get a ValueError.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

`Traceback (most recent call last):
  File "test.py", line 37, in <module>
    prob.model.list_outputs(residuals=True, residuals_tol=1e-5, prom_name=True, units=True)
  File "/openmdao/core/system.py", line 3635, in list_outputs
    if "resids" in m and n in states and m['resids'] > residuals_tol:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

### Example

```python
import openmdao.api as om
import numpy as np

prob = om.Problem()
model = prob.model

model.add_subsystem(
    "quad_1",
    om.ExecComp(
        "y = a * x ** 2 + b * x + c",
        a={"value": 2.0},
        b={"value": 5.0},
        c={"value": 3.0},
        x={"shape": (2,)},
        y={"shape": (2,)},
    ),
)

balance = model.add_subsystem("balance", om.BalanceComp())
balance.add_balance("x_1", val=np.array([1, -1]), rhs_val=np.array([0., 0.]))
model.connect("balance.x_1", "quad_1.x")
model.connect("quad_1.y", "balance.lhs:x_1")


prob.model.linear_solver = om.ScipyKrylov()
prob.model.nonlinear_solver = om.NewtonSolver(solve_subsystems=False, maxiter=100, iprint=2)


prob.setup()

prob.model.nonlinear_solver.options["maxiter"] = 0

prob.run_model()

prob.model.list_inputs()

prob.model.list_outputs(residuals=True, residuals_tol=1e-5)
```

DKilkenny:
### Summary

Fixed issue where resid_tol failed with vectorized resids

### Related Issues

- Resolves #1801

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/tests/test_exec_comp.py
+++ b/openmdao/components/tests/test_exec_comp.py
@@ -5,6 +5,7 @@ import math
 import numpy as np
 from numpy.testing import assert_almost_equal
 import scipy
+from io import StringIO
 
 from distutils.version import LooseVersion
 
@@ -1160,6 +1161,40 @@ class TestExecComp(unittest.TestCase):
 
         assert_near_equal(prob.get_val('comp.y'), [2., 4.], 0.00001)
 
+    def test_list_outputs_resids_tol(self):
+        prob = om.Problem()
+        model = prob.model
+
+        model.add_subsystem(
+            "quad_1",
+            om.ExecComp(
+                "y = a * x ** 2 + b * x + c",
+                a={"value": 2.0},
+                b={"value": 5.0},
+                c={"value": 3.0},
+                x={"shape": (2,)},
+                y={"shape": (2,)},
+            ),
+        )
+
+        balance = model.add_subsystem("balance", om.BalanceComp())
+        balance.add_balance("x_1", val=np.array([1, -1]), rhs_val=np.array([0., 0.]))
+        model.connect("balance.x_1", "quad_1.x")
+        model.connect("quad_1.y", "balance.lhs:x_1")
+
+        prob.model.linear_solver = om.ScipyKrylov()
+        prob.model.nonlinear_solver = om.NewtonSolver(solve_subsystems=False, maxiter=100, iprint=2)
+
+        prob.setup()
+        prob.model.nonlinear_solver.options["maxiter"] = 0
+        prob.run_model()
+
+        stream = StringIO()
+        outputs = prob.model.list_outputs(residuals=True, residuals_tol=1e-5, out_stream=stream)
+
+        text = stream.getvalue()
+        self.assertTrue("balance" in text)
+        self.assertTrue("x_1" in text)
 
 class TestExecCompParameterized(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


openmdao/core/system.py
```diff
@@ -3632,8 +3632,14 @@ def list_outputs(self,
...
-                    if "resids" in m and n in states and m['resids'] > residuals_tol:
-                        impl_outputs[n] = m
+                    if "resids" in m and n in states:
+                        if isinstance(m['resids'], np.ndarray):
+                            for i in m['resids']:
+                                if i > residuals_tol:
+                                    impl_outputs[n] = m
+                                    break
+                        else:
+                            impl_outputs[n] = m
```
naylor-b, line 3642: Shouldn't we have a check here against residuals_tol as well?

openmdao/components/tests/test_exec_comp.py
```diff
@@ -1160,6 +1161,40 @@ def test_feature_options(self):
 
         assert_near_equal(prob.get_val('comp.y'), [2., 4.], 0.00001)
 
+    def test_list_outputs_resids_tol(self):
```
naylor-b, line 1164: Do we have a test for residuals_tol when the residual is a scalar?

naylor-b (changes requested):

openmdao/components/tests/test_exec_comp.py
DKilkenny: Yes, it's `test_list_residuals_with_tol` in `test_impl_comp.py`

naylor-b: If that test works correctly then why didn't it fail when you were missing the residuals_tol check for scalars?

DKilkenny: Fair point. It was cause I was only checking for an array and not the length of the array

openmdao/core/system.py
```diff
@@ -3632,8 +3632,14 @@ def list_outputs(self,
             impl_outputs = {}
             if residuals_tol:
                 for n, m in outputs.items():
-                    if "resids" in m and n in states and m['resids'] > residuals_tol:
-                        impl_outputs[n] = m
+                    if "resids" in m and n in states:
+                        if len(m['resids']) > 1:
```
naylor-b, line 3636: Check to make sure that for a scalar variable, the value of m['resids'] ends up being a size 1 array, because calling len(x) where x is a scalar will fail.

naylor-b (changes requested):

naylor-b (approved):

swryan (approved):

PR #1818, opened 2021-01-05
Added warning when nonlinear solver stalls three times

### LINKED ISSUE #1811: Useful warning when Newton stall is detected.
### Summary of Issue

Line Search should print a warning when stalled out due to bounds enforcement. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM Not needed. Just adding a warning to existing functionality
### Description

Stall detection allows for a user selectable exit of of the solver when the residual stops changing. In most cases, stall happens when you are continually bumping up against a bound. 

Many users do not know about the `print_bound_enforce` option on the line search. When stall is detected (by the residual not changing for at least 3 iterations in a row) we can throw a warning to the screen on the same line as the latest iteration to tell them to turn on `print_bound_enforce`.  This warning can be very descriptive. It should tell them exactly how to turn on the option, and also give them bound enforcement information that they would normally get from print_bound enforce. 

The warning should include the full path of the solver, like this: `<something>.<something>.<something>.<something>.nonlinear_solver.linesearch.options['print_bound_enforce']=True. `

It should also give them the current bounded variables (just like print_bound_enforce). 

This will make it much easier for users to recognize the problem and debug it.

DKilkenny:
### Summary

Added warning when nonlinear solver stalls three times

### Related Issues

- Resolves #1811

None

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/solver.py
+++ b/openmdao/solvers/solver.py
@@ -611,9 +611,11 @@ class NonlinearSolver(Solver):
                             "now are:\n")
                     simple_warning(msg)
 
+                    self._single_iteration()
                     self.linesearch.options['print_bound_enforce'] = False
+                else:
+                    self._single_iteration()
 
-                self._single_iteration()
                 self._iter_count += 1
                 self._run_apply()
                 norm = self._iter_get_norm()
--- a/openmdao/solvers/solver.py
+++ b/openmdao/solvers/solver.py
@@ -604,11 +604,11 @@ class NonlinearSolver(Solver):
                     self.linesearch.options['print_bound_enforce'] = True
 
                     msg = ("Your model has stalled three times and may be violating the "
-                            "bounds. In the future, turn on print_bound_enforce in your "
-                            "solver options to see. For example:\n"
-                            "'prob.model.nonlinear_solver.linesearch.options"
-                            "['print_bound_enforce']=True'. \nThe bound(s) being violated "
-                            "now are:\n")
+                           "bounds. In the future, turn on print_bound_enforce in your "
+                           "solver options to see. For example:\n"
+                           "'prob.model.nonlinear_solver.linesearch.options"
+                           "['print_bound_enforce']=True'. \nThe bound(s) being violated "
+                           "now are:\n")
                     simple_warning(msg)
 
                     self._single_iteration()
--- a/openmdao/solvers/solver.py
+++ b/openmdao/solvers/solver.py
@@ -571,14 +571,6 @@ class NonlinearSolver(Solver):
         norm0 = norm if norm != 0.0 else 1.0
         return norm0, norm
 
-    def _run_solver(self):
-        """
-        Run a single iteration and increment.
-        """
-        self._single_iteration()
-        self._iter_count += 1
-        self._run_apply()
-
     def _solve(self):
         """
         Run the iterative solver.
@@ -606,7 +598,9 @@ class NonlinearSolver(Solver):
 
         while self._iter_count < maxiter and norm > atol and norm / norm0 > rtol and not stalled:
             with Recording(type(self).__name__, self._iter_count, self) as rec:
-                self._run_solver()
+                self._single_iteration()
+                self._iter_count += 1
+                self._run_apply()
                 norm = self._iter_get_norm()
 
                 # Save the norm values in the context manager so they can also be recorded.
@@ -634,8 +628,7 @@ class NonlinearSolver(Solver):
                                    "now are:\n")
                             simple_warning(msg)
 
-                            self._run_solver()
-
+                            self._single_iteration()
                             self.linesearch.options['print_bound_enforce'] = False
                     else:
                         stall_count = 0
--- a/openmdao/solvers/solver.py
+++ b/openmdao/solvers/solver.py
@@ -592,12 +592,27 @@ class NonlinearSolver(Solver):
         self._mpi_print(self._iter_count, norm, norm / norm0)
 
         stalled = False
+        stall_count = 0
         if stall_limit > 0:
-            stall_count = 0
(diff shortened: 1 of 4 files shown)


openmdao/solvers/solver.py
```diff
@@ -616,6 +617,19 @@ def _solve(self):
...
+                            msg = ("Your model has stalled three times and may be violating the "
+                                   "bounds. In the future, turn on print_bound_enforce in your "
+                                   "solver options to see. For example:\n"
+                                   "'prob.model.nonlinear_solver.linesearch.options"
+                                   "['print_bound_enforce']=True'. \nThe bound(s) being violated "
+                                   "now are:\n")
+                            simple_warning(msg)
+
+                            self._single_iteration()
+                            self.linesearch.options['print_bound_enforce'] = False
```
Kenneth-T-Moore, line 632: This line will always turn off `print_bound_enforce` after 3 stalls.  Instead, you should cache the value of that option, and restore it here with the cache.

Kenneth-T-Moore (changes requested):

openmdao/solvers/solver.py
Kenneth-T-Moore: Actually, a better solution is to change the conditional `if stall_count == 3:` into `if stall_count == 3 and self.linesearch.options['print_bound_enforce']:`.  That way you don't do any of this if they have the option turned on, and you don't have to cache it.

DKilkenny: Fixed

openmdao/solvers/solver.py
```diff
@@ -616,6 +617,19 @@ def _solve(self):
                         stall_count += 1
                         if stall_count >= stall_limit:
                             stalled = True
+                        if stall_count == 3 and not self.linesearch.options['print_bound_enforce']:
```
Kenneth-T-Moore, line 620: One more thing. I don't think we want to run `_single_iteration` twice in the same iteration, because it may be inefficient if the submodel contains solvers or slow components.  I think it would be better to move this new block to the point at the beginning of the recorder context where we already run _single_iteration.  If the conditional is met, we warn, set the option, and run single iteration, otherwise, we just run the single iteration.

Kenneth-T-Moore (changes requested):

Kenneth-T-Moore (approved):

openmdao/solvers/solver.py
```diff
@@ -591,13 +592,30 @@ def _solve(self):
...
+                    msg = ("Your model has stalled three times and may be violating the "
+                           "bounds. In the future, turn on print_bound_enforce in your "
+                           "solver options to see. For example:\n"
+                           "'prob.model.nonlinear_solver.linesearch.options"
+                           "['print_bound_enforce']=True'. \nThe bound(s) being violated "
+                           "now are:\n")
+                    simple_warning(msg)
+
+                    self._single_iteration()
+                    self.linesearch.options['print_bound_enforce'] = False
```
swryan, line 615: Warning messages are normally written to `stderr` or handled by `logging` while the bound violations are being printed to `stdout` (see  [_print_violations](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/linesearch/backtracking.py#L17)).  This raises the possibility that this output is not seen as intended.

- Maybe the proper thing to do is to make the bounds violation messages proper warnings instead of just prints? 
- Could also try to capture `stdout` and insert the violation info into the warning message, but there's probably a bunch of other output from that `_single_iteration`

openmdao/solvers/solver.py
```diff
@@ -591,13 +592,30 @@ def _solve(self):
...
-                self._single_iteration()
+
+                if stall_count == 3 and not self.linesearch.options['print_bound_enforce']:
+
+                    self.linesearch.options['print_bound_enforce'] = True
+
+                    msg = ("Your model has stalled three times and may be violating the "
+                           "bounds. In the future, turn on print_bound_enforce in your "
+                           "solver options to see. For example:\n"
+                           "'prob.model.nonlinear_solver.linesearch.options"
```
swryan, line 609: I think the intent of Issue #1811 was that the actual path to _this_ solver is included in the warning:

> The warning should include the full path of the solver, like this: <something>.<something>.<something>.<something>.nonlinear_solver.linesearch.options['print_bound_enforce']=True.

openmdao/solvers/solver.py
Kenneth-T-Moore: yeah, I think the bounds violation prints would be better as warnings (simple_warning) rather than prints (though maybe verify with Jubs).

robfalck (approved):

swryan (changes requested):
per previous comments:
- change the bounds_violation messages to warnings
- display the path to the solver instance

Kenneth-T-Moore (approved):

openmdao/solvers/solver.py
```diff
@@ -591,13 +592,34 @@ def _solve(self):
...
+                    self.linesearch.options['print_bound_enforce'] = True
+
+                    if self._system().pathname:
+                        pathname = f"{self._system().pathname}."
+                    else:
+                        pathname = ""
+
+                    msg = (f"Your model has stalled three times and may be violating the bounds. "
+                           f"In the future, turn on print_bound_enforce in your solver options "
+                           f"here: \n{pathname}nonlinear_solver.linesearch.options"
```
swryan, line 613: I disagree with hard-coding the assumed solver name here. I think we should only provide the copy/paste text if the solver does in fact know it's own name (by having it passed in via the system._setup_solvers call).

In the absence of a concurring opinion, I will approve the requested changes with this objection.

swryan (approved):

PR #1835, opened 2021-01-14
fix for driver.scaling_report

naylor-b:
### Summary

Problem was happening when scaling_report was called after driver totals had already been computed in some format other than 'array' format.

Also found and fixed a coloring bug where parts of the jacobian were ending up nonzero when they should have been zero based on the coloring.

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/total_jac.py
+++ b/openmdao/core/total_jac.py
@@ -950,10 +950,7 @@ class _TotalJacInfo(object):
 
         for color, ilist in enumerate(coloring.color_iter(mode)):
             if len(ilist) == 1:
-                if both:
-                    yield ilist, input_setter, jac_setter, None
-                else:
-                    yield ilist[0], self.single_input_setter, self.single_jac_setter, None
+                yield ilist, input_setter, jac_setter, None
             else:
                 # yield all indices for a color at once
                 yield ilist, input_setter, jac_setter, imeta['itermeta'][color]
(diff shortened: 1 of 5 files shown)


openmdao/core/total_jac.py
```diff
@@ -950,10 +950,7 @@ def simul_coloring_iter(self, imeta, mode):
 
         for color, ilist in enumerate(coloring.color_iter(mode)):
             if len(ilist) == 1:
-                if both:
```
Kenneth-T-Moore, line 953: I think you can get rid of the variable `both` now.

Kenneth-T-Moore (changes requested):

openmdao/core/total_jac.py
naylor-b: fixed

openmdao/visualization/scaling_viewer/scaling_table.html
```diff
@@ -466,31 +472,31 @@ <h3>Jacobian Info</h3>
...
-        svg.append("g")
-            .attr("transform", "translate(0," + height + ")")
-            .call(d3.axisBottom(x))
-            .selectAll("text")
-            .style("text-anchor", "end")
-            .attr("transform", "rotate(-45)")
-            .attr("dx", "-.5em")
-            .attr("dy", ".3em");
-    }
+    // if (is_top) {
```
robfalck, line 475: Is there a reason to leave these commented-out instead of just removing them?

naylor-b: Nope.  Removed.

swryan (approved):

Kenneth-T-Moore (approved):

robfalck (approved):

PR #1838, opened 2021-01-20
I1682 add a display to the N2 so that if an error occurs the user some idea of what happened

### LINKED ISSUE #1682: Indicate to the user when there are possible errors in the N2 display
### Summary of Issue

Currently, if there is an error in displaying the N2, either due to JavaScript errors, or missing resources,..., the user does not know about them unless the user facing display makes it obvious something is wrong. 

It would be helpful to put some kind of warning visible to the user when something is wrong. Also, possibly give them a way to get more info that they could put in a bug report. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Make some visual indication of errors when viewing the N2.

Other people have also tried doing something like this for their Web apps:

https://stackoverflow.com/questions/6604192/showing-console-errors-and-alerts-in-a-div-inside-the-page
https://stackoverflow.com/questions/23881716/show-javascript-console-errors-on-the-page
https://stackoverflow.com/questions/20253652/how-do-you-send-console-messages-and-errors-to-alert

This looks promising

```
window.addEventListener("error", handleError, true);
function handleError(evt) {
    if (evt.message) { // Chrome sometimes provides this
      alert("error: "+evt.message +" at linenumber: "+evt.lineno+" of file: "+evt.filename);
    } else {
      alert("error: "+evt.type+" from element: "+(evt.srcElement || evt.target));
    }
}
```

### Example

N/A

### Environment

N/A

hschilling:
### Summary

Currently, if there is an error in the N2 scripts on the JavaScript side, the user might just get a blank screen for the N2. This PR provides an indication of an error and also a way to get a little more info for use in debugging and reporting the issue. For minor errors, the user will still possibly see and be able to use the N2, but they will get a warning about it.

### Related Issues

- Resolves #1682 

None

### New Dependencies

None


openmdao/visualization/html_utils.py
```diff
@@ -507,7 +507,7 @@ class TemplateWriter(object):
         Contents of template file.
     """
 
-    def __init__(self, filename, embeddable=False, title=None, styles=None):
+    def __init__(self, filename, embeddable=False, title=None, styles=None,head_srcs=None):
         """
```
Kenneth-T-Moore, line 511: Need to add `head_srcs` to the doc string.

openmdao/visualization/html_utils.py
```diff
@@ -588,7 +592,7 @@ class DiagramWriter(TemplateWriter):
         String of HTML for the help dialog.
     """
 
-    def __init__(self, filename, embeddable=False, title=None, styles=None):
+    def __init__(self, filename, embeddable=False, title=None, styles=None, head_srcs=None):
         """
```
Kenneth-T-Moore, line 596: Here too.

Kenneth-T-Moore (changes requested):

openmdao/visualization/html_utils.py
hschilling: Fixed. Thanks

Kenneth-T-Moore (approved):

swryan (approved):

PR #1844, opened 2021-01-20
Added detection of notebook env, table formatting and in cell HTML output

### LINKED ISSUE #1825: Investigate modularization of HTML reports for integration with Jupyter
### Summary of Issue

Look into making our html reports work in both stand-alone manner, and integrated with jupyter

### Description

For things like the N2 and the scaling report, the utility of having a stand alone HTML file that can be shared is very high. We're also looking at building tighter jupyter intgegration. 

Can we re-structure the stand-alone HTML file as some kind of pseudo-jupyter like harness that HTML reports can be sloted into when we want stand-alone documents? 

The top level goal here, regardless of implementation, is to have one code base that can provide jupyter widgets (i think thats the right term) and also stand-alone HTML functionality.

### LINKED ISSUE #1843: Add cleaner formatting for list_* functions and embedded HTML for Jupyter Notebooks
### Summary of Issue

If a user wants to use Jupyter Notebooks, we should take advantage of the ability to have a cleaner format for table than just an ascii dump. Also, the N2 and view connections should display in the output cell. Whether or not the user is in a notebook environment should be determined automatically.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

See Summary 

### Example

N/A

### Environment

N/A

DKilkenny:
### Summary

Added automatic detection of notebook env, table formatting and in cell HTML output

### Related Issues

- Resolves #1843
- Resolves #1825

None

### New Dependencies

Tabulate


openmdao/core/notebook_mode.py
```diff
@@ -0,0 +1,23 @@
+"""Checking for interactive notebook mode."""
+
+
+def _notebook_mode():
+
+    try:
+        from IPython import get_ipython
+        ipy = get_ipython() is not None
+    except ImportError:
+        pass
```
robfalck, line 10: ipy = False

openmdao/core/notebook_mode.py
```diff
@@ -0,0 +1,23 @@
...
+    else:
+        try:
+            import tabulate
+        except ImportError:
+            print("Tabulate is not installed run `pip install openmdao[notebooks]` to install "
+                  "required dependencies")
+    return ipy
+
+
+if _notebook_mode():
```
robfalck, line 20: notebook = notebook_mode()

openmdao/core/notebook_mode.py
```diff
@@ -0,0 +1,23 @@
+"""Checking for interactive notebook mode."""
+
+
+def _notebook_mode():
```
Kenneth-T-Moore, line 4: Tried on my branch. I don't have tabulate installed. I get

`Tabulate is not installed run `pip install openmdao[notebooks]` to install required dependencies`

every time I run an openmdao model.

Kenneth-T-Moore (changes requested):

robfalck (changes requested):

robfalck (approved):

openmdao/core/system.py
```diff
@@ -17,7 +17,20 @@
 import numpy as np
 import networkx as nx
 
+try:
+    from IPython.display import HTML, display
```
Kenneth-T-Moore, line 21: Seems like we don't need both the `ipython` flag and the `tab_pkg` flag. They are always tested in conditional together (i.e., there is no special action when one it True and the other False). I think you can squash these down to one flag that is basically the same as `ipython and tab_pkg`

I think you could even get fancier and have no flags. Just set `tabulate` to None if both ipython and tabulate fail to import, and then use `if tabulate` for the conditionals.

openmdao/core/system.py
```diff
@@ -322,6 +335,8 @@ class System(object):
         If True, this is the first call to _linearize.
     _is_local : bool
         If True, this system is local to this mpi process.
+    _notebook : bool
+        If True, display output in Jupyter Notebook format
```
Kenneth-T-Moore, line 339: I guess I know this is just proof of concept, but in practice, do I have to add a notebook to every system in my model?

Kenneth-T-Moore (changes requested):

openmdao/core/system.py

openmdao/core/system.py
naylor-b: I agree with Ken on this.  Whether we're in a notebook or not is a global thing, right?  Can't you just use the 'notebook' flag defined in the notebook_mode module?

Kenneth-T-Moore (approved):

openmdao/core/system.py
Kenneth-T-Moore: Danny explained to me that this is automatically set if it detects that you are inside a ipython notebook environment.

DKilkenny: I talked with Bret about this and he was making the point that we don't need a global var but instead use the imported `notebook` reference. 
Steve offline thought it would be cleaner to have the `tabulate` import inside the `notebook_mode.py` file.

PR #1849, opened 2021-01-22
Fixed a bug where units weren't simplified in the get_val methods.

### LINKED ISSUE #1832: Additional unit checking is required for set_val and get_val
### Summary of Issue

Methods set_val and get_val need to check for valid units and need to simplify the given units.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The methods `set_val` and `get_val` and Problem, System, and Case currently do no checking that a given unit is valid.  They also don't simplify the given units.  This can cause problems when trying to get a variable that has units of `None` and the given user units are potentially determined computationally (`'m/m'`, for instance).

### Example

TBD

### Environment

OpenMDAO 3.5.1-dev

robfalck:
### Summary

`get_val` now uses `simplify_unit` in `Case`, `Problem`, and `System`.
`simplify_unit` now performs the same check as `valid_unit`, and calls to `valid_unit` just before simplify unit have been removed.
After removing some seemingly redundant imports, OpenMDAO was failing to import `_full_slice` from Component.  It is now imported from the module in which it is defined (vectors.vector).

### Related Issues

- Resolves #1832

None

### New Dependencies

None


naylor-b (approved):

openmdao/utils/units.py
```diff
@@ -1106,13 +1106,23 @@ def simplify_unit(old_unit_str):
...
         Simplified unit string.
     """
-    new_str = _find_unit(old_unit_str).name()
+    if old_unit_str is None:
+        return None
+
+    found_unit = _find_unit(old_unit_str)
+    if found_unit is None:
+        _msginfo = f'{msginfo}: ' if msginfo else ''
+        raise ValueError(f'{_msginfo}The units {old_unit_str} are invalid.')
```
Kenneth-T-Moore, line 1123: I think we should put single quotes around the units in the exception, so it would look like:

The units 'in' are invalid.

Kenneth-T-Moore (changes requested):

openmdao/core/tests/test_units.py
```diff
@@ -264,6 +264,7 @@ def setup(self):
             prob = om.Problem(model=Comp1())
             prob.setup()
         expected_msg = "The units 'junk' are invalid"
+        print(str(cm.exception))
```
Kenneth-T-Moore, line 267: Leftover debugging.

Kenneth-T-Moore (changes requested):

openmdao/core/tests/test_units.py
robfalck: @Kenneth-T-Moore fixed

Kenneth-T-Moore (approved):

PR #1878, opened 2021-02-05
Parallel analysis error fix

### LINKED ISSUE #1877: analysis error on non-convergence is broken in parallel runs with both linear and nonlinear solvers
### Summary of Issue

If the environment variable `USE_PROC_FILES` is `False` (or not set), then the logic of checking for solver failure is only done on the root proc in both `LinearSolver` (https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/solver.py#L873) and `NonLinearSolver` (https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/solver.py#L647). When the analysis error is raised, only the root proc raises it, and causes the driver to hang. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

All procs should raise the analysis error otherwise the driver hangs.

### Example

Will create test cases in the PR that fixes the issue.

anilyil:
### Summary

This PR fixes how the `AnalysisError` is raised in parallel runs with both linear and nonlinear solvers. Before this PR, the error was only raised on root unless the environment variable `USE_PROC_FILES` was set to True. The PR fixes it so that the convergence checks are performed on all procs in the comm, and similarly, all procs in the comm raise this error if desired. The behavior of the print statements are preserved; if `USE_PROC_FILES` is false or not set, only root prints, else, all procs print to their files.

I also added tests to check if the analysis error is raised properly on all procs when nonlinear or linear solvers fail. However, without this fix, these tests actually just hang and not explicitly fail. This is because the process hangs without the fix. Is there a way to trigger a failure for this case @JustinSGray? 

### Related Issues

- Resolves #1877

None

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/linear/tests/test_linear_block_gs.py
+++ b/openmdao/solvers/linear/tests/test_linear_block_gs.py
@@ -12,6 +12,12 @@ from openmdao.test_suite.components.expl_comp_simple import TestExplCompSimpleDe
 from openmdao.test_suite.components.sellar import SellarDerivatives
 from openmdao.utils.assert_utils import assert_near_equal
 
+from openmdao.utils.mpi import MPI
+try:
+    from openmdao.api import PETScVector
+except:
+    PETScVector = None
+
 
 class SimpleImp(om.ImplicitComponent):
     def setup(self):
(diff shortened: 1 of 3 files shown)


Kenneth-T-Moore (approved):
Looks good (code and tests.) Nice catch.

anilyil:
I think the lint test failed due to indentation. will push a fix in a second

swryan (approved):

PR #1904, opened 2021-02-16
New N2 capability to save and load views.

### LINKED ISSUE #1612: Add ability to save the state of N2 diagram navigation
### Summary of Issue

Currently, there's no way to save the navigation state (zoomed nodes, collapsed nodes, connection arrows, etc.) of an N2 diagram. This can make it cumbersome for users that are debugging or revising a model to recreate the view they were examining. 

### Issue Type

- [ ] Bug
- [X] Enhancement

### Description

Add the following capabilities to enable the N2 state to be saved and restored:
- Save the zoomed element, the minimized and hidden state of each node, the persistent arrow cache, and the linear/non-linear solver toggle state. Back/forward history is another candidate, but may be too large.
  - This probably needs to be done as a JSON file.
  - Storing cookies are another possibility, but cookie support for local files between browsers is inconsistent. Also, diagrams are frequently saved as `n2.html`, so it could be difficult to know whether a cookie applies to the current diagram or not.
- Create a checksum to be able to compare the saved state to the existing model in case elements have been changed.
- On loading the saved state, compare the checksum and give the user a warning that it may not be possible to completely restore everything. If an upper-level group name or relationship has changed, it may not be possible to restore anything.
- Saving/loading the state should be possible with toolbar buttons. Loading the state may also be an n2 command line argument.

Kenneth-T-Moore:
### Summary

New N2 capability to save and load views.
- Former "save svg" button is now a button group with 3 buttons: save svg, save view, load view.
- Removed some extra stuff in the icon.css that appeared to be left over from older versions (please check during review)
- Modified icon instructions so we don't lose one of the free icons.
- A hash of the model's data structure is now computed when the modeldata is saved/recorded so that we don't load a view from a different model

### Related Issues

- Resolves #1612

None

### New Dependencies

None


tadkollar:
The primary functionality of saving and restoring a view works well!

Some issues:
- the forward/back button history isn't restored after loading - I felt like I wanted to have it there, but it may not matter to others.
- there are two save dialogs that come up... one to just type in a filename and then another with the filesystem browser. But the same happens for `Save to SVG`, which hasn't been updated in awhile so I didn't notice when that began. Ideally we just want the filesystem browser. If this isn't trivial to fix, we should create a separate issue for it.
- if you try to load a file that isn't a state (or even a JSON file), it bombs. Is there a way to check the validity of the file first?
- the toolbar spacing and sizing got wonky. Some of the buttons shrank (orange annotation), the pullout arrows shifted right (purple annotation), and the pullout expansions are no longer flush (red annotation)
- pr_screenshotmaster_screenshot

Kenneth-T-Moore:
Thanks, Tad

1. _the forward/back button history isn't restored after loading_

Now that I have the basics working, I could probably add that.  Not sure how much that will impact the file size.

2. _there are two save dialogs that come up..._ 

Yeah, this is kind of clunky.  I'll look into this a little, but maybe you are right about it being a separate issue.

3. if you try to load a file that isn't a state (or even a JSON file), it bombs

Yeah, I should just check that the top level dictionary keys are there, and if they aren't, raise an error.

4. the toolbar spacing and sizing got wonky. Some of the buttons shrank (orange annotation), the pullout arrows shifted right (purple annotation), and the pullout expansions are no longer flush (red annotation)

Herb mentioned that his happened in the past, but didn't know how to fix it. Do you have any ideas? It changed when I created the new Icomoon font .woff file.

tadkollar:
> Herb mentioned that his happened in the past, but didn't know how to fix it. Do you have any ideas? It changed when I created the new Icomoon font .woff file.

We ran into this the last time we added a button to the toolbar... the problem is that portion of `icon.css` beneath the docs (i.e. the actual style definitions) is auto-generated by icomoon. It links the font's specific unicode characters to styles, so we copy and paste the css from the zip file that icomoon generates. If you need to make any mods to those styles, they need go into `toolbar.css` or one of the other style sheets.

openmdao/visualization/n2_viewer/src/N2Arrow.js
```diff
@@ -968,4 +980,86 @@ class N2ArrowManager {
...
     }
+
+    /* Save all pinnedArrows to a dictionary for saving the view. */
+    savePinnedArrows() {
+        let pinned = this.pinnedArrows;
+        let data = {};
+
+        for (var arrowId in pinned.arrows) {
+            let arrow = pinned.arrows[arrowId];
+
```
tadkollar, line 991: Using `var` makes a global, so it's usually avoided. Using `let` is OK because it limits the scope to the block, but if you're not going to reassign the variable, `const` is generally the way to go. It creates a constant reference (with block scope), so you can change its properties, but not reassign it. Here, `const arrowId in pinned.arrows` and `const arrow = pinned.arrows[arrowId]` would be good.

Kenneth-T-Moore: Thanks, I will fix these.

Kenneth-T-Moore:
_1. The forward/back button history isn't restored after loading_

Based on today's discussion, I modified the loadView code so that it saves the previous view to the back-button stack so the user can restore the view prior to load.

_2. there are two save dialogs that come up..._

I've created issue #1908 for this story.

_3. if you try to load a file that isn't a state (or even a JSON file), it bombs_

Fixed this. Now an alert box is generated to tell the user the file isn't a view.

_4. the toolbar spacing and sizing got wonky. Some of the buttons shrank (orange annotation), the pullout arrows shifted right (purple annotation), and the pullout expansions are no longer flush (red annotation)_

I didn't have any luck with this, and will hold off until your refactor of the icon system to remove the icomoon processing. Hopefully then it will be fully repeatable.

tadkollar (approved):

hschilling:
- Should the view also save the setting on what to show in the solver section?
- This might be a bug. I saved a view. Did a refresh of the page. Loaded in the view. Then did something to change the view. Hit the move back in history button. So far so good. But then hit the go forward button and nothing happens.
- Probably nothing you can do about this one but when you save a view twice in a row with the same name, the OS, at least on a Mac, saves it with a slightly modified name, which is good. But the you don't get the usual warning about overwriting and keeping both... It behaves just like a browser download.

Kenneth-T-Moore:
Hey, Herb. Thanks for checking it out.

- The view does save the solver setting. Seems to work for linear/nonlinear select as well as for just turning off the display of solvers.  I haven't been able to reproduce a case where it doesn't work.  I suppose there's also an argument that could be made that maybe I shouldn't save the solver state in the view, but I am not sure.
- I couldn't reproduce. One thing that can cause confusion is that forward/backward history only saves the zoomed node and collapsed state, so if your view has some highlighted some connections, those stay on when you go backwards in history.
- This is a good point. We have a story (#1908) up to completely redo the file interaction to use a browser (instead of having two dialog boxes), so I think (hope) it will be fixed with this story.

hschilling:
Now I can't make those two errors show up either so never mind! I was especially careful to document and do over and over the second one. Not sure why that isn't doing it now. 

One final comment. Since these are called "views" should the default file extension be ".view" instead of ".state" ?

swryan:
> One final comment. Since these are called "views" should the default file extension be ".view" instead of ".state" ?

or more specifically ".n2view"?

Kenneth-T-Moore:
Good points. I will make it ".n2view".

hschilling (approved):

DKilkenny (approved):

PR #1930, opened 2021-02-27
Add code excerpt tagging capability

bbrelje:
### Summary

Embedding longer scripts is ungainly. For certain tutorials and walkthroughs it doesn't make sense to break up a longer script into many functions or classes to make it possible to excerpt them using the existing sphinx build processes.

I added a capability to excerpt from scripts. In the script (or any code really, not just scripts), the developer can surround the code block with `# EXCERPT 1 #` tags, and then use a special `:EXCERPT:1` rst embed to access just that portion.

By default, these special excerpt comment tags are suppressed in all code printout unless the `:dont-strip-excerpt-tags:` option is used in that embed.

This is really for the pygeo project (we inherit from the openmdao sphinx extensions) but it could be a nice stopgap on the road to jupyter notebooks, which use this kind of flow.

None AFAIK, tests pass

### New Dependencies

None


robfalck:
We're moving towards using JupyterBooks for our docs, though we're still a ways away from a full implementation yet.  This should let us better document smaller code snippets, and allow users to immediately open our examples on a cloud-based service like Google Colab without having to install anything locally.  I think this is fine for our current docs, but it might not be in the code base for long.

JustinSGray:
we won't need similar kind of embedding with jupyter books? i guess not ... the code just goes directly into the doc itself... I hadn't thought about that.

openmdao/docs/_utils/docutil.py
```diff
@@ -282,6 +329,16 @@ class or None
     indent = 0
     class_obj = None
     method_obj = None
+    
+    # check if it's a script excerpt
+    if len(path.split(':EXCERPT:')) > 1:
```
Kenneth-T-Moore, line 334: I ultimately like this feature, and I think it would be a shame to limit it to just a single pair of tags in a function or file.  It should be workable even for tagged sections that are fully contained or cross boundaries with other tags.

With that said, I know that we are moving away from sphinx and towards notebooks (provided that all deficiencies can be solved), so maybe there isn't much justification for putting time into this...

Kenneth-T-Moore (changes requested):

openmdao/docs/_utils/docutil.py
JustinSGray: specificaly because we're moving away, I see no need to gold plate here. Limited functionality that can be grown later is fine with me.

JustinSGray: I had some discussions with outside folks who are using our docs tools about moving some of our sphinx embedding stuff into its own stand alone repo. So there will probably be life for this code after we move on anyway. I have no objection to including it in the codebase for now (assuming its apporpriately tested and passes code review)

bbrelje: @Kenneth-T-Moore I'm pretty sure that overlapping or nested tags should work as written? I'll give it a try when I update the closing tag

bbrelje:
OK, I think we can close this one without adding it

I did not know about the built-in sphinx literalinclude directive which can do the exact same thing as this PR on scripts (not with imported modules/functions/etc though, so I suppose there's a tiny benefit here)

These do the same thing:
```
.. embed-code::
    ../examples/ffd_cylinder/runFFDExample.py:EXCERPT:6
    :layout: code
```

```
.. literalinclude:: ../examples/ffd_cylinder/runFFDExample.py
    :start-after: # EXCERPT 6 #
    :end-before: # /EXCERPT 6#
```

But IMO it's not worth carrying along the cruft. I recommend closure but it's up to you

PR #1947, opened 2021-03-09
Overhaul internal storage of scaling vectors.

### LINKED ISSUE #1887: Look into excess scaling vector allocation
### Summary of Issue

We are potentially allocating memory for scaler/adder for both fwd and reverse modes

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

From discussion with users at DLR, they noted that in OpenMDAO V2.10 we were allocating separate vectors for fwd and rev modes. Im not sure if we are currently still doing this or not, but if so we don't need to. 

We can multiply by the scalers in fwd and divide by them in reverse mode. That will save 2 additional vector allocations. 

Investigate if we are still doing this un-needed allocation and update if necessary.

Kenneth-T-Moore:
### Summary

We have been storing scaling vectors for both physical->normalized and normalized->physical transformations because we suspected that the division might be costly. Benchmarking has shown that it isn't that significant, while the extra memory usage can be, so this PR modifies the scaling system to only store the scaling in one direction.

### Related Issues

- Resolves #1887

None

### New Dependencies

None


Kenneth-T-Moore:
Here is a simple, large problem that shows the effects of this change.

      from time import time
      import numpy as np
      import openmdao.api as om

      class BigComp(om.ExplicitComponent):

          def initialize(self):
              self.options.declare('num_nodes', default=1)

          def setup(self):
              nn = self.options['num_nodes']
              self.add_input('x', np.ones((nn, )))
              self.add_output('y', np.ones((nn, )), ref=1.5, ref0=0.5)

              outputs['y'] = 0.95 * inputs['x']

      N = 3000000
      prob = om.Problem()
      prob.model.add_subsystem('c1', BigComp(num_nodes=N))
      prob.model.add_subsystem('c2', BigComp(num_nodes=N))

      prob.model.connect('c1.y', 'c2.x')

      prob.setup()
      prob.final_setup()

      t0 = time()
      prob.run_model()
      print('run_model time', time() - t0)


Prior to this pull request:
memory: 1.71 GB  (cf. 839 MB if no vars have ref/ref0)
time: 0.5 sec

NEW
memory: 1.34 GB  (cf. 839 MB if no vars have ref/ref0)
time: 0.51 sec
'''

JustinSGray (approved):

naylor-b (approved):

JustinSGray (approved):

robfalck (approved):

PR #1956, opened 2021-03-12
Fixed N2 collapse-to-depth slider, added OpenMDAO version number

### LINKED ISSUE #1880: N2 collapse depth feature not working properly
### Summary of Issue

The N2 collapse feature on the toolbar does not work properly.

### Issue Type

- [X] Bug
- [ ] Enhancement  

### Description

The N2 collapse feature on the toolbar does not work properly. If you move the slider to the right, let the N2 redraw, and then move it back all the way to the left, you would expect the collapse depth to go back to where it was initially but it doesn't.

There are other odd behaviors like slightly changing the N2 at some depth settings. 

It also isn't very intuitive as to what it does. Would be helpful to see a number indicating the collapse depth as you slide it. 

### Example

You can see the behavior on the docs page https://openmdao.org/twodocs/versions/latest/features/model_visualization/n2_details.html

### Environment

NA

### LINKED ISSUE #1943: Include the OpenMDAO version in the generated N2
### Summary of Issue

Include the OpenMDAO version number in the N2 HTML output so we can tell what version generated it. Probably at the top of the file to make it easy to find. Maybe include it in the Help window so it can be found while the N2 is loaded.

### Issue Type

- [ ] Bug
- [X] Enhancement  
      POEM # N/A <N/A or POEM number at https://github.com/OpenMDAO/POEMs>

tadkollar:
### Summary

The collapse-to-depth slider was not being setup with the correct range of allowable depths, and it was difficult to determine what level was selected. The range was fixed and a small bubble is now next to the slider to display the depth. The selected level is correctly preserved in the back and forward history.

The version of OpenMDAO that generated the N2 is now added to the HTML file, and can be viewed in the diagram help window footer. Grepping the HTML text for `data-openmdao-version` will find it without loading up the N2 in a browser.

### Related Issues

- Resolves #1943, resolves #1880

None

### New Dependencies

None


hschilling (approved):

Kenneth-T-Moore (approved):
Tried this out. Seems a lot better now.

robfalck (approved):

Issue #1920, opened 2021-02-24
Improve docs for shape_by_conn


JustinSGray:
### Summary of Issue

Users work on parallel components are unsure of the behavior for `shape_by_conn`

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM # N/A
### Description

Users work on parallel components are unsure of the behavior for `shape_by_conn`. There is adaquate testing, but the lack of examples in the docs does not indicate what is and is not supported. 

Add two examples to these docs: 
http://openmdao.org/twodocs/versions/3.7.0/features/experimental/dynamic_shapes.html

1) serial to distributed connections 
2) distributed to distributed connections

anilyil:
It will also be nice to have a distributed to serial connection. I think by default it connects locally again, but I was not sure if it was going to allgather the distributed output and connect the full vector for each serial component.

JustinSGray:
superceeded by #1972

PR #1957, opened 2021-03-15
Test all combinations of dynamic variable sizing in parallel runs

anilyil:
### Summary

This PR adds a test to check if all possible connection sizes can be resolved with `shape_by_conn` feature in parallel runs. See the detailed explanation of the issue at: https://github.com/anilyil/OpenMDAO/blob/c760c3728ac403b163b0556aab4553322d8cb622/openmdao/core/tests/test_dyn_sizing.py#L858-L917 

To summarize, we have 3 different type of components in parallel runs: independent variable component, serial component, parallel component. IVCs only connect downstream (no inputs, just outputs), so we end up with 6 combinations of connections. We also have 2 directions the dynamic sizing can be resolved: fwd and bwd. As a result, we have 12 individual cases where the feature must be tested. 

The only current tests for parallel runs are defined in: https://github.com/anilyil/OpenMDAO/blob/c760c3728ac403b163b0556aab4553322d8cb622/openmdao/core/tests/test_dyn_sizing.py#L202 (there is a separate parallel group test but I am not including that here in this discussion). 
These parallel tests do not check some of the combinations. Here is a table to clarify which connections are tested in the current tests:
```
Connection:    Forward    Backward
ivc => ser       No         No
ivc => par       No         Yes
ser => ser       No         No
ser => par       Yes        No
par => ser       Yes        Yes
par => par       Yes        Yes
```

In this PR, I add a test that checks every single combination in the table. I am pretty sure some of these combinations result in an "unexpected" connection size. I think this PR is a good place to also fix these issues.

Once the code is modified and all tests work, we can also try to improve the documentation of `shape_by_conn` feature's parallel behavior by pointing the users to this test. 

### Related Issues

The documentation issue is in #1920. If we update the docs, we can also close that issue.

### DIFF
--- a/openmdao/core/tests/test_dyn_sizing.py
+++ b/openmdao/core/tests/test_dyn_sizing.py
@@ -946,6 +946,30 @@ class TestDistribDynShapeCombos(unittest.TestCase):
 
         # test all of the i/o sizes set by shape_by_conn
 
+        # ivc => serial
+        self.assertEqual(p.get_val('ser2.ivc_ser_fwd').size, var_shape)
+        self.assertEqual(p.get_val('ivc.ivc_ser_bwd').size, var_shape)
+
+        # ivc => parallel
+        self.assertEqual(p.get_val('par2.ivc_par_fwd').size, var_shape)
+        self.assertEqual(p.get_val('ivc.ivc_par_bwd').size, var_shape)
+
+        # serial => serial
+        self.assertEqual(p.get_val('ser2.ser_ser_fwd').size, var_shape)
+        self.assertEqual(p.get_val('ser1.ser_ser_bwd').size, var_shape)
+
+        # serial => parallel
+        self.assertEqual(p.get_val('par2.ser_par_fwd').size, var_shape)
+        self.assertEqual(p.get_val('ser1.ser_par_bwd').size, var_shape)
+
+        # parallel => serial
+        self.assertEqual(p.get_val('ser2.par_ser_fwd').size, var_shape)
+        self.assertEqual(p.get_val('par1.par_ser_bwd').size, var_shape)
+
+        # parallel => parallel
+        self.assertEqual(p.get_val('par2.par_par_fwd').size, var_shape)
+        self.assertEqual(p.get_val('par1.par_par_bwd').size, var_shape)
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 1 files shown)


anilyil:
The `p.setup()` call in the test does not even finish right now. Possibly because I screwed up something in the test. When I manually comment out all of the `_bwd` variables so that we only run the `_fwd` variables, at least one of the variable's size do not match my expectations. Also, there is a connection here from parallel to serial, and `get_value` does not work here with `get_remote=False`. I did not spend too much time on this but it would be great if we could check the local size of the variable on each proc. I think having `get_remote=True` would give us the total size.

joanibal:
Thanks for looking into this and adding more tests. 


I believe the test matrix should actually look like this (see below)
```
Connection:    Forward    Backward
ivc => ser       Yes        Yes
ivc => par       No         Yes
ser => ser       Yes        Yes
ser => par       Yes        Yes
par => ser       Yes        Yes
par => par       Yes        Yes
```

The reason there isn't complete test converge is because we never thought to treat ivc's different from other components. 
Could you explain why we need extra tests for ivc's?
What is the difference between the output of a independent variable component and output of a regular component? 


Here are the tests I think you may have overlooked

> Connection:    Forward    Backward
> ivc => ser       No         No

The forward is tested in `TestAdder:test_adder` and in the `TestDynShapes:test_baseline_series`

The Backward is tested in tested in `TestPassSize:test_serial` (A => B) and in `TestDynShapes:test_cycle_rev` (ivc is autoivc in this case)

> Connection:    Forward    Backward
> ser => ser       No         No

Both the forward and backward passing of variables is tested in `TestPassSize:test_serial`
See C => D for forward and B => C for backwards

This is tested extensively in `TestDynShapes`

> Connection:    Forward    Backward
> ser => par       Yes        No

This is tested in `TestPassSizeDistributed:test_serial_start`. C => D for the forward and B => C for the backward.

In `TestDistribDynShapes:test_remote_distrib` the backward is also tested. See 'par.G2.C2.y1' => 'sink.x2'.
It uses parallel Groups, but it is still testing this feature.

anilyil:
You are right, I thought ivcs were causing some extra issues initially because that is the first example that I ran into with mphys. Turns out they are no different than regular components. I was going to update this PR today, let me work that out and then I think we can have a discussion on it again. Sorry I could not post my updates after my discussion with justin yesterday.

You are right, the serial to serial stuff etc is tested, but only with a serial run. The behavior is not tested when we have 2 or more procs.

anilyil:
The new PR is up at #1964

PR #1964, opened 2021-03-17
Test all combinations of dynamic variable sizing in parallel runs, take 2

anilyil:
### Summary

This PR is a second try of the original PR in #1957. I wanted to start clean here because most of the discussion is outdated now.

This PR adds a test for the `shape_by_conn` feature with all combinations of serial/distributed components and the two modes of information transfer. The difference in this test compared to the existing ones is that we now test everything in a "parallel" run (i.e. more than one processor). So even though dynamic sizing of a serial component connected to another serial component is checked in other tests, it is only checked with a single processor. The test I am adding with this PR aims to check everything in a parallel run.

See the detailed explanation of the test at: https://github.com/anilyil/OpenMDAO/blob/4776e54ddf647846b93a3c47b04e399468636a31/openmdao/core/tests/test_dyn_sizing.py#L856-L916 

The checks marked with a TODO have issues currently: https://github.com/anilyil/OpenMDAO/blob/4776e54ddf647846b93a3c47b04e399468636a31/openmdao/core/tests/test_dyn_sizing.py#L979-L1003 

After discussing the possible pitfalls of propagating the dynamic shape in the "upstream" direction with @JustinSGray, we were moving towards disallowing the upstream mode. However, @joanibal is interested in maintaining it, so I kept the both directions in this test. In this PR, we should discuss if we will keep the upstream direction enabled, or if not, I will update this test to remove all examples of the upstream propagation.

Currently, I know of at least one bug with the dynamic shaping: In a serial to parallel component connection (see `ser_par_down` variable for an example), I would expect the local size of the serial output (which is the same on all procs) to be propagated to the parallel component in all procs. So say the serial component has an output of size 4 duplicated on all procs, I would have expected the parallel input to be sized 4 on all procs as well. However, currently the code just takes the size on one of the procs (size is 4) and "distributes" this vector among the rest of the processors. If running with 3 procs for example, the sizes each proc gets ends up being: 2,1,1, instead of 4,4,4. This is the case where shape by connection did not work in mphys (see: https://github.com/OpenMDAO/mphys/pull/51#discussion_r589012059)

Besides this, the "upstream" direction causes ambiguities with serial => parallel connections. In this case, we are sizing the input to the parallel component. This input size can be same on all procs with a variable that is duplicated, or the size on each proc can vary if we have a distributed input. 

Now, a serial component's output should have the same size on all processors. This is kinda the idea we had with the "duplicated serial components" in parallel runs. However, in this upstream dynamic shaping case, we cannot ensure that the input connected to the serial component's output will obey this. We can just throw an error if the user tries to connect a vector with different sizes on different procs this way between a serial and parallel component in the upstream mode. However, the user can have a distributed array, that happens to have the same size on all procs. In this case, we cannot throw the error because if the input to the parallel component has the same size on all procs, we have no way of knowing if the user expected this to be a distributed vector or a duplicated serial vector.

@joanibal and @JustinSGray, please discuss here if you want to keep the upstream mode of dynamic sizing. I personally do not care about removing it or not because my use cases do not rely on that feature. I have already spent way too much time on testing this, and would like to not spend more time on this. I am interested in fixing the dynamic sizing of serial to parallel connection in downstream mode.

### Related Issues

The documentation issue is in #1920. If we update the docs, we can also close that issue.

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1494,8 +1494,8 @@ class Group(System):
                     a2m = allprocs_abs2meta[abs_in]
                     # if (a2m['shape_by_conn'] or a2m['copy_shape']):
                     #     raise ValueError(f"{self.msginfo}: Setting of 'src_indices' along with "
-                    #                      f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
-                    #                      "is currently unsupported.")
+                    #                  f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
+                    #                  "is currently unsupported.")
 
                     if abs_in in abs2meta:
                         meta = abs2meta[abs_in]
@@ -1690,7 +1690,7 @@ class Group(System):
                         local_max = np.array([mx], dtype=INT_DTYPE)
                         global_max = np.zeros(1, dtype=INT_DTYPE)
                         self.comm.Allreduce(local_max, global_max, op=MPI.MAX)
-                        size = global_max[0] +1 # +1 becuase src_indices are 0 based
+                        size = global_max[0] + 1  # +1 becuase src_indices are 0 based
                     else:  # src_indices are not set, so just sum up the sizes
                         size = np.sum(distrib_sizes[from_var])
 
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1654,7 +1654,28 @@ class Group(System):
             elif from_io == 'output':
                 # from_var is an output.  assume to_var is an input
                 if from_dist and not to_dist:  # known dist output to serial input
-                    size = np.sum(distrib_sizes[from_var])
+                    # original
+                    # size = np.sum(distrib_sizes[from_var])
+
+                    # Anil's way
+                    rank = self.comm.rank
+                    # gets the local size
+                    size = distrib_sizes[from_var][rank]
+                    # gets the cumulative sum
+                    cumsum = np.cumsum(distrib_sizes[from_var])
+
+                    # AY: I dont know why we check to_meta here
+                    # but Josh's modification below checks it so I also checked it
+                    if to_meta:
+                        to_meta['src_indices'] = np.arange(
+                            cumsum[rank] - size,
+                            cumsum[rank],
+                            dtype=INT_DTYPE
+                        )
+
+                    # AY: do we need to set distrib_sizes here?
+                    distrib_sizes[to_var] = cumsum[-1]
+
                 else:  # known serial output to dist input
 
                     if all_to_meta['has_src_indices']:
@@ -1671,12 +1692,15 @@ class Group(System):
                         #                                     offsets[rank] + sizes[rank],
                         #                                     dtype=INT_DTYPE)
 
-                        # # Anil's way
+                        # Anil's way
                         # copy the serial output to all inputs
                         rank = self.comm.rank
                         size = from_size
                         if to_meta:
-                            to_meta['src_indices'] = np.arange(rank*size, (rank+1)*size, dtype=INT_DTYPE)
+                            to_meta['src_indices'] = np.arange(
+                                rank * size,
+                                (rank + 1) * size,
+                                dtype=INT_DTYPE)
 
                         distrib_sizes[to_var] = np.array([from_size]*self.comm.size)
 
@@ -1731,7 +1755,7 @@ class Group(System):
                     to_meta['value'] = np.full(size, to_meta['value'])
 
 
-            print(f'done connecting {from_var} to {to_var}', all_to_meta['size'] )
+            # print(f'done connecting {from_var} to {to_var}', all_to_meta['size'] )
 
 
         all_abs2prom_in = self._var_allprocs_abs2prom['input']
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1492,10 +1492,10 @@ class Group(System):
 
                 if src_indices is not None:
                     a2m = allprocs_abs2meta[abs_in]
-                    # if (a2m['shape_by_conn'] or a2m['copy_shape']):
-                    #     raise ValueError(f"{self.msginfo}: Setting of 'src_indices' along with "
-                    #                  f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
-                    #                  "is currently unsupported.")
+                    if (a2m['shape_by_conn'] or a2m['copy_shape']):
+                        raise ValueError(f"{self.msginfo}: Setting of 'src_indices' along with "
+                                     f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
+                                     "is currently unsupported.")
 
                     if abs_in in abs2meta:
                         meta = abs2meta[abs_in]
@@ -1638,6 +1638,7 @@ class Group(System):
 
             to_dist = nprocs > 1 and all_to_meta['distributed']
 
+            # print(f'connecting {from_var} to {to_var}', all_to_meta['size'] )
             # distrib to distrib or serial to serial
             if (from_dist and to_dist) or not (from_dist or to_dist):
                 # all copy_shapes (and some shape_by_conn) handled here
@@ -1661,14 +1662,25 @@ class Group(System):
                     else:
                         # there is not enough info to determine how the variable is split
                         # over the procs. for now we split the variable up equally
-                        rank = self.comm.rank
-                        sizes, offsets = self._evenly_distribute_sizes_to_locals(to_var, from_size)
-                        size = sizes[rank]
-                        distrib_sizes[to_var] = np.array(sizes)
+                        # rank = self.comm.rank
+                        # sizes, offsets = self._evenly_distribute_sizes_to_locals(to_var, from_size)
+                        # size = sizes[rank]
+                        # distrib_sizes[to_var] = np.array(sizes)
+                        # if to_meta:
+                        #     to_meta['src_indices'] = np.arange(offsets[rank],
+                        #                                     offsets[rank] + sizes[rank],
+                        #                                     dtype=INT_DTYPE)
+
+                        # Anil's way
+                        # copy the serial output to all inputs
+                        size = from_size
                         if to_meta:
-                            to_meta['src_indices'] = np.arange(offsets[rank],
-                                                            offsets[rank] + sizes[rank],
-                                                            dtype=INT_DTYPE)
+                            to_meta['src_indices'] = np.arange(0, size, dtype=INT_DTYPE)
+
+                        distrib_sizes[to_var] = np.array([from_size]*self.comm.size)
+
+
+
 
                 all_to_meta['size'] = size
                 all_to_meta['shape'] = (size,)
@@ -1681,6 +1693,19 @@ class Group(System):
                     sizes, _ = self._evenly_distribute_sizes_to_locals(to_var, from_size)
                     size = sizes[self.comm.rank]
                     distrib_sizes[to_var] = np.array(sizes)
+
+
+                    # Anil's way
+                    # copy the serial input to each output (assume the input is distubuted)
+                    # size = from_size
+
+                    # sizes = self.comm.allgather(from_size)
+                    # distrib_sizes[to_var] = np.array(sizes)
+                    # print(self.comm.rank, size, distrib_sizes[to_var])
+
+                    # all_to_meta['size'] =  np.sum(distrib_sizes[from_var])
+                    # all_to_meta['shape'] = (all_to_meta['size'],)
+
 
                 else:  # known dist input to serial output
                     if all_from_meta['has_src_indices']:
@@ -1696,10 +1721,15 @@ class Group(System):
 
                 all_to_meta['size'] = size
                 all_to_meta['shape'] = (size,)
+
                 if to_meta:
                     to_meta['size'] = size
                     to_meta['shape'] = (size,)
                     to_meta['value'] = np.full(size, to_meta['value'])
+
+
+            print(f'done connecting {from_var} to {to_var}', all_to_meta['size'] )
+
 
         all_abs2prom_in = self._var_allprocs_abs2prom['input']
         all_abs2prom_out = self._var_allprocs_abs2prom['output']
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1671,11 +1671,12 @@ class Group(System):
                         #                                     offsets[rank] + sizes[rank],
                         #                                     dtype=INT_DTYPE)
 
-                        # Anil's way
+                        # # Anil's way
                         # copy the serial output to all inputs
+                        rank = self.comm.rank
                         size = from_size
                         if to_meta:
-                            to_meta['src_indices'] = np.arange(0, size, dtype=INT_DTYPE)
+                            to_meta['src_indices'] = np.arange(rank*size, (rank+1)*size, dtype=INT_DTYPE)
 
                         distrib_sizes[to_var] = np.array([from_size]*self.comm.size)
 
@@ -1705,6 +1706,8 @@ class Group(System):
 
                     # all_to_meta['size'] =  np.sum(distrib_sizes[from_var])
                     # all_to_meta['shape'] = (all_to_meta['size'],)
+                    # if to_meta:
+                    #     to_meta['src_indices'] = np.arange(0, size, dtype=INT_DTYPE)
 
 
                 else:  # known dist input to serial output
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -1492,10 +1492,10 @@ class Group(System):
 
                 if src_indices is not None:
                     a2m = allprocs_abs2meta[abs_in]
-                    if (a2m['shape_by_conn'] or a2m['copy_shape']):
-                        raise ValueError(f"{self.msginfo}: Setting of 'src_indices' along with "
-                                         f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
-                                         "is currently unsupported.")
+                    # if (a2m['shape_by_conn'] or a2m['copy_shape']):
+                    #     raise ValueError(f"{self.msginfo}: Setting of 'src_indices' along with "
+                    #                      f"'shape_by_conn' or 'copy_shape' for variable '{abs_in}' "
+                    #                      "is currently unsupported.")
 
                     if abs_in in abs2meta:
                         meta = abs2meta[abs_in]
@@ -1655,16 +1655,21 @@ class Group(System):
                 if from_dist and not to_dist:  # known dist output to serial input
                     size = np.sum(distrib_sizes[from_var])
                 else:  # known serial output to dist input
-                    # there is not enough info to determine how the variable is split
-                    # over the procs. for now we split the variable up equally
-                    rank = self.comm.rank
-                    sizes, offsets = self._evenly_distribute_sizes_to_locals(to_var, from_size)
-                    size = sizes[rank]
-                    distrib_sizes[to_var] = np.array(sizes)
-                    if to_meta:
-                        to_meta['src_indices'] = np.arange(offsets[rank],
-                                                           offsets[rank] + sizes[rank],
-                                                           dtype=INT_DTYPE)
+
+                    if all_to_meta['has_src_indices']:
+                        size = len(to_meta['src_indices'])
+                    else:
+                        # there is not enough info to determine how the variable is split
+                        # over the procs. for now we split the variable up equally
+                        rank = self.comm.rank
+                        sizes, offsets = self._evenly_distribute_sizes_to_locals(to_var, from_size)
+                        size = sizes[rank]
+                        distrib_sizes[to_var] = np.array(sizes)
+                        if to_meta:
+                            to_meta['src_indices'] = np.arange(offsets[rank],
+                                                            offsets[rank] + sizes[rank],
+                                                            dtype=INT_DTYPE)
+
                 all_to_meta['size'] = size
                 all_to_meta['shape'] = (size,)
                 if to_meta:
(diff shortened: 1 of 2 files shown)


joanibal:
To get all the tests to run I fixed one bug, fixed the test cases, and applied one proposed change.
  
 ## Dynamic size setup bug fix 
When src indices are used to pass size info from dist input to a serial output, the size of the output is determined by the maximum value of the src_indices. In this case the size is equal to the max_src_inc +1 because the source indices start at zero. The +1 was missing. 


## Test Case Fixes
1) Before `Ser2` had different sized input for `par_ser_up` on each proc. This is because the variable `var_size` was different depending on the comm.rank.
```
class ser2(om.ExplicitComponent):
    def setup(self):
        rank = self.comm.rank
        var_shape = 2 * rank
```
This is why `par_ser_up` was not working in setup.

2) The parallel tests were failing because they were using a bad reference value.
 In `test_dyn_shape_combos` self.comm.rank was 0 for all procs. 
 `rank = MPI.COMM_WORLD.rank` fixed the issue.
  
3) To specify that all the inputs of Par2.ser_par_up come from the same source, you can add src_indices to the add_input or the connection. 
 
These changes along with the bugfix fixed all but one test (`ser1.ser_par_down` to `par2.ser_par_down`)
 
 
 ## Suggested prototype changes
 
>Currently, I know of at least one bug with the dynamic shaping: In a serial to parallel component connection (see ser_par_down variable for an example), I would expect the local size of the serial output (which is the same on all procs) to be propagated to the parallel component in all procs.

I wouldn't say this is a bug, although I think we can modify the code to fix your issue. 
There are no docs on the dynamic shape feature, so I'm not surprised it didn't work exactly how you expected. 
As you know, when going from a serial output to a distributed input the source indices can be set to either use the whole output or just a slice. Right now it assumed that the output should be evenly distributed among the procs, so the inputs use just a slice. 

I see the fact that the whole serial output lives on each proc as an implementation detail that I wouldn't expect the average person to know. So I don't think it is fair to say that it should be expected to propagate size information about the whole output to each distributed input. 

To resolve the ambiguity of the distributed input src_indices, I modified the code to allow src_indices to be specified for `shape_by_conn` sources in `connect` and `_setup_global_connections` to use the new information.

Let me know what you think about these changes.

anilyil:
1. I actually had a different sized input for `par_ser_up` on purpose. This is a common use case actually. Think of this example: ADflow solver component is distributed, and outputs the distributed state. We then connect this to ADflow functionals, which is a serial component. The input to the serial component will have different sizes on different procs. This should be tested, and that is why I added the test this way. I guess the downstream direction worked fine, but the upstream did not. If we will maintain the upstream mode of shape propagation, I would expect both versions to do the same, so the upstream should also be able to handle varying variable sizes on each proc. 

2. That is so weird. Do you know why that is happening? Good catch! 

3. I did not really understand what you meant here. 

Regarding the final modification; I just want to be able to use shape by conn to connect a serial output that is duplicated to a parallel component's input. I want this input to be also duplicated on all procs, in the sense that "we just propagate the local shape and values" will be enough to explain the situation.

So lets say I have 3 processors with a serial component outputting a variable size 4 on each proc. The total vector ends up being 12 entries, but it is really 4 entries duplicated on each proc. When I connect this duplicated output to an input of a serial group, it would just copy the local shape directly right? So it would put the same 4 duplicated entries on all of the procs. Now, the way I left the code, this had a different behavior if I connected the variable to a parallel component. In that case, it would pick the entries on the root proc (of size 4) and tried to evenly distribute it among the inputs to the parallel group. This is not only counter-intuitive for me, it is also not super useful imo. The expected behavior for me would be the same as if I connected this to a serial component. Now, I did not fully understand the result of the changes you made @joanibal , I am happy if it follows what I just outlined, I am not happy if the previous behavior is still the default.

JustinSGray:
> To resolve the ambiguity of the distributed input src_indices, I modified the code to allow src_indices to be specified for shape_by_conn sources in connect and _setup_global_connections to use the new information.

@joanibal my objection to this is that, connections are done at the containing group level, and during the configure of that group. So that means that you basically have all the size information you need at that point, and you can hand to to the upstream component so it can size its own variables. I don't think that you've actually simplified it, I think you've split key information up so it now lives in two places (some in the component definition and some in the connect statement)

joanibal:
@anilyil 

1) 
> We then connect this to ADflow functionals, which is a serial component. The input to the serial component will have different sizes on different procs

From the openMDAO docs. 

> A distributed component is a component that operates on distributed variables. A variable is distributed if each process contains only a part of the whole variable.

So if each of the functionals component operates on distributed variables, by the definition of the purpose of distributed components in the docs, it really ought to be a distributed component. 

@JustinSGray, Am I wrong in thinking serial components shouldn't be used like this?

3) 
>  Now, I did not fully understand the result of the changes you made @joanibal , I am happy if it follows what I just outlined, I am not happy if the previous behavior is still the default.

The ser1 => par2 passes now, so if you are happy with how the test case looks then it sounds like we are on the same page. 

It's not difficult to change the default behavior to work exactly how you envisioned, but we still need to agree on a way to resolve the ambiguity.

anilyil:
> So if each of the functionals component operates on distributed variables, by the definition of the purpose of distributed components in the docs, it really ought to be a distributed component.
I think that is for outputs. The functionals group outputs the same functionals duplicated on all procs, and thats why we keep it as a serial component.

naylor-b:
A distributed component has both distributed inputs and distributed outputs.  You're not allowed to have one without the other.  That's why the 'distributed' option is associated with the component and not with individual variables.

JustinSGray:
The docs are not specific enough here. 

The distributed-ness of a component affects it outputs, but not its inputs. It doesn't really make sense to talk about an input as distributed or not, because it does not provide any additional context. 

The functionals component should be serial, not distributed. because its outputs are serial (i.e. the local and global size of the outputs should be the same)

joanibal:
@JustinSGray 

>  So that means that you basically have all the size information you need at that point, and you can hand to to the upstream component so it can size its own variables.

I didn't follow what you are saying here.

>  I don't think that you've actually simplified it, I think you've split key information up so it now lives in two places (some in the component definition and some in the connect statement)

Would moving the src_indices info into `add_input` and `add_output` address the your concern?
I'd be happy to move the information about the src_indices to `add_input` and `add_output`. However, that way of doing things is deprecated.

anilyil:
>The ser1 => par2 passes now, so if you are happy with how the test case looks then it sounds like we are on the same page. 

I just checked but you specified the `src_indices` to get it working. Also, it should not just be 0,1,2,3 on all procs. the written out src indices should be more like rank+0, rank+1, rank+2, rank+3 on all procs in my opinion. Also, this should be the default behavior so that we dont need to specify src_indices

joanibal:
> The docs are not specific enough here.
> 
> The distributed-ness of a component affects it outputs, but not its inputs. It doesn't really make sense to talk about an input as distributed or not, because it does not provide any additional context.
> 

So you can have one without the other?


> The functionals component should be serial, not distributed. because its outputs are serial (i.e. the local and global size of the outputs should be the same)

So it doesn't matter that the local output values could be different? 
What would the value of get_val('my_comp.out', get_remote=True) return???

joanibal:
> > The ser1 => par2 passes now, so if you are happy with how the test case looks then it sounds like we are on the same page.
> 
> I just checked but you specified the `src_indices` to get it working. Also, it should not just be 0,1,2,3 on all procs. the written out src indices should be more like rank+0, rank+1, rank+2, rank+3 on all procs in my opinion. Also, this should be the default behavior so that we dont need to specify src_indices

@anilyil I changed the default behavior to match what you said

anilyil:
Thank you, I think it looks like what I expect. I am happy with the current version of serial to parallel connection in the downstream mode.

Regarding the upstream mode; since the downstream version works w/o src_indices, shouldn't the upstream version also work w/o src_indices? And it should also get the same outcome as if we are running the downstream version. There is also a caveat with this; the input to the parallel comp might not have the same size on all procs, which then results in an invalid connection since the serial components have identical outputs on all procs. This was the original issue that led me and @JustinSGray to think that removing the upstream mode might just be simpler here.

joanibal:
> Regarding the upstream mode; since the downstream version works w/o src_indices, shouldn't the upstream version also work w/o src_indices? 

The downstream does use src_indices. In this case, they are just assumed to be np.arange(0,size) to get the default behavior you wanted 

> There is also a caveat with this; the input to the parallel comp might not have the same size on all procs, which then results in an invalid connection since the serial components have identical outputs on all procs. This was the original issue that led me and @JustinSGray to think that removing the upstream mode might just be simpler here.

Could you explain this a bit more. An example would help me understand.

anilyil:
>The downstream does use src_indices. In this case, they are just assumed to be np.arange(0,size) to get the default behavior you wanted

Oh, didnt realize that. In that case, I think the assumed src_indices should be `np.arange(rank*size, (rank+1)*size)` This would locally connect all of the duplicated vector. The way you have it currently would only use the rank 0's output and distribute it to every proc for the parallel component, which is unnecessary communication. 

When I say "not using src_indices", I mean not using it in the runscript. So the assumed src indices should connect stuff locally on each processor. 

The upstream case should also "not use src_indices" in the runscript, and the assumed indices in openmdao should be the same as the downstream version. 

>Could you explain this a bit more. An example would help me understand. 

Say the user added an input to the parallel group and shaped it to have different sizes on each proc. Then the user connected this to an output of a serial component, with shape_by_conn=True for the output. This should raise an error because just propagating the local variable sizes would result in an output in the serial component with varying sizes on each proc, which is not allowed.

joanibal:
>     1. I actually had a different sized input for `par_ser_up` on purpose. This is a common use case actually. Think of this example: ADflow solver component is distributed, and outputs the distributed state. We then connect this to ADflow functionals, which is a serial component. The input to the serial component will have different sizes on different procs. This should be tested, and that is why I added the test this way. I guess the downstream direction worked fine, but the upstream did not. If we will maintain the upstream mode of shape propagation, I would expect both versions to do the same, so the upstream should also be able to handle varying variable sizes on each proc.
> 

In regard to 1 (from a while ago).
I haven't been able to get the setup to work with different inputs sizes for `Ser2.par_par_up` *even when not using shape_by_conn.*

The error always occurs here is _setup_connections
https://github.com/OpenMDAO/OpenMDAO/blob/e96aa063d04330ed0aedece365b7f0b8717aad3b/openmdao/core/group.py#L1983-L2007
I don't see how it would be possible for the different `meta_in['shape'] of `Ser2.par_par_up` on each proc to all match the all_meta_out['global_shape'] of `Par2.par_par_up`

anilyil:
we do this in adflow with mphys. the solver group is distributed: https://github.com/OpenMDAO/mphys/blob/master/mphys/mphys_adflow.py#L250 and each processor outputs the local state: https://github.com/OpenMDAO/mphys/blob/master/mphys/mphys_adflow.py#L283

The functionals group is serial: https://github.com/OpenMDAO/mphys/blob/master/mphys/mphys_adflow.py#L786 and takes in the local state as an input: https://github.com/OpenMDAO/mphys/blob/master/mphys/mphys_adflow.py#L819 

This actually works too if you remove the src_indices and change it to shape_by_conn I believe.

joanibal:
I saw that. 

But I can't figure out why it doesn't work for your test case.

anilyil:
I think the test works, the issue is with get_val. get_val probably has a bug with that type of inputs to the serial comp. I think we can check in the compute statement like the other tests to make sure it works. (I was dumb and didnt realize that was a better way to check the sizes and did it with get_val). Also, get_val returns the correct total size once you do get_remote = True. This is the par_ser_down connection.

joanibal:
> I think the test works, the issue is with get_val. get_val probably has a bug with that type of inputs to the serial comp. I think we can check in the compute statement like the other tests to make sure it works. (I was dumb and didnt realize that was a better way to check the sizes and did it with get_val). Also, get_val returns the correct total size once you do get_remote = True. This is the par_ser_down connection.

For me it always fails in the setup

>The error always occurs here is _setup_connections

I'm giving up on it at this point

joanibal:
> > The downstream does use src_indices. In this case, they are just assumed to be np.arange(0,size) to get the default behavior you wanted
> 
> Oh, didnt realize that. In that case, I think the assumed src_indices should be `np.arange(rank*size, (rank+1)*size)` This would locally connect all of the duplicated vector. The way you have it currently would only use the rank 0's output and distribute it to every proc for the parallel component, which is unnecessary communication.
> 
it's changed now.

joanibal:
> > Could you explain this a bit more. An example would help me understand.
> 
> Say the user added an input to the parallel group and shaped it to have different sizes on each proc. Then the user connected this to an output of a serial component, with shape_by_conn=True for the output. This should raise an error because just propagating the local variable sizes would result in an output in the serial component with varying sizes on each proc, which is not allowed.

I agree that case should raise an error.

I'd say the input of that distributed component would need `src_indices`, since each proc is dealing with different slices of the data. You would need `src_indices` whether you decided to use `shape_by_conn` or not. 
 With the info about the `src_indices`, you can figure out what size the output of the serial component should be when using `shape_by_conn` like we do now by finding the largest index.

anilyil:
>For me it always fails in the setup


I just pulled from the branch again and the test passes as is. I am referring to the check:
```
        # parallel => serial
        # TODO the get_val does not work on this parallel output. need get_remote=True
        # self.assertEqual(p.get_val('ser2.par_ser_down').size, var_shape)
        # so we do it with get_remote=True
        self.assertEqual(p.get_val('ser2.par_ser_down', get_remote=True).size,var_shape_all)
```
To me this does work in the test, and also in practice with mphys. 


> I agree that case should raise an error.

So we were leaning towards removing the upstream mode because while that case would lead to an error, how about a "distributed" input that happened to have the same size on all procs. This can happen, and we have no way of distinguishing if the user meant to add a distributed vector as an input to the distributed comp or if it is a regular vector with duplicated entries on all procs. 

> I'd say the input of that distributed component would need src_indices, since each proc is dealing with different slices of the data. You would need src_indices whether you decided to use shape_by_conn or not. With the info about the src_indices, you can figure out what size the output of the serial component should be when using shape_by_conn like we do now by finding the largest index

So I am not sure what you mean by still using src_indices even with shape_by_conn. Are you referring to the runscript only or internally assuming a src_indices. When I say "not using src_indices", I am referring to not specifying this in the runscript. Also in my opinion, if you are specifying src_indices already in the runscript, then shape_by_conn has no point; you should already have all the shape info you need if you can prescribe src_indices.

JustinSGray:
@naylor's comment isn't wrong, but it is a subtle issue that isn't easy to explain susinctly. 

When you have a serial component, by definition the inputs on each proc have the same size. Their source indices are set up to use local-data transfers, but you can say that they will always be interger offsets of eachother where the offset is equal to the local size.

In a distributed component you could set up inputs to behave exactly the same way (i.e. same local size with integer offsets equal to the local size), or you could set them up differently (i.e. varying local sizes, and non-correlated offsets) 

The challenge is with inputs that are set up as in a distributed fashion (i.e. varying local sizes, and non-correlated offsets), which can only happen in on distributed components (hence' Bret's comment). However, it doesn't have to happen. Some inputs for distributed components will behave in a serial fashion. Its really a function of the connection into them. 

And this is where `size_by_conn` gets tricky IMO. The serial-like behavior is the most intuitive to me. I realize that is not how it works right now, but I think thats a mistake and would like to change it. The problem is there is no unique solution for the distributed like behavior unless you know BOTH the size information and the src_index information. However, src_idex information doesn't belong in the component. Its a feature of the connection, so it gets messy. 

As a compromise, we could allow downstream and upstream `shape_by_conn` but throw an error if the src is serial, and the target has different sizes on different processors. We would allow upstream if the target had the same size on all procs and the source is serial. I personally feel that this is an acceptable option, but Im not convinced that we need the feature and hence I still want to take it away till we have a real use case. 

I can commit to fixing the problem with auto-ivc, so that we don't need upstream for any IVC situation. That just leaves the data-transfer. Currently the mphys refactor doesn't use the upstream stuff though, and I don't think Anil or Kevin agree with you on that detail. So Im trying to understand the use case better

joanibal:
> > For me it always fails in the setup
> 
> I just pulled from the branch again and the test passes as is. I am referring to the check:
> 
> ```
>         # parallel => serial
>         # TODO the get_val does not work on this parallel output. need get_remote=True
>         # self.assertEqual(p.get_val('ser2.par_ser_down').size, var_shape)
>         # so we do it with get_remote=True
>         self.assertEqual(p.get_val('ser2.par_ser_down', get_remote=True).size,var_shape_all)
> ```
> 
> To me this does work in the test, and also in practice with mphys. There is a possible issue with get_val but the code itself looks okay.

I was talking about `Ser2.par_ser_up`. If that variable is set to have a different shape on each proc (see comments in code). Then it fails during set up whether or not you use `shape_by_conn`.

How do you know it is working if you have to use get_remote==True?

I printed the sizes of the inputs in the compute and this is what I got

```

id, sizes in Ser2
0 ser_ser_down 4
0 ser_ser_up 4
0 par_ser_down 6
0 par_ser_up 6
id, sizes in Ser2
1 ser_ser_down 4
1 ser_ser_up 4
1 par_ser_down 6
1 par_ser_up 6
id, sizes in Ser2
2 ser_ser_down 4
2 ser_ser_up 4
2 par_ser_down 6
2 par_ser_up 6

```

all the inputs have the same size, they shouldn't right? 
This doesn't change it you remove 'shape_by_conn' either


> So I am not sure what you mean by still using src_indices even with shape_by_conn. Are you referring to the runscript only or internally assuming a src_indices. When I say "not using src_indices", I am referring to not specifying this in the runscript. Also in my opinion, if you are specifying src_indices already in the runscript, then shape_by_conn has no point; you should already have all the shape info you need if you can prescribe src_indices.

I mean using src_indices for the input of the dist comp which isn't using `shape_by_conn`

joanibal:
> The problem is there is no unique solution for the distributed like behavior unless you know BOTH the size information and the src_index information. However, src_idex information doesn't belong in the component. Its a feature of the connection, so it gets messy.

Ok, I think we are all on the same page here. I totally agree that you need the `src_index` info for a unique solution. 

Are you saying that with `shape_by_conn` the src_index in connections get particularly messy? Or just that using src_indices with connections is messy in general?

> As a compromise, we could allow downstream and upstream `shape_by_conn` but throw an error if the src is serial, and the target has different sizes on different processors. We would allow upstream if the target had the same size on all procs and the source is serial. I personally feel that this is an acceptable option, but Im not convinced that we need the feature and hence I still want to take it away till we have a real use case.

This is where you're losing me. I'm still not seeing the difference between upstream and downstream. In my mind both need src_indices to be fully described and neither can be used otherwise without making assumptions. 

I'm talking with Anil tomorrow, so hopefully that will help me understand your concerns better.

JustinSGray:
You're right that they both need src indices. The difference is that with downstream, it is always correct. The input gets sized by the serial source. With upstream, there is the possibility of an error if the inputs on each proc don't have the same size. While we could error check for this, the need for it means that there is more complexity in upstream than downstream. 

This is why I don't think we need to support upstream right now. I think, since were not currently using it for any known cases that we should dis-allow it. We can bring it back later if the need arises. In general, I feel that adding un-used functionality for the sake of future-proofing is a mistake, and I think that removing upstream capabality here falls under that category. 

I won't make the final call till I understand the difference between the way you want to handle the transfers and the way that its currently done in Mphys though.

anilyil:
>I was talking about Ser2.par_ser_up. If that variable is set to have a different shape on each proc (see comments in code). Then it fails during set up whether or not you use shape_by_conn.

I see, yeah that one is the messy one to begin with. I am not sure why its doing that. I don't think I ever used that case in mphys.

>How do you know it is working if you have to use get_remote==True?
>I printed the sizes of the inputs in the compute and this is what I got

I guess I was assuming it ended up being okay, but maybe I was wrong. I think it definitely worked in the mphys case where I use that type connection between adflow solver and functionals, but maybe I messed up that connection as well. I will check this again. I also had some issues with testflo printing under mpi in some cases but I could not look into it more...

anilyil:
@joanibal you were right, parallel to serial downstream is also not working. I thought it was fine but I missed it.  

I will check the current state of this test to guide the efforts to fix this. I can also try to fix this myself but my know-how is limited. I will try to at least check and update the test by the end of today.

anilyil:
I updated the test a bit. Currently, I look at only the downstream direction, will work on the upstream as well, but it is not a priority for me. 

I will try to explain the cases I modified in the openmdao code below. When I say v3.8.0, I mean the latest tagged version at this state. Previous versions possibly have these issues as well.

Among the downstream connection cases, I was happy on v3.8.0 with serial=>serial and parallel=>parallel connections.

The case serial=>parallel was doing sth unexpected for me in v3.8.0. I would expect all of the duplicated output of the serial component to be connected to the input of the parallel component locally on each proc. The src indices then look like: `np.arange(rank*size, (rank+1)*size)`. Josh fixed this here in this PR: https://github.com/OpenMDAO/OpenMDAO/pull/1964/files#diff-5545bcee157fc878e12c8cccee1a1386344ecdf769d7cbbcdcb5c1ce6b59415aR1695-R1705

Also, the case parallel=>serial was doing sth unexpected for me in v3.8.0. The code by default was gathering all of the distributed vector on all processors, and adding the full vector as input on all procs. Again, I was expecting the distributed output to be connected to the input of the serial components using the local shapes. The src_indices for my expectation looks sth like: `np.arange(np.cumsum(sizes)[rank]-sizes[rank], np.cumsum(sizes)[rank])` I fixed this here in this PR: https://github.com/OpenMDAO/OpenMDAO/pull/1964/files#diff-5545bcee157fc878e12c8cccee1a1386344ecdf769d7cbbcdcb5c1ce6b59415aR1660-R1677

I don't know about @joanibal's fix, but I have no idea about the details of the code in my fix, so my fix is likely wrong. I just had it in there to have a working example. 

I check the i/o sizes in 2 places:

1. Inside the compute method of all components. With Josh's and my fixes, all downstream checks work.
2. After run_model, via `get_val`. `get_val` does not work for the two variables: `ser_par_down` and `par_ser_down`. I have commented this in the test. This is not critically important for me; the fact that the tests pass inside the compute method is good enough. I may be doing something illegal with `get_val` to begin with.

I think the next steps here is to update the upstream tests to mimic the downstream tests, and also apply the fixes if necessary.

JustinSGray:
superseded by #2013

PR #1968, opened 2021-03-18
Suppress warnings for MPI/PETSc failed import

### LINKED ISSUE #1917: Make MPI/Petsc Warnings less annoying/confusing for users
### Summary of Issue

The warning: 

`Unable to import petsc4py. Parallel processing unavailable` 

confuses many users (or at least they find it annoying). Its not clear where the warnings come from (i.e. which package). They could be made more clear, and also be changed from warnings to errors that only get raised when some functionality is accessed that won't work without petsc4py or MPI4py. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM # N/A

### Description

The warnings are thrown here: 

https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/total_jac.py#L30-L36

https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/utils/mpi.py#L66-L77

https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/linear/petsc_ksp.py#L19-L26

You basically get all three any time you import OpenMDAO, regardless of whether or not you're using any mpi/petsc stuff. There should be a less annoying way to do this... 

Change warnings to errors, and raise them only when the functionality is accessed (rather than just imported)? 

Im open to other suggestions, but we need to clean this up somehow.

tadkollar:
### Summary

Previously, if `OPENMDAO_REQUIRE_MPI` was unset and MPI or petsc4py could not be imported, a warning would be printed. This has been changed so that the import fails silently, but there will be an error later on if MPI/petsc functionality is accessed. The user can still set `OPENMDAO_REQUIRE_MPI=1` (or other positive values such as `true`, `yes`, etc) to either cause the code to fail early, or set `OPENMDAO_REQUIRE_MPI=0` to not attempt an import at all.

The docs and a test were updated with the new behavior. The RuntimeError that used to print "PETSc is not available. " now prints "PETSc is not available. Set shell variable OPENMDAO_REQUIRE_MPI=1 to detect earlier.

### Related Issues

- Resolves #1917

None

### New Dependencies

None


openmdao/core/total_jac.py
```diff
@@ -17,23 +17,27 @@
...
+    # If OPENMDAO_REQUIRE_MPI is set to a recognized positive value, import
+    # petsc4py and raise ImportError on failure.
+    if os.environ['OPENMDAO_REQUIRE_MPI'].lower() in ['always', '1', 'true', 'yes', 'y', 'on']:
+        try:
+            from petsc4py import PETSc
+
+        except ImportError:
+            raise ImportError("Importing petsc4py failed and OPENMDAO_REQUIRE_MPI is true.")
     else:
+        # OPENMDAO_REQUIRE_MPI is set to something else, so no import is attempted.
```
swryan, line 31: It seems like `OPENMDAO_REQUIRE_MPI` means it's okay either way, as opposed to "I don't want MPI"

I would expect the test to be:

> if `OPENMDAO_REQUIRE_MPI` exists and is truthy, then failure to import MPI/PETSc is an error, otherwise try and fail silently

tadkollar: I wanted to keep an option to prevent the attempt, in case it _is_ installed but you don't want to load it for some reason. Maybe it's broken (can happen after updates) or you just want to be confident that you're doing a serial run.

swryan: I guess then it's the var name that just doesn't convey the meaning to me..   
This sounds more like `OPENMDAO_USE_MPI`  
(True to use it and fail if you can't, False to not even try)

tadkollar: Agree, but the problem with changing the var name at this point is that `OPENMDAO_REQUIRE_MPI` has been around for almost 2 years. The only difference with this PR is what happens when it's not set.

swryan (approved):

Kenneth-T-Moore (approved):

PR #1984, opened 2021-03-29
Added notebook utils for Jupyter Book

DKilkenny:
### Summary

Added notebook utils for Jupyter Book. Added function to 
- Show the source code 
- Show the options table of a class
- Show a citation

### Related Issues

- Resolves #N/A

None

### New Dependencies

None


openmdao/api.py
```diff
@@ -95,6 +95,9 @@
 # Units
 from openmdao.utils.units import convert_units, unit_conversion
 
+# Docs Util
+from openmdao.docs.jb_utils import display_source, show_options_table, cite
+
```
swryan, line 100: seems like we don't need these in the `api` module.. these aren't really user-facing?

openmdao/docs/jb_utils.py
```diff
@@ -0,0 +1,75 @@
+import importlib
+import inspect
+from IPython.display import display, HTML, Code
+
+
+def get_object_from_reference(reference):
```
swryan, line 6: could use a docstring here, and on the module itself 
(it's not  really clear what `jb` stands for... I assume jupyter book, but elsewhere we use `notebook`.. maybe `notebook_utils` is better?)

openmdao/api.py
DKilkenny: They are technically user facing since contributors can use that to display source code, options, etc

openmdao/api.py
```diff
@@ -95,6 +95,9 @@
 # Units
 from openmdao.utils.units import convert_units, unit_conversion
 
+# Docs Util
```
Kenneth-T-Moore, line 98: General comment, would this api be preferable:
    `om.Component.show_options_table()`
(or in long form:)
    `openmdao.core.component.Component.show_options_table()`
rather than:
    `om.show_options_table("openmdao.core.component.Component")`

openmdao/docs/jb_utils.py
DKilkenny: I'll add that docstring. I felt that since these function should only be used for docs, jb_utils is more appropriate than `notebook_utils`.

openmdao/api.py
swryan: I would consider contributors to be developers rather than users..

DKilkenny: The other reason I did it this way was that at the top of every notebook we `import openmdao.api as om` instead of just `import openmdao` since this is in a separate repo and users opening these books in Colab don't automatically have the full import. I don't think @robfalck wants to add the full import just to access those func so that's why I put the funcs in the API

openmdao/docs/jb_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        Source code of the given class or function.
+    """
+    obj = inspect.getsource(_get_object_from_reference(reference))
+
+    if hide_doc_string:
+        obj = obj.split('"""')
+        del obj[1]
+        obj = ''.join(obj)
+
+    return Code(obj, language='python')
```
robfalck, line 55: this Code object isn't strictly a string

Kenneth-T-Moore (approved):

openmdao/docs/jb_utils.py
robfalck, line 55: It seems like this shouild actually call `display(Code(...` rather than just returning the code object.

robfalck (changes requested):
Should jb_utils.py be called notebook_utils.py - they're general notebook features, not specific to JupyterBook?  Also could move the existing notebook testing functionality there.

openmdao/docs/jb_utils.py
DKilkenny: Added display

openmdao/core/notebook_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    module = None
+    while split:
+        try:
+            module = importlib.import_module('.'.join(split))
+            break
+        except ModuleNotFoundError:
+            right.append(split.pop())
+    if module:
+        for entry in reversed(right):
+            module = getattr(module, entry)
```
naylor-b, line 43: Naming is a little confusing here.  You're not returning a module, but an object in a module, correct?

openmdao/core/notebook_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    ipy = False
+    try:
+        from IPython import get_ipython
+        ipy = get_ipython() is not None
+    except ImportError:
+        pass
+
+    if ipy and tabulate is None:
+        simple_warning("Tabulate is not installed run `pip install openmdao[notebooks]` to "
```
naylor-b, line 125: Need a period after 'installed'.

openmdao/core/notebook_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns
+    -------
+    IPython.display
+        Source code of the given class or function.
+    """
+    obj = inspect.getsource(_get_object_from_reference(reference))
+
+    if hide_doc_string:
+        obj = obj.split('"""')
```
naylor-b, line 66: Not sure if 'reference' can ever be a class here, but if so, getsource would return all of its source including methods and their docstrings, and your split on '"""' and rejoin on '.' would print all of the function docstrings with '.'s replacing the '"""'s.  If reference can't be a class, maybe put something in the docstring stating that (or stating that reference must be a function or whatever).

naylor-b (changes requested):

openmdao/core/notebook_utils.py
DKilkenny: Fixed

openmdao/core/notebook_utils.py
```diff
@@ -40,18 +40,18 @@ def _get_object_from_reference(reference):
             right.append(split.pop())
     if module:
         for entry in reversed(right):
-            module = getattr(module, entry)
-    return module
+            obj = getattr(module, entry)
```
naylor-b, line 43: This won't work because it's doing a getattr repeatedly using 'module' which now never updates.

naylor-b (changes requested):

openmdao/core/notebook_utils.py
DKilkenny: Missed that commit. Changed

naylor-b (approved):

robfalck (approved):

swryan (approved):

PR #1986, opened 2021-03-31
Fixed a bug where the global_size was incorrect for a design variable that is specified with indices on an IVC output declared as a distributed component.

### LINKED ISSUE #1965: Check Totals Bug with distributed components and DV indices.
### Summary of Issue

When doing `check_totals` with distributed components and indices for the design variables, an OM Jacobian size isn't correct during the approximation computation.

### Issue Type

- [x] Bug

- [ ] Enhancement  
      POEM # N/A 

### Description
The example below works when either the distributed option is False in `DVS` or the indices option is removed from the `add_design_var`. However, as written, There is a error:

```
    .../openmdao/core/total_jac.py", line 1630, in compute_totals_approx
    totals[prom_out, prom_in][:] = _get_subjac(approx_jac[output_name, input_name],
ValueError: could not broadcast input array from shape (1,2) into shape (1,4)
```

### Example

```python
import openmdao.api as om
import numpy as np

class DVS(om.IndepVarComp):
    def initialize(self):
        self.options['distributed'] = True
    def setup(self):
        self.add_output('state', np.ones(4))

class SolverComp(om.ExplicitComponent):
    def initialize(self):
        self.options['distributed'] = True
    def setup(self):
        self.add_input('state',shape_by_conn=True)
        self.add_output('func')
        self.declare_partials('func','state',method='fd')

        outputs['func'] += np.sum(inputs['state'])

prob = om.Problem()
dvs = prob.model.add_subsystem('dvs',DVS())
prob.model.add_subsystem('solver', SolverComp())
prob.model.connect('dvs.state','solver.state')
prob.model.add_design_var('dvs.state', indices=[0,2])
prob.model.add_objective('solver.func')

prob.setup()
prob.run_model()
prob.check_totals(wrt='dvs.state')
```

### Environment

Operating System: OS X 10.15.7

Python environment: Python 3.9.0 (homebrew)

Packages: 
openmdao==3.7.0
petsc==3.14.5
petsc4py==3.14.1

Kenneth-T-Moore:
### Summary

Fixed a bug where the global_size was incorrect for a design variable that is specified with indices on an IVC output declared as a distributed component.

### Related Issues

- Resolves #1965

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -3049,12 +3049,12 @@ class System(object):
                 if src_name in abs2idx:
                     indices = meta['indices']
                     meta = abs2meta_out[src_name]
+                    out[name]['distributed'] = meta['distributed']
                     if indices is not None:
                         # Index defined in this response.
                         out[name]['global_size'] = len(indices) if meta['distributed'] \
                             else meta['global_size']
                     else:
-                        out[name]['distributed'] = meta['distributed']
                         out[name]['global_size'] = meta['global_size']
                 else:
                     out[name]['global_size'] = 0  # discrete var
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -3041,9 +3041,15 @@ class System(object):
                 meta['size'] = int(meta['size'])  # make default int so will be json serializable
 
                 if src_name in abs2idx:
+                    indices = meta['indices']
                     meta = abs2meta_out[src_name]
-                    out[name]['distributed'] = meta['distributed']
-                    out[name]['global_size'] = meta['global_size']
+                    if indices is not None:
+                        # Index defined in this response.
+                        out[name]['global_size'] = len(indices) if meta['distributed'] \
+                            else meta['global_size']
+                    else:
+                        out[name]['distributed'] = meta['distributed']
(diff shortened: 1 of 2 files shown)


openmdao/core/system.py
```diff
@@ -3041,9 +3041,15 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
...
 
                 if src_name in abs2idx:
+                    indices = meta['indices']
                     meta = abs2meta_out[src_name]
-                    out[name]['distributed'] = meta['distributed']
-                    out[name]['global_size'] = meta['global_size']
+                    if indices is not None:
+                        # Index defined in this response.
+                        out[name]['global_size'] = len(indices) if meta['distributed'] \
+                            else meta['global_size']
```
swryan, line 3049: It looks like you are no longer setting `out['name']['distributed']` in this case, where we were before... 
Just want to check that this is not a potential problem...

Kenneth-T-Moore: Good catch. I think we do want to keep 'distributed' set for both cases, so I will modify this.

swryan (approved):

robfalck (approved):

PR #1988, opened 2021-03-31
Generate more informative error message if a src_indices containing slicing operators was incorrectly set

### LINKED ISSUE #1987: Need better error message when connections with src_indices includes colon or ellipsis operator
### Summary of Issue

A better error message needs to be generated when there is an error in a src_indices that has a slicing operator in it. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

If the src_indices uses a slicing operator like a colon or ellipsis, and the user makes an error with the src_indices such that an exception occurs, the user only gets a very unhelpful error message like this:

```IndexError: index 4 is out of bounds for axis 0 with size 4```

A better error message needs to be generated so the user has some hint as to where the problem is.

### Example

```
arr_large_4x4 = np.array([[0, 1, 2, 3], [10, 11, 12, 13], [20, 21, 22, 23], [30, 31, 32, 33]])

class SlicerComp(om.ExplicitComponent):
   def setup(self):
       self.add_input('x', np.ones(4))
       self.add_output('y', 1.0)

       outputs['y'] = np.sum(inputs['x'])**2.0

   p = om.Problem()

   p.model.add_subsystem('indep', om.IndepVarComp('x', arr_large_4x4))
   p.model.add_subsystem('row4_comp', SlicerComp())

   p.model.connect('indep.x', 'row4_comp.x', src_indices=om.slicer[4, ...])

   p.setup()
```

### Environment

N/A

### PR DESCRIPTION
**Pull request by hschilling**
### Summary

If the src_indices uses a slicing operator like a colon or ellipsis, and the user makes an error with the src_indices such that an exception occurs, the user only gets a very unhelpful error message like this:

IndexError: index 4 is out of bounds for axis 0 with size 4

A better error message needs to be generated so the user has some hint as to where the problem is.

This PR does that and also includes a test for it.

### Related Issues

- Resolves #1987 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1548,7 +1548,14 @@ class Component(System):
                                                 meta['src_indices'], src_shape)
                 elif _is_slicer_op(src_inds):
                     meta['src_slice'] = src_inds
-                    src_inds = _slice_indices(src_inds, np.prod(parent_src_shape), parent_src_shape)
+                    try:
+                        src_inds = _slice_indices(src_inds, np.prod(parent_src_shape),
+                                                  parent_src_shape)
+                    except IndexError as err:
+                        raise IndexError(f"In component {self.msginfo}: \nError '{err}'\n  in "
+                                         f"resolving source indices in connection "
+                                         f"between source='{oldprom}' "
+                                         f"and target='{tgt}'\n  with src_indices='{src_inds}'")
                     meta['flat_src_indices'] = True
                 elif src_inds.ndim == 1:
                     meta['flat_src_indices'] = True
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1552,7 +1552,7 @@ class Component(System):
                         src_inds = _slice_indices(src_inds, np.prod(parent_src_shape),
                                                   parent_src_shape)
                     except IndexError as err:
-                        raise IndexError(f"In component {self.msginfo}: \nError '{err}'\n  in "
+                        raise IndexError(f"{self.msginfo}:\nError '{err}'\n  in "
                                          f"resolving source indices in connection "
                                          f"between source='{oldprom}' "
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by Kenneth-T-Moore — APPROVED** [review]

File: openmdao/core/component.py [hunk 1]
```diff
@@ -1548,7 +1548,14 @@ def _resolve_src_inds(self, my_tdict, top):
                                                 meta['src_indices'], src_shape)
                 elif _is_slicer_op(src_inds):
                     meta['src_slice'] = src_inds
-                    src_inds = _slice_indices(src_inds, np.prod(parent_src_shape), parent_src_shape)
+                    try:
+                        src_inds = _slice_indices(src_inds, np.prod(parent_src_shape),
+                                                  parent_src_shape)
+                    except IndexError as err:
+                        raise IndexError(f"In component {self.msginfo}: \nError '{err}'\n  in "
```
swryan, line 1555: No need for the "In component" part of this message.   All of our other error messages just start with `{self.msg_info}:`, which gives the class and name of the system/component..   This message would be inconsistent with that common format.

openmdao/core/tests/test_group.py
```diff
@@ -2506,6 +2506,22 @@ def test_src_indices_shape_bad_idx_flat_neg(self):
...
+        p.model.add_subsystem('row4_comp', SlicerComp())
+
+        # the 4 should be a 3
+        p.model.connect('indep.x', 'row4_comp.x', src_indices=om.slicer[4, ...])
+
+        with self.assertRaises(IndexError) as e:
+            p.setup()
+
+        self.assertTrue(str(e.exception).startswith("In component 'row4_comp'"))
+        self.assertTrue(str(e.exception).endswith("with src_indices='(4, Ellipsis)'"))
```
swryan, line 2523: seems like you should just check the entire message, which would also test the value of `{oldprom}`...

swryan (approved):

PR #1989, opened 2021-03-31
Compress JSON & BLOBs in case db, create separate metadata db for parallel runs

### LINKED ISSUE #1874: reduce size of case-db
### Summary of Issue

User reported a 60 mb case db, with only a single case in it, and only a few scalar variables in that case. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

60 mbs for just one case is crazy. The user happens to be running a large DOE in parallel, with one case on each proc (hence just one case) but that means that he gets 60 * n_procs data... with one case each. 

Is all of that data just the meta-data being saved? The case-db can be zipped down to 4 mb ish (see attached file) ... perhaps we should be compressing the data before we save it? 

[log_opt.sql_0.zip](https://github.com/OpenMDAO/OpenMDAO/files/5929095/log_opt.sql_0.zip)

As another note, when running in parallel we don't need the metadata for every case. Just one case db has to have it. The 0th one maybe? Can this be adjusted? 

### Example

case processing script: 
```
import openmdao.api as om

cr = om.CaseReader('log_opt.sql_0')
print(sorted(cr.list_sources(out_stream=None)))
cases = cr.get_cases(source='driver')
for case in cases:
    case.list_outputs()
```

### LINKED ISSUE #1997: n2 diagram breaks when class names contain 'nan' and other letters before and after
### Summary of Issue

N2 diagram fails for class names containing nan or Nan

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The n2 diagram does not load when I have a class name that contains either `*nan` or `nan*` or `*nan*`, where * is one or more letters. The bug does not appear to be sensitive to capitalization (I only tried changing the first letter though). I originally found this when I wanted to name a class `OperationsAndMaintenanceCost`, which I can rename to `OMCosts` easily enough, but there is serious tokamak physics that has the word `Banana` in it which I might like to describe someday.

When this bug occurs, the whole n2 screen comes up error-red and the diagram does not load. Of the normal diagram one 'pull-outer' arrow appears but that's all. There's no friendly n2 error message either.

This happens in both Firefox and in Chromium.

### Example
```
import openmdao.api as om

# breaks
class Banana(om.ExplicitComponent):
    pass

# breaks
class IpodNano(om.ExplicitComponent):
    pass

# breaks
class Nanana(om.ExplicitComponent):
    pass

# breaks
class Bananas(om.ExplicitComponent):
    pass

# breaks
class Ananas(om.ExplicitComponent):
    pass

# breaks
class Dynanite(om.ExplicitComponent):
    pass

# okay
class Montana(om.ExplicitComponent):
    pass

# okay
class Banunu(om.ExplicitComponent):
    pass

# okay
class Banan(om.ExplicitComponent):
    pass

# okay
class Nano(om.ExplicitComponent):
    pass

# okay
class Nan(om.ExplicitComponent):
    pass

# okay
class Nanite(om.ExplicitComponent):
    pass

# okay
class Dynamite(om.ExplicitComponent):
    pass

# okay
class Mamama(om.ExplicitComponent):
    pass

if __name__ == '__main__':
    prob = om.Problem()

    class test(om.Group):
        def setup(self):
            self.add_subsystem("test", Banana(),
                    promotes=["*"])

    prob.model = test()
    prob.setup()
    prob.run_driver()
```

```
openmdao n2 test.py
```

### Environment

Operating System: Ubuntu 20.04

Python environment:
Python 3.8.5 (default, Aug  3 2020, 14:44:06)
[GCC 9.3.0] on linux

Packages:

`OpenMDAO 3.8.0`

tadkollar:
### Summary

Two methods to reduce the size of case dbs are used:
- For parallel runs, a separate metadata db with a `_meta` suffix is created to store the shared metadata for all ranks. The individual rank databases (still named `.sql_[rank]`) no longer contain any metadata. When these files are used with `CaseReader`, `metadata_filename` can be supplied to the constructor. If it's not, and metadata is missing in the db, `SqliteCaseReader.__init__()` will automatically attempt to find the `_meta` file based on the path and filename of the rank db
- Pickle BLOBs in metadata are now compressed with `zlib` before storing in the db. JSON strings are also compressed - these were by far the largest data structures in the example db from the issue.

Notes:
- `sqlite_recorder.format_version` incremented to 14
- A metadata db filename can be specified to `n2()` by using a comma in the `datasource` string, e.g. `datasource="cases.sql_2,cases.sql_meta"`. On the command line, `openmdao n2 cases.sql_2,cases.sql_meta` works (docs updated). If not supplied, `SqliteCaseReader` will automatically search for the `_meta` file.
- The default pickle version was updated to 4 from 2. Version 4 is supported by Python 3.4+. In the quick tests I did, files generated with version 4 were exactly the same size as with version 2.
- Tests were updated to support the new format.
- During testing, I discovered that the regex used to convert `NaN` to `"NaN"` would also convert `finance` to `fi"nan"ce` which caused errors. For a better solution, I switched to using the JSON5 javascript module instead of the internal one. It supports bare NaN, positive and negative infinity, so no regex was necessary.
- Before deciding to use zlib, I looked into built-in compression capabilities for `sqlite3` but found the extension was commercial-only.
- I also explored using native SQL types instead of JSON strings, as well as NoSQL databases, but found they only increased the file size.

### Related Issues

- Resolves #1874
- Resolves #1997 

None

### New Dependencies

None


openmdao/utils/om.py
```diff
@@ -60,7 +60,10 @@ def _n2_setup_parser(parser):
     parser : argparse subparser
         The parser we're adding options to.
     """
-    parser.add_argument('file', nargs=1, help='Python script or recording containing the model.')
+    parser.add_argument('file', nargs=1,
+                        help='Python script or recording containing the model. '
+                        'For separate recording metadata from a parallel run, specify both '
```
swryan, line 65: I think this phrasing is a bit unclear... suggest:
"If metadata was recorded in a separate file (e.g. for a parallel run), specify both "

swryan (approved):

openmdao/utils/om.py
tadkollar: Updated

openmdao/recorders/sqlite_reader.py
```diff
@@ -80,11 +82,16 @@ def __init__(self, filename, pre_load=False):
             The path to the filename containing the recorded data.
         pre_load : bool
             If True, load all the data into memory during initialization.
+        metadata_filename : str
```
Kenneth-T-Moore, line 85: The metadata_filename argument doesn't seem to be covered by a test (aside from indirect use for the N2).  Is it actually needed? Seems like everything is working with OpenMDAO's guess at a filename.  I suppose the user could rename one and not the other, so then you would need this. Still needs a test to cover this scenario I think.

Kenneth-T-Moore (changes requested):

swryan:
> I think we should probably just put this in as a standalone PR.. I had to do the same on my branch
> probably should do it in line 156, before networkx, since that's the incompatibility
> (see #1992)

per comment & closure of #1992, this is no longer necessary

robfalck (approved):
Approved but please address Ken's coverage recommendation.

openmdao/recorders/sqlite_reader.py
tadkollar: I added a test where `metadata_filename` is used explicitly and expected to work, and another where the filename is bogus so it expects to fail.

Kenneth-T-Moore (approved):

PR #1992, opened 2021-04-02
New decorator package v5 breaks networkx

swryan:
### Summary

install older version of `decorator` before `network` due to incompatibility with new decorator v5
### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -152,6 +152,7 @@ setup(
     },
     python_requires=">=3.6",
     install_requires=[
+        'decorator<5',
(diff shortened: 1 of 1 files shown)


robfalck (approved):

dschult:
V5.0.5 of decorator now works with the v4 syntax.  This should no longer be needed.

swryan:
> V5.0.5 of decorator now works with the v4 syntax. This should no longer be needed.

Thanks

PR #2012, opened 2021-04-19
Fix bug where the rel_err on a case was actually the abs_err.

### LINKED ISSUE #2011: Bug in outputting relative solver error with case reader
### Summary of Issue

Recording/reading of the relative error (for solvers) is incorrect. The data is correct in the SQLite database, but incorrect when accessed from a case reader.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When using a case recorder/reader to display an implicit component's solver history the relative error shows up as the absolute error (the absolute error is correct). I believe I have found a typo in the openmdao source that might explain it (see below).

### Potential Solution

From what I can tell, it seems like there is a typo in "openmdao/recorders/case.py" (line 126):

    self.rel_err = data['abs_err'] if 'rel_err' in data.keys() else None

I believe that it should be:

    self.rel_err = data['rel_err'] if 'rel_err' in data.keys() else None

Changing this line solved the issue for me. I have changed this locally and can submit a pull request, I just wasn't sure if there was an additional process to run through other than making the change and committing it.

### Environment

(should not be the source of the error)

Operating System: Ubuntu 20.04

Python environment: Anaconda Python 3.7.4

Packages: OpenMDAO 3.8.0

Kenneth-T-Moore:
### Summary

Fix bug where the rel_err on a case was actually the abs_err.  Simply a copy-paste error in the code.

### Related Issues

- Resolves #2011

None

### New Dependencies

None

### DIFF
--- a/openmdao/recorders/tests/test_sqlite_reader.py
+++ b/openmdao/recorders/tests/test_sqlite_reader.py
@@ -2778,7 +2778,7 @@ class TestSqliteCaseReader(unittest.TestCase):
         prob.cleanup()
 
         cr = om.CaseReader(self.filename)
-        case = cr.get_cases()[-1]
+        case = cr.get_case(cr.list_cases()[-1])
 
         norm =  nl._iter_get_norm()
         norm0 = nl._norm0
--- a/openmdao/recorders/tests/test_sqlite_reader.py
+++ b/openmdao/recorders/tests/test_sqlite_reader.py
@@ -2758,6 +2758,33 @@ class TestSqliteCaseReader(unittest.TestCase):
         for case in expected_cases:
             self.assertTrue(case in all_driver_cases)
 
+    def test_abs_rel_error(self):
+        prob = om.Problem()
+        model = prob.model
+        model.add_subsystem('comp', ImplCompTwoStates())
+
+        # mda solver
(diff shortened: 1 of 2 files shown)


naylor-b (approved):

openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -2758,6 +2758,33 @@ def test_reading_all_case_types(self):
...
+        nl.add_recorder(self.recorder)
+
+        prob.setup()
+        prob.set_val('comp.y', 8.0)
+        prob.set_val('comp.z', 5.0)
+        prob.run_model()
+        prob.cleanup()
+
+        cr = om.CaseReader(self.filename)
+        case = cr.get_cases()[-1]
```
swryan, line 2781: Not a big deal maybe, but this is instantiating a Case object for every recorded case just to get the last one.
Would probably be better to do something like `cr.get_case(cr.list_cases()[-1])`

Kenneth-T-Moore: There is only one case in the db, but I will change it.

swryan (approved):

PR #2013, opened 2021-04-19
Changed default behavior when computing src_indices between variables when one side is distributed.

### LINKED ISSUE #1916: Add a test to verify the fix from PR#1914

### LINKED ISSUE #1972: Cleanup of `shape_by_conn`
### Summary of Issue

Change some behaviors of `shape_by_conn` based on user feedback. 

### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM [#046](https://github.com/OpenMDAO/POEMs/pull/103)
      PR #1964

### Description

Users of `shape_by_conn` reported non-intuitive behavior for serial output -> parallel input cases of `shape_by_conn`. 

The rule with shape_by_conn should always be a direct local copy of all data. In the case of serial-to-serial or parallel-to-parallel this is always obviously the case. 

The tricky situation arises when you have serial output -> parallel input. In this case, we will choose to follow the same convention as if both inputs were manually sized and a simple connection was made with no additional src indices given. 
This means that the parallel input side must have the same sized variable on each processor, and a local data transfer is set up between the serial copy of the output on each proc, and that proc's local input. If the inputs are different sizes on different processors, then it should be an error. 

As part of this cleanup, we should investigate using the `shape_by_conn` feature for the sizing of auto_ivc variables. This should be possible, as long as the `set_input_defaults` command causes the proper sizing of input variables before the `shape_by_conn` resolution needs to happen. It seems possible that this would yield a cleaner design for auto-ivcs and less code, since the sizing logic could now be handled by `shape_by_conn` completely. If this turns out to be a bad idea, we don't have to move auto_ivc's over to shape_by_conn but it would be worth understanding why it doesn't work well. 

Another potentially unclear case is parallel-to-serial connections, with shape by conn. Conceptually this isn't fundamentally confusing, but not all users realize that a serial component can still have a parallel input. The expected behavior in this case is again, a local data transfer with sizes matching exactly. However, in this case its possible that the sizes are different on different processors. 

Documentation updates need to be made to give clear examples of serial-parallel, parallel-parallel, and parallel-serial cases as well so users know what is supported and what to expect. The parallel-serial example should clearly show that different sizes on different processors are allowed. 

### Example

A detailed set of tests are given in PR#1964 from Anil Yildirim

naylor-b:
### Summary

- changed default src_indices computation when distributed vars are involved
- deprecated connections between serial outputs and distributed inputs
- no longer allow dynamically shaped connection between distributed outputs and serial inputs
- deprecated component option 'distributed'
- added 'distributed' arg to add_input/add_output
- added 'run_root_only' option to Component

See: [POEM 046](https://github.com/OpenMDAO/POEMs/pull/103)

### Related Issues

- Resolves #1916
- Resolves #1972

- dynamically shaped connection between distributed outputs and serial inputs are no longer allowed
- dynamically shaped connections now by default assume local data transfer to distributed inputs, i.e., size/shape and src_indices in the input correspond to the local part of the output.
- old behavior of connecting distributed outputs to serial inputs was to have the serial input in each proc contain the full distributed size of the output.  To achieve this same behavior, user must now specify src_indices for the input of om.slicer[:] (recommended) or an index array matching the full distrib size of the output.

### New Dependencies

None


openmdao/core/component.py
```diff
@@ -373,6 +388,29 @@ def setup_partials(self):
...
+        Return the value of the run_root_only option and check for possible errors.
+
+        Returns
+        -------
+        bool
+            True if run_root_only is active.
+        """
+        if self.options['run_root_only']:
+            if self.comm.size > 1 or (self._full_comm is not None and self._full_comm.size > 1):
+                if self._has_distrib_vars:
```
Kenneth-T-Moore, line 402: Can't all of these checks be done during final setup, instead of during every api call on the component?

openmdao/core/component.py
```diff
@@ -1580,5 +1629,5 @@ def items(self):
         return [(key, self._dict[key]['value']) for key in self._dict]
 
     def iteritems(self):
-        for key, val in self._dict.iteritems():
+        for key, val in self._dict.items():
```
Kenneth-T-Moore, line 1632: How did we miss this? Must not be covered.

openmdao/core/implicitcomponent.py
```diff
@@ -67,11 +67,22 @@ def _apply_nonlinear(self):
         """
         with self._unscaled_context(outputs=[self._outputs], residuals=[self._residuals]):
             with self._call_user_function('apply_nonlinear', protect_outputs=True):
+                args = [self._inputs, self._outputs, self._residuals]
                 if self._discrete_inputs or self._discrete_outputs:
-                    self.apply_nonlinear(self._inputs, self._outputs, self._residuals,
-                                         self._discrete_inputs, self._discrete_outputs)
+                    args += [self._discrete_inputs, self._discrete_outputs]
```
Kenneth-T-Moore, line 72: Not sure why we didn't create `_apply_nonlinear_wrapper` here like we did with all the other api functions.

openmdao/core/tests/test_distrib_adder.py
```diff
@@ -89,7 +89,8 @@ def test_distributed_adder(self):
 
         prob.model.add_subsystem('des_vars', IndepVarComp('x', np.ones(size)), promotes=['x'])
         prob.model.add_subsystem('plus', DistributedAdder(size=size), promotes=['x', 'y'])
-        summer = prob.model.add_subsystem('summer', Summer(size=size), promotes=['y', 'sum'])
+        summer = prob.model.add_subsystem('summer', Summer(size=size), promotes_outputs=['sum'])
+        prob.model.promotes('summer', inputs=['y'], src_indices=slicer[:])
```
Kenneth-T-Moore, line 93: Is this test modification just so that you can also test 'y' with the promotes method, or is there some reason it didn't work anymore when done in add_subsystem?

Kenneth-T-Moore (changes requested):

openmdao/core/component.py
naylor-b: My reason for doing it this way was that there isn't an existing setup function where these checks seemed to fit, and I didn't really want to create another function.  I could have crammed the checks into some existing setup function, but it didn't seem worth it, considering that these checks are not expensive.

openmdao/core/component.py
naylor-b: Turns out iteritems was being referenced from some dead code so I removed it.

openmdao/core/tests/test_distrib_adder.py
naylor-b: Since the default behavior has changed when we connect a distrib output to a serial input, to get the same behavior as before, i.e. the serial variable is sized according to the full distrib size of the output, you have to set src_indices for that input to slicer[:] (or some equivalent slice or index array).

openmdao/core/implicitcomponent.py
naylor-b: Yeah I agree it's not totally consistent, but I did it that way because if I move all of that stuff into a wrapper function, there's almost nothing left of the original function so it didn't seem worth it.

Kenneth-T-Moore (approved):

JustinSGray (approved):

anilyil:
The changes, and the tests under `TestDistribDynShapeCombos` look good. Thanks to everyone who worked on this. I know it was quite a bit of work but I think it will significantly improve the distributed behavior with codes like mphys.

One "nice to have" suggestion; when we use `shape_by_conn` for an output of an IVC, and the input's size downstream is defined, would it be possible to also copy the default value if it is available? I am guessing this is similar to `set_input_defaults` functionality. This way, besides getting the shape and src indices for its outputs, IVCs can also get the default values for these outputs. I think the behavior here would be the same for distributed or serial, because the src_indices are defined with the new convention, it should be straightforward to add this change. Let me know what you think. This would simplify adding geometric DVs to large models, but even without this, it should be pretty straightforward. Like I said, this is a "nice to have" and is not a necessity for any use case I can think of.

JustinSGray:
I would suggest you just use set_input_defaults and rely on auto_ivc instead. I don't love the idea of making IVC outputs behave differently than normal outputs. 

Whats wrong with just using auto-ivc here?

anilyil:
Okay, that makes sense. Then I am fine with this PR as is. Thanks again!

robfalck (approved):

PR #2045, opened 2021-05-06
Fix thread.isAlive

fzahle:
### Summary

Changed deprecated `thread.isAlive` to `thread.is_alive`.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/devtools/iprofile_app/iprofile_app.py
+++ b/openmdao/devtools/iprofile_app/iprofile_app.py
@@ -225,5 +225,5 @@ else:
             serve_thread = _startThread(tornado.ioloop.IOLoop.current().start)
             launch_thread = _startThread(lambda: _launch_browser(options.port))
 
-            while serve_thread.isAlive():
+            while serve_thread.is_alive():
                 serve_thread.join(timeout=1)


JustinSGray (approved):

JustinSGray:
@naylor-b want to check this one?

swryan:
https://bugs.python.org/issue37804

JustinSGray:
@swryan thanks! Does that mean that this PR is not backwards compatible with  python < 3.9?

fzahle:
it should be backward compatible:

https://docs.python.org/3.9/library/threading.html#threading.Thread.is_alive

naylor-b:
Thread.is_alive has been around since at least python 3.6 so we should be good.

naylor-b (approved):

swryan:
right.. it was a 2.x carryover, so good catch:
```
is_alive()
isAlive()
Return whether the thread is alive.

This method returns True just before the run() method starts until just after the run() method terminates. The module function enumerate() returns a list of all alive threads.

Changed in version 2.6: Added is_alive() spelling.```

PR #2049, opened 2021-05-07
Got rid of separate storage for vectors and right hand sides for parallel deriv coloring

### LINKED ISSUE #1854: Internal refactor to remove the multile-vectors for parallel derivatives
### Summary of Issue

Due to a prior refactor of the way serial components are handled, we no longer need the machinery to track multiple linear vectors in order to support parallel derivatives. 

### Issue Type

- [ ] Bug
- [ ] Enhancement  
- [x] Refactor

### Description

The original implementation of this feature dates back to 2015: Hwang, J., Roy, S., Kao, J., Martins, J., and Crossley, W. A., Simultaneous aircraft allocation and mission optimization using a modular adjoint approach, 56th AIAA/ASCE/AHS/ASC Structures, Structural Dynamics, and Materials Conference. (AIAA 2015-0900)

The basic idea is very similar to total-derivative coloring, except that in the strictest sense the various rows of the jacobian are not separable because there is a serial bottleneck at the start of the calculation. This situation can happen commonly in multi-point models: 

[geometry] (cheap/searial)  -> point 1 (expensive/parallel)
                                               -> point 2 (expensive/parallel)

If you a separate constraint on p1 and p2, if you naively solved for total derivatives (assuming p1 and p2 were on separate processors) then p2 would sit idle while p1 solved its local constraint and propagated it back to geometry. Then p1 would sit idle while p2 took its turn. 

So the original idea was to duplicate geometry separately on both proccessors and handle the two adjoints at the same time. You pay a small memory duplication cost for the serial component, but thats trivial compared to the cost savings you get by solving both adjoints at the same time. The linear algebra of it is documented in the OpenMDAO paper, section 6.2 (picture reproduced here). 

Screen Shot 2021-01-27 at 9 15 02 AM

In original OpenMDAO, we defaulted to using the rank 0 copy of geometry for normal forward execution (nonlinear solves). in the toy example, rank 0 geometry would broadcast its outputs to both procs. This meant that normal reverse mode solves would accumulate from both procs to rank0. In parallel derivative mode we needed to change that xfer so that p1 accumulated to rank 0 and p2 accumulated to rank 1. Hence we ended up with two different sets of xfers for normal and parallel derivatives. 

However, recent refactoring of the MPI structure has changed the default behavior. Now, geometry is always duplicated on rank0 and rank1 in normal operation. Thus normal adjoint derivatives no longer accumulate in geometry itself (via the xfer). Instead, we manually do a gather after the solve_linear is complete. OpenMDAO knows to do the gather because it knows that geometry is serial and hence duplicated. 

So in order to support parallel derivatives now, we simply need to not do the gather after the solve is complete. We can rely on the duplicate copies of geometry and the non-broadcasting xfers to handle the situation without the need for multiple rhs linear vectors and multiple xfers. 

The user facing APIs can stay exactly the same. These APIs will now need to trigger slightly different behavior for the end of the solve_linear. Where normally we would gather (because geometry is serial), the API would tell us now to not gather. The only other issue is that we need to make sure to properly seed a 1 in the right places at the start of solve_linear. Again, the API would tell us where to put the 1's. 

So the goal of this refactor is to achieve a significant simplification of the linear solver code. It will remove a for loop in the solve_linear methods and make the code more readable. It will also remove setup code to compute special xfers for this situations since they are not needed.

naylor-b:
### Summary

Also removed all code involving vectorized derivatives (aka matrix-matrix).

This is the implementation branch of a yet to be written POEM.  Once the POEM number is know, this description will be updated.

### Related Issues

- Resolves #1854

- vectorized derivatives are no longer available
- certain use cases for parallel deriv coloring are no longer supported, e.g., multiple variables on the same component having the same parallel deriv color.

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -1539,16 +1507,13 @@ def _setup_vec_names(self, mode):
             for system in self.system_iter(include_self=True, recurse=True):
                 for name, meta in system._get_vec_names_from_vois(mode):
                     vois.add(system.get_source(name))
-                    if meta['vectorize_derivs']:
```
Kenneth-T-Moore, line 1542: Should the argument `vectorized_derivs` also be removed from add_design_var, etc? 

Note: there is also use and mention of this feature in the docs, but that will have to be removed after the new notebooks are merged in.

Kenneth-T-Moore (changes requested):

openmdao/core/system.py
naylor-b: Removed all refs to 'vectorize_derivs'.

naylor-b:
Closing this for now and will issue a PR that just deprecates vectorized_derivs.  This PR will be resubmitted after these new deprecations have been removed.

PR #2051, opened 2021-05-07
Update pyOptSparseDriver to handle an API change in pyOptSparse in version 2.5.1

### LINKED ISSUE #1990: Update pyOptSparse wrapper for latest version
### Summary of Issue

Latest pyOptSparse version is 2.5.1, but there have been some backwards incompatible changes to its APIs. 

We should find a way to update, but preferrably in a way that supports both the older API and the newer one. If there is no clean way to do it, then lets talk about how to handle this in a dev meeting. 

### Issue Type

- [ ] Bug
- [x] Enhancement

hschilling:
### Summary

There have been two changes to the pyOptSparse API recently. 

- removed support for the "openmdao style" sensitivity dictionary, which is a tuple. Now the only supported way is to use a nested dict
- renamed the functions finalizeDesignVariables() and finalizeConstraints() and made them private (prepending with underscore). The preferred way now is to call finalize()

The first one doesn't affect OpenMDAO, as the code only used dicts, and not tuples, so no changes were needed.

The second one, which occurred in 2.5.1, requires detecting the version of pyOptSparse and calling the appropriate API call of pyOptSparse

### Related Issues

- Resolves #1990

None

### New Dependencies

None

### DIFF
--- a/.github/workflows/openmdao_test_workflow.yml
+++ b/.github/workflows/openmdao_test_workflow.yml
@@ -36,7 +36,7 @@ jobs:
             NUMPY: 1
             SCIPY: 1
             # PETSc: 3
-            PYOPTSPARSE: 'v2.1.5'
+            PYOPTSPARSE: 'v2.6.1'
             SNOPT: 7.7
             OPTIONAL: '[test]'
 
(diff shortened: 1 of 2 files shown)


JustinSGray:
@nwu63 just FYI!

nwu63:
Thanks for the work :)

Kenneth-T-Moore (approved):

swryan (approved):

PR #2077, opened 2021-05-17
Added rank check to only print driver summary to single rank

### LINKED ISSUE #2066: Driver summary should only print on a single rank when using MPI
### Summary of Issue

Optimization drivers print out the optimization summary on every rank, but they should only print it on one.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When running under MPI, an optimization will terminate with something like:

```
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 6.669236596900299
            Iterations: 26
            Function evaluations: 29
            Gradient evaluations: 26
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
Optimization Complete
-----------------------------------
```

We should only print these optimization methods on a single rank.

### Example

TBD

### Environment

OpenMDAO 3.9.1

DKilkenny:
### Summary

Added rank check to only print driver summary to single rank when running under MPI

### Related Issues

- Resolves #2066

None

### New Dependencies

None


openmdao/drivers/scipy_optimizer.py
```diff
@@ -449,20 +448,27 @@ def run(self):
...
-                                  method=opt,
-                                  jac=jac,
-                                  hess=hess,
-                                  # hessp=None,
-                                  bounds=bounds,
-                                  constraints=constraints,
-                                  tol=self.options['tol'],
-                                  # callback=None,
-                                  options=self.opt_settings)
+                if comm.rank == 0:
```
naylor-b, line 458: This doesn't seem correct since you'd only be doing the optimization on one rank and I'd expect it to hang when collective MPI calls are made in the model.

openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -152,6 +155,42 @@ def test_opt_distcomp(self):
                           1e-5)
 
 
+@unittest.skipUnless(MPI, "MPI is required.")
+class TestScipyOptimizeDriverMPI(unittest.TestCase):
+
```
naylor-b, line 160: This TestCase doesn't set N_PROCS so it's not actually running under MPI.

naylor-b (changes requested):

robfalck:
Would we be better off making a utility function like `print_mpi(*print_args, rank)` that does all of the logic internally?  Then we'd only need to change the print calls.

DKilkenny:
> Would we be better off making a utility function like `print_mpi(*print_args, rank)` that does all of the logic internally? Then we'd only need to change the print calls.

In this case, `minimize` is a function from Scipy so we have to set the options differently per rank. This utility function idea could be helpful on internal prints

openmdao/drivers/scipy_optimizer.py
```diff
@@ -449,20 +448,38 @@ def run(self):
                               f"than min allowed ({info['min_improve_pct']:.1f}%)."
                         issue_warning(msg, prefix=self.msginfo, category=DerivativesWarning)
 
+        if MPI:
```
Kenneth-T-Moore, line 451: Rather than doing all of this, you can just check rank with `self._problem_comm.rank`.  That's what we do in DOE and GA drivers.

openmdao/drivers/scipy_optimizer.py
```diff
@@ -544,22 +561,23 @@ def accept_test(f_new, x_new, f_old, x_old):
 
         self.result = result
 
-        if hasattr(result, 'success'):
-            self.fail = False if result.success else True
-            if self.fail:
-                print('Optimization FAILED.')
+        if comm.rank == 0:
```
Kenneth-T-Moore, line 564: You don't want to change how we set `self.fail` on any rank because that may be being used to tell some outer-level problem or workflow to continue.  The story is just to turn off the output printing, so the simplest thing is just to put an extra `if` check around the prints.

openmdao/drivers/scipy_optimizer.py
```diff
@@ -449,20 +448,38 @@ def run(self):
...
-                                  jac=jac,
-                                  hess=hess,
-                                  # hessp=None,
-                                  bounds=bounds,
-                                  constraints=constraints,
-                                  tol=self.options['tol'],
-                                  # callback=None,
-                                  options=self.opt_settings)
+                if comm.rank == 0:
+                    result = minimize(self._objfunc, x_init,
```
Kenneth-T-Moore, line 459: Unless my eyes betray me, the call to minimize is exactly the same in both parts of the if branch.

openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -1,6 +1,8 @@
 """ Unit tests for the ScipyOptimizeDriver."""
-
+#import wingdbstub
```
Kenneth-T-Moore, line 2: Cleanup the WING dbstub

Kenneth-T-Moore (changes requested):

openmdao/drivers/tests/test_scipy_optimizer.py
DKilkenny: Fixed

openmdao/drivers/scipy_optimizer.py
```diff
@@ -15,7 +14,7 @@
 import openmdao.utils.coloring as coloring_mod
 from openmdao.core.driver import Driver, RecordingDebugging
 from openmdao.utils.class_util import weak_method_wrapper
-from openmdao.utils.mpi import MPI
+from openmdao.utils.mpi import MPI, FakeComm
```
Kenneth-T-Moore, line 17: Don't think this import is needed now.

openmdao/drivers/scipy_optimizer.py
```diff
@@ -177,6 +177,24 @@ def _declare_options(self):
         self.options.declare('singular_jac_tol', default=1e-16,
                              desc='Tolerance for zero row/column check.')
 
+    def _setup_comm(self, comm):
```
naylor-b, line 180: This function isn't necessary.  Same goes for self._problem_comm.  All drivers have access to self._problem().comm so you should just use that.

naylor-b (changes requested):

openmdao/drivers/tests/test_scipy_optimizer.py
```diff
@@ -152,6 +155,48 @@ def test_opt_distcomp(self):
                           1e-5)
 
 
+@unittest.skipUnless(MPI, "MPI is required.")
```
naylor-b, line 158: Should add a check for PETSc here as well since the MPI stuff won't work without PETSc (we use PETSc to do our scatters between Vectors under mpi)

naylor-b (changes requested):

naylor-b (approved):

openmdao/drivers/scipy_optimizer.py
Kenneth-T-Moore: Is FakeComm still needed as an import?

Kenneth-T-Moore (approved):

PR #2089, opened 2021-05-20
Added flag to System to tell when under finite difference

### LINKED ISSUE #2083: OpenMDAO will set an `under_FD` flag when doing finite difference approximations on components
### Summary of Issue

Users can already test if they are under_complex_step, but we should also have `under_finite_difference` to match. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Some users have asked to know when their components are being finited-differenced. We need to add a flag to indicate it. 

Make sure there is a good example shown in the docs for both use cases (complex-step and finite-difference) so users can find it. Add it to the `Features`; `working with derivatives` sections.

naylor-b:
### Related Issues

- Resolves #2083

None

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/finite_difference.py
+++ b/openmdao/approximation_schemes/finite_difference.py
@@ -195,13 +195,13 @@ class FiniteDifference(ApproximationScheme):
         else:
             self._results_tmp = self._starting_resids.copy()
 
-        # Turn on complex step.
+        # Turn on finite difference.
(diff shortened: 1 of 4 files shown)


Kenneth-T-Moore (approved):

JustinSGray (approved):
I checked out the docs with ReviewNB. They looked good. Thanks for adding the new section

PR #2092, opened 2021-05-20
moved asserts in a notebook into hidden cells

naylor-b:


swryan (approved):

Kenneth-T-Moore (approved):
There is still an assert in each cell block.

However, approving since I will be away for the next 2 days. Merge after fix.

PR #2095, opened 2021-05-21
Updated `show_options_table` func

DKilkenny:
### Summary

Updated the `show_options_table` func for Dymos cases where the options are held in the `__init__` class instead of an options dict.

### Related Issues

- Resolves # N/A

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/tests/test_notebook_utils.py
+++ b/openmdao/utils/tests/test_notebook_utils.py
@@ -49,6 +49,9 @@ class TestNotebookUtils(unittest.TestCase):
         self.assertEqual(options, None)
 
     def test_show_options_table_warning(self):
+        from openmdao.utils import notebook_utils
+        notebook_utils.ipy = False
+
         msg = ("IPython is not installed. Run `pip install openmdao[notebooks]` or `pip install "
                "openmdao[docs]` to upgrade.")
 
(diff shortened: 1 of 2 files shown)


swryan:
should probably add a test

openmdao/utils/tests/test_notebook_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_show_options_wo_attr(self):
+        from openmdao.utils import notebook_utils
+        notebook_utils.ipy = True
+
+        options = om.show_options_table("openmdao.components.cross_product_comp.CrossProductComp")
+
+        self.assertEqual(options, None)
+
+    def test_show_options_w_attr(self):
+        from openmdao.utils import notebook_utils
```
naylor-b, line 29: It's not clear to me what the difference is between this test and the previous one.  Maybe put some comments in here to explain what these tests are doing.

naylor-b (changes requested):

openmdao/utils/tests/test_notebook_utils.py
DKilkenny: I missed the declared options in the base component so there is no difference. Deleted

naylor-b (approved):

openmdao/utils/tests/test_notebook_utils.py
swryan: It doesn't look like you are actually testing the new functionality? `BalanceComp` has an `options` dict, right?  
Don't you need an object ```where the options are held in the __init__ class instead of an options dict.```?

DKilkenny: I added a better test for a case where the options dictionary is being passed directly

robfalck (approved):

PR #2106, opened 2021-05-27
Whenever get_io_metadata is called, update metadata and promoted names to account for any new i/o that was added.

### LINKED ISSUE #1978: get_io_metadata on child does not stay up to date with calls to `add_output` on that child.
### Summary of Issue

Users working with get_io_metadata are confused when they add an output, but don't see it show up in the metadata right away. 

### Issue Type

- [x] Bug (sort of)
- [ ] Enhancement  

### Description

I think the current behavior is not technically a bug, but rather an intentional design choice. IIRC the metadata is updated by _setup_var_data which is not re-called every time add_output happens. The basic idea was that we only call _setup_var_data once on each level of the hierarchy as we recurse back up the tree. Hence `get_io_metadata` would not keep up to date with changes made during the configure you were currently in, but would show up in the next level's up configure. 

This was deemed acceptable because, in theory, any changes being made at the current configure level are known by whoever is coding it. In other words, you don't need get_io_metadata to tell you that you added a variable `b`, because you just added it! 

However, at least one user has said they found this non-ituitive. **if** we wanted to "fix" this, so get_io_metadata always gave an accurate (as of the point it was called) return value with all the metadata we know at that moment, I see two options: 

1) internally call _setup_var_data if anything has changed. Im not sure how expensive _setup_var_data really is, so no idea how bad this would be. I suppose it would depend totally on how many variables you had underneath. Could maybe do something fancy like only calling setup_var_data on new stuff... 

2) push the metadata that you know from the call to add_output directly into the metadata dictionary as we get it. Other metadata (e.g. promoted names) could be added later in setup_var_data, but you could at least query the things you just set from it it this way. 

(2) seems doable to me, but im probably missing something. Open to discussion on **if** we should fix this, and how to do it. 

### Example

```
import openmdao.api as om
from pprint import pprint as pp


class Middle(om.Group):
    def setup(self):
        b = self.add_subsystem("bottom", om.ExplicitComponent())

        b.add_output("a", val=1, tags=["my_tag"])

    def configure(self):

        # the only output is a, and get_io_metadata works here.
        print("\nconfigure of middle, before adding b")
        self.nom_get_metadata()

        # add another output
        self.bottom.add_output("b", val=1, tags=["my_tag"])

        # we have both a and b now, but get_io_metadata only returns a
        # TODO I would expect get_io_metadata to return both a and b, but this issue is not critical
        print("\nconfigure of middle, after adding b")
        self.nom_get_metadata()

    def nom_get_metadata(self):
        # generic function to return metadata. we dont really need to add this method,
        # but this is just an example of how the code would look like in mphys
        metadata = self.bottom.get_io_metadata(iotypes=["output"], metadata_keys=["tags"], tags="my_tag")
        pp(metadata)

    def nom_add_output(self):
        # method to add another output after this component's configure is done
        self.bottom.add_output("c", val=1, tags=["my_tag"])


class Top(om.Group):
    def setup(self):
        self.add_subsystem("middle", Middle())

    def configure(self):
        print("\nconfigure of top, before adding c")
        self.middle.nom_get_metadata()

        # now add the c output and try again
        self.middle.nom_add_output()

        # we should have all a, b, and c as outputs, but get metadata only returns a and b
        # TODO I would expect get metadata to return all a, b, and c. This is a critical issue for my use case that would help a lot if fixed
        print("\nconfigure of top, after adding c")
        self.middle.nom_get_metadata()


p = om.Problem()
p.model = Top()

p.setup()
```

Kenneth-T-Moore:
### Summary

1. Whenever get_io_metadata is called, update metadata and promoted names to account for any new i/o that was added. This only updates subsystems that have changes.
2. Added `@use_tempdirs` to some more tests to clean up some output that is dumped in the test execution directory.

### Related Issues

- Resolves #1978

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -3372,6 +3372,16 @@ def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
...
         rel_idx = len(prefix)
 
+        # Setup any modified subsystems so the metadata dicts are up-to-date.
+        if self._problem_meta:
+            conf_info = self._problem_meta['config_info']
+            if conf_info:
+                if self._subsystems_allprocs:
+                    for s in conf_info._modified_system_iter(self):
+                        s._setup_var_data()
+                    conf_info._reset
```
naylor-b, line 3382: _reset should be a function call.

Kenneth-T-Moore: Oops, fixed.

openmdao/core/system.py
```diff
@@ -3372,6 +3372,16 @@ def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
...
         rel_idx = len(prefix)
 
+        # Setup any modified subsystems so the metadata dicts are up-to-date.
+        if self._problem_meta:
+            conf_info = self._problem_meta['config_info']
+            if conf_info:
+                if self._subsystems_allprocs:
+                    for s in conf_info._modified_system_iter(self):
+                        s._setup_var_data()
+                    conf_info._reset()
```
naylor-b, line 3382: Actually I would move the _reset() call to after the self._setup_var_data, and would put the self._setup_var_data inside a `if config_info._modified_systems:` conditional so it won't get called unnecessarily.

naylor-b (changes requested):

naylor-b (approved):

robfalck (approved):

PR #2112, opened 2021-06-01
Fixed invalid value in QuadraticComp

### LINKED ISSUE #2023: "RuntimeWarning: invalid value encountered in sqrt" in test test_feature_cache_linear
### Summary of Issue

The test 

`openmdao.core.tests.test_feature_cache_linear_solution.CacheLinearTestCase.test_feature_cache_linear`

generates a warning

`<string>:47: RuntimeWarning: invalid value encountered in sqrt`

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The test 

openmdao.core.tests.test_feature_cache_linear_solution.CacheLinearTestCase.test_feature_cache_linear

which is used in the doc:

https://openmdao.org/twodocs/versions/latest/features/core_features/working_with_derivatives/linear_restart.html

generates a warning

`<string>:47: RuntimeWarning: invalid value encountered in sqrt`

when running this line of code which is in the solve_nonlinear method in the ImplicitComponent defined in the test.

`outputs['states'][0] = (-b + (b ** 2 - 4 * a * c) ** 0.5) / (2 * a)`

### Example

See description.

### Environment

NA

DKilkenny:
### Summary

Added upper bound to prevent RuntimeWarning

### Related Issues

- Resolves #2023

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_feature_cache_linear_solution.py
+++ b/openmdao/core/tests/test_feature_cache_linear_solution.py
@@ -80,7 +80,7 @@ class CacheLinearTestCase(unittest.TestCase):
         p.model.add_subsystem('obj', om.ExecComp('y = (x[1]-x[0])**2', x=np.ones(2)))
         p.model.connect('states', 'obj.x')
 
-        p.model.add_design_var('a', cache_linear_solution=True)
+        p.model.add_design_var('a', cache_linear_solution=True, upper=4.)
         p.model.add_constraint('states', upper=10)
         p.model.add_objective('obj.y')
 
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore (approved):

Kenneth-T-Moore (changes requested):
Need a little re-write on the note after the code to reflect the changes. Something like

```{Note}
In the example shown above, the model is solving for derivatives in ‘fwd’ mode, specified by `p.setup(mode=’fwd’)`. In that case the `cache_linear_solution` arg may be passed when adding design variables as shown above with `p.model.add_design_var(‘a’, cache_linear_solution=True, upper=4.)`. However, if the model were to solve for derivatives in ‘rev’ mode instead, then the `cache_linear_solution` arg would be applied to the objective and/or the constraint variables. For example, `p.model.add_constraint(‘states’, upper=10, cache_linear_solution=True)`.```

Actually, seeing all of this, I recommend changing the order to:

```
p.model.add_design_var(‘a’, upper=4., cache_linear_solution=True)
```

So that it looks like we add the "cache_linear_solution" at a later time.

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2118, opened 2021-06-04
Fixed inconsistency of `val` vs `value` in OM

### LINKED ISSUE #1834: Inconsistent application of Val vs. Value
### Summary of Issue

Inconsistent use of `val` vs. `value` 

### Issue Type

- [X] Bug
- [ ] Enhancement  
      POEM #N/A

### Description

In `ExecComp()` initial values are specified using `value=` 
However in component definitions, initial conditions are specified using `val=`

Should have a single nomenclature for setting values in all of OpenMDAO.

### Example

ExecComp() Example:

```
p.model.add_subsystem('LengthCable', om.ExecComp(['length_cable = 2*((L_arm+D_sphere)*4+L_Aux+L_TMS)'],
                                               length_cable = {'value':0, 'units':'m'},
                                               L_arm={'value':0, 'units':'m'}, 
                                               L_Aux={'value':0, 'units':'m'},
                                               L_TMS={'value':0, 'units':'m'},
                                               D_sphere={'value':0, 'units':'m'}),
                  promotes=['*'])
```

Component Definition Example:
```
class mach(ExplicitComponent):

    def setup(self):

        # Inputs
        self.add_input('V_inf', val=20, desc='free stream velocity', units='m/s')
        
        # Outputs
        self.add_output('Mach', val=np.zeros(nn), desc='mach number', units=None)
```
        
### Environment

N/A

DKilkenny:
### Summary

Fixed inconsistency of `val` vs `value` in OM

### Related Issues

- Resolves #1834

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -342,7 +342,8 @@ class ExecComp(ExplicitComponent):
                                        (self.msginfo, arg, sorted(diff)))
 
                 if 'val' in val and 'value' in val:
-                    raise RuntimeError("Cannot use 'val' and 'value' at the same time, use 'val'.")
+                    raise RuntimeError(f"{self.msginfo}: 'val' and 'value' at the same time, use "
+                                       "'val'.")
                 elif 'value' in val and not warned:
                     warn_deprecation(f"{self.msginfo}: 'value' will be deprecated in 4.0. Please "
                                      "use 'val' in the future.")
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -344,7 +344,8 @@ class ExecComp(ExplicitComponent):
                 if 'val' in val and 'value' in val:
                     raise RuntimeError("Cannot use 'val' and 'value' at the same time, use 'val'.")
                 elif 'value' in val and not warned:
-                    warn_deprecation("'value' will be deprecated in 4.0. Please use 'val' instead.")
+                    warn_deprecation("'value' will be deprecated in 4.0. Please use 'val' in the"
+                                     "future.")
 
                 if 'value' in val:
                     val['val'] = val.pop('value')
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -16,7 +16,7 @@ from openmdao.warnings import issue_warning, DerivativesWarning, warn_deprecatio
 VAR_RGX = re.compile(r'([.]*[_a-zA-Z]\w*[ ]*\(?)')
 
 # Names of metadata entries allowed for ExecComp variables.
-_allowed_meta = {'value', 'shape', 'units', 'res_units', 'desc',
+_allowed_meta = {'value', 'val', 'shape', 'units', 'res_units', 'desc',
                  'ref', 'ref0', 'res_ref', 'lower', 'upper', 'src_indices',
                  'flat_src_indices', 'tags', 'shape_by_conn', 'copy_shape'}
 
@@ -161,7 +161,7 @@ class ExecComp(ExplicitComponent):
             Initial values of variables can be set by setting a named
             arg with the var name.  If the value is a dict it is assumed
             to contain metadata.  To set the initial value in addition to
-            other metadata, assign the initial value to the 'value' entry
+            other metadata, assign the initial value to the 'val' entry
             of the dict.
 
         Notes
@@ -189,7 +189,7 @@ class ExecComp(ExplicitComponent):
         .. code-block:: python
 
             excomp = ExecComp('y=sum(x)',
-                              x={'value': numpy.ones(10, dtype=float),
+                              x={'val': numpy.ones(10, dtype=float),
                                  'units': 'ft'})
         """
         options = {}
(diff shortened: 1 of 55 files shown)


naylor-b (approved):

openmdao/components/exec_comp.py
```diff
@@ -339,6 +341,16 @@ def _setup_expressions(self):
                                        "recognized for variable '%s': %s" %
                                        (self.msginfo, arg, sorted(diff)))
 
+                if 'val' in val and 'value' in val:
+                    raise RuntimeError("Cannot use 'val' and 'value' at the same time, use 'val'.")
+                elif 'value' in val and not warned:
+                    warn_deprecation("'value' will be deprecated in 4.0. Please use 'val' in the "
```
swryan, line 347: I think you need `self.msginfo` in this warning message to point the user to the component in question.
And I think you mean to say that the argument *is* deprecated and will be *removed* in 4.0?
Also, I think you could clean this logic up a bit by merging with the following if statement, e.g.:
```
elif 'value' in val:
    if not warned:
        warn_deprecation(f"{self.msginfo}: the 'value' argument is deprecated and will be removed in version 4.0. Please use 'val'.")
        warned = True
    val['val'] = val.pop('value')
```

Kenneth-T-Moore (approved):

openmdao/components/exec_comp.py
```diff
@@ -424,7 +436,10 @@ def _setup_expressions(self):
                     dct = self.add_input(var, val, **meta)
 
             if var not in init_vals:
-                init_vals[var] = dct['value']
+                if 'value' in dct:
+                    dct['val'] = dct.pop('value')
```
swryan, line 440: not clear that this is needed?  `value` has been replaced by `val` in `kwargs2`/`meta` right?

openmdao/components/exec_comp.py
```diff
@@ -339,6 +341,16 @@ def _setup_expressions(self):
                                        "recognized for variable '%s': %s" %
                                        (self.msginfo, arg, sorted(diff)))
 
+                if 'val' in val and 'value' in val:
+                    raise RuntimeError("Cannot use 'val' and 'value' at the same time, use 'val'.")
```
swryan, line 345: I think you need `self.msginfo` in this warning message to point the user to the component in question.

openmdao/core/problem.py
```diff
@@ -408,23 +408,31 @@ def __setitem__(self, name, value):
...
-            Indices or slice to set to specified value.
+            Indices or slice to set to specified val.
+        value : float or ndarray or list or None
+            This argument is deprecated. Use 'val' instead.
         """
+        if val is None and value is None:
+            raise RuntimeError("Val is required.")
+        elif value is not None:
+            val = value
+            warn_deprecation("'value' will be deprecated in 4.0. Please use 'val' in the future.")
```
swryan, line 434: I think you need some additional info in this warning message to point the user to the call in question
And I think you mean to say that the argument *is* deprecated and will be *removed* in 4.0?
e.g.:
```
warn_deprecation(f"{self.msginfo} when calling `set_val` on {name}, the 'value' argument is deprecated and will be removed in version 4.0. Please use 'val'.")
```

swryan (changes requested):

robfalck (approved):

openmdao/core/system.py
```diff
@@ -3511,7 +3511,7 @@ def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
         return result
 
     def list_inputs(self,
-                    values=True,
+                    vals=True,
```
robfalck, line 3514: I think this should be singular `val` to match `shape`, as well as the use of `val` everywhere else in the code.

openmdao/core/system.py
```diff
@@ -3616,7 +3633,7 @@ def list_inputs(self,
 
     def list_outputs(self,
                      explicit=True, implicit=True,
-                     values=True,
+                     vals=True,
```
robfalck, line 3636: Singular `val` instead of `vals` here also.

openmdao/recorders/case.py
```diff
@@ -392,7 +393,7 @@ def get_responses(self, scaled=True, use_indices=True):
         return self._get_variables_of_type('response', scaled, use_indices)
 
     def list_inputs(self,
-                    values=True,
+                    vals=True,
```
robfalck, line 396: `val` instead of `vals`

openmdao/recorders/case.py
```diff
@@ -488,7 +501,7 @@ def list_inputs(self,
 
     def list_outputs(self,
                      explicit=True, implicit=True,
-                     values=True,
+                     vals=True,
```
robfalck, line 504: `val` instead of `vals`

robfalck (changes requested):

openmdao/recorders/case.py
DKilkenny: Fixed

robfalck (approved):

swryan (approved):

PR #2130, opened 2021-06-14
Replaced `value` with `val` in `ExecComp`'s across the documentation

DKilkenny:
### Summary

Replaced `value` with `val` in `ExecComp`'s across the documentation to remove deprecation warnings

### Related Issues

- Resolves #N/A

None

### New Dependencies

None


swryan (approved):

naylor-b (approved):

openmdao/core/component.py
```diff
@@ -582,6 +582,7 @@ def add_input(self, name, val=1.0, shape=None, src_indices=None, flat_src_indice
 
         metadata = {
             'val': val,
+            'value': val,
```
swryan, line 585: unfortunately, since this a return value for a user API function, I think we need to "do this right" and make it throw a deprecation if the user tries to access "value" from these metadata dicts.

suggestion: make a special wrapper for that metadata object that looks for "value" and throws a deprecation warning, and turns it into 'val'

swryan (changes requested):

openmdao/core/component.py
DKilkenny: Added

swryan (approved):

PR #2146, opened 2021-06-17
fix deprecation warning; remove extraneous 'value' from metadata

swryan:
### Summary

- metadata returned by `get_io_metadata`  will deprecate the `value` key
- fixed the message and warning type for the  deprecated `values` argument to the `list_inputs` and `list_outputs` functions
- removed redundant 'value' key from `list_inputs` & `list_outputs` return values

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -3318,7 +3318,7 @@ class System(object):
                     ret_meta = None
                 else:
                     if metadata_keys is None:
-                        ret_meta = _MetadataDict(meta.copy())
+                        ret_meta = _MetadataDict(meta)
                     else:
                         ret_meta = _MetadataDict()
                         for key in metadata_keys:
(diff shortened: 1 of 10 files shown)


openmdao/core/system.py
```diff
@@ -3299,9 +3318,9 @@ def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
                     ret_meta = None
                 else:
                     if metadata_keys is None:
-                        ret_meta = meta.copy()
+                        ret_meta = _MetadataDict(meta.copy())
```
Kenneth-T-Moore, line 3321: I don't think you need the .copy here anymore.

Kenneth-T-Moore (changes requested):

openmdao/core/system.py
swryan: hmm yeah, I guess you're right.. the copy constructor and the `copy()` function both do a shallow copy

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2174, opened 2021-07-09
Refactored DOEDriver to eliminate recorder pointer copies

### LINKED ISSUE #2098: DOEDriver unnecessarily managing separate pointers to its case recorders
### Summary of Issue

The DOEDriver code is keeping a shadow copy of pointers to case recorders. But it does not look like it needs to do that. This is just a code clean up issue. Nothing is broken.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

This code keeps track of the recorders in `DOEDriver._recorders` so it can adjust some attributes on the recorders if running in parallel. But it has access to the recorders through `DOEDriver._rec_mgr` so doesn't need to keep those shadow copies. 

```
    def add_recorder(self, recorder):
        """
        Add a recorder to the driver.

        Parameters
        ----------
        recorder : CaseRecorder
           A recorder instance.
        """
        # keep track of recorders so we can flag them as parallel
        # if we end up running in parallel
        self._recorders.append(recorder)

        super().add_recorder(recorder)

    def _setup_recording(self):
        """
        Set up case recording.
        """
        if MPI:
            procs_per_model = self.options['procs_per_model']

            for recorder in self._recorders:
                recorder._parallel = True

                # if SqliteRecorder, write cases only on procs up to the number
                # of parallel DOEs (i.e. on the root procs for the cases)
                if isinstance(recorder, SqliteRecorder):
                    if procs_per_model == 1:
                        recorder._record_on_proc = True
                    else:
                        size = self._problem_comm.size // procs_per_model
                        if self._problem_comm.rank < size:
                            recorder._record_on_proc = True
                        else:
                            recorder._record_on_proc = False

        super()._setup_recording()
```

The refactor is to remove the `DOEDriver.add_recorder` method and in the `DOEDriver._setup_recording` method, use `self._rec_mgr` instead of `self._recorders` in the for loop ( the RecordingManager class has an `__iter__` defined so you can do that ). 

### Environment

N/A

N/A

DKilkenny:
### Summary

Refactored DOEDriver to eliminate recorder pointer copies

### Related Issues

- Resolves #2098

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/doe_driver.py
+++ b/openmdao/drivers/doe_driver.py
@@ -24,8 +24,6 @@ class DOEDriver(Driver):
     ----------
     _name : str
         The name used to identify this driver in recorded cases.
-    _recorders : list
-        List of case recorders that have been added to this driver.
     _problem_comm : MPI.Comm or None
         The MPI communicator for the Problem.
     _color : int or None
--- a/openmdao/drivers/doe_driver.py
+++ b/openmdao/drivers/doe_driver.py
@@ -71,7 +71,6 @@ class DOEDriver(Driver):
             self.options['generator'] = generator
 
         self._name = ''
-        self._recorders = []
         self._problem_comm = None
         self._color = None
 
--- a/openmdao/drivers/doe_driver.py
+++ b/openmdao/drivers/doe_driver.py
@@ -14,7 +14,6 @@ from openmdao.drivers.doe_generators import DOEGenerator, ListGenerator
 from openmdao.utils.mpi import MPI
 
 from openmdao.recorders.sqlite_recorder import SqliteRecorder
-from openmdao.recorders.recording_manager import RecordingManager
 
 
 class DOEDriver(Driver):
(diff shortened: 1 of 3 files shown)


openmdao/drivers/doe_driver.py
```diff
@@ -240,29 +240,14 @@ def _parallel_generator(self, design_vars, model=None):
...
-        ----------
-        recorder : CaseRecorder
-           A recorder instance.
-        """
-        # keep track of recorders so we can flag them as parallel
-        # if we end up running in parallel
-        self._recorders.append(recorder)
-
-        super().add_recorder(recorder)
-
```
swryan, line 257: should be able to get rid of the `_recorders` attribute as well...

DKilkenny: Agreed. Removed

Kenneth-T-Moore (approved):

openmdao/drivers/doe_driver.py
swryan: I notice that the attribute is still included in the class docstring...  I thought linting would catch this but maybe it only detects undocumented attributes... anyway, I think we need to delete this in line 27:
```
    _recorders : list
        List of case recorders that have been added to this driver.
```

DKilkenny: Ah, another good catch. Removed

swryan (approved):

PR #2182, opened 2021-07-14
Removed warning and allowed global_shape to be set

### LINKED ISSUE #2176: Distributed Array Shape Assumption
### Summary of Issue

For distributed variables, OM [assumes all dimensions of the array except the first one are the same across procresses.](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/system.py#L1493). If this isn't true as in the example below, the error message says:

```
RuntimeError: <model> <class Group>: Global size of output 'adder.x' (900) does not agree with local shape (3, 200)
```
which isn't descriptive of the actual problem. Someone did comment that this assumption is made at the linked part of the code, but I don't see anything about this assumption in the docs. So, rather than a bug, it may be an undocumented expected behavior that doesn't give a clear error message?

### Issue Type

- [x] Bug

- [ ] Enhancement  

### Description

The example below shows the error described in the summary. Switching n0 and n1 allows the code to run. For most arrays, it is probably safe to assume everything but the first dimension is the same, however there are some cases where that isn't true. The place I ran into this is structural mode shapes where I want the dimensions to be [num_modes ,num_points,n_dof], where the points are the variable that is different sizes across processors, in order to most efficiency access the data with an outer loop over the modes. So with either C or F ordering, I'd want to middle dimension to be distributed meaning I can't just switch to F order to align with OM's assumption about the shape. However, I could flatten everything to meet the assumption.

### Example

```python
import numpy as np
import openmdao.api as om
from mpi4py import MPI

class TwoDArrayAdder(om.ExplicitComponent):
    def initialize(self):
        self.options.declare('n0')
        self.options.declare('n1')

    def setup(self):
        n0 = self.options['n0']
        n1 = self.options['n1']
        self.add_input('x', shape=((n0, n1)), distributed=True)
        self.add_output('x_sum', shape=((n0, n1)), distributed=True)

        outputs['x_sum'] = np.sum(inputs['x'], axis=0)

n0 = 3
n1 = 100 if MPI.COMM_WORLD.rank ==0 else 200
prob = om.Problem()
ivc.add_output('x', val = np.ones((n0, n1)), distributed=True)

prob.model.add_subsystem('adder',TwoDArrayAdder(n0=n0, n1=n1))
prob.model.connect('ivc.x','adder.x')

prob.setup()
prob.run_model()
```

### Environment

OpenMDAO 3.9.3.dev0

DKilkenny:
### Summary

Removed warning and allowed global_shape to be set

### Related Issues

- Resolves #2176

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1489,10 +1489,6 @@ class System(object):
                     if high_dims:
                         high_size = np.prod(high_dims)
                         dim1 = global_size // high_size
-                        if global_size % high_size != 0:
-                            raise RuntimeError("%s: Global size of output '%s' (%s) does not agree "
-                                               "with local shape %s" % (self.msginfo, abs_name,
-                                                                        global_size, local_shape))
                         mymeta['global_shape'] = tuple([dim1] + list(high_dims))
                     else:
                         mymeta['global_shape'] = (global_size,)


Kenneth-T-Moore (changes requested):
Definitely need a test here.

Also, is getting rid of the error-checking the right solution? Question for Bret perhaps.

Kenneth-T-Moore (changes requested):

PR #2183, opened 2021-07-15
Fixed errors in multifi_cokriging documentation

### LINKED ISSUE #2168: Errors in multifi_cokriging documentation
I noticed that there are some errors in the documentation of the methods of the `MultiFiCoKrigingSurrogate` class (http://openmdao.org/newdocs/versions/latest/_srcdocs/packages/surrogate_models/multifi_cokriging.html):

The documentation for the `train()` method says that it will "Calculate a predicted value of the response based on the current trained model." This is the same line of documentation as the `predict()` method. I notice that this error is also present in the base class `MultiFiSurrogateModel`.

The documentation for the `train_multifi()` method says that its arguments `X` and `Y` are "An array with shape (n_samples_X, n_features)". However, this is only true if  `train_multifi()` is used for the *single* fidelity case. In the multi-fidelity case, the arguments are similar to those of the `fit()` method of the `MultiFiCoKriging` class, except that whereas the lists of arrays passed to the `fit()` method are ordered from lowest to highest fidelity, the lists of arrays passed to `train_multifi()` method are from the highest to the lowest fidelity.

(BTW, I notice that the documentation for the `train_multifi()` method of the base class `MultiFiSurrogateModel` doesn't specify the order of fidelity in the lists passed to it. Not sure if that's intentional or not.)

hschilling:
### Summary

User reported some doc string errors in MultiFiSurrogateModel and MultiFiCoKrigingSurrogate

### Related Issues

- Resolves #2168 

None

### New Dependencies

None

### DIFF
--- a/openmdao/surrogate_models/multifi_cokriging.py
+++ b/openmdao/surrogate_models/multifi_cokriging.py
@@ -1004,7 +1004,7 @@ class MultiFiCoKrigingSurrogate(MultiFiSurrogateModel):
             with X[i] = np.vstack([..., X[i+1])
         Y : list of double array_like elements
             A list of arrays with the observations of the scalar output to be predicted,
-            from lowest fidelity to highest fidelity.
+            from highest fidelity to lowest fidelity.
         """
         opt = self.options
         if not self.model:
--- a/openmdao/surrogate_models/multifi_cokriging.py
+++ b/openmdao/surrogate_models/multifi_cokriging.py
@@ -998,12 +998,13 @@ class MultiFiCoKrigingSurrogate(MultiFiSurrogateModel):
 
         Parameters
         ----------
-        X : array_like
-            An array with shape (n_samples_X, n_features) with the input at which observations
-            were made.
-        Y : array_like
-            An array with shape (n_samples_X, n_features) with the observations of the scalar
-            output to be predicted.
+        X : list of double array_like elements
+            A list of arrays with the input at which observations were made, from highest
+            fidelity to lowest fidelity. Designs must be nested
+            with X[i] = np.vstack([..., X[i+1])
+        Y : list of double array_like elements
+            A list of arrays with the observations of the scalar output to be predicted,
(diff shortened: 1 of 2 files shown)


openmdao/surrogate_models/multifi_cokriging.py
```diff
@@ -998,12 +998,13 @@ def train_multifi(self, X, Y):
...
-        X : array_like
-            An array with shape (n_samples_X, n_features) with the input at which observations
-            were made.
-        Y : array_like
-            An array with shape (n_samples_X, n_features) with the observations of the scalar
-            output to be predicted.
+        X : list of double array_like elements
+            A list of arrays with the input at which observations were made, from highest
+            fidelity to lowest fidelity. Designs must be nested
+            with X[i] = np.vstack([..., X[i+1])
```
Kenneth-T-Moore, line 1004: Wait, is x really ordered from high to low fidelity, but y is from low to high?

Kenneth-T-Moore (changes requested):

openmdao/surrogate_models/multifi_cokriging.py
hschilling: Thanks for catching that!

Kenneth-T-Moore: I think the lowest fidelity is the first one. See the docstring for the `fit` method and notice that in the code they normalize by the lowest fidelity, which is in index 0.

jjramsey: The `train_multifi()` method of the `MultiFiCoKrigingSurrogate` class reverses the order of the lists of arrays before feeding them to the `fit()` method of the `MultiFiCoKriging` class. That means that the *highest* fidelity is the first one, not the lowest.

Kenneth-T-Moore (approved):

swryan (approved):

PR #2184, opened 2021-07-15
Some speed improvements to final_setup when recorders are present.

### LINKED ISSUE #2144: Performance issue related to case recorder
### Summary of Issue

User provides  test case showing that adding a case recorder can drastically increase the run time of a model. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

User provided pyCycle example case where adding recorder causes massive delay in the start of model execution. No actual recording is done, just the addition of the recorder itself. The slowdown is shown to be triggered by run_model (i.e. setup and final_setup don't see the slow down). 

The performance issue is NOT related to the cost of case recording itself. It seems to be related to the cost of case recorder setup. Maybe the cost of finding and saving all the metadata? 

### Example
[foo.zip](https://github.com/OpenMDAO/OpenMDAO/files/6671376/foo.zip)

provided zip file has runable example. 

Requires OpenMDAO, pycycle. 

Execute the full_sweep_10x.py file. 

See lines 357 and 358. These can be commented/uncommented to cause the speed issue. 

### Environment

User reported the following environment: 

- pycycle commit 7ac701b8044106a778ee49f26f04b9e6944d6cb6 from pycycle master github
- openmdao 3.9.2 as installed via pip

@JustinSGray  saw same problem with master versions of OpenMDAO and pyCycle

naylor-b:
- also removed a little code from Vector that's no longer needed now that vectorized derivs are gone.

### Related Issues

- Resolves #2144

None

### New Dependencies

None

### DIFF
--- a/openmdao/vectors/vector.py
+++ b/openmdao/vectors/vector.py
@@ -57,9 +57,6 @@ class Vector(object):
         Mapping of var name to slice.
     _under_complex_step : bool
         When True, this vector is under complex step, and data is swapped with the complex data.
-    _icol : int or None
-        If not None, specifies the 'active' column of a multivector when interfaceing with
-        a component that does not support multivectors.
     _do_scaling : bool
         True if this vector performs scaling.
     _do_adder : bool
--- a/openmdao/vectors/vector.py
+++ b/openmdao/vectors/vector.py
@@ -102,7 +102,6 @@ class Vector(object):
         self._name = name
         self._typ = _type_map[kind]
         self._kind = kind
-        self._icol = None
         self._len = 0
 
         self._system = weakref.ref(system)
@@ -319,10 +318,7 @@ class Vector(object):
         """
         abs_name = self._name2abs_name(name)
         if abs_name is not None:
-            if self._icol is None:
-                val = self._views[abs_name]
-            else:
-                val = self._views[abs_name][:, self._icol]
+            val = self._views[abs_name]
         else:
             raise KeyError(f"{self._system().msginfo}: Variable name '{name}' not found.")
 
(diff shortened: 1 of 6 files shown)


JustinSGray:
@naylor-b did you record any timing data for the speedup here?

JustinSGray (approved):

naylor-b:
The setup time dropped by about 25%.  The changes were mostly in _get_viewer_data, where we collect metadata for use in the n2 viewer.  The biggest cost was the computation of the cycle_arrows, so I sped that up as much as I could.

openmdao/vectors/vector.py
```diff
@@ -102,7 +102,6 @@ def __init__(self, name, kind, system, root_vector=None, alloc_complex=False):
         self._name = name
         self._typ = _type_map[kind]
         self._kind = kind
-        self._icol = None
```
swryan, line 105: need to remove class docstring entry for `_icol`   
:docstring_police:

openmdao/visualization/n2_viewer/n2_viewer.py
```diff
@@ -77,73 +77,87 @@ def _convert_ndarray_to_support_nans_in_json(val):
...
+            'dtype': type(val).__name__,
+            'is_discrete': False,
+            'distributed': is_dist,
+            'shape': str(meta['shape']),
+        }
+
+        if typ == 'output':
+            var_dict['implicit'] = is_implicit
+            vec = system._outputs
+        else:  # input
```
swryan, line 110: could we get a code comment here explaining the `if MPI` ?

swryan (changes requested):

openmdao/vectors/vector.py
naylor-b: done

openmdao/visualization/n2_viewer/n2_viewer.py
naylor-b: done

swryan (approved):

PR #2187, opened 2021-07-19
Switch from Pyppeteer to Playwright for N2 GUI tests

### LINKED ISSUE #1969: Investigate Playwright as a replacement for Pyppeteer for GUI testing
### Summary of Issue

Playwright supports the 3 major browsers, has a Python API, and is supported by Microsoft. Investigate it as a replacement for Pyppeteer, which only supports Chromium and is not updated frequently.

### Issue Type

- [ ] Bug
- [X] Enhancement  
      POEM #N/A <N/A or POEM number at https://github.com/OpenMDAO/POEMs>

### Description

The Playwright site is https://playwright.dev/ and the GitHub repo is https://github.com/microsoft/playwright.

Need to make sure it supports at least the same functionality as Pyppeteer, is not any slower, and not too cumbersome to install.

tadkollar:
### Summary

Playwright is better supported and has more functionality than Pyppeteer. Playwright can support Chromium, Firefox, and Webkit browsers, although only Chromium is used in this PR. Tests complete in about 60 seconds instead of 260 seconds on the same system. This is mostly due to Playwright's `Page.wait_for_selector()`, which makes it unnecessary to always wait for animated transitions to complete. It could probably be sped up even more with additional attention.

### Related Issues

- Resolves #1969

None

### New Dependencies

None

### DIFF
--- a/.github/workflows/openmdao_test_workflow.yml
+++ b/.github/workflows/openmdao_test_workflow.yml
@@ -160,7 +160,6 @@ jobs:
           echo "Install additional packages for testing/coverage"
           echo "============================================================="
           pip install psutil objgraph git+https://github.com/mdolab/pyxdsm
-          type playwright > /dev/null && playwright install
 
       - name: Display conda info
         shell: bash -l {0}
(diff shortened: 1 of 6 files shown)


hschilling:
In the n2_gui_test.py, the docstring at the top still refers to Pyppeteer 

`"""Test N2 GUI with multiple models using Pyppeteer."""
`

hschilling (approved):

Kenneth-T-Moore (approved):

PR #2192, opened 2021-07-22
Improved error message for global shape errors

### LINKED ISSUE #2188: Need a more informative message for global shape errors.
### Summary of Issue

This issue is related to #2176.

For distributed variables, OM [assumes all dimensions of the array except the first one are the same across procresses.](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/system.py#L1493). If this isn't true as in the example below, the error message says:

```
RuntimeError: <model> <class Group>: Global size of output 'adder.x' (900) does not agree with local shape (3, 200)
```

While this is currently the expected behavior, the error message is not descriptive.

The code assumes that all but the first dimension of the shapes local parts in a distributed variable must match, and the error message should be changed to reflect that.

### Issue Type

- [x] Bug

- [ ] Enhancement  

### Description

See above.

### Example

```python
import numpy as np
import openmdao.api as om
from mpi4py import MPI

class TwoDArrayAdder(om.ExplicitComponent):
    def initialize(self):
        self.options.declare('n0')
        self.options.declare('n1')

    def setup(self):
        n0 = self.options['n0']
        n1 = self.options['n1']
        self.add_input('x', shape=((n0, n1)), distributed=True)
        self.add_output('x_sum', shape=((n0, n1)), distributed=True)

        outputs['x_sum'] = np.sum(inputs['x'], axis=0)

n0 = 3
n1 = 100 if MPI.COMM_WORLD.rank ==0 else 200
prob = om.Problem()
ivc.add_output('x', val = np.ones((n0, n1)), distributed=True)

prob.model.add_subsystem('adder',TwoDArrayAdder(n0=n0, n1=n1))
prob.model.connect('ivc.x','adder.x')

prob.setup()
prob.run_model()
```

### Environment

OpenMDAO 3.10.0.dev0

tadkollar:
### Summary

The old error message looked like:
`RuntimeError: <model> <class Group>: Global size of output 'adder.x' (900) does not agree with local shape (3, 200)`

The new message looks like:
`RuntimeError: <model> <class Group>: All but the first dimension of the shape's local parts in a distributed variable must match across processes. For output 'adder.x', local shape (3, 200) in MPI rank 1 has a higher dimension that differs in another rank.`

### Related Issues

- Resolves #2188

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1527,10 +1527,11 @@ class System(object):
                         high_size = np.prod(high_dims)
                         dim1 = global_size // high_size
                         if global_size % high_size != 0:
-                            raise RuntimeError(f"{self.msginfo}: Array dimensions above the "
-                                f"first must match across processes. For output '{abs_name}', "
-                                f"local shape {local_shape} has an upper dimension that "
-                                "differs in another process.")
+                            raise RuntimeError(f"{self.msginfo}: All but the first dimension of "
+                                f"the shape's local parts in a distributed variable must match "
+                                f"across processes. For output '{abs_name}', local shape "
+                                f"{local_shape} in MPI rank {MPI.COMM_WORLD.rank} has a higher "
+                                "dimension that differs in another rank.")
                         mymeta['global_shape'] = tuple([dim1] + list(high_dims))
                     else:
                         mymeta['global_shape'] = (global_size,)
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1527,9 +1527,10 @@ class System(object):
                         high_size = np.prod(high_dims)
                         dim1 = global_size // high_size
                         if global_size % high_size != 0:
-                            raise RuntimeError("%s: Global size of output '%s' (%s) does not agree "
-                                               "with local shape %s" % (self.msginfo, abs_name,
-                                                                        global_size, local_shape))
+                            raise RuntimeError(f"{self.msginfo}: Array dimensions above the "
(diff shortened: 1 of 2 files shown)


naylor-b (approved):

openmdao/core/system.py
```diff
@@ -1527,9 +1527,11 @@ def _setup_global_shapes(self):
...
                         dim1 = global_size // high_size
                         if global_size % high_size != 0:
-                            raise RuntimeError("%s: Global size of output '%s' (%s) does not agree "
-                                               "with local shape %s" % (self.msginfo, abs_name,
-                                                                        global_size, local_shape))
+                            raise RuntimeError(f"{self.msginfo}: All but the first dimension of "
+                                f"the shape's local parts in a distributed variable must match "
+                                f"across processes. For output '{abs_name}', local shape "
+                                f"{local_shape} in MPI rank {MPI.COMM_WORLD.rank} has a higher "
+                                "dimension that differs in another rank.")
```
swryan, line 1534: there should be test for this...

swryan (changes requested):

openmdao/core/system.py
tadkollar: In the process of writing the test, I found that `_setup_global_shapes()` would not raise errors in all ranks. Since the exception had to be caught for the test, it would hang waiting for the other process instead of aborting. I added the `_comm_global_shape_status()` function so that all ranks would know whether the global shape was valid and raise an exception if not.

swryan (approved):

PR #2195, opened 2021-07-23
Use backward difference for check options on structured mm with slinear.

### LINKED ISSUE #2194: FD direction for structured metamodel should be backward for 'slinear' to match the direction the bracketing chooses when interpolating on a grid point.
### Summary of Issue

FD direction for structured metamodel should be backward for 'slinear' to match the direction the bracketing chooses when interpolating on a grid point.

### Issue Type

- [x ] Bug
- [ ] Enhancement  

### Description

Forum user saw derivatives reported in check partials were incorrect for a point in his table when he used slinear. The point was on a gridpoint, so the derivative is discontinuous there.  Two lines meet there with two different slopes. The discrepancy in the check happens because the bracketing algorithm, which figures out which bin you are interpolating, chooses the "left" line when you are on a grid point, but the finite difference or complex step chooses the "right" line because the default check direction is "forward".  The fix, to alleviate future confusion, is to use 'backward' difference for slinear.

### Example

See https://stackoverflow.com/questions/68497621/error-when-checking-partial-derivatives-of-slinear-structured-metamodel/68499697#68499697

### Environment

<For bugs, otherwise N/A>

Kenneth-T-Moore:
### Summary

Use backward difference for check options on structured mm when the method is slinear.

This is to avoid confusion when checking partials on this component. Since bracketing looks left and forward difference checks look right, check derivatives would report an erroneous derivative if you tried to interpolate on a grid point.

### Related Issues

- Resolves #2194

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/meta_model_structured_comp.py
+++ b/openmdao/components/meta_model_structured_comp.py
@@ -176,8 +176,8 @@ class MetaModelStructuredComp(ExplicitComponent):
         if self.options['method'].startswith('scipy'):
             self.set_check_partial_options('*', method='fd')
 
-        # Our bracketing album picks the bin behind it if you are interpolating exactly on one of
-        # the grid points, so we needd to set the derivative check to look backwards.
+        # Our bracketing algorithm picks the bin behind it if you are interpolating exactly on one
+        # of the grid points, so we needd to set the derivative check to look backwards.
         elif self.options['method'] == 'slinear':
             self.set_check_partial_options('*', form='backward')
 
--- a/openmdao/components/meta_model_structured_comp.py
+++ b/openmdao/components/meta_model_structured_comp.py
@@ -176,6 +176,11 @@ class MetaModelStructuredComp(ExplicitComponent):
         if self.options['method'].startswith('scipy'):
             self.set_check_partial_options('*', method='fd')
 
+        # Our bracketing album picks the bin behind it if you are interpolating exactly on one of
+        # the grid points, so we needd to set the derivative check to look backwards.
+        elif self.options['method'] == 'slinear':
+            self.set_check_partial_options('*', form='backward')
+
     def compute(self, inputs, outputs):
         """
         Perform the interpolation at run time.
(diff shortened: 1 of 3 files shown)


swryan (approved):

openmdao/components/meta_model_structured_comp.py
```diff
@@ -176,6 +176,11 @@ def _setup_partials(self):
         if self.options['method'].startswith('scipy'):
             self.set_check_partial_options('*', method='fd')
 
+        # Our bracketing album picks the bin behind it if you are interpolating exactly on one of
```
naylor-b, line 179: 'bracketing album'?

openmdao/components/tests/test_meta_model_structured_comp.py
```diff
@@ -1126,10 +1126,36 @@ def test_analysis_error_warning_msg(self):
         p.set_val('x', 0.75)
 
         msg = "Analysis Error: 'interp' <class MetaModelStructuredComp> " \
-              "Line 205 of file {}".format(inspect.getsourcefile(om.MetaModelStructuredComp))
+              "Line 210 of file {}".format(inspect.getsourcefile(om.MetaModelStructuredComp))
```
naylor-b, line 1129: Would be better to rewrite this check so it isn't so fragile.  Adding a line to MetaModelStructuredComp.py shouldn't break this test.

naylor-b (changes requested):

openmdao/components/meta_model_structured_comp.py
Kenneth-T-Moore: haha, should be algorithm

openmdao/components/tests/test_meta_model_structured_comp.py
Kenneth-T-Moore: The fault is the exception message actually, but to make the test more robust, I've added the capability to the `assert_warning` to optionally check if the warning contains the msg rather than check equality.

naylor-b (approved):

PR #2198, opened 2021-07-28
Fix for list_outputs with includes on case object

### LINKED ISSUE #2193: list_outputs on case object is (sometimes) not respecting the `includes` argument.
### Summary of Issue

list_outputs on case object is (sometimes) not respecting the `includes` argument. ### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Title pretty much says it all... one more detail though. 

But only seems to show up when dymos phases are in the model. Simple toy models without dymos phases don't seem to trip it up. 

```
import openmdao.api as om
import dymos as dm
from dymos.examples.plotting import plot_results
from dymos.examples.brachistochrone import BrachistochroneODE
import matplotlib.pyplot as plt
#
# Initialize the Problem and the optimization driver
#
p = om.Problem()

rec = om.SqliteRecorder('testfile.db')
p.model.add_recorder(rec)
#
# Create a trajectory and add a phase to it
#
traj = p.model.add_subsystem('traj', dm.Trajectory())
phase = traj.add_phase('phase0',
                       dm.Phase(ode_class=BrachistochroneODE,
                                transcription=dm.GaussLobatto(num_segments=10)))
#
# Set the variables
#
phase.set_time_options(fix_initial=True, duration_bounds=(.5, 10))
phase.add_state('x', fix_initial=True, fix_final=True)
phase.add_state('y', fix_initial=True, fix_final=True)
phase.add_state('v', fix_initial=True, fix_final=False)
phase.add_control('theta', continuity=True, rate_continuity=True,
                  units='deg', lower=0.01, upper=179.9)
phase.add_parameter('g', units='m/s**2', val=9.80665)
#
# Minimize time at the end of the phase
#
phase.add_objective('time', loc='final', scaler=10)
p.model.linear_solver = om.DirectSolver()
#
# Setup the Problem
#
p.setup()
#
# Set the initial values
#
p['traj.phase0.t_initial'] = 0.0
p['traj.phase0.t_duration'] = 2.0
p.set_val('traj.phase0.states:x', phase.interp('x', ys=[0, 10]))
p.set_val('traj.phase0.states:y', phase.interp('y', ys=[10, 5]))
p.set_val('traj.phase0.states:v', phase.interp('v', ys=[0, 9.9]))
p.set_val('traj.phase0.controls:theta', phase.interp('theta', ys=[5, 100.5]))
#
# Solve for the optimal trajectory
#
# dm.run_problem(p,solution_record_file='testfile.db')
p.run_model()

# Check the results
read_p = om.CaseReader('testfile.db').get_case(-1)
print()

# these two lists of variables should be same, but are not! 
model_outs = p.model.list_outputs(val=False, includes='traj.phases.phase0.timeseries.*')
case_outs = read_p.list_outputs(val=False, includes='traj.phases.phase0.timeseries.*')
```

tadkollar:
### Summary

The example problem was using a `str` arg instead of a list of `str`. Since this could be a common problem, case `list_outputs` now converts `str` args used for `include` and `excludes` to lists before continuing. Also added a test to compare the results from system and case `list_outputs`.

### Related Issues

- Resolves #2193

None

### New Dependencies

None

### DIFF
--- a/openmdao/recorders/case.py
+++ b/openmdao/recorders/case.py
@@ -590,12 +590,12 @@ class Case(object):
             values = True
         else:
             values = val
-        
+
         if isinstance(includes, str):
-            includes = [includes,]
+            includes = [includes, ]
 
         if isinstance(excludes, str):
-            excludes = [excludes,]
+            excludes = [excludes, ]
 
         for var_name in self.outputs.absolute_names():
             if not list_autoivcs and var_name.startswith('_auto_ivc.'):
--- a/openmdao/recorders/case.py
+++ b/openmdao/recorders/case.py
@@ -560,10 +560,10 @@ class Case(object):
             User defined tags that can be used to filter what gets listed. Only outputs with the
             given tags will be listed.
             Default is None, which means there will be no filtering based on tags.
-        includes : str, iter of str, or None
+        includes : str, list of strs, or None
             Glob patterns for pathnames to include in the check. Default is None, which
             includes all.
-        excludes : str, iter of str, or None
+        excludes : str, list of strs, or None
             Glob patterns for pathnames to exclude from the check. Default is None, which
             excludes nothing.
         list_autoivcs : bool
@@ -590,12 +590,12 @@ class Case(object):
             values = True
         else:
             values = val
-
+        
         if isinstance(includes, str):
-            includes = iter(includes)
+            includes = [includes,]
 
         if isinstance(excludes, str):
-            excludes = iter(excludes)
+            excludes = [excludes,]
 
         for var_name in self.outputs.absolute_names():
             if not list_autoivcs and var_name.startswith('_auto_ivc.'):
--- a/openmdao/recorders/case.py
+++ b/openmdao/recorders/case.py
@@ -560,10 +560,10 @@ class Case(object):
             User defined tags that can be used to filter what gets listed. Only outputs with the
             given tags will be listed.
             Default is None, which means there will be no filtering based on tags.
-        includes : iter of str or None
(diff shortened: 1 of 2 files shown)


openmdao/recorders/case.py
```diff
@@ -560,10 +560,10 @@ def list_outputs(self,
             User defined tags that can be used to filter what gets listed. Only outputs with the
             given tags will be listed.
             Default is None, which means there will be no filtering based on tags.
-        includes : iter of str or None
+        includes : str, list of strs, or None
```
naylor-b, line 563: This is a nitpick, but includes and excludes could be any iterator of strs rather than just a list of strs.

openmdao/recorders/case.py
```diff
@@ -591,6 +591,12 @@ def list_outputs(self,
         else:
             values = val
 
+        if isinstance(includes, str):
+            includes = [includes, ]
+
+        if isinstance(excludes, str):
+            excludes = [excludes, ]
+
```
swryan, line 599: Does `Case.list_inputs()`  have the same problem then?

(Note that these methods are meant to mirror `list_inputs`/`list_outputs` on `System`, so the docstrings should probably match)

naylor-b (approved):

openmdao/recorders/case.py
tadkollar: I'll make the change to `Case.list_inputs()` as well. I didn't have to change anything with `System`, but it handles a `str` just fine, so I'll update its docstrings

swryan (approved):

PR #2203, opened 2021-08-02
Added MissingImport checks for various packages

### LINKED ISSUE #1870: Add testing of environments without optional dependencies
### Summary of Issue

Add testing of environments without optional dependencies to ensure OpenMDAO is as functional as expected.

### Issue Type

- [ ] Bug
- [ ] Enhancement  
- [x] Miscellaneous

### Description

When optional dependencies are added to OpenMDAO they are often common dependencies that the developers have. It's easy to accidentally use an `import module` statement and have it slip through the cracks, but we should be testing that these imports are protected in `try` blocks.

There's a context manager at https://gist.github.com/robfalck/ea479099c84f29edd415a1ddf8d28802 which allows users to emulate missing package dependencies, and we should probably be testing to make sure OpenMDAO is at least somewhat functional without these dependencies (MPI, IPython, etc)

### Example

The following code should not raise an exception (OpenMDAO should import cleanly in the absence of IPython)

```python
    with MissingImports('IPython'):
        import openmdao.api as om
```

### Environment

N/A

DKilkenny:
### Summary

Added MissingImport checks for various packages

### Related Issues

- Resolves #1870

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/tests/test_testing_utils.py
+++ b/openmdao/utils/tests/test_testing_utils.py
@@ -6,8 +6,7 @@ from openmdao.utils.testing_utils import use_tempdirs, MissingImports
 @use_tempdirs
 class TestMissingImports(unittest.TestCase):
 
-    def test_missing_imports(self):
-
+    def test_missing_imports_cm(self):
         with self.assertRaises(ImportError) as e:
             with MissingImports('pyoptsparse'):
                 import pyoptsparse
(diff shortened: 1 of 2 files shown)


openmdao/utils/tests/test_testing_utils.py
```diff
@@ -0,0 +1,103 @@
+import unittest
+
+from openmdao.utils.testing_utils import use_tempdirs, MissingImports
+
+
+@use_tempdirs
+class TestMissingImports(unittest.TestCase):
```
swryan, line 7: It looks like you are just testing the Missing Imports context manager here...

I think the intent of the Issue was to make sure that openmdao features don't blow up if some optional dependency is not installed, eg. (from the issue description):

```
    with MissingImports('IPython'):
        import openmdao.api as om
```

Just replacing all your imports below with `import openmdao.api` may be a good baseline test...  
Could maybe also try importing some class related to a specific dependency as a further test...

swryan (changes requested):

swryan (approved):

robfalck (approved):

PR #2209, opened 2021-08-05
Modifications to relative step sizing in finite difference

### LINKED ISSUE #2133: Improvement needed for relative FD step size
### Summary of Issue

The current FD step size calculation is highly unorthodox and gives nonintuitive behavior. It needs to be adjusted to better match expectations. 

### Issue Type

- [x] Bug
- [x] Enhancement  

### Description

A POEM needs to be written to detail this change! 

The current FD step size calculation looks like: 

``` python
 if step_calc == 'rel':
            if system._outputs._contains_abs(wrt):
                step *= np.linalg.norm(system._outputs._abs_get_val(wrt))
            elif system._inputs._contains_abs(wrt):
                step *= np.linalg.norm(system._inputs._abs_get_val(wrt))
```

the use of a norm there, causes weird behavior where the step size varies with vector length. The step for an input of `np.ones(10)` would be different than the step for an input of `np.ones(100)`. This i highly non-intuitive. 

Part of the motivation of this approach might have been to use some kind of a vector average value because you might have a vector with some incidental zeros in it. This would make computing element-wise relative step sizes difficult. You could use a avg(abs(input)) instead to combat this. 

However, the most intuitive behavior would be an element wise relative step. This would suffer the incidental-zero weekness, but users could select absolute value instead. 

A POEM should be written to discuss both options pros/cons. A final decision on which option to use can be reached by the dev-team after getting some user input. 

Changing the relative step size calculation is going to introduce some numerical noise to many users programs. We'll need to give them a heads up about this. We should also make a transition release that gives a warning to anyone using relative step size about the change in step-calculation method in the following release (that way no one gets caught totally off guard). 

Finally, the docs need to be updated to detail whatever method we use for relative step size.

Kenneth-T-Moore:
### Summary

This is the implementation for accepted POEM 051.

1. New step_calc modes added for computing the relative finite difference step size.
- 'rel_avg' scale by average absolute value of the vector variable.
- 'rel_element' for each element, scale by its own current value.
- 'rel_legacy' scale by the norm of the vector variable. This is the current behavior for 'rel'
2. The step_calc 'rel' mode will be changed to the behavior of 'rel_avg' in the next release. A deprecation warning has been added.
3. New 'minimum_step' argument has been added to prevent scaling the step size by 0. The default is 1e-12.

### Related Issues

- Resolves #2133

Behavior of the 'rel' step_calc will change, though current behavior is considered a mild bug. A deprecation warning is raised if you use this. The behavior of 'rel' will be fixed in the next release. The 'rel_legacy' choice keeps the old behavior and will continue to be available for those who need it.

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/finite_difference.py
+++ b/openmdao/approximation_schemes/finite_difference.py
@@ -181,30 +181,34 @@ class FiniteDifference(ApproximationScheme):
         fd_form = _generate_fd_coeff(form, order, system)
 
         if step_calc != 'abs':
-            if system._inputs._contains_abs(wrt):
+            var_local = True
+            if system._outputs._contains_abs(wrt):
+                wrt_val = system._outputs._abs_get_val(wrt)
+            elif system._inputs._contains_abs(wrt):
                 wrt_val = system._inputs._abs_get_val(wrt)
             else:
-                wrt_val = system._outputs._abs_get_val(wrt)
+                var_local = False
 
-            # TODO - behavior or 'rel' will switch to 'rel_avg' in an upcoming release.
-            if step_calc == 'rel_legacy' or step_calc == 'rel':
-                step *= np.linalg.norm(wrt_val)
+            if var_local:
+                # TODO - behavior or 'rel' will switch to 'rel_avg' in an upcoming release.
+                if step_calc == 'rel_legacy' or step_calc == 'rel':
+                    step *= np.linalg.norm(wrt_val)
 
-                if step < minimum_step:
-                    step = minimum_step
+                    if step < minimum_step:
+                        step = minimum_step
 
-            elif step_calc == 'rel_avg':
-                step *= np.sum(np.abs(wrt_val)) / len(wrt_val)
+                elif step_calc == 'rel_avg':
+                    step *= np.sum(np.abs(wrt_val)) / len(wrt_val)
 
-                if step < minimum_step:
-                    step = minimum_step
+                    if step < minimum_step:
+                        step = minimum_step
 
-            else:  # 'rel_element'
-                step = np.abs(wrt_val) * step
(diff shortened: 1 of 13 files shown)


openmdao/approximation_schemes/finite_difference.py
```diff
@@ -117,6 +121,23 @@ def add_approximation(self, abs_key, system, kwargs, vector=None):
...
+            elif options['directional'] and step_calc == 'rel_element':
+                raise ValueError(f"{system.msginfo}: Option 'directional' is not supported when "
+                                 "'step_calc' is set to 'rel_element.'")
+
+            elif step_calc == 'rel':
+                warn_deprecation("When using 'rel' as the step_calc, the fd stepsize is currently "
+                                 "scaled by the norm of the vector variable. This is not ideal for"
+                                 " larger vectors, and this behavior is being changed in "
+                                 "OpenMDAO 3.12.0. To preserve the older way of doing this "
+                                 "calculation, set step_calc to 'rel_legacy'.")
```
swryan, line 139: It seems like we could make use of an `OptionsDictionary` to handle some of these checks (validity & deprecation)... is there a reason we wouldn't want to do that?

Kenneth-T-Moore: I think the main reason is that the default values branch and some options become valid/invalid depending on whether you choose 'cs' or 'fd'.

Kenneth-T-Moore: Though, if you mean specifically on the `finite_difference` object, we could do it.  I reckon this class was crafted before we started using the OptionsDictionary for call arguments.

robfalck (approved):

swryan (approved):

PR #2210, opened 2021-08-05
Added support for multidimensional shapes in `EQConstraintComp` and `BalanceComp`

### LINKED ISSUE #2175: `normalize` option in BalanceComp and EQConstraintComp incompatible with multidimensional variables
### Summary of Issue

Compute in EQConstraintComp and BalanceComp will fail when presented with a multidimensional input.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The following code in compute only pulls the first index of rhs where the absolute value is close to/not close to zero.

```
idxs_nz = np.where(cs_safe.abs(rhs) < 2)[0]
idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)[0]
```

For multidimensional inputs, this results in an indexing error when attempting to apply the normalization to the scale factor:

```
# scale factor that normalizes by the rhs, except near 0
self._scale_factor[idxs_nnz] = 1.0 / cs_safe.abs(rhs[idxs_nnz])
self._scale_factor[idxs_nz] = 1.0 / (.25 * rhs[idxs_nz] ** 2 + 1)
```

`_scale_factor` and `_dscale_drhs` need to be stored per-variable (probably via a dict keyed by `name` in the same way that `_output_vars` is).

The storing of indices and the assignment back into `_scale_factor` and `_dscale_drhs` need to be compatible with n-dimensional variables.

Alternatively,  the compute and partials methods could be written such that `_scale_factor` and `_dscale_drhs` aren't preallocated.

### Example

```
    def test_shaped(self):
        prob = om.Problem()
        model = prob.model

        n = 15

        defect_comp = model.add_subsystem('eq_comp', om.EQConstraintComp())

        defect_comp.add_eq_output(name='diff_r',
                                  shape=(n, 3),
                                  eq_units='km',
                                  lhs_name='r1', rhs_name='r2',
                                  use_mult=False,
                                  normalize=True,
                                  add_constraint=True,
                                  ref=100.0, ref0=None, adder=None, scaler=None)

        defect_comp.add_eq_output(name='diff_k',
                                  shape=(n,),
                                  eq_units='km',
                                  lhs_name='k1', rhs_name='k2',
                                  use_mult=False,
                                  normalize=True,
                                  add_constraint=True,
                                  ref=100.0, ref0=None, adder=None, scaler=None)


        prob.setup()

        prob.set_val('eq_comp.r1', np.random.rand(n, 3) * 10)
        prob.set_val('eq_comp.r2', np.random.rand(n, 3) * 10)

        prob.set_val('eq_comp.k1', np.random.rand(n) * 10)
        prob.set_val('eq_comp.k2', np.random.rand(n) * 10)

        prob.run_model()
```

### Environment

N/A

DKilkenny:
### Summary

Added support for multidimensional shapes when using `add_constraint` with `EQConstraintComp` or `add_balance` with `BalanceComp`. This applies when using/not using the `normalize` arg

### Related Issues

- Resolves #2175

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/balance_comp.py
+++ b/openmdao/components/balance_comp.py
@@ -159,7 +159,6 @@ class BalanceComp(ImplicitComponent):
         residuals : Vector
             unscaled, dimensional residuals written to via residuals[key]
         """
-
         for name, options in self._state_vars.items():
             lhs = inputs[options['lhs_name']]
             rhs = inputs[options['rhs_name']]
--- a/openmdao/components/balance_comp.py
+++ b/openmdao/components/balance_comp.py
@@ -170,11 +170,12 @@ class BalanceComp(ImplicitComponent):
 
             if options['normalize']:
                 # Indices where the rhs is near zero or not near zero
-                idxs_nz = np.where(cs_safe.abs(rhs) < 2)[0]
-                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)[0]
+                idxs_nz = np.where(cs_safe.abs(rhs) < 2)
+                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)
 
                 # Compute scaling factors
                 # scale factor that normalizes by the rhs, except near 0
+                self._scale_factor = np.ones((rhs.shape))
                 self._scale_factor[idxs_nnz] = 1.0 / cs_safe.abs(rhs[idxs_nnz])
                 self._scale_factor[idxs_nz] = 1.0 / (.25 * rhs[idxs_nz] ** 2 + 1)
             else:
--- a/openmdao/components/balance_comp.py
+++ b/openmdao/components/balance_comp.py
@@ -159,14 +159,11 @@ class BalanceComp(ImplicitComponent):
         residuals : Vector
             unscaled, dimensional residuals written to via residuals[key]
         """
-        if inputs._under_complex_step:
-            self._scale_factor = self._scale_factor.astype(complex)
-        else:
-            self._scale_factor = self._scale_factor.real
 
         for name, options in self._state_vars.items():
             lhs = inputs[options['lhs_name']]
             rhs = inputs[options['rhs_name']]
+            _scale_factor = np.ones((rhs.shape))
 
             if options['normalize']:
                 # Indices where the rhs is near zero or not near zero
@@ -175,16 +172,13 @@ class BalanceComp(ImplicitComponent):
 
                 # Compute scaling factors
                 # scale factor that normalizes by the rhs, except near 0
-                self._scale_factor = np.ones((rhs.shape))
-                self._scale_factor[idxs_nnz] = 1.0 / cs_safe.abs(rhs[idxs_nnz])
-                self._scale_factor[idxs_nz] = 1.0 / (.25 * rhs[idxs_nz] ** 2 + 1)
-            else:
-                self._scale_factor[:] = 1.0
+                _scale_factor[idxs_nnz] = 1.0 / cs_safe.abs(rhs[idxs_nnz])
+                _scale_factor[idxs_nz] = 1.0 / (.25 * rhs[idxs_nz] ** 2 + 1)
 
             if options['use_mult']:
-                residuals[name] = (inputs[options['mult_name']] * lhs - rhs) * self._scale_factor
+                residuals[name] = (inputs[options['mult_name']] * lhs - rhs) * _scale_factor
             else:
-                residuals[name] = (lhs - rhs) * self._scale_factor
+                residuals[name] = (lhs - rhs) * _scale_factor
 
     def linearize(self, inputs, outputs, jacobian):
         """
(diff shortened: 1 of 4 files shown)


openmdao/components/tests/test_eq_constraint_comp.py
```diff
@@ -540,6 +540,43 @@ def test_renamed_vars(self):
 
         assert_check_partials(cpd, atol=1e-5, rtol=1e-5)
 
+    def test_multidimentional_shape(self):
```
Kenneth-T-Moore, line 543: Need a test for BalanceComp as well.

openmdao/components/eq_constraint_comp.py
```diff
@@ -130,9 +130,10 @@ def compute(self, inputs, outputs):
             # scale factor that normalizes by the rhs, except near 0
             if options['normalize']:
                 # Indices where the rhs is near zero or not near zero
-                idxs_nz = np.where(cs_safe.abs(rhs) < 2)[0]
-                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)[0]
+                idxs_nz = np.where(cs_safe.abs(rhs) < 2)
+                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)
 
+                self._scale_factor = np.ones((rhs.shape))
```
Kenneth-T-Moore, line 136: The missing part of this story is that we are saving `_scale_factor ` and `_dscale_drhs` in the class, but we are only saving the value from the last variable in the for loop. Choices were: 1) save in dictionary keyed by varname, or 2) don't save at all and remove it everywhere except where it is calculated.

Your job on this story is to evaluate whether we really *need* to save it, and make the call between 1 and 2.  I think Rob has already indicated that we don't need to query it after the model runs, so it's just a matter of whether the values are re-used internally.

openmdao/components/balance_comp.py
```diff
@@ -170,11 +170,12 @@ def apply_nonlinear(self, inputs, outputs, residuals):
```
Kenneth-T-Moore, line 170: You are not finished with the story.  Saw this in the last PR too, so if you are pushing up early to get a good CI run in, we can talk to Steve about doing CI on your local branch, and also about fixing any problems in your local env.

openmdao/components/tests/test_eq_constraint_comp.py
```diff
@@ -540,6 +540,43 @@ def test_renamed_vars(self):
...
+        prob.setup()
+
+        prob.set_val('eq_comp.r1', np.random.rand(n, 3) * 10)
+        prob.set_val('eq_comp.r2', np.random.rand(n, 3) * 10)
+
+        prob.set_val('eq_comp.k1', np.random.rand(n) * 10)
+        prob.set_val('eq_comp.k2', np.random.rand(n) * 10)
+
+        prob.run_model()
+
```
Kenneth-T-Moore, line 579: One more thing. We need better descriptions for the PR so that the public understands what this fixed, and Rob can put together the patch notes more easily. 

"Added support for multidimentional support for EQConstraintComp and BalanceComp"

Spelling and grammar aside, this doesn't tell me anything about what bug was fixed.  It isn't even technically correct.

Kenneth-T-Moore (changes requested):

openmdao/components/balance_comp.py
```diff
@@ -170,11 +170,12 @@ def apply_nonlinear(self, inputs, outputs, residuals):
...
             if options['normalize']:
                 # Indices where the rhs is near zero or not near zero
-                idxs_nz = np.where(cs_safe.abs(rhs) < 2)[0]
-                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)[0]
+                idxs_nz = np.where(cs_safe.abs(rhs) < 2)
+                idxs_nnz = np.where(cs_safe.abs(rhs) >= 2)
 
                 # Compute scaling factors
                 # scale factor that normalizes by the rhs, except near 0
+                self._scale_factor = np.ones((rhs.shape))
```
robfalck, line 178: If we're reallocating _scale_factor here on each pass through, then it should just be a local variable and not an attribute of the class.  Move it up out of the `if` clause here and remove the else-clause so that it's correct when `normalize` is `False`.

openmdao/components/eq_constraint_comp.py
robfalck, line 136: Same comment as above applies here.  Don't bother making `_scale_factor` an attribute if we're just using it locally and not saving it.

openmdao/components/tests/test_eq_constraint_comp.py
robfalck, line 543: We should have an equivalent test for BalanceComp, and we should probably also test this with `normalize=False`.

robfalck (changes requested):

Kenneth-T-Moore (approved):

PR #2212, opened 2021-08-06
POEM_053: Make `src_indices` consistent with other indices

### LINKED ISSUE #2140: Users don't get useful error messages with src_indices errors in promotes.
### Summary of Issue

src_indices errors that result in useful error messages when the src_indices were provided on `add_input` do not result in useful errors in `promotes`.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

In the following code, the src_indices are not valid (9 is out of range for the flattened indies of the src_shape (3, 3)).
When src_indices was applied to the input, this gave a clear error message.
Now that it is applied in promotes, the resulting error message is generic and harder to diagnose.

### Example

```
    def test_case_for_issue(self):
        src_shape = (3, 3)
        tgt_shape = (2, 2)
        src_indices = [[4, 5], [7, 9]]
        flat_src_indices = True

        class MyComp(om.ExplicitComponent):
            def __init__(self, input_shape):
                super().__init__()
                self._input_shape = input_shape

            def setup(self):
                self.add_input('x', val=np.zeros(self._input_shape))
                self.add_output('y', val=np.zeros(self._input_shape))

                outputs['y'] = 2.0 * inputs['x']

        prob = om.Problem()
        prob.model.add_subsystem('indeps', om.IndepVarComp('x', shape=src_shape))
        prob.model.add_subsystem('C1', MyComp(tgt_shape))

        prob.model.promotes('C1', any=['x'], src_indices=src_indices,
                            flat_src_indices=flat_src_indices)
        prob.model.set_input_defaults('x', src_shape=src_shape)

        prob.setup()
```

### Environment

OpenMDAO 3.9.3-dev

### LINKED ISSUE #2148: Behavior of src_indices vs. indices defined for design vars and constraints is currently inconsistent.
### Summary of Issue

Currently, indices defined for design vars and constraints behave in the same way that indexing into a numpy array behaves, but src_indices allows a different syntax where the indices are defined as a possibly nested list of tuples where tuples of length 'n' indicates that the array being indexed is of dimension 'n'.  The equivalent numpy indexing syntax would be a tuple of index arrays, one for each dimension of the array being indexed.

The goal is to make src_indices behave (like design var and constraint indices) in the same way as indices applied to a normal numpy array.  This change will be made on a branch and tested with openmdao and dymos and assuming there are no major problems, there will be a POEM written describing the issue and advocating for the removal of the non-standard indexing syntax from src_indices.

swryan:
### Summary

Make `src_indices` behave (like design var and constraint indices) in the same way as indices applied to a normal numpy array. 

### Related Issues

- Resolves #2148 
- Resolves #2140

None

### New Dependencies

None


openmdao/core/group.py
```diff
@@ -1992,101 +2004,27 @@ def _setup_connections(self):
...
-                                else:
-                                    issue_warning(msg, category=SetupWarning)
-                                    fail = True
-                        if src_indices.ndim > 1:
-                            meta_in['src_indices'] = src_indices.ravel()
+                            issue_warning(msg, category=SetupWarning)
+                            fail = True
+                            continue
+
+                    if src_indices.size != shape_to_len(in_shape):
```
naylor-b, line 2020: Is this the same check we're already doing on line 1990?

openmdao/utils/mpi.py
```diff
@@ -191,6 +191,8 @@ def multi_proc_exception_check(comm):
     """
     Raise an exception on all procs if it is raised on one.
 
+    Exception raised with be the one from the lowest rank where an exception occurred.
```
naylor-b, line 194: with --> will

naylor-b (changes requested):

openmdao/core/group.py
swryan: yup.. removed

openmdao/utils/mpi.py
swryan: fixed

naylor-b (approved):

robfalck (approved):

swryan:
This PR to be replaced with new and improved one from @naylor-b

PR #2228, opened 2021-08-16
I2219 refactor playwright code to enable re-use with other GUI testing

### LINKED ISSUE #2219: Refactor testing code using Playwright so it can be used for tests beyond those for N2
### Summary of Issue

There are already are and might also be more "GUIs" to test in OpenMDAO. To avoid duplicating code, parts of the current Playwright code for testing the N2 should be broken out so that they can be re-used.

### Issue Type

- [ ] Bug
- [X] Enhancement  

### Description

The current testing class, `n2_gui_test_case` has some code that is not specific to N2 diagram testing and can be broken out into a separate class that can be inherited from so that it can be re-used for other "GUI" testing. One example is testing the search capabilities of the docs.

More specifically, the methods in that class that handle errors, logging, interacting with the GUI, and setup can be re-used. 

### Example

N/A

### Environment

N/A

hschilling:
### Summary

The existing code in the N2 GUI testing code couldn't be re-used for other GUI testing applications  in OpenMDAO. This PR creates a class that can be inherited and re-used in other GUI testing applications.

Also fixed some PEP errors.

### Related Issues

- Resolves #2219 

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/gui_testing_utils.py
+++ b/openmdao/utils/gui_testing_utils.py
@@ -1,10 +1,21 @@
 """Define utils for use in testing GUIs with Playwright."""
 import unittest
 
+
 class gui_test_case(unittest.TestCase):
+    """
+    Class that can be inherited from and used for GUI tests using Playwright.
+    """
 
     def handle_console_err(self, msg):
-        """ Invoked any time that an error or warning appears in the log. """
+        """
+        Invoke any time that an error or warning appears in the log.
+
+        Parameters
+        ----------
+        msg : string
+            message string from console.
+        """
         if msg.type == 'warning':
             self.console_warning = True
             print('    Console Warning: ' + msg.text)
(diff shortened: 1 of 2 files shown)


tadkollar (approved):

openmdao/utils/gui_testing_utils.py
```diff
@@ -0,0 +1,104 @@
+"""Define utils for use in testing GUIs with Playwright."""
+import unittest
+
+
+class gui_test_case(unittest.TestCase):
```
swryan, line 5: Maybe we should "CamelCase" this class name (i.e. GUITestCase) to conform with PEP8:
https://www.python.org/dev/peps/pep-0008/#class-names

hschilling: Changed. Thanks.

swryan (approved):

PR #2236, opened 2021-08-20
Some improvements to MetaModelStructuredComp documentation for dynamically generated training values.

### LINKED ISSUE #2048: Improve docs for StructuredMetaModel
### Summary of Issue

Include description and example of structured metamodel with derivatives w.r.t training data. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Docs on structured meta-model are not clear on the behavior when you turn on derivatives for training data. We need to 
1) include an example of how that works that shows the total derivative computation
2) provide clear definition of which derivatives are supported. Do we allow changes in just the training data, or also in the locations of the training data points as well?

Kenneth-T-Moore:
### Summary

1. Rewrote the notebook example for the MetaModelStructuredComp for dynamically generated training values.
2. Modified MetaModelStructuredComp and MetaModelSemiStructuredComp so that the training values aren't required in `add_output` if training_data_gradients is True.

### Related Issues

- Resolves #2048 

None

### New Dependencies

None


openmdao/components/meta_model_semi_structured_comp.py
```diff
@@ -118,11 +118,20 @@ def add_output(self, name, training_data, **kwargs):
         n = self.options['vec_size']
         super().add_output(name, np.ones(n), **kwargs)
 
-        self.training_outputs[name] = training_data
-
         if self.options['training_data_gradients']:
+            if training_data is None:
+                for item in self.training_inputs.values():
+                    n_train = len(item)
+                    break
```
swryan, line 125: maybe a code comment here?  
in isolation, it's not clear why there is an iterator with an immediate break...

swryan (approved):

naylor-b (approved):

robfalck (approved):

PR #2237, opened 2021-08-23
I2113 Provide a way for users to exclude search results from source docs

### LINKED ISSUE #2113: Let the user limit the search of the docs so it doesn't show results from the source docs
### Summary of Issue

If you do a search in the new Jupyter Book based docs, the results include hits on the source docs. Often the user really only wants to see results from the actual docs. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Let the user have the option to include or exclude search results from the source docs. 

It's probable that in the old docs work was done to modify the sphinx search, but that didn't get transferred over to the new doc system.

### Example

There are many examples, but one is searching for "distributed"

### Environment

N/A

hschilling:
### Summary

The current doc search returns results from the actual docs and source docs. Often, users really only want to search the actual docs. This PR provides a UI element to turn off and on including source docs in the search results.

### Related Issues

- Resolves #2113 

None

### New Dependencies

None


swryan (approved):

Kenneth-T-Moore (approved):
Tested. Seems to work well.

PR #2248, opened 2021-09-02
Transitional release prior to full POEM054 implementation.

### LINKED ISSUE #2247: Need transitional release prior to full POEM054 implementation.
### Summary of Issue

We're changing the default way that indices into openmdao source arrays are interpreted when the indices appear to be flat,
but this will break some existing models, so we first need a transitional release that issues warnings in cases where the behavior will change without actually changing the behavior.

### Issue Type

- [ ] Bug
- [ x] Enhancement  
      POEM #054 (part 1)

naylor-b:
### Summary

This PR issues warnings in cases where apparently flat indices index into non-flat source arrays when the flat_src_indices or flat_indices flags have not been set to True.

### Related Issues

- Resolves #2247

None

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/linear/tests/test_direct_solver.py
+++ b/openmdao/solvers/linear/tests/test_direct_solver.py
@@ -488,7 +488,7 @@ class TestDirectSolver(LinearSolverTests.LinearSolverTestCase):
         model.connect('c1.y', 'c2.x')
         model.connect('c2.y', 'c3.x')
         model.connect('c3.y', 'c4.x')
-        model.connect('c4.y', 'c5.x', src_indices=([0]))
+        model.connect('c4.y', 'c5.x', src_indices=([0]), flat_src_indices=True)
 
         model.linear_solver = om.DirectSolver(assemble_jac=True)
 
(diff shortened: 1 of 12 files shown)


openmdao/solvers/linear/tests/test_direct_solver.py
```diff
@@ -546,7 +546,7 @@ def test_error_on_singular_with_densejac_bug(self):
         model.connect('c1.y', 'c2.x')
         model.connect('c2.y', 'c3.x')
         model.connect('c3.y', 'c4.x')
-        model.connect('c4.y', 'c5.x', src_indices=([0]))
+        model.connect('c4.y', 'c5.x', src_indices=([0]), flat_src_indices=True)
```
Kenneth-T-Moore, line 549: Have you checked through the notebook docs to see if we need to add the flat_src_indices argument there to get rid of warnings?

Kenneth-T-Moore (changes requested):

openmdao/solvers/linear/tests/test_direct_solver.py
naylor-b: Yeah I checked.  No changes were needed.

Kenneth-T-Moore (approved):

swryan (approved):

PR #2269, opened 2021-09-14
pin ipyparallel because of some problems in v 7

Kenneth-T-Moore:
### Summary

Recent updates to the ipyparallel package's config file API caused our CI to fail. Pinning to 7 fixes this and any future changes they make. We can revisit in the future whether we need any new features they add.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore:
Fixed in #2267

PR #2279, opened 2021-09-28
This completes the transition to numpy style indexing described in POEM_054.

naylor-b:
### Summary

Numpy style indexing is now assumed and the old openmdao specific indexing syntax is no longer supported.


openmdao/core/system.py
```diff
@@ -5333,3 +5321,17 @@ def _get_jac_col_scatter(self):
             tarr -= toffset
 
         return sarr, tarr, tsize, has_dist_data
+
+    def _resolve_src_shape(self, input_name):
```
Kenneth-T-Moore, line 5325: Am I missing something, or does this do nothing?

Kenneth-T-Moore (changes requested):

openmdao/core/system.py
naylor-b: Oops, leftover code from some aborted early iteration.  I removed it.

Kenneth-T-Moore (approved):

openmdao/utils/general_utils.py
```diff
@@ -175,16 +174,21 @@ def ensure_compatible(name, value, shape=None, indices=None):
             shape = tuple(shape)
     elif not np.isscalar(value):
         shape = np.atleast_1d(value).shape
-    if indices is not None and indices.shaped_instance() is not None:
-        if indices.src_ndim > 1 and shape is None:
+
+    if indices is not None:  # and indices.shaped_instance() is not None:
```
swryan, line 178: clean up commented code?

openmdao/utils/general_utils.py
```diff
@@ -175,16 +174,21 @@ def ensure_compatible(name, value, shape=None, indices=None):
...
                                "shape must be provided." % name)
-        indshape = shape2tuple(indices.shape)
-        if shape is not None and indshape != shape:
-            raise ValueError("Shape of indices %s does not match shape of %s for '%s'." %
-                             (indshape, shape, name))
-        if shape is None:
-            shape = indshape
+        try:
+            indshape = indices.indexed_src_shape
+        except Exception:
```
swryan, line 184: what kind of exception are we expecting here?  should be more specific...

naylor-b: Could be different kinds (TypeError, ValueError, IndexError) so I just used a general exception.

openmdao/utils/general_utils.py
naylor-b: done

openmdao/utils/general_utils.py
swryan: Perhaps it would be worth adding the tuple, so we could  pass this check:
https://bandit.readthedocs.io/en/latest/plugins/b110_try_except_pass.html

swryan (approved):

PR #2281, opened 2021-10-01
Implementation PR for POEM 056 (function metadata API)

naylor-b:
### Summary

This implements the function metadata API.

Also:
  - fixes a bug in Vector where calling set_var on a scalar variable wasn't updating the shared array.
  - fixes a bug in the docstring linter (was returning class name instead of method name)
  - fixes an incorrect ExecComp test name

None

### New Dependencies

None


openmdao/func_api.py
```diff
@@ -0,0 +1,826 @@
...
+            else:
+                try:
+                    shp = meta['shape']
+                except KeyError:
+                    raise RuntimeError(f"Can't determine shape of input '{name}'.")
+                if jax is not None:
+                    args.append(jax.ShapedArray(_shape2tuple(shp), dtype=np.float64))
+
+        # compute shapes as a check against annotated value (if any)
+        if jax is not None:
```
Kenneth-T-Moore, line 521: Why isn't this `self._use_jax` instead? Seems like the user might have jax imported, but they aren't using it and haven't specified it as the declare_partials method.  I guess it just seems weird that importing jax into your local env changes the behavior of this component.  Even though that is a major motivation for this component, there might be a reason it's not being used at a time.

openmdao/func_api.py
```diff
@@ -0,0 +1,826 @@
...
+from numbers import Number
+import ast
+import inspect
+import textwrap
+import warnings
+import numpy as np
+from contextlib import contextmanager
+
+try:
+    import jax
```
Kenneth-T-Moore, line 14: I think to test certain behaviors, we'll need to add jax to the testing dependencies in setup.py.

and additionally, add some tests were jax is present.

openmdao/vectors/vector.py
```diff
@@ -538,20 +548,26 @@ def set_var(self, name, val, idxs=_full_slice, flat=False):
...
                 try:
-                    value = value.reshape(self._views[abs_name][idxs()].shape)
+                    value = value.reshape(view[idxs()].shape)
                 except Exception:
                     raise ValueError(f"{self._system().msginfo}: Failed to set value of "
                                      f"'{name}': {str(err)}.")
-                self._views[abs_name][idxs()] = value
+                view[idxs()] = value
 
     def dot(self, vec):
```
Kenneth-T-Moore, line 572: Normally I'd say we need an example for docs, but in this case, the function API doesn't have any practical use until the FuncComp is added.

Kenneth-T-Moore (changes requested):

openmdao/func_api.py
naylor-b: This block uses jax only to compute output shapes automatically, not to compute derivatives or anything, so the reasoning for doing it this way is that if jax is available we should always use it as a check on the output shapes.  Unless we're worried about performance or something due to calling jax.make_jaxpr, I don't see why we wouldn't just always do this, assuming jax is installed.

openmdao/vectors/vector.py
naylor-b: Yeah there isn't much reason for someone to use the function API until we have function comps, and even if they do want to play with it, they could look at POEM 056 which explains it pretty thoroughly.

openmdao/func_api.py
Kenneth-T-Moore: I guess one reason might be: they are doing something weird in this function that jax doesn't like, but they imported it for their other functions.

naylor-b: Actually I just noticed that _compute_out_shapes doesn't even get called unless self._use_jax is True, so disregard my earlier comment.

openmdao/func_api.py
naylor-b: I added a couple of tests for the jax based output shape computation.  I'm not sure about making jax a hard testing dependency though.

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2290, opened 2021-10-06
Prevent mpi4py & petsc4py from importing when they're not wanted

### LINKED ISSUE #2288: mpi still initialized in corner use case, even when `OPENMDAO_REQUIRE_MPI=False` is set as an env-var
### Summary of Issue

even when `OPENMDAO_REQUIRE_MPI=False` is set as an env-var, OpenMDAO still tries to initialize mpi4py if a user has it installed on their system. 

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Reported by Ryan Durscher and Nathan Wukie 

> "... if I have mpi4py installed. I have discovered that the constructor for om.Problem() has the following code:
>  
> 146         if comm is None:
> 147             try:
> 148                 from mpi4py import MPI
> 149                 comm = MPI.COMM_WORLD
> 150             except ImportError:
> 151                 comm = FakeComm()
>  
> That is, in creating the om.Problem() object in our example case, it will attempt to initialize mpi4py regardless of the value of OPENMDAO_USE_MPI, causing the fun3d.system(CAPS method) call to fail for the same reason as before.
>  
> I’ve found this can be kludged by passing in openmdao’s FakeComm() object to the om.Problem constructor.
>  
> 215 omProblem = om.Problem(comm=FakeComm())
>  
> Which requires the import
>  
> 10 from openmdao.utils.mpi import FakeComm

Additional bug data from users: 

> We have discovered that if a user has petsc4py built/installed then we observe the same issue that subsequent mpi-application spawns fail. PetscInitialize says it will call MPI_Init if it has not yet been called. Presumably importing petsc4py calls PetscInitialize. I don't currently see a way to prevent this through the petsc4py api.

> It seems that if the user requests OPENMDAO_USE_MPI=False then petsc4py should be disabled as well.

tadkollar:
### Summary

In problem.py, mpi4py was still being imported without checking the value of `OPENMDAO_USE_MPI`. Also, petsc_vector.py and petsc_transfer.py were importing petsc4py which called `MPI_Init()` without checking `OPENMDAO_USE_MPI`.

This fix uses the value of `MPI` as set by `openmdao.utils.mpi` in problem.py to determine whether mpi4py is in use. In the other two files, `check_mpi_env()` is called to determine the value of `OPENMDAO_USE_MPI` before importing petsc4py and defining the `PETScVector` and `PETScTransfer` classes.

The commits make it look like a lot was changed due to the indentation, but it was only a few lines toward the top of each file.

### Related Issues

- Resolves #2288

### DIFF
--- a/openmdao/docs/openmdao_book/features/building_blocks/solvers/petsc_krylov.ipynb
+++ b/openmdao/docs/openmdao_book/features/building_blocks/solvers/petsc_krylov.ipynb
@@ -3,23 +3,24 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "source": [
-    "try:\n",
-    "    from openmdao.utils.notebook_utils import notebook_mode\n",
-    "except ImportError:\n",
-    "    !python -m pip install openmdao[notebooks]"
-   ],
-   "outputs": [],
    "metadata": {
     "tags": [
      "remove-input",
      "active-ipynb",
      "remove-output"
     ]
-   }
+   },
+   "outputs": [],
+   "source": [
+    "try:\n",
+    "    from openmdao.utils.notebook_utils import notebook_mode\n",
+    "except ImportError:\n",
+    "    !python -m pip install openmdao[notebooks]"
+   ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "# PETScKrylov\n",
     "\n",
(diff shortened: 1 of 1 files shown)


tadkollar:
The Actions doc build is failing with these messages:
```
WARNING: Execution Failed with traceback saved in /home/runner/work/OpenMDAO/OpenMDAO/openmdao/docs/openmdao_book/_build/html/reports/linear_user_defined.log
WARNING: Execution Failed with traceback saved in /home/runner/work/OpenMDAO/OpenMDAO/openmdao/docs/openmdao_book/_build/html/reports/petsc_krylov.log
```

I don't think I can get to those log files. But when I build locally, it fails with this:
```
sphinx.errors.SphinxWarning: Execution Failed with traceback saved in /Users/tkollar/work/om/mpi2288/openmdao/docs/openmdao_book/_build/html/reports/doe_driver.log
```

End of doe_driver.log:
```
CompositeError: one or more exceptions from call to method: execute
[0:execute]: OperationalError: table global_iterations already exists
[2:execute]: OperationalError: attempt to write a readonly database
```

Kenneth-T-Moore:
Do the docs build if you get rid of that third condition?

PR #2310, opened 2021-10-15
Fix for output of reverse mode partials if they were divided by comm size

### LINKED ISSUE #2262: Bug in check_partials when running reverse mode matrix-free apis in parallel
### Summary of Issue

See this SO question: 

https://stackoverflow.com/questions/69110573/optimizing-distributed-i-o-with-serial-output/69111105#69111105

Note that when run in parallel with 2 procs, the reverse mode partials of the serial wrt the distributed input look to be divided by 2 (the comm size). I checked the in-line partial calculations and they are correct. So it looks like the check_partials call is dividing by comm size before showing to the user? 

I think that scaling is a hold over from before the recent seed refactor? 

### Issue Type

- [x] Bug
- [ ] Enhancement  

tadkollar:
In cases where reverse mode partials are multiplied by `1.0 / comp.comm.size in` in `check_partials()`, a flag is now set. When outputting, if the flag is found, the reverse mode partials are divided by the same amount to display the original values.

### Summary

Original output with bug, via `mpirun -n 2`:
```
...
    Raw Forward Derivative (Jfor)
[[62.5 62.5 62.5 62.5]]

    Raw Reverse Derivative (Jrev)
[[31.25 31.25 31.25 31.25]]

    Raw FD Derivative (Jfd)
[[62.49998444 62.49998444 62.49998444 62.49998444]]
...
```

Fixed output:
```
...
    Raw Forward Derivative (Jfor)
[[62.5 62.5 62.5 62.5]]

    Raw Reverse Derivative (Jrev)
[[62.5 62.5 62.5 62.5]]

    Raw FD Derivative (Jfd)
[[62.49998444 62.49998444 62.49998444 62.49998444]]
...
```

### Related Issues

- Resolves #2262

### DIFF
--- a/.github/workflows/openmdao_test_workflow.yml
+++ b/.github/workflows/openmdao_test_workflow.yml
@@ -33,7 +33,7 @@ jobs:
 
           # latest versions
           - NAME: Latest
-            PY: 3
+            PY: <3.9
             NUMPY: 1
             SCIPY: 1
             PETSc: 3
@@ -43,7 +43,7 @@ jobs:
 
           # minimal install
           - NAME: Minimal
-            PY: 3
+            PY: <3.9
             NUMPY: 1
             SCIPY: 1
             PYOPTSPARSE: 'conda-forge'


swryan:
> @swryan : Look at the number of tests that are being skipped too. It is more than usual. One of the new skips is `test_pyoptsparse_driver.py:TestPyoptSparseSnoptFeature.test_snopt_fd_solution`, which only skips if it can't import SNOPT.

SNOPT is not available on a PR, so SNOPT related skips would be expected

PR #2330, opened 2021-11-02
CaseRecorder Variable Viewer

DKilkenny:
### Summary

This PR introduces a new visualizer to OpenMDAO for users to be able to visualize the recorded aspects of their model to the progression and result of their model. Users can:
- Plot variable vs variable
- Plot variable vs variable  over a given number of cases
- Plot a variable vs case iterations (with normalization, min/max, or vector lines)
- Plot a variable vs the length of the vector 
- Explore the variables by source and inputs, outputs, and residuals

### Related Issues

- Resolves #N/A

None

### New Dependencies

None


openmdao/visualization/record_viewer.py
```diff
@@ -0,0 +1,401 @@
...
+        y_len = new_data['y_vals'].shape[1]
+        new_data['color'] = self._line_color_list(new_data['x_vals'])
+        new_data['cases'] = [self.case_options[int(case)][1] for case in self.case_select.value]
+        case_len = len(new_data['cases'])
+
+        if new_data['x_vals'].shape[1] > 1:
+            if (new_data['x_vals'].shape[0], new_data['y_vals'].shape[0]) == (1, 1) and \
+               (set(new_data['x_vals'][0]), set(new_data['y_vals'][0])) == ({0.}, {0.}):
+                self.warning_box.text = """NOTE: Both X and Y values contain zeros for values,
+                                        unable to plot"""
```
swryan, line 351: looks like you will get a bunch of spaces between `values,` and `unable`... is that intentional?

swryan (commented):
just some observations when browsing the code...

DKilkenny:
Good suggestions Steve. Fixed

openmdao/visualization/record_viewer.py
```diff
@@ -0,0 +1,427 @@
+"""Define the CaseViewer class."""
```
robfalck, line 1: change filename to case_viewer.py

robfalck (changes requested):

robfalck (approved):

robfalck (changes requested):
Putting this on hold for now.

Kenneth-T-Moore:
You could add a section on the case viewer to the Command Line Tools doc: https://openmdao.org/newdocs/versions/latest/other_useful_docs/om_command.html

Kenneth-T-Moore (changes requested):

Kenneth-T-Moore:
I got this warning at one point, and from then on, the plot window glitched, and I was getting frames within frames.  I wonder if there is a better way to handle this.

```
BokehUserWarning: ColumnDataSource's columns must be of the same length. Current lengths: ('cases', 46), ('color', 46), ('x_vals', 8), ('y_vals', 8)```

Kenneth-T-Moore:
z
A couple of things confuse me on the axis selection menus. The "Case Iterations" and "Number of Points" can be selected here, but they aren't technically OpenMDAO inputs.  I'd prefer to see those separate from inputs, maybe standalone above the inputs section.  Also, I don't know why they are also listed in the "residuals" section.

Kenneth-T-Moore:
I just pulled down the latest, and I can't get Number of Cases or Case Iterations to show up at all on the "X" axis.  Also, i can't get plot a variable vs another variable anymore. Y axis is fixed at just Cases and Iterations.

Kenneth-T-Moore:
I think that Case Iteration as the x-axis value isn't working correctly. I'd expect them all to have different x values, instead of all being 0.

z1

Kenneth-T-Moore (commented):
Things are working smoothly now.

Kenneth-T-Moore (approved):

JustinSGray (approved):
Im approving this PR. I think its not quite ready for release, but the things Danny developed are complete enough to merge. 

Further development can be handled by subsequent PRs

ehariton:
Below are some interface revision suggestions.

Original Implementation:
view_cases_original

Revised Implementation:
view_cases

- Case Selection should be limited to `First`, `Last`, `All`, or `Other`, using a single item selection box. If Other is checked, the sliders along the bottom, which are normally greyed out, should become active and allow the user to slide left and right to make a selection. The original implementation allowed the user to select using the whole word string for that iteration (`rank0.pyOptSparse_SNOPT|1`), which is unnecessarily detailed.  
- X Values & Y Values should have input boxes that allow users to filter the drop down menu to only those items of interest. This is especially important because typically the users knows "I want to graph alpha in ascent". But the user doesn't necessarily remember the full path name for it. Displaying the sub-set of options that contain the search term will assist the user in quickly identifying the full path name.
- Case Iteration Plot Options should be renamed to Options for simplicity. Single item selection box should be used to select one of the three valid options. There are not that many options here so using a drop-down list seems overkill. 
- Help Icon on the bottom right corner that opens up examples. This should have at least two examples showing how to make a graph of a single-variable control and graph it over the major iterations, and a graph of a vector control graphed over all the major iteration. We may also want to include here an example of ill-posed problems that don't converge vs. well-posed problems that do converge as a reminder to the user of things that might be of interest. 
- If we want users to be able to select major iterations that are not sequential (e.g. 4, 8, 27) then we may need an additional "Advanced Options" that brings up the original interface which can allow for such selections.

robfalck:
Superseded by #2470, which combines the ideas in this PR with jupyter widgets for now.

PR #2334, opened 2021-11-05
Allow arguments to be passed to show_options_table

### LINKED ISSUE #2333: Documentation util `show_options_table` fails if the object has required options.
### Summary of Issue

Documenting the options on an object will fail if it has required options.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

The `show_options_table` method takes a string reference that is used to instantiate the referenced class, but passes no values to the instantiation.  If there are options without a default value, this will result in:

```
RuntimeError: Option 'num_segments' is required but has not been set.
```

To fix this, `show_options_table` should take a list of args to pass to the instantiation using *args and a dictionary of keyword arguments that are passed as **kwargs.

### Example

This bug can be demonstrated with dymos by calling `show_options_table('dymos.transcriptions.Rdau')`.

### Environment

N/A

robfalck:
### Summary

Documentation build would fail in some use-cases of `show_options_table`.
The method instantiates an object and then queries its options, but provides no arguments to the instantiation.
During instantiation, some objects with options (transcriptions in Dymos) will raise an exception if a required option is not provided.  This change allows args and kwargs to be passed to the instantiation of the object.

### Related Issues

- Resolves #2333

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/notebook_utils.py
+++ b/openmdao/utils/notebook_utils.py
@@ -94,27 +94,31 @@ def display_source(reference, hide_doc_string=False):
         display(get_code(reference, hide_doc_string))
 
 
-def show_options_table(reference, recording_options=False):
+def show_options_table(reference, recording_options=False, init_args=None, init_kwargs=None):
     """
     Return the options table of the given reference path.
 
     Parameters
     ----------
-    reference : str or object
-        Dot path of desired class or function or an instance.
+    reference : str
+        Dot path of desired class or function.
 
     recording_options : bool
         If True, display recording options instead of options.
+    init_args : list or None
+        Arguments to be passed to the initialization of the referenced object.
+    init_kwargs : dict or None
+        Keyword arguments to be passed to the initialization of the referenced object.
 
     Returns
     -------
     IPython.display
         Options table of the given class or function.
     """
-    if isinstance(reference, str):
-        obj = _get_object_from_reference(reference)()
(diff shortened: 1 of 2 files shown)


naylor-b:
This change will break explicit_func_comp.ipynb.  I'm ok with the change, but you'll have to fix it to get the doc build to pass.

openmdao/utils/notebook_utils.py
```diff
@@ -94,27 +94,31 @@ def display_source(reference, hide_doc_string=False):
...
     -------
     IPython.display
         Options table of the given class or function.
     """
-    if isinstance(reference, str):
-        obj = _get_object_from_reference(reference)()
-    else:
-        obj = reference
+    _init_args = init_args if init_args is not None else []
+    _init_kwargs = init_kwargs if init_kwargs is not None else {}
```
swryan, line 119: curious why you wouldn't just set the defaults to `[]` and `{}`

robfalck: @naylor-b  I didn't realize you'd made this change.  I originally was going to do it in the exact way you did here.  I'll cancel this

@swryan It's sort of been drummed into my brain that mutable default arguments are bad, just a habit of avoiding them regardless of the situation.  Moot now anyway.

robfalck:
Issue previously resolved by #2309

PR #2341, opened 2021-11-16
Added warning when user calls list_inputs before final_setup

### LINKED ISSUE #2324: Give warning about output of list_inputs if called before final_setup
### Summary of Issue

If `list_inputs` is called before final setup, any value given will just display the default value. It will not show values set by `set_val`

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Because of the way caching is done before `final_setup`, `list_inputs' will only display the default values of variables, and not show the result of any `set_val` calls. This might confuse users. 

To help with that, display a warning about this when calling `list_inputs` before final setup.

### Example

Here is some code that shows the issue

```
import openmdao.api as om

class SpeedComp(om.ExplicitComponent):
    """Simple speed computation from distance and time with unit conversations."""

    def setup(self):
        self.add_input('distance', val=1.0, units='km')
        self.add_input('time', val=1.0, units='h')
        self.add_output('speed', val=1.0, units='km/h')

        outputs['speed'] = inputs['distance'] / inputs['time']

prob = om.Problem()
prob.model.add_subsystem('c1', SpeedComp(), promotes=['*'])
prob.model.add_subsystem('c2', om.ExecComp('f=speed',speed={'units': 'm/s'}), promotes=['*'])

prob.setup()

prob.set_val('distance', 1.0, units='m')

prob.model.list_inputs(units=True, prom_name=True)   # Still lists distance as having a value of 1 km

prob.final_setup()

prob.model.list_inputs(units=True, prom_name=True)   # Still lists distance as having a value of 1 km

prob.run_model()

prob.model.list_inputs(units=True, prom_name=True)   # Now distance has a value of 1 m, which is 0.001 km
prob.model.list_outputs(units=True, prom_name=True) 
```

### Environment

All

DKilkenny:
### Summary

Added check inside of system's `list_inputs` func to see if user called it before final_setup/run_model

### Related Issues

- Resolves #2324 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -18,7 +18,8 @@ import numpy as np
 import networkx as nx
 
 from openmdao.core.configinfo import _ConfigInfo
-from openmdao.core.constants import _DEFAULT_OUT_STREAM, _UNDEFINED, INT_DTYPE, INF_BOUND
+from openmdao.core.constants import _DEFAULT_OUT_STREAM, _UNDEFINED, INT_DTYPE, INF_BOUND, \
+    _SetupStatus
 from openmdao.jacobians.assembled_jacobian import DenseJacobian, CSCJacobian
 from openmdao.recorders.recording_manager import RecordingManager
 from openmdao.vectors.vector import _full_slice
@@ -3489,7 +3490,7 @@ class System(object):
         else:
             values = val
 
-        if self._problem_meta['setup_status'] == 2:
+        if self._problem_meta['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
             issue_warning("Calling `list_inputs` before `final_setup` will only "
                           "display the default values of variables and will not show the result of "
                           "any `set_val` calls.")
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -3490,8 +3490,8 @@ class System(object):
             values = val
 
         if self._problem_meta['setup_status'] == 2:
-            issue_warning("Calling `list_inputs` before `final_setup` or `run_model` will only "
-                          "display the default values variables and will not show the result of "
+            issue_warning("Calling `list_inputs` before `final_setup` will only "
+                          "display the default values of variables and will not show the result of "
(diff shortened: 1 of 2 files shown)


openmdao/core/system.py
```diff
@@ -3489,6 +3489,11 @@ def list_inputs(self,
         else:
             values = val
 
+        if self._problem_meta['setup_status'] == 2:
+            issue_warning("Calling `list_inputs` before `final_setup` or `run_model` will only "
+                          "display the default values variables and will not show the result of "
```
naylor-b, line 3494: "default values variables"  --> "default values of variables"

openmdao/core/system.py
```diff
@@ -3489,6 +3489,11 @@ def list_inputs(self,
         else:
             values = val
 
+        if self._problem_meta['setup_status'] == 2:
+            issue_warning("Calling `list_inputs` before `final_setup` or `run_model` will only "
```
naylor-b, line 3493: I would just mention 'final_setup' and wouldn't include 'run_model'.  'final_setup' is also called by 'run_driver', 'check_partials' and 'check_totals', but we probably don't want to mention them all, so just mentioning 'final_setup' seems cleaner.

naylor-b (changes requested):

openmdao/core/system.py

naylor-b (approved):

Kenneth-T-Moore (approved):

hschilling (changes requested):
Shouldn't

`        if self._problem_meta['setup_status'] == 2:`

make use of the IntEnum defined for that ? 

And shouldn't it be a < instead a specific equality test ? So

`        if self._problem_meta['setup_status'] < _SetupStatus.POST_FINAL_SETUP:`

hschilling:
Also, for the test, `test_list_inputs_before_final_setup`, it only tests the warning message. If it only does that, you really don't need the `set_val` in there.

hschilling (approved):

PR #2350, opened 2021-11-18
Added a RuntimeError When User Connects Design Variable to Output

### LINKED ISSUE #2295: When an input is a design variable and connected to an output other than an AutoIVC, an error should be raised.
### Summary of Issue

An input of a subsystem can be made a design variable.  When provided by an AutoIVC, this is a valid situation.  But if the user connects an output to that desvar input, an error should be raised.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

See the example below.  Subsystem `initial_comp` provides a connection to `x`, but `x` is also a design variable.
This situation should raise an error.

Removing subsystem `initial_comp` will mean `x` is provided by an AutoIVC and the problem works as expected.

### Example

```
        prob = om.Problem()
        root = prob.model

        root.add_subsystem('initial_comp', om.ExecComp(['x = 10']), promotes_outputs=['x'])

        outer_group = root.add_subsystem('outer_group', om.Group(), promotes_inputs=['x'])
        inner_group = outer_group.add_subsystem('inner_group', om.Group(), promotes_inputs=['x'])

        c1 = inner_group.add_subsystem('c1', om.ExecComp(['y = x * 2.0',  'z = x ** 2']),  promotes_inputs=['x'])

        c1.add_design_var('x', lower=0, upper=5)
        c1.add_constraint('y', lower=1.5)
        c1.add_objective('z')

        prob.setup()

        prob.run_driver()
        prob.list_problem_vars()
```

### Environment

OpenMDAO 3.13.0

DKilkenny:
### Summary

Added a RuntimeError when user connects design variable to output.

Substory: Found a bug in get_source where it was returning the name of the input even if it was found. Since all inputs now have a source, a warning is issued if it cannot find the variable input in prom2abs. Also added include_self=True to two instances of system_iter because it was not including the top level system in prom2abs.

### Related Issues

- Resolves #2295

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -464,8 +464,7 @@ class Driver(object):
 
         myoutputs = set(myoutputs)
         if recording_options['record_desvars']:
-            myoutputs.update(model.get_source(n) for n in self._designvars if model.get_source(n)
-                             is not None)
+            myoutputs.update(model.get_source(n) for n in self._designvars)
         if recording_options['record_objectives'] or recording_options['record_responses']:
             myoutputs.update(self._objs)
         if recording_options['record_constraints'] or recording_options['record_responses']:
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -464,7 +464,8 @@ class Driver(object):
 
         myoutputs = set(myoutputs)
         if recording_options['record_desvars']:
-            myoutputs.update(model.get_source(n) for n in self._designvars)
+            myoutputs.update(model.get_source(n) for n in self._designvars if model.get_source(n)
+                              is not None)
         if recording_options['record_objectives'] or recording_options['record_responses']:
             myoutputs.update(self._objs)
         if recording_options['record_constraints'] or recording_options['record_responses']:
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -465,7 +465,7 @@ class Driver(object):
         myoutputs = set(myoutputs)
         if recording_options['record_desvars']:
             myoutputs.update(model.get_source(n) for n in self._designvars if model.get_source(n)
-                              is not None)
+                             is not None)
         if recording_options['record_objectives'] or recording_options['record_responses']:
             myoutputs.update(self._objs)
         if recording_options['record_constraints'] or recording_options['record_responses']:
--- a/openmdao/core/implicitcomponent.py
+++ b/openmdao/core/implicitcomponent.py
@@ -5,6 +5,7 @@ import numpy as np
 from openmdao.core.component import Component
 from openmdao.recorders.recording_iteration_stack import Recording
 from openmdao.utils.class_util import overrides_method
+from openmdao.utils.general_utils import make_set
 from openmdao.utils.om_warnings import warn_deprecation
 
 _inst_functs = ['apply_linear']
@@ -30,6 +31,11 @@ class ImplicitComponent(Component):
         Store some bound methods so we can detect runtime overrides.
         """
         super().__init__(**kwargs)
+
+        if 'tags' not in kwargs:
+            kwargs['tags'] = {'allow_desvar'}
+        else:
+            kwargs['tags'] = make_set(kwargs['tags'], name='tags') | {'allow_desvar'}
 
         self._inst_functs = {name: getattr(self, name, None) for name in _inst_functs}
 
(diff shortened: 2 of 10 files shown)


openmdao/core/problem.py
```diff
@@ -940,6 +940,14 @@ def setup(self, check=False, logger=None, mode='auto', force_alloc_complex=False
         # set static mode back to True in all systems in this Problem
         self._metadata['static_mode'] = True
 
+        des_vars = self.model.get_design_vars()
+        prom2abs = self._metadata['prom2abs']
+        if des_vars:
+            for key in des_vars.keys():
+                if key in prom2abs['output'] and key in prom2abs['input']:
```
naylor-b, line 947: Won't this flag even legitimate cases, like when the output is an IndepVarComp?

openmdao/core/problem.py
```diff
@@ -940,6 +940,14 @@ def setup(self, check=False, logger=None, mode='auto', force_alloc_complex=False
         # set static mode back to True in all systems in this Problem
         self._metadata['static_mode'] = True
 
+        des_vars = self.model.get_design_vars()
```
naylor-b, line 943: Rather than doing this here and having a whole extra traversal down the system tree, it might be better to do this down in the `System._get_design_vars` method.

naylor-b (changes requested):

openmdao/core/problem.py
naylor-b: Never mind.  Apparently it doesn't.

openmdao/core/problem.py
DKilkenny: I needed to do it in the setup function because the system level `_get_design_vars` method loops through each subsystem and I need the final values all at once.

robfalck (changes requested):

openmdao/core/system.py
```diff
@@ -3055,6 +3056,14 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
                             else:
                                 out[name] = meta
 
+        if out:
+            for conn_in, conn_out in conns.items():
```
naylor-b, line 3060: Couldn't this check just be done once at the top level? 'conns' in this case is the top level connections dict, which can be quite large, and you're iterating over it at every level in the system tree.

naylor-b (changes requested):

openmdao/core/system.py
```diff
@@ -3056,13 +3056,12 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
...
                                 out[name] = meta
 
-        if out:
-            for conn_in, conn_out in conns.items():
-                conn_out_attr = conn_out in abs2meta_out
-                if conn_out in out and conn_out_attr:
-                    if "allow_desvar" not in abs2meta_out[conn_out]['tags']:
-                        raise RuntimeError(f"Cannot connect the design variable '{conn_out}' to "
-                                           f"'{conn_in}' as an input.")
+        if out and 'model' in self.msginfo:
```
naylor-b, line 3059: instead of checking if `'model' in self.msginfo`, why not just check if `self is model`?

DKilkenny: Changed

openmdao/core/driver.py
```diff
@@ -464,7 +464,8 @@ def _get_vars_to_record(self, recording_options):
 
         myoutputs = set(myoutputs)
         if recording_options['record_desvars']:
-            myoutputs.update(model.get_source(n) for n in self._designvars)
+            myoutputs.update(model.get_source(n) for n in self._designvars if model.get_source(n)
```
naylor-b, line 467: It's not a huge deal, but it would be nice if we didn't call model.get_source(n) twice in the same loop.

openmdao/core/implicitcomponent.py
```diff
@@ -31,6 +32,11 @@ def __init__(self, **kwargs):
         """
         super().__init__(**kwargs)
 
+        if 'tags' not in kwargs:
```
naylor-b, line 35: This doesn't seem like it would actually do anything.  'tags' isn't passed to component __init__ methods except in the case of IndepVarComp which allows it because the user can add an output as part of the __init__ args.  I think you can get rid of this whole block.  You're already taking care of the 'allow_desvar' tag in the add_output method.

openmdao/core/implicitcomponent.py
```diff
@@ -289,6 +295,37 @@ def _linearize(self, jac=None, sub_do_ln=True):
...
+            metadata for the input name that matches that key.
+
+        Returns
+        -------
+        dict
+            Metadata for added variable.
+        """
+        metadata = super().add_output(name, val, **kwargs)
+
+        if "tags" in metadata:
```
naylor-b, line 319: You don't need to do any of this checking here.  The base class calls `make_set` on tags, so you're guaranteed that tags exists and is a set by the time it gets here.  You can just do `metadata['tags'].add('openmdao:allow_desvar')` and that's it.

DKilkenny: Fixed

naylor-b (approved):

robfalck (approved):

PR #2355, opened 2021-11-23
Provide a better error message when setting an invalid value to an auto_ivc output

### LINKED ISSUE #2336: Provide a better error message when setting an invalid value to an auto_ivc output.
### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When the user sets the value of an auto_ivc output via the `Problem.set_val` method, they do so using a promoted name.
If the value is invalid for some reason (such as having the wrong shape), they receive an error message giving the absolute path of the auto_ivc output.  This error message should contain the promoted name used in the `set_val` call.

### Example

In the following test, the given values for `b` and `c` in the `set_val` calls have the wrong shape.

```
    def test_set_val_auto_ivc(self):

        p = om.Problem()
        p.model.add_subsystem('exec',
                              om.ExecComp('z = a + b**2 + c**3',
                                          z={'shape': (100,)},
                                          a={'shape': (100,)},
                                          b={'shape': (100,)},
                                          c={'shape': (100,)}))

        p.setup()

        p.set_val('exec.a', np.linspace(5, 10, 100))
        p.set_val('exec.b', np.linspace(10, 20, 15))
        p.set_val('exec.c', np.linspace(1, 3, 23))

        p.final_setup()
```

The resulting stack error message ends with

```
  File "...openmdao/core/problem.py", line 1013, in final_setup
    self._set_initial_conditions()
  File ".../openmdao/core/problem.py", line 612, in _set_initial_conditions
    self.set_val(name, value)
  File ".../openmdao/core/problem.py", line 584, in set_val
    model._outputs.set_var(src, value, indices)
  File ".../openmdao/vectors/vector.py", line 583, in set_var
    raise ValueError(f"{self._system().msginfo}: Failed to set value of "
ValueError: <model> <class Group>: Failed to set value of '_auto_ivc.v1': could not broadcast input array from shape (15) into shape (100).
```

The ValueError message should be

```
ValueError: <model> <class Group>: Failed to set value of 'exec.b': could not broadcast input array from shape (15) into shape (100).
```

### Environment

N/A

swryan:
### Summary

Add an optional `var_name` argument the to `Vector.set_val()` method to override the variable name used in error messages.  When setting an Auto-IVC output value via `Problem.set_val()`, error messages will refer to the matching promoted input name, which is known to the user. 

### Related Issues

- Resolves #2336 

None

### New Dependencies

None

### DIFF
--- a/openmdao/vectors/vector.py
+++ b/openmdao/vectors/vector.py
@@ -580,17 +580,16 @@ class Vector(object):
             If True, set into flattened variable.
         var_name : str or None
             If specified, the variable name to use when reporting errors. This is useful
-            when setting an IVC value that the user only knows by a connected input name.
+            when setting an AutoIVC value that the user only knows by a connected input name.
         """
-        if not var_name:
-            var_name = name
-
         abs_name = self._name2abs_name(name)
         if abs_name is None:
-            raise KeyError(f"{self._system().msginfo}: Variable name '{var_name}' not found.")
+            raise KeyError(f"{self._system().msginfo}: Variable name "
+                           f"'{var_name if var_name else name}' not found.")
 
         if self.read_only:
-            raise ValueError(f"{self._system().msginfo}: Attempt to set value of '{var_name}' in "
+            raise ValueError(f"{self._system().msginfo}: Attempt to set value of "
+                             f"'{var_name if var_name else name}' in "
                              f"{self._kind} vector when it is read only.")
 
         if idxs is _full_slice:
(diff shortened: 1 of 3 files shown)


openmdao/vectors/vector.py
```diff
@@ -578,13 +578,19 @@ def set_var(self, name, val, idxs=_full_slice, flat=False):
             The locations where the data array should be updated.
         flat : bool
             If True, set into flattened variable.
+        var_name : str or None
+            If specified, the variable name to use when reporting errors. This is useful
+            when setting an IVC value that the user only knows by a connected input name.
         """
+        if not var_name:
```
naylor-b, line 585: Since `set_var` gets called a lot, we could get rid of this block and just replace `{name}` in the error messages with `{var_name if var_name else name}`.  That way there's no additional overhead introduced unless there's an actual error.

naylor-b (approved):

Kenneth-T-Moore (approved):

PR #2358, opened 2021-11-24
Fix bug where subjacs declared with scipy.sparse cause check_partials to fail.

### LINKED ISSUE #2354: Check partials stopped working with sparse subjacobian when upgrading from 3.8.0 to >3.9.0
### Summary of Issue

Check partials stopped working with sparse subjacobian when upgrading from 3.8.0 to >3.9.0

### Issue Type

- [x] Bug
- [ ] Enhancement

### Description

There is an assignment problem when using FD to compute derivatives with a component that has a sparse Jacobian. I speculate it is due to a missing conversion from a diagonal sparse format to the COO sparse format, but I tried adding .tocoo() to all sparse matrices I created and it did not help.

### Example

This  [standalone example](https://gist.github.com/bbahiam/606e66d367ed0433f65a2689a4229f02) runs on OpenMDAO 3.8.0 but not on 3.9.0 (or 3.14.0). In that case it fails with the following output:
```
$ python test_cross_section_properties.py 
./home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/jacobians/jacobian.py:315: DeprecationWarning: Calling np.sum(generator) is deprecated, and in the future will give a different result. Use np.sum(np.fromiter(generator)) or the python sum builtin instead.
  ncols = np.sum(end - start for _, start, end, _, _ in col_var_info.values())
EE
======================================================================
ERROR: test_partials_against_cs (__main__.TestCrossSectionProperties)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_cross_section_properties.py", line 249, in test_partials_against_cs
    assert_partials_match_fd(
  File "test_cross_section_properties.py", line 187, in assert_partials_match_fd
    data = prob.check_partials(method=method, out_stream=None)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/core/problem.py", line 1386, in check_partials
    approximation.compute_approximations(comp, jac=approx_jac)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/approximation_schemes/approximation_scheme.py", line 512, in compute_approximations
    jac.set_col(system, ic, col)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/jacobians/dictionary_jacobian.py", line 265, in set_col
    subjac['value'][:, loc_idx] = column[start:end]
TypeError: 'dia_matrix' object does not support item assignment

======================================================================
ERROR: test_partials_against_fd (__main__.TestCrossSectionProperties)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_cross_section_properties.py", line 244, in test_partials_against_fd
    assert_partials_match_fd(
  File "test_cross_section_properties.py", line 187, in assert_partials_match_fd
    data = prob.check_partials(method=method, out_stream=None)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/core/problem.py", line 1386, in check_partials
    approximation.compute_approximations(comp, jac=approx_jac)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/approximation_schemes/approximation_scheme.py", line 512, in compute_approximations
    jac.set_col(system, ic, col)
  File "/home/bbahia/tmp/venv/lib/python3.8/site-packages/openmdao/jacobians/dictionary_jacobian.py", line 265, in set_col
    subjac['value'][:, loc_idx] = column[start:end]
TypeError: 'dia_matrix' object does not support item assignment

----------------------------------------------------------------------
Ran 3 tests in 0.023s

FAILED (errors=2)
```

### Environment

Operating System: Ubuntu 20.04.3 LTS Linux 5.4.0-89-generic

Python environment: Python 3.8.10

Packages: 
```
$ pip freeze
certifi==2021.10.8
charset-normalizer==2.0.7
idna==3.3
networkx==2.6.3
numpy==1.21.4
openmdao==3.9.0
pyDOE2==1.3.0
pyparsing==3.0.6
requests==2.26.0
scipy==1.7.2
urllib3==1.26.7
```

Kenneth-T-Moore:
### Summary

Fix bug where subjacs declared with scipy.sparse cause check_partials to fail.

This bug arose when we moved over to using a `_CheckingJacobian` to perform the fd/cs checking. Subjacs are copied from the model when it is created, so we can just convert them to a regular dense array. 

### Related Issues

- Resolves #2354

None

### New Dependencies

None

### DIFF
--- a/openmdao/jacobians/dictionary_jacobian.py
+++ b/openmdao/jacobians/dictionary_jacobian.py
@@ -168,10 +168,13 @@ class _CheckingJacobian(DictionaryJacobian):
         super().__init__(system)
         self._subjacs_info = self._subjacs_info.copy()
 
-        # Convert any scipy.sparse subjacs to dense.
+        # Convert any scipy.sparse subjacs to OpenMDAO's interal COO specification.
         for key, subjac in self._subjacs_info.items():
             if sp.issparse(subjac['val']):
-                self._subjacs_info[key]['val'] = subjac['val'].toarray()
+                coo_val = subjac['val'].tocoo()
+                self._subjacs_info[key]['rows'] = coo_val.row
+                self._subjacs_info[key]['cols'] = coo_val.col
+                self._subjacs_info[key]['val'] = coo_val.data
 
         self._errors = []
 
--- a/openmdao/jacobians/dictionary_jacobian.py
+++ b/openmdao/jacobians/dictionary_jacobian.py
@@ -1,5 +1,6 @@
 """Define the DictionaryJacobian class."""
 import numpy as np
+import scipy.sparse as sp
 
 from openmdao.jacobians.jacobian import Jacobian
 from openmdao.core.constants import INT_DTYPE
@@ -166,6 +167,12 @@ class _CheckingJacobian(DictionaryJacobian):
     def __init__(self, system):
         super().__init__(system)
         self._subjacs_info = self._subjacs_info.copy()
+
+        # Convert any scipy.sparse subjacs to dense.
+        for key, subjac in self._subjacs_info.items():
+            if sp.issparse(subjac['val']):
+                self._subjacs_info[key]['val'] = subjac['val'].toarray()
+
         self._errors = []
 
     def __iter__(self):
--- a/openmdao/jacobians/tests/test_jacobian.py
+++ b/openmdao/jacobians/tests/test_jacobian.py
@@ -11,7 +11,7 @@ from openmdao.api import IndepVarComp, Group, Problem, \
                          ExplicitComponent, ImplicitComponent, ExecComp, \
                          NewtonSolver, ScipyKrylov, \
                          LinearBlockGS, DirectSolver
-from openmdao.utils.assert_utils import assert_near_equal
+from openmdao.utils.assert_utils import assert_near_equal, assert_check_partials
 from openmdao.utils.array_utils import rand_sparsity
 from openmdao.test_suite.components.paraboloid import Paraboloid
 from openmdao.api import ScipyOptimizeDriver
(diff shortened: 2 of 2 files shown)


JustinSGray (approved):

openmdao/jacobians/dictionary_jacobian.py
```diff
@@ -166,6 +167,12 @@ class _CheckingJacobian(DictionaryJacobian):
     def __init__(self, system):
         super().__init__(system)
         self._subjacs_info = self._subjacs_info.copy()
+
+        # Convert any scipy.sparse subjacs to dense.
+        for key, subjac in self._subjacs_info.items():
+            if sp.issparse(subjac['val']):
+                self._subjacs_info[key]['val'] = subjac['val'].toarray()
```
naylor-b, line 174: This will definitely work, but in most cases if a user is using sparse matrices it's because the dense version would be really big, so I'm not sure we want to do this. What if we just converted subjac['val'] to coo format and then set 'rows' and 'cols' metadata based on the coomat.row and coomat.col and set the subjac['val'] based on coomat.data?  That would use a lot less memory and we'd still be checking that the sparsity was correct.

naylor-b (changes requested):

openmdao/jacobians/dictionary_jacobian.py
Kenneth-T-Moore: yeah, that's probably wiser. I've made the change.

Kenneth-T-Moore: Yeah, that's probably wiser.  Made the change.

naylor-b (approved):

PR #2363, opened 2021-11-29
Removed Deprecation Warning For Serial to Distributed IO Connections  

### LINKED ISSUE #2346: Connections between serial inputs and distributed outputs
### Summary of Issue

I get a deprecation warning every time I try to make a connection between a serial output and distributed input

Why is this feature being deprecated and is there another method for making these connections that will be supported in the future?

### Issue Type

- [ X] Bug

### Description

I have a model that needs to connect a serial IVC output to a distributed input in another component. I use the `src_indices` argument to split the serial across processors when defining the connection between these components, but get the following deprecation warning when running the model:

`OMDeprecationWarning:Connection between serial output 'serial_ivc.global_a' and distributed input 'parallel_sum.local_a' is deprecated and will become an error in a future release.`

### Example

```
import numpy as np
import openmdao.api as om

"""
This example creates a vector, with each entry split across one processor
The entries are then summed together using an mpi allreduce
This example generates a deprecation warning
about connections between serial and distributed inputs not being supported
NOTE: Must be run in parallel to generate warning
"""

class ParallelSum(om.ExplicitComponent):
    """Simple parallel summation."""

    def setup(self):
        self.add_input('local_a', shape=1, distributed=True, desc='local slice of parallel a vector.')
        self.add_output('sum')

        outputs['sum'] = self.comm.allreduce(inputs['local_a'])

class SerialIVC(om.IndepVarComp):
    """Simple serial vector independent variable component"""
    def setup(self):
        size = self.comm.size
        self.add_output('global_a', val=np.arange(size), shape=size, desc='global a vector.')

class SimpleGroup(om.Group):
    def setup(self):
        indx = self.comm.rank
        self.add_subsystem('serial_ivc', SerialIVC())
        self.add_subsystem('parallel_sum', ParallelSum())
        # Split serial vector across each processor
        self.connect('serial_ivc.global_a', 'parallel_sum.local_a', src_indices=[indx])

p = om.Problem()
p.model = SimpleGroup()

p.setup()
p.run_model()

print(p['parallel_sum.sum'])
```

### Environment

OS X 10.14.6

Python environment: Anaconda Python 3.9

Packages: OpenMDAO, mpi4py

DKilkenny:
### Summary

Removed the deprecation warning as we have decided to keep the functionality.

### Related Issues

- Resolves #2346

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -876,8 +876,7 @@ class Group(System):
                 elif in_dist and not out_dist:
                     issue_warning(f"Connecting a serial output '{abs_out}' and distributed input "
                                   f"'{abs_in}' is not common. While legal in OM, verify "
-                                  f"connections are as expected.",
-                                  category=DistributedComponentWarning)
+                                  "connections are as expected.", category=DistributedComponentWarning)
 
     def _get_group_input_meta(self, prom_in, meta_name):
         if prom_in in self._group_inputs:
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -31,7 +31,7 @@ from openmdao.utils.mpi import MPI, check_mpi_exceptions, multi_proc_exception_c
 import openmdao.utils.coloring as coloring_mod
 from openmdao.utils.indexer import indexer, Indexer
 from openmdao.utils.om_warnings import issue_warning, UnitsWarning, UnusedOptionWarning, \
-    SetupWarning, PromotionWarning, MPIWarning, DistributedComponentWarning
+    SetupWarning, PromotionWarning, MPIWarning
 from openmdao.core.constants import _SetupStatus
 from openmdao.utils.om_warnings import warn_deprecation
 
@@ -873,11 +873,6 @@ class Group(System):
                         raise RuntimeError(f"{self.msginfo}: Can't connect distributed output "
                                            f"'{abs_out}' to serial input '{abs_in}' without "
                                            "specifying src_indices.")
-                elif in_dist and not out_dist:
-                    issue_warning(f"Connecting a serial output '{abs_out}' and distributed input "
-                                  f"'{abs_in}' is not common. While legal in OM, verify "
-                                  "connections are as expected.",
-                                  category=DistributedComponentWarning)
 
     def _get_group_input_meta(self, prom_in, meta_name):
         if prom_in in self._group_inputs:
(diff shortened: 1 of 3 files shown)


naylor-b (changes requested):

openmdao/core/tests/test_group.py
```diff
@@ -1412,6 +1413,39 @@ def test_double_set_input_defaults(self):
 class TestGroupMPISlice(unittest.TestCase):
     N_PROCS = 2
 
+    def test_serial_input_to_distrib_output(self):
```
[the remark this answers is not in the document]
DKilkenny: fixed

openmdao/core/group.py
```diff
@@ -874,9 +874,10 @@ def _top_level_post_sizes(self):
                                            f"'{abs_out}' to serial input '{abs_in}' without "
                                            "specifying src_indices.")
                 elif in_dist and not out_dist:
-                    warn_deprecation(f"Connection between serial output '{abs_out}' and distributed"
-                                     f" input '{abs_in}' is deprecated and will become an error "
-                                     "in a future release.")
+                    issue_warning(f"Connecting a serial output '{abs_out}' and distributed input "
```
naylor-b, line 877: What exactly are we asking the user to do here?  If I were a user and saw this message I would be worried that something was wrong but would have no idea what the 'expected' connections would be.  I wonder if we should even warn the user at all if we're not going to provide them with any useful information.

robfalck (approved):

naylor-b (approved):

openmdao/core/tests/test_group.py
```diff
@@ -17,7 +17,8 @@
 from openmdao.utils.mpi import MPI
 from openmdao.utils.assert_utils import assert_near_equal, assert_warning, assert_no_warning
 from openmdao.utils.logger_utils import TestLogger
-from openmdao.utils.om_warnings import PromotionWarning, OMDeprecationWarning
+from openmdao.utils.om_warnings import PromotionWarning, OMDeprecationWarning, \
+    DistributedComponentWarning
```
swryan, line 21: new import no longer needed?

swryan (changes requested):
I think we should change to description of the PR to match what was ultimately done (e.g. for the release notes) and also add a comment to the issue explaining the outcome to the original poster

openmdao/core/tests/test_group.py
DKilkenny: Removed

swryan (approved):

PR #2368, opened 2021-12-06
I2338 if no explicit name given to a Problem, assign one automatically

### LINKED ISSUE #2338: Set Problem name automatically if user doesn't set it
### Summary of Issue

Some complex OpenMDAO scripts may create multiple Problem instances, and users are currently able to specify a name for each Problem instance, but it would be nice if we automatically provided a locally unique name for each Problem instance if the user didn't provide one.  This would allow OpenMDAO command line tools like the connection viewer and the n2 to target a specific Problem instance and the user wouldn't have to worry about it.  Problem instances would be named 'problem1', problem2', etc. based on their order of instantiation.

### Issue Type

- [ ] Bug
- [x] Enhancement

hschilling:
### Summary

Currently, if the user doesn't provide an explicit name for a Problem, the name is the empty string. If a model has sub problems, and the user wants to use one of the command line tools that take problem name as an argument, there is no way to indicate which Problem to apply the command to.

This PR assigns default values to Problems that are explicitly given a name.

### Related Issues

- Resolves #2338 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -56,6 +56,10 @@ from openmdao.utils.name_maps import rel_key2abs_key, rel_name2abs_name
 
 _contains_all = ContainsAll()
 
+# Used for naming Problems when no explicit name is given
+# Also handles sub problems
+_problem_names = []
+_problem_counter = 0
 
 CITATION = """@article{openmdao_2019,
     Author={Justin S. Gray and John T. Hwang and Joaquim R. R. A.
@@ -123,7 +127,9 @@ class Problem(object):
     _logger : object or None
         Object for logging config checks if _check is True.
     _name : str
-        Problem name.
+        Problem name. If no name given, a default name of the form 'problemN', where N is an
+        integer, will be given to the problem so it can be referenced in command line tools
+        that have an optional problem name argument
     _system_options_recorded : bool
         A flag to indicate whether the system options for all the systems have been recorded
     _metadata : dict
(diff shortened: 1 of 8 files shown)


openmdao/core/problem.py
```diff
@@ -138,8 +144,30 @@ def __init__(self, model=None, driver=None, comm=None, name=None, **options):
         """
         Initialize attributes.
         """
+        global _problem_counter
```
Kenneth-T-Moore, line 147: Do you also need to global `_problem_names ` since you are appending to it?

openmdao/core/problem.py
```diff
@@ -138,8 +144,30 @@ def __init__(self, model=None, driver=None, comm=None, name=None, **options):
...
+
         self.cite = CITATION
-        self._name = name
+
+        # Code to give non-empty names to Problems so that they can be
+        # referenced from command line tools (e.g. check) that accept a Problem argument
+        _problem_counter += 1
```
Kenneth-T-Moore, line 153: Do we actually need `_problem_counter`, if the number of problems is always the length of `_problem_names`?

openmdao/core/problem.py
```diff
@@ -138,8 +144,30 @@ def __init__(self, model=None, driver=None, comm=None, name=None, **options):
...
+        if name:  # if name hasn't been used yet, use it. Otherwise, error
+            if name not in _problem_names:
+                self._name = name
+            else:
+                raise ValueError(f"The problem name '{name}' already exists")
+        else:  # No name given: look for a name, of the form, 'problemN', that hasn't been used yet
+            _name = f"problem{_problem_counter}"
```
Kenneth-T-Moore, line 160: I think I understand the logic here now. If we are the second problem, but "problem2" is taken, we just start with "problem2.1", and keep working until we find one that isn't taken.

openmdao/core/tests/test_parallel_groups.py
```diff
@@ -308,13 +308,11 @@ def test_setup_messages_bad_vec_type(self):
 
         # check that error is thrown if not using PETScVector
         if MPI:
-            msg = ("Problem: The `distributed_vector_class` argument must be a distributed vector "
+            msg = ("Problem .*: The `distributed_vector_class` argument must be a distributed vector "
```
Kenneth-T-Moore, line 311: I am not sure where this ".*" comes from for the problem name, but it doesn't look so great. Seems like it should be "Problem Problem1", or maybe just "Problem".  There must be a bug in the print here that wasn't uncovered until you made the change.

Kenneth-T-Moore (changes requested):

openmdao/core/problem.py
hschilling: Good point. Fixed.

openmdao/core/problem.py
hschilling: Also a great point. Will modify code to do that

openmdao/core/tests/test_parallel_groups.py
Kenneth-T-Moore: My bad, that's regex.

Kenneth-T-Moore (approved):

naylor-b (approved):

PR #2370, opened 2021-12-06
fwd/rev coloring support for function comps (using jax)

### LINKED ISSUE #2256: Add partial derivative coloring support for JAX'd function components.
### Summary of Issue

Support simple and advanced colorings for JAX'd function comps 

### Issue Type

- [ ] Bug
- [x] Enhancement : POEM-TBD

### Description

Current partial derivative coloring is fwd only, because complex-step is fwd only. 

JAX'd components could have fwd or rev mode calls, depending on if the number of inputs or outputs are larger. However, from total derivative coloring we know that its possible that the direction changes when you take into account either unidirectional or bi-directional coloring.

A POEM needs to be written for the API extensions to cover these cases. That should happen before impelementation

Details: 

- If the user has specificed either ad-fwd or ad-rev mode, then obviously we can only do uni-directional coloring
- If the user has specified both ad-auto, then we could use bi-directional coloring instead

- Could you use something similar to JAX's size-only data types to get the sparsity? Something that tells you which inputs affect which outputs? This would avoid the incidental non-zero problem.

naylor-b:
### Summary

Forward and reverse coloring works with ExplicitFuncComp and ImplicitFuncComp.

### Related Issues

- Resolves #2256

None

### New Dependencies

None


openmdao/docs/openmdao_book/features/building_blocks/components/explicit_func_comp.ipynb
```diff
@@ -26,10 +26,9 @@
     "\n",
```
robfalck, line 26: <p>I think it would be worthwhile to include a small example of <code>use_jit</code>  here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMDAO/OpenMDAO/pull/2370/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

naylor-b: added

Kenneth-T-Moore (changes requested):

openmdao/components/func_comp_common.py
```diff
@@ -60,3 +76,162 @@ def _add_options(comp):
...
+
+    Returns
+    -------
+    function
+        A function that returns rows of the jacobian grouped by output variable, e.g., if there were
+        2 output variables of size 3 and 4, the function would return a list with two entries. The
+        first entry would contain the first 3 rows of J and the second would contain the next
+        4 rows of J.
+    """
+    f = linear_util.wrap_init(fun)
```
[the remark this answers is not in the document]
naylor-b: There wouldn't be much difference either way but I think the code looks cleaner the way it is.

Kenneth-T-Moore (approved):

JustinSGray (approved):

PR #2408, opened 2022-01-12
implement CaseViewerWidget

### LINKED ISSUE #2259: Jupyter widget to do plotting of data in case reader
### Summary of Issue

Widget to query for case or cases from a reader, then plot the results. 
### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

This widget should work without requiring an openmdao model, only a case reader object. 

User need to be able to pick a single case or a select a series of cases. 

For a single case, they will then select one or more variable to plot from that case. In this situation you're not looking to plot "over time" (i.e. over an iteration history) but instead just data from the case itself. User will select values for both the x and y axes. 

If there are a series of cases, then its assumed you're looking to plot "over time" (i.e. over the iteration history). So the x-axis will always be case #, and the y axis will be one or more variables from the case-set. 

Can draw heavily from ovis for user interface starting point.

swryan:
### Summary

Implement Jupyter widget to do plotting of data in case reader

### Related Issues

- Resolves #2259

None

### New Dependencies

None

### DIFF
--- a/openmdao/api.py
+++ b/openmdao/api.py
@@ -88,6 +88,7 @@ from openmdao.utils.options_dictionary import OptionsDictionary
 # Recorders
 from openmdao.recorders.sqlite_recorder import SqliteRecorder
 from openmdao.recorders.case_reader import CaseReader
+from openmdao.recorders.case_viewer_widget import CaseViewerWidget
 
 # Visualizations
 from openmdao.visualization.n2_viewer.n2_viewer import n2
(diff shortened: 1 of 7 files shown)


robfalck (approved):

robfalck:
I recommend we close this.  While the widget-ness is good, it's lacking some features that are missing from the other PR:
- Assumes cases are number and selects with a slider, while Dymos saves out single problem-level cases by default.  It does pick up the case, seemingly, because its the only one present though.
- I currently assumes the x-axis is scalar and will not allow a vector on the x-axis.  For dymos, for instance, we might want to see the evolution of a trajectory over iterations (altitude vs time with multiple lines representing various iterations of a driver recorder).  This implementation does not  support that.
- 
Screen Shot 2022-01-18 at 11 49 51 AM

swryan:
closing PR per @robfalck comment, issue #2259 should probably also be closed

PR #2422, opened 2022-01-24
Represent filtered variables as collapsed instead of hidden

### LINKED ISSUE #2343: Update variable selection in N2 so deselected variables are collapsed instead of hidden
### Summary of Issue

When the variable selection dialog is used to choose which variables to display, currently the deselected ones are not represented on the diagram at all when the parent component is expanded. Update the N2 diagram to show the deselected variables as collapsed, while displaying the selected variables as individuals. This was not implemented in the original variable selection dialog PR because it will likely affect a large amount of code in a different area.

### Issue Type

- [ ] Bug
- [X] Enhancement  
      POEM # N/A <N/A or POEM number at https://github.com/OpenMDAO/POEMs>

tadkollar:
### Summary

Filtered variables were initially completely hidden from the diagram. This update shows filtered variables as one of two nodes per component, _Inputs_ or _Outputs_, along with their aggregated connections.

To achieve this, two filter quasi-nodes (`N2FilterNode`) are added as children to each component node. They are hidden unless the component is visible and has had its variables filtered through the variable selection dialog. When a variable is filtered, its `draw.filtered` property is set true, it's added as a child to the appropriate filter node, and its `draw.filterParent` is set to point to that filter node. When unfiltered, the `draw.filtered` property is false, and `draw.filterParent` is `null`. Since these properties are new and filtering was previously based on simply hiding the variable, significant changes were made to `N2Layout`, `N2Matrix`, and `N2Diagram`.

The diagram save/load state functions needed to be updated to accommodate the changes, as well as the GUI tests. Some cleanups were also performed, including relocating each node's display properties to its `draw` object property (e.g. `varIsHidden` became `draw.hidden`, `isMinimized` became `draw.minimized`). This helps to separate diagram-related info from model-based info.

### Related Issues

- Resolves #2343 

None

### New Dependencies

None


swryan:
merge out to master when you get a chance, to get the fix to the docstring test

robfalck:
Recommend you change the label on the `inputs` and `outputs` nodes to `filtered inputs` and `filtered outputs`?

swryan (approved):

robfalck (approved):

PR #2443, opened 2022-02-11
Fixed a bug when a flat slice is specified as the src_indices for a non-flat source, and added usage of slicer to connection documentation.

### LINKED ISSUE #2437: Add examples of using om.slicer in connections to our connection documentation.
### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

[Connection documentation in core features](https://openmdao.org/newdocs/versions/latest/features/core_features/working_with_groups/connect.html) should discuss the use of `om.slicer` when issuing connections.

A user on stack overflow pointed out the deficiency here:  https://stackoverflow.com/questions/70993067/src-indices-best-practices

### Example

[This response to the stackoverflow question](https://stackoverflow.com/a/70998783/754536) points out two example use-cases of om.slicer when issuing connections.

### Environment

OpenMDAO 3.16.1-dev

### LINKED ISSUE #2446: `om.slicer` src_indices is broken when the number of dimensions does not match those of the variable to which it is applied.
### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

Using om.slicer on src_indices fails and results in incorrect values when the slicer is created with a single dimension but the variable to which it is applied has multiple dimensions.  See examples below.

### Example to test the slicer itself
Comparison of `om.slicer` to numpy slicing:

```
import numpy as np
import openmdao.api as om

M = np.arange(25, dtype=int).reshape((5,5))
print('numpy first row:', M[:1])
print('openmdao first row:', M[om.slicer[:1]])
```

```
numpy: [[0 1 2 3 4]]
openmdao: [[0 1 2 3 4]]
```

So the slicer itself seems to be working as expected.

### Example to test the slicer used as src_indices in a connection

```
        p = om.Problem()

        ivc = p.model.add_subsystem('ivc', om.IndepVarComp())
        M_np = np.arange(9)[::-1].reshape((3, 3))
        ivc.add_output('M', val=M_np)

        exec = p.model.add_subsystem('exec', om.ExecComp())

        exec.add_expr('A = B', A={'shape': (3,)}, B={'shape': (3,)})

        p.model.connect('ivc.M', 'exec.B', src_indices=om.slicer[:1])

        p.setup()

        p.run_model()

        print('M')
        print(p.get_val('ivc.M'))

        print()
        print('p.get_val("exec.A")')
        print(p.get_val('exec.A'))
        print('p.get_val("exec.B")')
        print(p.get_val('exec.B'))
        print('p.model.exec.get_val("B")')
        print(p.model.exec.get_val('B'))
        print()
        print('numpy M[:1]')
        print(p.get_val('ivc.M')[:1])

        p.model.list_inputs(print_arrays=True)
```

output:
```
M
[[8. 7. 6.]
 [5. 4. 3.]
 [2. 1. 0.]]

p.get_val("exec.A")
[8. 8. 8.]
p.get_val("exec.B")
[8. 7. 6.]
p.model.exec.get_val("B")
[8. 7. 6.]

numpy M[:1]
[[8. 7. 6.]]
1 Input(s) in 'model'

varname  val                 
-------  --------------------
exec
  B      |13.85640646|       
         val:
         array([8., 8., 8.])
```

In this case, the call to `prob.get_val('exec.B')` shows the expected result (get_val has the indices applied correctly).
However, printing the value of `A` shows that, inside compute, the values are wrong (a print statement inside the compute method will prove this.

The `list_inputs` call will show that `B` has incorrect values.  So
1. The actual model input is wrong `[8, 8, 8]` instead of `[8, 7, 6]`
2. `list_inputs` prints the values as received by `exec.compute`
3. `get_val` (on both the problem and on exec) shows the expected value of [8, 7, 6], but this is _not_ the input value sent to `exec.compute`.
4. `list_inputs` shows the values as received in `inputs` by `exec.compute`.

### Example to test the slicer used as src_indices in promotion

If we do the exact same thing, but use promotion instead of explicit connections, we get the same result:

```
        p = om.Problem()

        ivc = p.model.add_subsystem('ivc', om.IndepVarComp())
        M_np = np.arange(9)[::-1].reshape((3, 3))
        ivc.add_output('M', val=M_np)

        exec = p.model.add_subsystem('exec', om.ExecComp())

        exec.add_expr('A = B', A={'shape': (3,)}, B={'shape': (3,)})

        p.model.promotes('ivc', outputs=['M'])
        p.model.promotes('exec', inputs=[('B', 'M')], src_indices=om.slicer[:1], src_shape=(3, 3))

        p.setup()

        p.run_model()

        print('M')
        print(p.get_val('ivc.M'))

        print()
        print('p.get_val("exec.A")')
        print(p.get_val('exec.A'))
        print('p.get_val("exec.B")')
        print(p.get_val('exec.B'))
        print('p.model.exec.get_val("B")')
        print(p.model.exec.get_val('B'))
        print()
        print('numpy M[:1]')
        print(p.get_val('ivc.M')[:1])

        p.model.list_inputs(print_arrays=True)
```

### Environment

OpenMDAO 3.16.1-dev

robfalck:
### Summary

The core features documentation on connections did not demonstrate the use of `om.slicer` for `src_indices`.
This adds two examples of its use.
It also fixes a big that occurs when a flat slice is applied as non-flat_src_indices.

If `M` is a 3 x 3 matrix, then in numpy's convention, `M[:1]` retrieves the first _row_ of M.
In OpenMDAO, applying this slice to a connection would just give the first element.
This has been fixed.

There may be some setup performance hits due to this PR, but the fact that this bug was not encountered previously means that is probably unlikely that the slightly slower way of sorting out the src_indices from the indexer is commonly used.

### Related Issues

- Resolves #2437 
- Resolves #2446 

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/core_features/working_with_groups/connect.ipynb
+++ b/openmdao/docs/openmdao_book/features/core_features/working_with_groups/connect.ipynb
@@ -45,9 +45,19 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[1. 1. 1. 1. 1.]\n",
+      "[12. 12. 12. 12. 12.]\n",
+      "[60.]\n"
+     ]
+    }
+   ],
    "source": [
     "import numpy as np\n",
     "import openmdao.api as om\n",
@@ -72,14 +82,25 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {
     "tags": [
      "remove-input",
      "remove-output"
     ]
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "1.1842378929335003e-16"
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
    "source": [
     "from openmdao.utils.assert_utils import assert_near_equal\n",
     "\n",
@@ -97,9 +118,19 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[10.]\n",
+      "[20.]\n",
+      "[30.]\n"
+     ]
+    }
+   ],
    "source": [
     "p = om.Problem()\n",
     "\n",
@@ -118,14 +149,25 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
    "metadata": {
     "tags": [
      "remove-input",
      "remove-output"
     ]
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "0.0"
+      ]
+     },
+     "execution_count": 4,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
    "source": [
     "assert_near_equal(p.get_val('C1.y'), 10.)\n",
     "assert_near_equal(p.get_val('C2.y'), 20.)\n",
@@ -142,9 +184,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[1. 1. 1.]\n",
+      "[6.]\n",
+      "[1. 1.]\n",
+      "[8.]\n"
+     ]
+    }
+   ],
    "source": [
     "p = om.Problem()\n",
     "\n",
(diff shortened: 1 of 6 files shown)


openmdao/docs/openmdao_book/features/core_features/working_with_groups/connect.ipynb
```diff
@@ -229,21 +236,130 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Using `om.slicer` to connect non-scalar variables\n",
+    "\n",
+    "The `om.slicer` object was introduced to OpenMDAO to make connecting non-scalar variables easier by allowing the `src_indices` to be specified as slices.\n",
+    "This way of thinking about `src_indices` is probably more intuitive for users who are experienced with numpy.\n",
+    "Note when `src_indices` is provided as a slicer object, `flat_src_indices` has no effect.\n",
```
naylor-b, line 249: This is no longer true since we refactored how indexing works a while back.  For example, for om.slicer[:1] where the src is a 3x3 array.  With flat_src_indices=False, result after indexing is size 3 (first row of the array), but if flat_src_indices=True, result is size 1 (first entry of the flattened array).

naylor-b (changes requested):

robfalck:
This PR exposed the bug at #2446.  That bug should be addressed before merging this.

openmdao/docs/openmdao_book/features/core_features/working_with_groups/connect.ipynb
robfalck: Fixed, and added new examples.

naylor-b (approved):

openmdao/utils/indexer.py
```diff
@@ -544,14 +544,22 @@ def as_array(self, copy=False, flat=True):
         ndarray
             The index array.
         """
-        # use maxsize here since a shaped slice always has positive int start and stop
-        arr = np.arange(*self._slice.indices(sys.maxsize), dtype=int)
-        if flat:
-            return arr
-
-        if self._orig_shape is None:
```
Kenneth-T-Moore, line 552: Wow, this was totally broken. The variable `_orig_shape` isn't defined anywhere.

Kenneth-T-Moore (approved):

PR #2445, opened 2022-02-11
Bug fix and some changes to recording under MPI

### LINKED ISSUE #2434: Running an OpenMDAO problem on one proc in MPI environment
### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

This issue is a kind of a follow-up of #2416. I want to run a program using MPI but an initial openmdao problem is used to generate a doe then the program continues using all procs.  With the version b0b8c3d which fixes #2416 the following program (which is a simplified version of what I try to achieve), blocks when run with: `mpirun -n 2 python test_om.py` 

### Example

```python
import os
from mpi4py import MPI
import openmdao.api as om
from openmdao.test_suite.components.paraboloid import Paraboloid

def run_sequential():
    prob = om.Problem()

    prob.model.add_subsystem('comp', Paraboloid(), promotes=['x', 'y', 'f_xy'])
    prob.model.add_design_var('x', lower=0.0, upper=1.0)
    prob.model.add_design_var('y', lower=0.0, upper=1.0)
    prob.model.add_objective('f_xy')

    prob.driver = om.DOEDriver(om.FullFactorialGenerator(levels=3))
    prob.driver.options['run_parallel'] = False
    prob.driver.options['procs_per_model'] = 1
    prob.driver.add_recorder(om.SqliteRecorder("cases.sql"))

    prob.setup()
    prob.run_driver()

def run_parallel():
    print(f"rank = {MPI.COMM_WORLD.rank}")
    # process cases.sql

# Run only by proc 0
if MPI.COMM_WORLD.rank == 0:
    run_sequential()

# Synchronize as generated doe will be used by all procs
MPI.COMM_WORLD.Barrier()

# Run by all procs
run_parallel()  
```

When run with `mpirun -n 2 python test_om.py`, the program blocks.
Actually if I remove the `if MPI.COMM_WORLD.rank==0` condition it would work with the previous #2430 fix but
my problem is the general case of using an OpenMDAO problem on one proc while MPI is active. 
A second example, even simpler of blocking situation is 

```python
# test_om2.py
import os
from mpi4py import MPI
import openmdao.api as om
from openmdao.test_suite.components.paraboloid import Paraboloid

def run_sequential():
    prob = om.Problem()
    prob.model.add_subsystem('comp', Paraboloid(), promotes=['x', 'y', 'f_xy'])
    prob.setup()
    prob.run_model()

def run_parallel():
    print(f"rank = {MPI.COMM_WORLD.rank}")

# Run only by proc 0
if MPI.COMM_WORLD.rank == 0:
    run_sequential()

# Synchronize as previously generated result is used by all procs
MPI.COMM_WORLD.Barrier()

# Run by all procs
run_parallel() 
```

### Environment

Operating System: CentOS 7
Python environment: 3.8.5
Packages: OpenMDAO 3.16.1.dev0 b0b8c3d

swryan:
### Summary

Fixes a bug in `SqliteRecorder`, where it was blocking with a `Barrier` regardless of whether recording was in parallel or not, and adds a test for this condition, where the model is run and recorded on a single process in a multi-process MPI context.

Also:
- The Problem `comm` is now passed to all recorders at startup
- The `parallel` property of CaseRecorder is now computed based on the `record_on_process` property (parallel recording under MPI is activated by specifying `record_on_process = True` on the ranks which are to record)
- Added parallel recording logic/tests for GA and Differential Evolution drivers
- Got rid of some references to `MPI.COMM_WORLD`, some of which actually wanted the Problem `comm`.
- Fixed a bug in the `variable_table` logic where it didn't handle an empty list of variables
- Cleaned up some unused code

### Related Issues

- Resolves #2434

Some small changes to the `CaseRecorder` and `RecordingManager`  APIs, which should not be visible to users but may effect any developers that have implemented parallel recording on top of those APIs.

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/genetic_algorithm_driver.py
+++ b/openmdao/drivers/genetic_algorithm_driver.py
@@ -214,7 +214,7 @@ class SimpleGADriver(Driver):
 
             for recorder in self._rec_mgr:
                 if run_parallel:
-                    # write cases only on procs up to the number of parallel DOEs
+                    # write cases only on procs up to the number of parallel models
                     # (i.e. on the root procs for the cases)
                     if procs_per_model == 1:
                         recorder.record_on_process = True
--- a/openmdao/drivers/genetic_algorithm_driver.py
+++ b/openmdao/drivers/genetic_algorithm_driver.py
@@ -44,6 +44,8 @@ class SimpleGADriver(Driver):
 
     Attributes
     ----------
+    _problem_comm : MPI.Comm or None
+        The MPI communicator for the Problem.
     _concurrent_pop_size : int
         Number of points to run concurrently when model is a parallel one.
     _concurrent_color : int
@@ -176,6 +178,8 @@ class SimpleGADriver(Driver):
         MPI.Comm or <FakeComm> or None
             The communicator for the Problem model.
         """
+        self._problem_comm = comm
+
         procs_per_model = self.options['procs_per_model']
         if MPI and self.options['run_parallel']:
 
@@ -199,6 +203,33 @@ class SimpleGADriver(Driver):
         self._concurrent_pop_size = 0
         self._concurrent_color = 0
         return comm
+
+    def _setup_recording(self):
+        """
+        Set up case recording.
+        """
+        if MPI:
+            run_parallel = self.options['run_parallel']
+            procs_per_model = self.options['procs_per_model']
+
+            for recorder in self._rec_mgr:
+                if run_parallel:
+                    # write cases only on procs up to the number of parallel DOEs
+                    # (i.e. on the root procs for the cases)
+                    if procs_per_model == 1:
+                        recorder.record_on_proc = True
+                    else:
+                        size = self._problem_comm.size // procs_per_model
+                        if self._problem_comm.rank < size:
+                            recorder.record_on_proc = True
+                        else:
+                            recorder.record_on_proc = False
+
+                elif self._problem_comm.rank > 0:
+                    # if not running cases in parallel, then just record on proc 0
+                    recorder.record_on_proc = False
+
+        super()._setup_recording()
 
     def _get_name(self):
         """
--- a/openmdao/drivers/genetic_algorithm_driver.py
+++ b/openmdao/drivers/genetic_algorithm_driver.py
@@ -217,17 +217,15 @@ class SimpleGADriver(Driver):
                     # write cases only on procs up to the number of parallel DOEs
                     # (i.e. on the root procs for the cases)
                     if procs_per_model == 1:
-                        recorder.record_on_proc = True
+                        recorder.record_on_process = True
                     else:
                         size = self._problem_comm.size // procs_per_model
                         if self._problem_comm.rank < size:
-                            recorder.record_on_proc = True
-                        else:
-                            recorder.record_on_proc = False
+                            recorder.record_on_process = True
 
-                elif self._problem_comm.rank > 0:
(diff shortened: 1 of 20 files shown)


openmdao/recorders/case_recorder.py
```diff
@@ -164,25 +187,22 @@ def record_iteration(self, recording_requester, data, metadata, **kwargs):
...
-            self.record_iteration_driver(recording_requester, data, metadata)
-        elif isinstance(recording_requester, System):
-            self.record_iteration_system(recording_requester, data, metadata)
-        elif isinstance(recording_requester, Solver):
-            self.record_iteration_solver(recording_requester, data, metadata)
-        elif isinstance(recording_requester, Problem):
-            self.record_iteration_problem(recording_requester, data, metadata)
-        else:
-            raise ValueError("Recorders must be attached to Drivers, Systems, or Solvers.")
+        if self.record_on_proc:
```
Kenneth-T-Moore, line 190: Just to be clear, _Undefined triggers this condition, and that's expected? (As in the case where we have multiple procs but aren't running parallel, and the distributed drivers set all other procs to False, but don't set this one.)

openmdao/drivers/genetic_algorithm_driver.py
```diff
@@ -200,6 +204,33 @@ def _setup_comm(self, comm):
...
+                    else:
+                        size = self._problem_comm.size // procs_per_model
+                        if self._problem_comm.rank < size:
+                            recorder.record_on_proc = True
+                        else:
+                            recorder.record_on_proc = False
+
+                elif self._problem_comm.rank > 0:
+                    # if not running cases in parallel, then just record on proc 0
+                    recorder.record_on_proc = False
```
Kenneth-T-Moore, line 230: So GE is telling their recorder "Don't record on on procs 1-N, but do whatever your default behavior is on rank 0." (and current default behavior value is `_Undefined`, which behaves as True.) I feel like there is a slight disconnect here. I think the driver should either 1.) explicitly tell the recorder to record on 0 and not to record on 1-N, or 2.) just let openmdao pick one single rank to record on, since it is mostly irrelevant to the driver when run_parallel is False.  What we have here is a hybrid and doesn't feel consistent.

Kenneth-T-Moore (changes requested):

openmdao/drivers/genetic_algorithm_driver.py
swryan: The idea was that if you don't touch the `record_on_proc` flag then you get the default behavior, which is to only record on proc 0.  If you set `record_on_proc` to True on any procs then those procs and only those procs will record.  

I guess setting the flag to False is pointless.. since we only care if the flag is True or not.

I tried to express this in line 82 of case_recorder.py..   looks like I should embellish the explanation.

openmdao/recorders/case_recorder.py
swryan: right.. this is a bug.. need a test.

swryan: reworked the logic and got rid of _UNDEFINED

openmdao/drivers/genetic_algorithm_driver.py
swryan: reworked the logic..  no longer set False, since it only matters if it's True..  updated explanations in CaseRecorder

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2447, opened 2022-02-14
POEM 063 implementation: Allow multiple constraints on an output via constraint aliases.

### LINKED ISSUE #2367: Allow multiple constraints/objectives on the same variable as long as they use different indices
### Summary of Issue

Currently OpenMDAO will not allow multiple constraints to be applied to the same variable name.
While this is desirable behavior for scalars, for array-valued outputs we may wish to apply different constraints to different indices (or equality constraints to some indices, and inequality constraints to others).
OpenMDAO should allow this if the indices involved are different.

### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

If the same variable name appears in multiple responses, OpenMDAO should check the indices associated with each response and make sure they do not overlap.

**Currently, this test is done at the time the constraint/objective is added, but this may have to be delayed until setup, since slicer indices will not be resolved until then.  This will add the additional complication that responses are keyed by variable name.  If we now allow multiple responses to be applied to the same variable, we'll have to address this.**

### Example

The following contrived example attempts to set two constraints on different indices of `exec.y`.

```
    def test_two_constraints_different_indices(self):

        import openmdao.api as om
        from openmdao.utils.assert_utils import assert_near_equal

        p = om.Problem()

        exec = om.ExecComp(['y = x**2',
                            'z = a + x**2'],
                           a={'shape': (1,)},
                           y={'shape': (101,)},
                           x={'shape': (101,)},
                           z={'shape': (101,)})

        p.model.add_subsystem('exec', exec)

        p.model.add_design_var('exec.a', lower=-1000, upper=1000)
        p.model.add_objective('exec.y', index=50)
        p.model.add_constraint('exec.z', indices=[0], equals=25)
        p.model.add_constraint('exec.z', indices=[-1], lower=20)

        p.driver = om.pyOptSparseDriver(optimizer='SNOPT')
        p.driver.opt_settings['iSumm'] = 6

        p.setup()

        p.set_val('exec.x', np.linspace(-10, 10, 101))

        p.run_driver()

        assert_near_equal(p.get_val('exec.z')[0], 25)
        assert_near_equal(p.get_val('exec.z')[50], -75)
```

This fails when the second constraint is added.

```
  File "...test_multiple_constraints.py", line 519, in test_two_constraints_different_indices
    p.model.add_constraint('exec.z', indices=[-1], lower=20)
  File "...openmdao/core/system.py", line 2824, in add_constraint
    self.add_response(name=name, type_='con', lower=lower, upper=upper,
  File "...openmdao/core/system.py", line 2667, in add_response
    raise RuntimeError(msg.format(name))
RuntimeError: <class Group>: Constraint 'exec.z' already exists.
```

Similarly, adding a constraint to the same output as the constraint results in a similar error.

```
    def test_obj_and_con_same_var_different_indices(self):

        import openmdao.api as om
        from openmdao.utils.assert_utils import assert_near_equal

        p = om.Problem()

        exec = om.ExecComp(['y = x**2',
                            'z = a + x**2'],
                           a={'shape': (1,)},
                           y={'shape': (101,)},
                           x={'shape': (101,)},
                           z={'shape': (101,)})

        p.model.add_subsystem('exec', exec)

        p.model.add_design_var('exec.a', lower=-1000, upper=1000)
        p.model.add_objective('exec.z', index=50)
        p.model.add_constraint('exec.z', indices=[0], equals=25)

        p.driver = om.pyOptSparseDriver(optimizer='SNOPT')
        p.driver.opt_settings['iSumm'] = 6

        p.setup()

        p.set_val('exec.x', np.linspace(-10, 10, 101))

        p.run_driver()

        assert_near_equal(p.get_val('exec.z')[0], 25)
        assert_near_equal(p.get_val('exec.z')[50], -75)
```

### Environment

N/A

Kenneth-T-Moore:
### Summary

Allow multiple contraints to be declared on an output via constraint aliases.
- Constraints must be decaled with non-colliding indices.

This completes POEM 063.

### Related Issues

- Resolves #2367

None

### New Dependencies

None


openmdao/core/total_jac.py
```diff
@@ -1519,6 +1540,12 @@ def check_total_jac(self, raise_error=True, tol=1e-16):
         if np.any(col):  # there's at least 1 row that's zero across all columns
             zero_rows = []
             for name, tup in self.of_meta.items():
+
+                # Support for constraint aliases.
+                if name in self.responses and 'path' in self.responses[name] and \
```
naylor-b, line 1545: This same logic appears in a bunch of places around the code.  Might be worth replacing with a function.

naylor-b (changes requested):

openmdao/core/group.py
```diff
@@ -3071,12 +3079,21 @@ def _setup_approx_partials(self):
 
         approx_keys = self._get_approx_subjac_keys()
         for key in approx_keys:
+            left, right = key[0], key[1]
```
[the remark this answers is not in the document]
Kenneth-T-Moore: Fixed

openmdao/core/system.py
```diff
@@ -1612,6 +1629,73 @@ def _init_relevance(self, mode):
...
+
+                if name in discrete['input'] or name in discrete['output']:
+                    continue
+
+                if responses[name]['path'] is not None:
+                    path_name = responses[name]['path']
+                else:
+                    path_name = None
+
+                if name not in responses or \
```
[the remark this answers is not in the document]
Kenneth-T-Moore: Fixed

openmdao/core/system.py
```diff
@@ -1612,6 +1629,73 @@ def _init_relevance(self, mode):
...
+                        (path_name is not None and path_name not in responses and
+                         prom2abs_names[path_name][0] not in responses):
+                    raise RuntimeError(f"Alias '{name}' is not needed when only "
```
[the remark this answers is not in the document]
Kenneth-T-Moore: I vaguely recall a discussion about this, and we decided that aliases should only be used when you have multiple constraints on a source.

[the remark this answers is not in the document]
Kenneth-T-Moore: Wait, Bret's memory is  correct???  Let me mark may calendar.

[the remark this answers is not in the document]
naylor-b: I know!  Amazing isn't it?

openmdao/utils/indexer.py
```diff
@@ -11,6 +11,37 @@
 from openmdao.utils.om_warnings import issue_warning, OMDeprecationWarning
 
 
+def _index_sort(indices):
```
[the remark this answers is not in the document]
Kenneth-T-Moore: This function was removed in my rewrite of the overlap check.

openmdao/core/system.py
```diff
@@ -1615,6 +1616,53 @@ def _init_relevance(self, mode):
...
+                                used_idx[path][:] += 1
+                            else:
+                                used_idx[path][idx.flat()] += 1
+                        else:
+                            # If an alias is in responses, but the path isn't, then we need to
+                            # make sure the path is present for the relevancy calculation.
+                            responses[path] = responses[name]
+
+                    indices = responses[name].get('indices')
+                    if indices is None:
```
naylor-b, line 1655: It seems like with this logic that we could potentially be adding +1 for the indices twice.

openmdao/core/system.py
```diff
@@ -1615,6 +1616,53 @@ def _init_relevance(self, mode):
...
+                    if name in discrete['input'] or name in discrete['output']:
+                        continue
+
+                    path = responses[name]['path']
+                    if path is None:
+                        path = name
+
+                    if name not in used_idx and path not in used_idx:
+                        size = self._var_allprocs_abs2meta['output'][path]['global_size']
+                        used_idx[path] = np.zeros(size, dtype=int)
```
naylor-b, line 1638: I still think that at some point it would be worth it to investigate using the sparse matrix trick to check the overlapping indices, because the current method could be fairly expensive in terms of memory at least (creating arrays of full global size for every alias and keeping them all simultaneously in a dict) and most likely the sparse method would be considerably faster for large response variables.

openmdao/core/system.py
```diff
@@ -1615,6 +1616,53 @@ def _init_relevance(self, mode):
...
+                            # make sure the path is present for the relevancy calculation.
+                            responses[path] = responses[name]
+
+                    indices = responses[name].get('indices')
+                    if indices is None:
+                        used_idx[path][:] += 1
+                    else:
+                        used_idx[path][indices.flat()] += 1
+
+                    overlap = np.where(used_idx[path] > 1)
```
naylor-b, line 1660: Since you're not using the actual overlapping indices for anything, you could instead just use `overlap = np.any(used_idx[path] > 1)`

openmdao/drivers/tests/test_pyoptsparse_driver.py
```diff
@@ -2270,6 +2270,413 @@ def test_singular_jac_tol(self):
 
         prob.run_driver()
 
+    def test_constraint_alias(self):
```
naylor-b, line 2273: Do we have tests to check that we'll catch it if a user chooses an alias name that collides with a real variable name or with another alias?  I may have missed them.

naylor-b (changes requested):

openmdao/core/driver.py
```diff
@@ -341,13 +342,27 @@ def _setup_driver(self, problem):
...
+                        vpath = responses[vname]['path']
                     indices = responses[vname].get('indices')
+
                 elif vname in src_design_vars:
                     indices = src_design_vars[vname].get('indices')
-                else:
-                    continue
+
+                meta = model._var_allprocs_abs2meta['output'][vpath]
+                i = abs2idx[vpath]
```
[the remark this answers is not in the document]
Kenneth-T-Moore: It is also in the else branch.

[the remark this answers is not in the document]
naylor-b: Ah, missed that, sorry.

openmdao/core/system.py
Kenneth-T-Moore: It shouldn't.  The first part under the conditional only adds the indices from any constraint that was added to the variable with no alias name (if there is one). The second part adds the indices for the alias. If they both have overlapping indices, it really is a dup.

openmdao/core/system.py
Kenneth-T-Moore: Replaced the code.

openmdao/core/system.py
Kenneth-T-Moore: It is probably worth looking into at some point. I don't know how we would handle partial or full slices though.

naylor-b: Fair point.  You'd have to convert the indexers to arrays, which might cost you some performance, but you could still potentially save a lot of memory if the indices are taking a smallish chunk of the full array.

openmdao/drivers/tests/test_pyoptsparse_driver.py
Kenneth-T-Moore: It turns out that it is a bit dangerous to allow them to use existing openmdao names, so I've added the test and the check.

naylor-b (approved):

Kenneth-T-Moore:
Pulling this down, now that it is in Bret's hands.

PR #2489, opened 2022-03-31
move MPI scaling report test to it's own file

swryan:
### Summary

This test has a history of timing out and failing when run as part of the test suite via testflo, so move it to it's own file and set it to run in ISOLATED mode.

### Related Issues

- Resolves # N/A

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_scaling_report_mpi.py
+++ b/openmdao/core/tests/test_scaling_report_mpi.py
@@ -1,0 +1,14 @@
+"""Define the units/scaling tests."""
+import unittest
+
+from openmdao.utils.testing_utils import use_tempdirs
+from openmdao.core.tests.test_scaling_report import TestDriverScalingReport
+
+
+class TestDriverScalingReportMPI(TestDriverScalingReport):
+    N_PROCS = 2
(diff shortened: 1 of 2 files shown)


openmdao/core/tests/test_scaling_report_mpi.py
```diff
@@ -0,0 +1,14 @@
+"""Define the units/scaling tests."""
+import unittest
+
+from openmdao.utils.testing_utils import use_tempdirs
+from openmdao.core.tests.test_scaling_report import TestDriverScalingReport
+
+
+class TestDriverScalingReportMPI(TestDriverScalingReport):
+    N_PROCS = 2
+    ISOLATED = True
```
naylor-b, line 10: I forgot somehow when I suggested using ISOLATED that the test was already an MPI test.  TestCases with N_PROCS defined already run isolated in their own subprocess, so the ISOLATED thing shouldn't be necessary.  Just moving the test to its own file is probably what got rid of the hang.

Kenneth-T-Moore (approved):

naylor-b (approved):

PR #2492, opened 2022-04-04
Print pyOptSparse opt_prob before the optimization

friedenhe:
### Summary

Added a feature to print the opt_prob in the pyOptSparse driver before the optimization runs. This allows users to confirm their optimization configurations (e.g., design variable bounds). The print can be controlled by the driver's `print_opt_prob` option (default True). Also fixed a problem for the pyOptSparse driver's `print_result` behavior in parallel. Before, each processor will print an identical optimization summary to the screen when the optimization finishes. This becomes annoying when running with a lot of CPU cores because users will see a long list of duplicated opt summaries at the end of the log file. Now only the root processor 0 will print the info.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/pyoptsparse_driver.py
+++ b/openmdao/drivers/pyoptsparse_driver.py
@@ -29,6 +29,7 @@ from openmdao.utils.class_util import WeakMethodWrapper
 from openmdao.utils.mpi import FakeComm
 from openmdao.utils.om_warnings import issue_warning, DerivativesWarning
 from openmdao.utils.general_utils import _src_or_alias_name
+from openmdao.utils.mpi import MPI
 
 
 # names of optimizers that use gradients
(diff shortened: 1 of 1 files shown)


openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -211,6 +212,8 @@ def _declare_options(self):
                              desc='Name of optimizers to use')
         self.options.declare('title', default='Optimization using pyOpt_sparse',
                              desc='Title of this optimization run')
+        self.options.declare('print_opt_prob', types=bool, default=True,
```
robfalck, line 215: I would prefer if this had a default of `False`.

We're in the process of trying to move away from having so much standard output, and moving towards reports instead.  In fact, one of our team members is working on an optimization report that will summarize everything about the optimization problem.

Sometimes instant feedback via stdout is good, but I think the user should have to ask for it rather than enabling it by default.

robfalck (changes requested):

openmdao/drivers/pyoptsparse_driver.py
friedenhe: @robfalck Thanks for the feedback. I have changed it to default false and pushed it to the PR.

robfalck (approved):

naylor-b (approved):

PR #2499, opened 2022-04-15
Align the population initialization of SimpleGA and DifferentialEvolution drivers

gbarter:
### Summary

Currently the two evolutionary drivers (SimpleGA and DifferentialEvolution) initialize their populations differently.  This was pointed out in a StackOverflow [exchange](https://stackoverflow.com/questions/71818035/how-are-the-design-variables-in-the-simplega-or-differentialevolution-drivers-in).  This PR better aligns the two drivers by using Latin Hypercube Sampling for npop-1 points and the user specified initial condition for the final npop entry.

I did not see any dedicated unit tests for these drivers, although I likely just missed them in the file structure.  I did test out the changes on the Hohmann transfer problem example and the overall behavior seems unchanged.

Note: I am skeptical about the rounding [here](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/drivers/genetic_algorithm_driver.py#L651), but I am hesitant to change that in this PR without understanding the GA encoding better.

### Related Issues

- StackOverflow [question](https://stackoverflow.com/questions/71818035/how-are-the-design-variables-in-the-simplega-or-differentialevolution-drivers-in)

None

### New Dependencies

None


Kenneth-T-Moore:
There are some unit tests in `openmdao/drivers/tests`.  According to the CI check, there are some failing tests: 1 in the differential evolution tests and several in the SimpleGA tests.  I haven't looked too deeply into them, but for the numerical ones, some are off by just a little bit (which is plausible), and some are off by a lot.  There is a size mismatch exception in one test too, but it's not clear how that happens with this change.

There are also some linting errors. The linting test is in `openmdao/code_review`, but we always recommend running all of the tests from the top directory using something like testflo.

Kenneth-T-Moore:
There are some unit tests in `openmdao/drivers/tests`.  According to the CI check, there are some failing tests: 1 in the differential evolution tests and several in the SimpleGA tests.  I haven't looked too deeply into them, but for the numerical ones, some are off by just a little bit (which is plausible), and some are off by a lot.  There is a size mismatch exception in one test too, but it's not clear how that happens with this change.

There are also some linting errors. The linting test is in `openmdao/code_review`, but we always recommend running all of the tests from the top directory using something like testflo.

Kenneth-T-Moore (changes requested):
See post above.

gbarter:
Thank you @Kenneth-T-Moore for the patient explanation to an OpenMDAO PR newbie.  To be honest, I am unsure why the small change to the SimpleGA code created such big changes to the unit test performance.  To play it safe, I reverted my changes back to the original approach.  

For the DifferentialEvolution unit test that was failing, the truth value in the assert was unaffected by the optimization problem being tested.  I therefore considered it safe to change the truth value to the result obtained with the revised initial population.

Kenneth-T-Moore (approved):

naylor-b (approved):

PR #2512, opened 2022-04-28
updated shape_to_len and made it call math.prod if python >= 3.8

### LINKED ISSUE #2506: Add a utility function for computing the size of variables.
### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #N/A

### Description

OpenMDAO often needs the total size of a variable and typically computes this via `np.prod(meta['shape'], dtype=int)`
The `prod` function in Python's math library is significantly faster, but only supported in 3.8 and later.

We should add a utility function that we can consistently use that uses `math.prod` when available and resorts to `np.prod` when it is not, and make changes to OpenMDAO to use this wrapped size function wherever we currently use `np.prod`.

### Example

N/A

### Environment

N/A

naylor-b:
### Summary

`np.prod` is quite slow for small tuples like shapes, and `math.prod` is faster.  We had a preexisting function called `shape_to_len` that was already faster than `np.prod` for shapes, but `math.prod` is still faster, so I just updated `shape_to_len` if python >= 3.8 to use math.prod.

### Related Issues

- Resolves #2506

None

### New Dependencies

None


openmdao/utils/array_utils.py
```diff
@@ -355,6 +374,26 @@ def get_input_idx_split(full_idxs, inputs, outputs, use_full_cols, is_total):
         return [(inputs, full_idxs)]
 
 
+def convert_neg(arr, size):
```
swryan, line 377: curious why this function exists since it's only used in the test... debugging?
also, why not get the size from the ndarray?

swryan (approved):

Kenneth-T-Moore (approved):

PR #2534, opened 2022-05-25
refactor of _apply_linear/_solve_linear and block linear solvers

### LINKED ISSUE #2405: Simple Caching for Linear Operators to improve performance
### Issue Type

- [ ] Bug
- [x] Enhancement  
      POEM #TBD

### Description

Offer a matrix-free-caching option for components that allows OpenMDAO to check if the seed is the same and avoid calling the linear operator multiple times for the same seed. When declared (during the `setup_partials` phase) the user is telling OpenMDAO that it will ignore relevance information in `d_` vectors (i.e. ignore whether or not a seed is present in `d_inputs`, `d_outputs`, or `d_residuals`) and will compute a full matrix-vector-product regardless. Hence, if multiple successive calls to the linear operator are made with the same seed but different relevant varialbles in `d_` vectors, OpenMDAO can no-op subsequent calls to the linear operator. 

Why is this needed? 

OpenMDAO's design for matrix-free linear operators inherently assumes that component authors would want to be as stingy as possible about computing matrix-vector products. If specific variables were not needed, then it would be more efficient to skip the parts of the product associated with them. This is generally a good assumption if the structure of the computation has a lot of inputs and funnels directly down to a small number of outputs. For example, if only `y` is relevant then you would only want to do the part of the product associated with `y` but not `x`. 

Screen Shot 2022-01-10 at 2 45 45 PM

When the computation is funnel shaped (i.e. starts wide and gets narrower) If `x` is not relevant then any computational work to propagate the derivatives through `x` if wasted. So if you only need the `y` part now, then you should only compute the `y` part. When/if you need the `x` part later, you can compute it then.  Even if you called the matrix-vector-product with the same exact seed twice, just with `y` relevant once and `x` relevant the next time it would still be better to only compute the relevant parts. 
(note: Due to algorithmic specifics of linear block gauss-seidel and the hierarchical nature of OpenMDAO, it is VERY probably that matrix-vector products will get called with the same seed but different relevant variables). 

 
However, if you have a modest number of outputs that first fans out to a larger number of intermediate variables before collapsing back down to a small number of outputs then the most efficient way to handle the derivatives changes. 

Screen Shot 2022-01-10 at 2 53 40 PM

With a fan-out/fan-in compute structure, splitting up the matrix-vector product into two parts (one for `x` and one for `y` ) is less efficient. You end up duplicating the derivative computations through the entire middle block for both of them regardless of what is relevant or not. Here it is better to compute the entire matrix-vector-product any time a given seed if provided, and then cache the result in the memory and re-use them if/when you get asked to do multiple products with the same seed but different relevant variables. 

In the extreme case you could cache every single case you got just in case you saw it again, but this is not needed. Since we know that algorithmic specifics of OpenMDAO will often yield successive cases of the same seed, we only need to cache the last matrix-vector-product and some checksum/hash of the seed. Any time we see the checksum/hash change, we can safely recompute the matrix-vector-product. 

This behavior has to be a user selected option at the component. They need to tell us if they are going to use/respect relevance information or not and if they want us to cache the last linear operator result.

naylor-b:
### Summary

When nested groups had block linear solvers (LinearBlockGS and LinearBlockJac), there were unnecessary _apply_linear calls being made on components under those groups.  

The _solve methods of the block linear solvers were modified so that direct calls to _apply_linear on subsystems would only be called if those subsystems were either components or were groups not being solved by block linear solvers themselves.  Additional args were added to _solve_linear so that the variables relevant to a block linear solve could be passed down the system tree, accumulating at each group level until (possibly) reaching a component.  This allows a component _apply_linear to execute only once in a given recursive block linear solve where the set of 'relevant' variables will be the full set from the group originating the linear solve down the tree to the component, instead of the old way, where the component apply linear would be called once for each group level linear solve, with different 'relevant' variables at each level.

The issue this PR is based on, #2405, described using caching to avoid unnecessary computations in compute_jacvec_product or apply_linear, but it was determined that we could refactor _apply_linear/_solve_linear instead and the caching wouldn't be required.

This PR also adds the get_hash method to Vector, and POEM 067 describes that method.

### Related Issues

- Resolves #2405

None

### New Dependencies

None


openmdao/vectors/vector.py
```diff
@@ -691,3 +685,20 @@ def set_complex_step_mode(self, active):
             Complex mode flag; set to True prior to commencing complex step.
         """
         self._under_complex_step = active
+
+    def get_hash(self, alg=hashlib.sha1):
```
Kenneth-T-Moore, line 689: I don't think this method is used anywhere.

Kenneth-T-Moore (changes requested):

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2537, opened 2022-05-31
I2474 optimization report

### LINKED ISSUE #2474: Add an HTML report for detailed optimization results.
### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

pyOptSparseDriver, by default, gives a detailed ASCII report to standard output after it is finished.
Now that we can automatically generate richer reports, OpenMDAO should generate an HTML optimization report for optimization drivers, using as much information from the Problem as possible, to be driver-independent.  The report should include:

- The script.
- The problem name
- A timestamp of the execution of the script
- The driver used.
- The metadata/options of the driver.
- Tables of the design variables, objectives, and constraints (include scaling information, bounds, current values vs bounds, and plots showing the relation of the current values vs the bounds.
- The exit status of the optimization (success, failure, inform, etc)

### Example

I previously wrote this implementation when thinking about this:
 
https://gist.github.com/robfalck/e5d4932cfb3560f70dc5ceab2bc27942

It produces the following output (github won't allow the html to be attached here):

Screen Shot 2022-03-17 at 8 30 59 AM

It captures most of the requirements above.  It might also be good to have some way to view all of the values in an array (a collapsable view).

### Environment

N/A

hschilling:
### Summary

Created a function to generate an HTML report to document the metrics and results of an optimization run.

Used this function to create a new report that was added to the list of default reports to run.

Added an attribute to drivers, `opt_result` that contains some basic metrics about the driver run. 

### Related Issues

- Resolves #2474 

None

### New Dependencies

None


openmdao/core/driver.py
```diff
@@ -820,6 +878,8 @@ def run(self):
         with RecordingDebugging(self._get_name(), self.iter_count, self):
             self._problem().model.run_solve_nonlinear()
 
+        # end_time = time.time()
```
naylor-b, line 881: debugging code?

naylor-b (approved):

openmdao/core/driver.py
hschilling: yes, removed

PR #2547, opened 2022-06-14
New feature that lets users specify constants in ExecComp expressions

### LINKED ISSUE #1774: Allow users to mark variables as constants in exec-comp
For example, the input vector for the following: `ExecComp('x = a + b', a={'val': 6, 'input':False})`,
would contain only 'b'.  'a' would be treated as an internal constant attribute.  Specifying 'input' of False without specifying 'val' would be an error.
The default value of 'input' will be True.

hschilling:
### Summary

Lets the user specify that a variable in an ExecComp expression is a constant.

### Related Issues

- Resolves #1774 

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/exec_comp.py
+++ b/openmdao/components/exec_comp.py
@@ -18,10 +18,11 @@ VAR_RGX = re.compile(r'([.]*[_a-zA-Z]\w*[ ]*\(?)')
 # Names of metadata entries allowed for ExecComp variables.
 _allowed_meta = {'value', 'val', 'shape', 'units', 'res_units', 'desc',
                  'ref', 'ref0', 'res_ref', 'lower', 'upper', 'src_indices',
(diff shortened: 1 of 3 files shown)


openmdao/components/exec_comp.py
```diff
@@ -441,6 +462,8 @@ def _setup_expressions(self):
                 if var in outs:
                     current_meta = self.add_output(var, val, **meta)
                 else:
+                    if 'constant' in meta:
+                        meta.pop('constant', None)
```
naylor-b, line 466: I assume that if 'constant' is in meta here that it has to be False, right?

naylor-b (approved):

openmdao/components/exec_comp.py
hschilling: yes

robfalck (approved):

PR #2553, opened 2022-06-23
Bug when pyoptsparse used with coloring and scaling report

### LINKED ISSUE #2552: Buggy interaction between pyoptsparse, coloring, and scaling report
### Issue Type

- [x] Bug
- [ ] Enhancement  

### Description

When pyoptsparse was used with coloring and scaling report active, the format of the total jacobian in the driver was being set to 'array' which is incompatible with pyoptsparse.

naylor-b:
### Summary

When used with coloring, the scaling report was setting the return format to 'array' when computing totals, which is incompatible with pyoptsparse, so the scaling report was updated to set the format to the preferred format for the driver.

Also added a decorator to allow temporary setting of environment variables, and tweaked the report system's use of the TESTFLO_RUNNING environment var to make it easier to turn reports on selectively during testing.

### Related Issues

- Resolves #2552

Added a new attribute to Driver called `_total_jac_format` that defaults to 'flat_dict'.  All drivers that support derivatives should set this to the appropriate value or it may cause errors when using the scaling report and coloring.

### New Dependencies

None

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -78,7 +78,8 @@ class Driver(object):
     _total_jac_sparsity : dict, str, or None
         Specifies sparsity of sub-jacobians of the total jacobian. Only used by pyOptSparseDriver.
     _total_jac_format : str
-        Specifies the format of the total jacobian.
+        Specifies the format of the total jacobian. Allowed values are 'flat_dict', 'dict', and
+        'array'.
     _res_subjacs : dict
         Dict of sparse subjacobians for use with certain optimizers, e.g. pyOptSparseDriver.
         Keyed by sources and aliases.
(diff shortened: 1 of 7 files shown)


Kenneth-T-Moore (approved):

openmdao/core/driver.py
```diff
@@ -77,6 +77,8 @@ class Driver(object):
         Metadata pertaining to total coloring.
     _total_jac_sparsity : dict, str, or None
         Specifies sparsity of sub-jacobians of the total jacobian. Only used by pyOptSparseDriver.
+    _total_jac_format : str
+        Specifies the format of the total jacobian.
```
swryan, line 81: should probably list the allowable values here, or point to where it's documented in `compute_totals`

swryan (approved):

openmdao/core/driver.py
naylor-b: done

PR #2578, opened 2022-07-20
Fixed some issues with reports dirs and the index page displayed by 'openmdao view_reports'

naylor-b:
### Summary

- lots of empty report directories were being created erroneously while running testflo
- said empty report directories were cluttering up the index page
- tree displayed on the index page was not collapsible


openmdao/utils/reports_system.py
```diff
@@ -22,7 +22,7 @@
 _reports_registry = {}
 _default_reports = ['scaling', 'total_coloring', 'n2', 'optimizer']
 _active_reports = set()  # these reports will actually run (assuming their hook funcs are triggered)
-_cmdline_reports = set()  # cmdline reports can be registered here to prevent default reports
+_cmdline_reports = set()  # cmdline reports registered here to so default reports aren't modified
```
swryan, line 25: typo: "to so"

openmdao/utils/reports_system.py
```diff
@@ -46,7 +46,7 @@ def reports_active():
     bool
         Return True if reports are active.
     """
-    return os.environ.get('TESTFLO_RUNNING', '').lower() not in ('1', 'true')
+    return os.environ.get('TESTFLO_RUNNING', '').lower() not in _truthy
```
swryan, line 49: I wonder if we need `_truthy` and `_falsey` to exist here in `reports_system`... it seems redundant with `general_utils.env_truthy`.
This version of `_truthy` is now being imported and used in `Problem` as well... seems like we should just be using the existing logic from `general_utils`...

openmdao/api.py
```diff
@@ -112,6 +112,9 @@
     MPIWarning, UnitsWarning, SolverWarning, OMDeprecationWarning, \
     OMInvalidCheckDerivativesOptionsWarning
 
+# Utils
+from openmdao.utils.file_utils import image2html, text2html
+
```
Kenneth-T-Moore, line 117: Were these added to api intentionally? Will people use these?

Kenneth-T-Moore (approved):

openmdao/api.py
naylor-b: The idea was that if someone is writing their own reports, they could use these functions to easily wrap their text output or image files. I put this in here when I as adding dymos reports since dymos was already importing openmdao.api, but it's not a big deal to move it out of here if it doesn't seem like it belongs.

openmdao/utils/reports_system.py
naylor-b: fixed

openmdao/utils/reports_system.py
naylor-b: I changed over to using general_utils functions.

robfalck (approved):

PR #2580, opened 2022-07-26
Allowed total derivatives to be approximated when there is discrete coupling

lukasmu:
### Summary

It is often possible to approximate total derivatives even there there is discrete coupling in a system. However, without the changes proposed in this pull request OpenMDAO would not attempt to approximate the total derivatives. These small changes make it possible to use OpenMDAO to investigate a large set of problems where it is unavoidable to pass discrete inputs and outputs between components.

### Related Issues

None

None

### New Dependencies

None

### Example

```
import numpy as np
from openmdao.test_suite.components.sellar_feature import SellarDis2

import openmdao.api as om


class SellarDis1Modified(om.ExplicitComponent):

    def setup(self):
        self.add_input('z', val=np.zeros(2))
        self.add_input('x', val=0.)
        self.add_input('y2', val=1.0)
        self.add_discrete_output('y1_discrete', val='1.0')

    def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
        z1 = inputs['z'][0]
        z2 = inputs['z'][1]
        x1 = inputs['x']
        y2 = inputs['y2']
        y1 = z1 ** 2 + z2 + x1 - 0.2 * y2
        discrete_outputs['y1_discrete'] = str(y1[0])


class StrToFloat(om.ExplicitComponent):
    """
    Component containing Discipline 1 -- no derivatives version.
    """

    def setup(self):
        self.add_discrete_input('y1_discrete', val='1.0')
        self.add_output('y1', val=1.0)

    def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
        outputs['y1'] = complex(discrete_inputs['y1_discrete'])


class SellarMDA(om.Group):
    """
    Group containing the Sellar MDA.
    """

    def setup(self):
        cycle = self.add_subsystem('cycle', om.Group(), promotes=['*'])
        cycle.add_subsystem('d1', SellarDis1Modified(), promotes=['x', 'z', 'y2', 'y1_discrete'])
        cycle.add_subsystem('dx', StrToFloat(), promotes=['y1_discrete', 'y1'])
        cycle.add_subsystem('d2', SellarDis2(), promotes=['z', 'y1', 'y2'])

        cycle.set_input_defaults('x', 1.0)
        cycle.set_input_defaults('z', np.array([5.0, 2.0]))

        # Nonlinear Block Gauss Seidel is a gradient free solver
        cycle.nonlinear_solver = om.NonlinearBlockGS()

        self.add_subsystem('obj_cmp', om.ExecComp('obj = x**2 + z[1] + y1 + exp(-y2)',
                                                  z=np.array([0.0, 0.0]), x=0.0),
                           promotes=['x', 'z', 'y1', 'y2', 'obj'])

        self.add_subsystem('con_cmp1', om.ExecComp('con1 = 3.16 - y1'), promotes=['con1', 'y1'])
        self.add_subsystem('con_cmp2', om.ExecComp('con2 = y2 - 24.0'), promotes=['con2', 'y2'])


prob = om.Problem()
prob.model = SellarMDA()

prob.driver.options['optimizer'] = 'SLSQP'
prob.driver.options['tol'] = 1e-8

prob.model.add_design_var('x', lower=0, upper=10)
prob.model.add_design_var('z', lower=0, upper=10)
prob.model.add_objective('obj')
prob.model.add_constraint('con1', upper=0)
prob.model.add_constraint('con2', upper=0)

prob.model.approx_totals()

prob.setup()

prob.run_driver()

print('minimum found at')
print(prob.get_val('x')[0])
print(prob.get_val('z'))

print('minumum objective')
print(prob.get_val('obj')[0])
```

### DIFF
--- a/openmdao/core/total_jac.py
+++ b/openmdao/core/total_jac.py
@@ -213,10 +213,15 @@ class _TotalJacInfo(object):
                 inter = discrete_outs.intersection(model._relevant[inp]['@all'][0]['output'])
                 if inter:
                     kind = 'of' if self.mode == 'rev' else 'with respect to'
-                    raise RuntimeError("Total derivative %s '%s' depends upon "
-                                       "discrete output variables %s." %
-                                       (kind, inp, sorted(inter)))
-
+                    msg = "Total derivative %s '%s' depends upon " \
+                          "discrete output variables %s." % (kind, inp, sorted(inter))
+                    if not approx:
+                        raise RuntimeError(msg)
+                    else:
+                        msg += " Since this total derivative is approximated this might " \
+                               "not be a problem."
+                        issue_warning(msg, category=DerivativesWarning)
+                        
         self.of = of
         self.wrt = wrt
         self.prom_of = prom_of


JustinSGray:
Thanks for the toy demonstration case. It makes it simpler to discuss the matter at a high level. The details may get a lot more complex though. I should start by saying that I am not personally in favor of a change like this. We put that error condition in because in general approximating total or semi-total derivatives across discrete outputs is mathematically invalid and incorrect. Perhaps you can change my mind though :) So here goes: 

I am going to separate out discrete variables into two sub-categories: 

1) Psuedo-discrete outputs 
2) true-discrete outputs 

In the toy example problem you have given, you created psuedo-discrete outputs (I acknowledge that you likely did this for the sake of a simple test problem). These are values that are in fact real and continuous, but which are being encapulated into a non floating point data format. Here yo simply wrapped a float as a string. In a more general case, you might have an output file from a code containing many many numbers in it; your might also have some object that contains a whole set of numbers and an associated computational state. Regardless, the core mathematical values you're passing along are continuous and hence you can validly approximate derivatives across them. 

true-discrete outputs are values that do not have an underlying continuous representation at all, such as (red, green, blue) or (1, 2, 3). In this case, although its possible to "approximate" a derivative using finite difference, this is not a mathematically valid thing to do. In the case of (1,2,3) you might be tempted to say that at least the discrete values have the concept of order and the relaxed versions make sense to differentiate. In the case of (reg, green, blue) you can't even say that though. Either way though, if the model requires true discrete outputs then I don't think you should be taking fake approximations of the derivative. If you think its valid to approximate the discrete numbers as continuous values, then you should do that at the model level as well and make the computations actually continuous. 

So overall, I would be personally opposed to allowing finite difference approximations of true-discrete variables at all, on the grounds that its mathematically invalid. The pseudo-discrete variables warrant a bit more careful consideration though. If the values are pseudo-discrete then why do they have to be passed as discrete values at all? 

Are you trying to pass an output file around? Can you instead parse the values from that file and place them into continuous outputs? I know that in many commercial frameworks, passing files around is the standard operating procedure, but I highly dislike it because, even though you can validly finite-difference across this, it locks you out of ever graduating to analytic derivatives (which absolutely must us continuous variables). So I strongly encourage you to consider parsing the values out of the file and passing them as floats. 

If instead you have a computational object, thats a slightly more tricky situation. Often codes output objects/structures that are useful to them, and you pretty reasonably want to pass that between openMDAO components. If those objects are mainly containers full of floating point data then you can again think of this as mathematically valid to FD across. I suppose you could lump "files" into this same class of information.

So should OpenMDAO let you FD across this class of data? Maybe. the problem is that its not possible for the framework to distinguish between a container like object and a true-discrete object at all. Im personally against OpenMDAO allowing any mathematically invalid operations at all, so we'd need to find a way to distinguish between them. The only thing that comes to mind is a kind of `to_float` method that you could implement on an object/file that would tell OpenMDAO how to convert the data to simple arrays of floats, and then perhaps OpenMDAO could hide the fact that its moving data around in that way from you and let you connect the objects. Of course, `to_float` is 90% of the work of just using normal outputs anyway. So that mostly defeats the point. 

So overall, I still think its a bad idea to let users FD across discrete variables. We have no way of know if they are true-discrete or pseudo-discrete, and allowing this style of model building cuts you off from other more advanced ways of taking derivatives (complex-step, and analytic methods). 

If you think there is a need for a more comprehensive consideration here though, I encourage you to submit a [POEM](https://github.com/OpenMDAO/POEMs) that gives more complex examples and talks about how you would handle the valid and invalid classes of discrete variable.

lukasmu:
Thanks @JustinSGray for your elaborate answer!

A bit of background on my real case:

I have a component that generates an object representing the geometry of an aircraft wing. Furthermore, I have several other components that use this object as an input. 

Initially, I tried to map this object into a number of continuous variables/arrays. However, always converting the object into an array and back into an object is quite computationally expensive. Especially, given that most components almost always require the entire object anyhow. Therefore, I am now simply passing the entire object between the components. This also looks more elegant on a N2 chart because I just have one node labelled _wing geometry_ and not a whole lot of different nodes like _wing trailing edge points_, _wing leading edge points_, _airfoil inner points_, _airfoil outer points_, ....

So we are talking about an object here that consist of only continuous variables (in fact it also contains discrete variables, but these are fixed during a single optimization run). I think mathematically there should be no problem to take finite-differences here.

I also don't think that I can obtain analytic derivatives for many of the components which are using the wing object because they perform rather complex geometric tasks (making lofts etc.) that are difficult/impossible to differentiate in an analytical way.

So for my particular case passing the wing object between components just works nicely. Actually, using the object enabled me to split up some larger components and reorder them. Interestingly, this resulted in better convergence and slightly better optimization results as well. 

However, I also get your point, especially with regards to the mathematical validness that would be difficult to guarantee. I will close this request for now but will continue thinking about it.

JustinSGray:
well, at the very least, you're obviously welcome to patch OM to make it do what you need it to do :) That's wonder of open source software. 

The computational object is at least a valid approach. if its literally just about grouping things into a data structure (and not about carrying along the extra methods and additional state data you need), then OpenMDAO allows you to organize variables using `<category>:<sub-category>:<etc.>` syntax in the variable names. 

The N2 currently splits all that out into individual variables, but I think it might be a good idea to allow it (at least optionally) to collapse things based on shared namespaces. 

None of that would fix the "all my components take this object as an input" problem though. I don't have a great suggestion for how to deal with this at the ready, but it seems like something that is at least worth considering via POEM. There are a few places were we have things im going to call "port" which are collections of variables that are meant to be passed all as one "object" around. Having an explicit construct in OpenMDAO to say "this is all one port" might be helpful. If any inspiration comes to me, I'll write my own POEM i guess.

PR #2591, opened 2022-08-05
Put `setup()` call before setting solvers in doc page

johnjasa:
### Summary

Simple fix for doc behavior. We say we're using the Newton solver but because we don't call setup() beforehand we don't actually change the solver away from NLBGS. This resolves that.


naylor-b (approved):

johnjasa:
I don't think that metamodel slinear fd test is failing because of this change and I don't think I have the permissions to rerun the test. But let me know if I can help here!

Kenneth-T-Moore:
I think I would classify this as a problem with the test model `DoubleSellar`, which should respect solvers that we add before setup instead of replacing them. DoubleSellar was written to be flexible for use in unit tests, so you can pass in solvers to use or you can just go with the ones that are default. Maybe was goldplated this system beyond what we should have done.

In terms of teaching, I think most of our examples show adding the solvers before setup. Perhaps Bret could confirm, but I *think* most, if not all, solvers are fine to add after setup, but I don't know that we want to present that as the "proper" way like this.  Maybe needs some discussion.  I don't think we should make this decision based on the quirks of the component we use for the demonstration. I'd recommend modifying DoubleSellar, or coding a new onw that doesn't get fancy with default solvers.

Kenneth-T-Moore (changes requested):
See front-page comment.

naylor-b:
`_setup_solvers` isn't called until `final_setup` so setting the solver after `setup` should be fine.

naylor-b:
I didn't see an issue with `DoubleSellar`, although there are multiple versions floating around and I may have missed it, but I did see this issue with `SellarDerivatives`.  We should change the default of the `nonlinear_solver` option to None and only set it to `NonlinearBlockGS` if the option isn't None.

swryan:
I think the `SellarDerivatives` model was just intended to be a configurable model for use in tests.. It was probably not a good choice to use in the docs, since all that configurability via the options is superfluous here.

My thought would be to just make a simpler version of the class...  there is a simple `SellarMDA` model in the `sellar_feature` module, maybe we should put a stripped down version of the `SellarDerivatives` model there and use that
(the `_feature` suffix is an artifact from the old docs, indicating these classes were used in the feature tests, i.e. the docs)

PR #2596, opened 2022-08-09
prevent hang in opt_report when using MPI

naylor-b:
### Summary

The opt_report was returning if MPI.WORLD_COMM != 0 but making collective calls after that, which was causing it to hang.

### Related Issues

- This fixes issue 777 in dymos.

None

### New Dependencies

None


swryan (approved):
should we have a test that exercises the previously blocking scenario?

Kenneth-T-Moore (approved):

naylor-b:
Yeah I'll add a test.

robfalck (approved):

PR #2624, opened 2022-08-29
I2275 show more helpful error message when user calls Problem.list_problem_vars before final_setup

### LINKED ISSUE #2275: Need better error message for when Problem.list_problem_vars is called before final_setup
### Summary of Issue

If `Problem.list_problem_vars` is called before `final_setup` is called, the user gets an error message which is not very helpful. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

Should provide an error message telling user that `final_setup` needs to be called first, either explicitly, or by running `run_driver`. 

Can make use of the `setup_status` value in Problem

```
        if self._metadata['setup_status'] >= _SetupStatus.POST_FINAL_SETUP:
```

### Example
```   
    prob = om.Problem()
    prob.model.add_subsystem('parab', Paraboloid(), promotes_inputs=['x', 'y'])

    # define the component whose output will be constrained
    prob.model.add_subsystem('const', om.ExecComp('g = x + y'), promotes_inputs=['x', 'y'])

    # Design variables 'x' and 'y' span components, so we need to provide a common initial
    # value for them.
    prob.model.set_input_defaults('x', 3.0)
    prob.model.set_input_defaults('y', -4.0)

    # setup the optimization
    prob.driver.options['optimizer'] = 'COBYLA'

    prob.model.add_design_var('x', lower=-50, upper=50)
    prob.model.add_design_var('y', lower=-50, upper=50)
    prob.model.add_objective('parab.f_xy')

    # to add the constraint to the model
    prob.model.add_constraint('const.g', lower=0, upper=10.)

    prob.setup()
    prob.list_problem_vars()
```

### Environment

NA

hschilling:
### Summary

Instead of failing with an unhelpful error message, let the user know what went wrong if they call `Problem.list_problem_vars` before `final_setup` is called.

### Related Issues

- Resolves #2275 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -2020,7 +2020,6 @@ class Problem(object):
             ['ref', 'ref0', 'indices', 'adder', 'scaler', 'units',
             'parallel_deriv_color', 'cache_linear_solution'].
         """
-
         if self._metadata['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
             raise RuntimeError(
                 "list_problem_vars can only be called after final_setup is called, either "
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -2020,6 +2020,12 @@ class Problem(object):
             ['ref', 'ref0', 'indices', 'adder', 'scaler', 'units',
             'parallel_deriv_color', 'cache_linear_solution'].
         """
+
+        if self._metadata['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
+            raise RuntimeError(
+                "list_problem_vars can only be called after final_setup is called, either "
+                "explicitly, or by running run_driver")
+
         default_col_names = ['name', 'val', 'size']
 
         # Design vars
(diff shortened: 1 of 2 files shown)


openmdao/core/problem.py
```diff
@@ -2020,6 +2020,11 @@ def list_problem_vars(self,
             ['ref', 'ref0', 'indices', 'adder', 'scaler', 'units',
             'parallel_deriv_color', 'cache_linear_solution'].
         """
+        if self._metadata['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
+            raise RuntimeError(
+                "list_problem_vars can only be called after final_setup is called, either "
+                "explicitly, or by running run_driver")
```
swryan, line 2026: also `run_model()`... maybe use message format consistent with https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/problem.py#L976

swryan (approved):

naylor-b (approved):

PR #2628, opened 2022-09-02
Implement --dependency_versions to print versions of all installed dependencies

### LINKED ISSUE #2042: Add `--include_dependencies` arg to show versions for debugging
### Summary of Issue

When users submit issues, their dependencies might be out of date which can be the source of an error. 

### Issue Type

- [ ] Bug
- [x] Enhancement  

### Description

To fix this issue, we should add the arg `--include_dependencies` to our CLI so we can tell users to run `openmdao --version --include_dependencies` when submitting a bug report on Github. This would be more specific than `pip freeze`

### Example

`openmdao --version --include_dependencies`

### Environment

N/A

tadkollar:
### Summary

The new `--dependency_versions` command line arg prints the version of OpenMDAO and all of its dependencies, discovered recursively using `importlib.metadata`. This is better than `pip freeze` because packages unrelated to OM are not listed. Some packages have optional dependencies which may not be installed; these are ignored.

### Related Issues

- Resolves #2042 

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/om.py
+++ b/openmdao/utils/om.py
@@ -460,7 +460,7 @@ def _cite_cmd(options, user_args):
 
 def _get_deps(dep_dict: dict, package_name: str) -> None:
     """
-    Recursively determine all installed depenency versions and add newly found ones to dep_dict
+    Recursively determine all installed depenency versions and add newly found ones to dep_dict.
 
     Parameters
     ----------
--- a/openmdao/utils/om.py
+++ b/openmdao/utils/om.py
@@ -5,6 +5,8 @@ A console script wrapper for multiple openmdao functions.
 import sys
 import os
 import argparse
+import importlib.metadata as ilmd
+import re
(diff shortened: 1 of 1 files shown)


swryan (commented):
I'm not really convinced of the value of this vs just pip list..

I tried it in a full dev environment with [all], mpi/petscy and pyoptsparse.. the list was only shorter by 7 dependencies, which were basically those extra but I think relevant things...  
mpi/petsc, pyoptsparse,  jaxlib, etc.

Maybe if a user has an environment that's super loaded up with unrelated stuff, this helps, but seems like it's not good to tell a user to use this if they are reporting an MPI or pyOS related bug?

tadkollar:
Well, it does find dependencies recursively... would it be more useful if it just printed the ones OM depends on directly?

swryan (approved):

robfalck (approved):

PR #2632, opened 2022-09-09
Fixed  to allow a shape mismatch in AutoIVC if all promoted inputs have a size of 1.

### LINKED ISSUE #2631: AutoIVC fails due to shape mismatch.
### Description

When two inputs are promoted in a model and they have different shapes, but are both size=1, a ValueError is raised in the AutoIVC setup code.

For instance, it should be valid to promote an input with shape (1, 1) to the same name as an input of shape (1,), since there is no ambiguity in the storage order of the underlying arrays.

### Example

```python
import unittest
import openmdao.api as om


class TestAutoIVCShapeMismatch(unittest.TestCase):

    def test_shape_mismatch(self):

        p = om.Problem()

        c1 = p.model.add_subsystem('c1', om.ExecComp())
        c2 = p.model.add_subsystem('c2', om.ExecComp())

        c1.add_expr('a = 5.0 * x1', a=dict(shape=(1,)), x1=dict(shape=(1,)))
        c2.add_expr('b = 5.0 * x2', b=dict(shape=(1,)), x2=dict(shape=(1, 1)))

        p.model.promotes('c1', inputs=[('x1', 'x')])
        p.model.promotes('c2', inputs=[('x2', 'x')])

        p.setup()

        p.set_val('x', 4.0)

        p.run_model()
```

### OpenMDAO Version

3.20.1-dev

N/A

robfalck:
### Summary

If multiple inputs have different shapes, are promoted to the same name, and are connected to AutoIVC, OpenMDAO will now allow this if they all have a size of 1.

### Related Issues

- Resolves #2631 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -3371,12 +3371,13 @@ class Group(System):
                     if val.size > 1:
                         raise ValueError(f"Shape of input '{tgt}', (), doesn't match shape "
                                          f"{val.shape}.")
-                elif val.shape != value.shape and val.size > 1 and value.size > 1:
+                elif np.squeeze(val).shape != np.squeeze(value).shape:
                     raise ValueError(f"Shape of input '{tgt}', {value.shape}, doesn't match shape "
                                      f"{val.shape}.")
+
                 if val is not value:
                     if val.shape:
-                        val[:] = value
+                        val = np.reshape(value, newshape=val.shape)
                     else:
                         val = value
 
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -3371,7 +3371,7 @@ class Group(System):
                     if val.size > 1:
                         raise ValueError(f"Shape of input '{tgt}', (), doesn't match shape "
                                          f"{val.shape}.")
-                elif val.shape != value.shape:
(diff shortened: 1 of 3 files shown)


openmdao/core/group.py
```diff
@@ -3371,7 +3371,7 @@ def _get_auto_ivc_out_val(self, tgts, vars_to_gather, abs2meta_in):
                     if val.size > 1:
                         raise ValueError(f"Shape of input '{tgt}', (), doesn't match shape "
                                          f"{val.shape}.")
-                elif val.shape != value.shape:
+                elif val.shape != value.shape and val.size > 1 and value.size > 1:
                     raise ValueError(f"Shape of input '{tgt}', {value.shape}, doesn't match shape "
                                      f"{val.shape}.")
                 if val is not value:
```
Kenneth-T-Moore, line 3377: Instead of doing this, what would you think about checking the "squeezed" shape (np.squeeze), so that inputs with shapes (1, 3, 1) and (3, ) could also be connected together?

robfalck: Good idea, I'll make that change.

Kenneth-T-Moore (approved):

PR #2648, opened 2022-09-23
POEM 073 draft implementation

tuckerbabcock:
### Summary

Add implementation to `DOEDriver` that will compute and record total derivatives if the recording option `record_derivaties` is set to `True`.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/doe_driver.py
+++ b/openmdao/drivers/doe_driver.py
@@ -35,6 +35,10 @@ class DOEDriver(Driver):
         The MPI communicator for the Problem.
     _color : int or None
         In MPI, the cached color is used to determine which cases to run on this proc.
+    _indep_list : list
+        List of design variables.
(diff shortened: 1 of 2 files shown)


swryan:
Could you maybe make a test case out of the example in the POEM? (minus plot, but asserting the derivs are recorded)

tuckerbabcock:
@swryan I've added that test at the end of the `TestDOEDriver` class, and another one that asserts that the derivs are not recorded when the `record_derivatives` option is `False`.

swryan:
> @swryan I've added that test at the end of the `TestDOEDriver` class, and another one that asserts that the derivs are not recorded when the `record_derivatives` option is `False`.

perfect.. tanks!

swryan (approved):

robfalck (approved):

PR #2677, opened 2022-10-19
move filterwarnings into catch_warnings context

### LINKED ISSUE #2676: Warnings treated as errors when running test suite
### Task Description

There is an intermittent issue when running the OpenMDAO test suite in which a large number of tests fail due to warnings being raised as errors.  This appears to be due to a test in test_warnings.py that calls `warnings.filterwarnings('error', category=om.OpenMDAOWarning)` outside of a `catch_warnings()` context manager.

see: https://github.com/OpenMDAO/OpenMDAO/actions/runs/3273935283/jobs/5386793177

swryan:
### Summary

Move `filterwarnings` call into `catch_warnings` context so it doesn't persist outside of test.

### Related Issues

- Resolves #2676

None

### New Dependencies

None

### DIFF
--- a/openmdao/test_suite/tests/test_warnings.py
+++ b/openmdao/test_suite/tests/test_warnings.py
@@ -95,9 +95,9 @@ class TestWarnings(unittest.TestCase):
         p.model.connect('a_comp.z', 'exec_comp.z')
         p.driver.declare_coloring()
 
-        warnings.filterwarnings('ignore', category=om.UnitsWarning)
-
         with warnings.catch_warnings(record=True) as w:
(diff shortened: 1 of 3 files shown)


naylor-b (approved):

robfalck (approved):

Kenneth-T-Moore (approved):

swryan:
Note:  This seems to have fixed the issue, but there is this caveat in the docs:

> Note: this can only be guaranteed in a single-threaded application. If two or more threads use the [catch_warnings](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager at the same time, the behavior is undefined.

We'll see how it goes...

PR #2683, opened 2022-10-23
Allow `indep_var` tag to work with inputs report for backward compatibility.

### LINKED ISSUE #2678: Non backward-compatibility of PR #2655
### Description

Better safe than sorry, I would want to address here a problem with the development version of OpenMDAO, though I don't know if it is planned to address them already or not.

It happens that PR #2655 does 2 modifications that would slightly break the backward compatibility of the next OpenMDAO version:
  - The `OptionsDictionary.to_table()` method does not accept anymore `"grid"` as `fmt` argument.
  - The variables in an IndepVarComp are now tagged with `"openmdao:indep_var"` instead of `"indep_var"`

I understand the rationale behind these changes, and that's indeed a good thing to put consistency in your tag names, but I simply would want the OpenMDAO team to be well aware that even slight modifications like these will break the backward-compatibility we would want to rely on.

There had been a very well done announcement before the "`value` to `val`" modification, and I can only hope that any new non-backward compatible change will be advertised the same in the future.

### Example

```python
# Not sure the example is worth providing, given the addressed issue :), but here it is:
# this example works with OpenMDAO 3.21.0 but not with current dev version
import openmdao.api as om

problem = om.Problem(om.IndepVarComp("my_var"))

problem.setup()
assert list(problem.model.get_io_metadata(tags="indep_var").keys()) == ["my_var"]
```

### OpenMDAO Version

3.21.1-dev

Tested on revision 4673d0b4356a716403a1826365050ba5ce66f36a

### LINKED ISSUE #2682: Vulnerability in Pillow, required by matplotlib, is blocking CI.
### Task Description

pip-audit is reporting a new vulnerability OSV-2022-715 in Pillow, which is required by matplotlib.

Information on the vulnerability is [here](https://osv.dev/vulnerability/OSV-2022-715).

It appears that this vulnerability has been present for a while and is not particularly severe, so for now I think we want to allow this one to be skipped.  It appears that it will be fixed in a forthcoming version of Pillow.

robfalck:
### Summary

We changed the tag `indep_var` to `openmdao:indep_var`, and while this wasn't being used by anything prior to the creation of the inputs report, the fact that some users may have had components which used this tag will cause incorrect behavior in the inputs report.

This tag is now recognized and generates an `OMDeprecationWarning` during the creation of the inputs report.

Tests were added to test the basic behavior of the inputs report and the behavior of this warning.

This PR also temporarily allows pip-audit to pass on CI given the vulnerability described in #2682 

### Related Issues

- Resolves #2682 
- Resolves #2678 

OMDeprecationWarning is now raised if a user builds an inputs_report and an input is found with the `'indep_var'` tag.

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/inputs_report/tests/test_inputs_report.py
+++ b/openmdao/visualization/inputs_report/tests/test_inputs_report.py
@@ -1,5 +1,6 @@
 """Test the Newton nonlinear solver. """
 import unittest
+import tempfile
 
 import openmdao.api as om
 from openmdao.test_suite.components.double_sellar import DoubleSellar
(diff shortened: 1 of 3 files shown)


openmdao/visualization/inputs_report/tests/test_inputs_report.py
```diff
@@ -0,0 +1,105 @@
+"""Test the Newton nonlinear solver. """
```
swryan, line 1: incorrect module doc string

swryan (approved):

naylor-b (approved):

PR #2689, opened 2022-10-26
Make pyDOE2 an optional dependency.

### LINKED ISSUE #2688: Do not require pyDOE2 as a dependency.

pyDOE2 is a required dependency in `setup.py` but is only used in the DOE driver and the genetic algorithm driver.
We should move pyDOE2 into the optional dependencies and alert users that it needs to be installed when they attempt to use a feature of OpenMDAO that requires it.

### Associated POEM

N/A

swryan:
### Summary

Changed __pyDOE2__ into an optional dependency.  

Some of the DOE generators will not be available unless it is installed.

__SimpleGADriver__ and __DifferentialEvolutionDriver__ depend on the Latin hypercube sampling ([lhs](https://github.com/clicumu/pyDOE2/blob/master/pyDOE2/doe_lhs.py)) module from __pyDOE2__, so they will also not be available unless it is installed.

#### Details

-  __pyDOE2__ is now installed only if __OpenMDAO__ is installed with `pip install openmda[doe]` or `pip install openmda[all]`
- updated the doe generators that require __pyDOE2__ to raise a user friendly error if it is not installed
- updated __SimpleGADriver__ and __DifferentialEvolutionDriver__ to raise a user friendly error if  __pyDOE2__ is not installed
- updated/added tests

Also:
- updated the __pydocstyle__ dependency to use the latest version, since 2.0.0 is not compatible with newer versions of Python

### Related Issues

- Resolves #2688

None

### New Dependencies

None


openmdao/code_review/test_lint_peps.py
```diff
@@ -122,4 +122,4 @@ def test_pep257(self):
 
 
 if __name__ == "__main__":
-    unittest.main()
\ No newline at end of file
+    unittest.main()
```
robfalck, line 125: Is this a whitespace change?  Doesn't really matter just curious

robfalck (approved):

openmdao/code_review/test_lint_peps.py
swryan: yeah.. missing newline at end of file..  VS Code autocorrected it...

naylor-b (approved):

Kenneth-T-Moore (approved):

Kenneth-T-Moore:
If we are interested in making more dependencies optional, we could do the same thing with pyparsing, which is only used in the filewrap utils.

PR #2709, opened 2022-11-09
Implementation PR for POEM 69

### LINKED ISSUE #2692: Implementation for POEM 69

This allows residual names to be specified for implicit components rather than the current practice of forcing residual names to match output names.

### Associated POEM

69

naylor-b:
### Summary

Adds the 'add_residual' method to ImplicitComponent.  Residual renaming is active only inside of the ImplicitComponent. Elsewhere in the framework, behavior is as before, where residual and output names are the same.

### Related Issues

- Resolves #2692

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/implicitcomponent.py
+++ b/openmdao/core/implicitcomponent.py
@@ -538,21 +538,41 @@ class ImplicitComponent(Component):
             outs_w_partial_res_overlap = defaultdict(list)
             for lst in self._resid2out_subjac_map.values():
                 for oname, wrt, outslc, resslc, dct in lst:
+                    newdct = dct
                     if 'val' in dct and dct['val'] is not None:
-                        val = dct['val']
-                        if isinstance(val, np.ndarray):
-                            dct = dct.copy()
-                            val = val[resslc]
-                            if outslc is _full_slice:
-                                dct['val'] = val
-                            else:
-                                # resid subjac is only a subset of the output subjac
-                                outs_w_partial_res_overlap[oname].append((lst, val))
-                                continue
-                    super()._declare_partials([oname], wrt, dct)
+                        if outslc is not _full_slice:
+                            # resid subjac is only a subset of the output subjac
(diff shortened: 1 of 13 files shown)


openmdao/core/implicitcomponent.py
```diff
@@ -317,6 +356,258 @@ def add_output(self, name, val=1.0, **kwargs):
...
+            Name of the residual in this component's namespace.
+        shape : int or tuple
+            Shape of this residual.
+        units : str or None
+            Units in which this residual will be given to the user when requested.
+            Default is None, which means it has no units.
+        desc : str
+            Description of the residual.
+        ref : float or ndarray
+            Scaling parameter. The value in the user-defined units of this residual
```
Kenneth-T-Moore, line 378: Is this 'ref' the same as 'res_ref' on add_output? What happens if the user declares a 'res_ref' on one of the outputs, and then a different 'ref' here?

openmdao/core/implicitcomponent.py
```diff
@@ -317,6 +356,258 @@ def add_output(self, name, val=1.0, **kwargs):
...
+        the outputs vector for this component.
+
+        Parameters
+        ----------
+        name : str
+            Name of the residual in this component's namespace.
+        shape : int or tuple
+            Shape of this residual.
+        units : str or None
+            Units in which this residual will be given to the user when requested.
```
Kenneth-T-Moore, line 373: Same with 'units' -- what if there is a conflict between units declared in the output and the residual?

Kenneth-T-Moore (changes requested):

openmdao/core/implicitcomponent.py
naylor-b: I added checks for ref/res_ref and units/res_units for residuals and outputs that overlap.

Kenneth-T-Moore (approved):

robfalck (approved):

PR #2715, opened 2022-11-16
change default for Driver supports['optimization'] to False

### LINKED ISSUE #2714: Change default for Driver option supports['optimization'] to False

The `supports['optimization']` option currently  [defaults](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/driver.py#L163) to `True` in the Driver base class.

This is actually incorrect, as the base class itself does not support optimization.

The default should be changed to `False` and Driver sub-classes should set it appropriately.

This is technically a backward incompatible change for outside developers that may have developed their own drivers.  However, this option was only added 4 months ago, specifically to support the optimization report, so the impact should only be that non-OpenMDAO drivers that rely on the default would not auto-generate an Optimization report unless updated to set the option to `True`.

### Associated POEM

N/A

swryan:
### Summary

The `supports['optimization']` option previously [defaulted](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/driver.py#L163) to True in the Driver base class.

Since the base Driver class does not in fact support optimization, the default has been changed to `False` and Driver sub-classes have been updated set the option as appropriate rather than relying on the default.

### Related Issues

- Resolves #2714

For outside developers that may have developed their own drivers, the optimization report will not be automatically generated unless the `supports['optimization']` option is set to `True`.  Previously this was the default.

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/opt_report/tests/test_opt_report.py
+++ b/openmdao/visualization/opt_report/tests/test_opt_report.py
@@ -163,8 +163,14 @@ class TestOptimizationReport(unittest.TestCase):
                                           )
         expect = {'obj_calls': 0, 'deriv_calls': 0}
         self.check_opt_result(expected=expect)
-        opt_report(self.prob)
-        self.check_opt_report(expected=expect)
+
+        expected_warning_msg = "The optimizer report is not applicable for the Driver Driver " \
+                               "which does not support optimization"
+        with assert_warning(DriverWarning, expected_warning_msg):
+            opt_report(self.prob)
+
+        outfilepath = str(pathlib.Path(self.prob.get_reports_dir()).joinpath(_default_optimizer_report_filename))
+        self.assertFalse(os.path.exists(outfilepath))
 
     def test_opt_report_scipyopt_SLSQP(self):
         self.setup_problem_and_run_driver(om.ScipyOptimizeDriver,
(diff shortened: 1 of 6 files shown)


naylor-b (approved):

openmdao/visualization/opt_report/tests/test_opt_report.py
```diff
@@ -163,8 +163,14 @@ def test_opt_report_run_once_driver(self):
                                           )
         expect = {'obj_calls': 0, 'deriv_calls': 0}
         self.check_opt_result(expected=expect)
-        opt_report(self.prob)
-        self.check_opt_report(expected=expect)
+
+        expected_warning_msg = "The optimizer report is not applicable for the Driver Driver " \
```
Kenneth-T-Moore, line 167: Tangential issue, but we probably want to put the driver name in single quotes in this error message, so that it looks like `for the Driver 'Driver'`.

Kenneth-T-Moore (approved):

openmdao/visualization/opt_report/tests/test_opt_report.py
swryan: Sorry meant to make this fix before merging... adding it to #2716

PR #2716, opened 2022-11-16
Deprecate Component as a model, handle gracefully

### LINKED ISSUE #2684: AutoIVC incorrect behavior if system is a single component. Various other issues when `Problem.model` is a Component.
### Description

A user reported that the AutoIVC feature doesn't work correctly when the system is a single component, as there is no Group in which to put the automatically created IndepVarComp.

While testing this, I uncovered a few other issues with problems in which the model was a single system.
Our reliance on AutoIVC has me tempted to say that we should just deprecate this behavior and in the future require that `Problem.model` be a Group, or at least be derived from Group.

Some of these cases are captured below.  Others include the N2 being wonky (but still generated as a report) in this situation.

### Example

```python
import unittest

import openmdao.api as om


class TestComponentAsSystem(unittest.TestCase):
    """
    This code shows various issues we encounter when the System _is_ a component.
    """
    def test_component_as_system(self):

        prob = om.Problem()

        prob.model = om.ExecComp('z = x + y')

        prob.setup()

        prob.set_val('x', 2)
        prob.set_val('y', 5)

        prob.run_model()

        assert_near_equal(7.0, prob.get_val('z'))

    def test_promotes_component_as_system(self):
        """
        Promotes error because theres nowhere to promote to, but that isn't necessarily clear.

        The error is

        RuntimeError: <model> <class ExecComp>: arg 'promotes_inputs' in call to ExecComp() does
        not refer to any variable in the expressions ['z = (x - 2)**2']

        Clearly that's not the case.
        """
        prob = om.Problem()

        prob.model = om.ExecComp('z = (x - 2)**2', promotes_inputs=['x'], promotes_outputs=['z'])

        prob.setup()

        prob.set_val('x', 5)

        prob.run_model()

        assert_near_equal(9.0, prob.get_val('z'))

    def test_optimize_component_as_system(self):
        """
        Promotes error because theres nowhere to promote to, but that isn't necessarily clear.

        Errors with 'Output not found for design variable 'x'.
        """
        prob = om.Problem()

        prob.model = om.ExecComp('z = (x - 2)**2')

        prob.model.add_design_var('x')
        prob.model.add_objective('z')

        prob.setup()

        prob.set_val('x', 5)

        prob.run_driver()

        assert_near_equal(2.0, prob.get_val('x'))
```

### OpenMDAO Version

3.21.1-dev

_No response_

swryan:
### Summary

- issue a deprecation warning if Problem setup() is called with a model that is not a Group
- improve error message when model is not a Group and Problem expects AutoIVCs
- improve error message when user accidentally uses 'promotes' arg in ExecComp arg list
- add tests for component as a model, consolidated existing tests into common test module
- update N2 logic to handle component as a model in more user friendly way, generate html file with message

Some of this is uglier than I would like.. open to suggestions.

### Related Issues

- Resolves #2684 

None

### New Dependencies

None


Kenneth-T-Moore (approved):

openmdao/recorders/recording_manager.py
```diff
@@ -190,7 +192,12 @@ def record_viewer_data(problem):
...
     if recorders:
         from openmdao.visualization.n2_viewer.n2_viewer import _get_viewer_data
-        viewer_data = _get_viewer_data(problem)
+        try:
+            viewer_data = _get_viewer_data(problem)
+        except TypeError as err:
+            viewer_data = {}
+            issue_warning(str(err))
+
         viewer_data['md5_hash'] = problem.model._generate_md5_hash()
```
naylor-b, line 201: If a TypeError is caught above, won't this just raise a KeyError?

swryan: Not sure I understand your thought here..
If there is a TypeError, the dict is initialized to empty and the 'md5_hash' key is added to that empty dict
(There is a test for this case btw: https://github.com/swryan/OpenMDAO/blob/2684_model_not_group/openmdao/recorders/tests/test_sqlite_reader.py#L1298)

naylor-b (approved):

PR #2774, opened 2023-01-13
Fixed KeyError when total coloring done on model with an unconnected input constraint in a ParallelGroup

### LINKED ISSUE #2740: Bug with total coloring involving AutoIVC and MPI
### Description

The example below requires [this dymos branch](https://github.com/robfalck/dymos/tree/mpi_autoivc_bug), and works correctly in serial operation.

When run under MPI, it fails with the end of the traceback being

```
  File ".../openmdao/core/total_jac.py", line 369, in __init__
    self.sol2jac_map[mode] = self._get_sol2jac_map(self.output_list[mode],
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../openmdao/core/total_jac.py", line 849, in _get_sol2jac_map
    local_idx, sizes_idx, _ = self._dist_driver_vars[name]
                              ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: '_auto_ivc.v1'
    local_idx, sizes_idx, _ = self._dist_driver_vars[name]
                              ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: '_auto_ivc.v1'
```

Where `self._dist_driver_vars` has keys that contain the user-friendly input names, but `name` is the true AutoIVC output name. OpenMDAO needs to map that name to the appropriate user-friendly input name in the MPI-relevant branch of logic in `_get_sol2jac_map`.

This particular branch of Dymos was an attempt to allow control inputs to be fed into Phases, instead of being outputs of explicit IndepVarComps within the phases, hence the reason this hasn't been found before.

### Example

```python
from openmdao.utils.assert_utils import assert_near_equal

from dymos.examples.finite_burn_orbit_raise.finite_burn_orbit_raise_problem import two_burn_orbit_raise_problem


p = two_burn_orbit_raise_problem(transcription='gauss-lobatto', transcription_order=3,
                                 compressed=False, optimizer='IPOPT', simulate=False,
                                 show_output=False)

if p.model.traj.phases.burn2 in p.model.traj.phases._subsystems_myproc:
    assert_near_equal(p.get_val('traj.burn2.states:deltav')[-1], 0.3995,
                      tolerance=2.0E-3)
```

### OpenMDAO Version

3.22.1-dev

_No response_

naylor-b:
### Related Issues

- Resolves #2740

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/pyoptsparse_driver.py
+++ b/openmdao/drivers/pyoptsparse_driver.py
@@ -433,6 +433,9 @@ class pyOptSparseDriver(Driver):
                 rels = relevant[path]
                 wrt = [v for v in indep_list if dv_meta[v]['source'] in rels]
 
+            if not wrt:
+                continue
+
             if meta['linear']:
                 jac = {w: _lin_jacs[name][w] for w in wrt}
                 opt_prob.addConGroup(name, size,
@@ -466,6 +469,9 @@ class pyOptSparseDriver(Driver):
             else:
                 rels = relevant[path]
                 wrt = [v for v in indep_list if dv_meta[v]['source'] in rels]
+
+            if not wrt:
+                continue
 
             if meta['linear']:
                 jac = {w: _lin_jacs[name][w] for w in wrt}
@@ -846,12 +852,11 @@ class pyOptSparseDriver(Driver):
         if total_sparsity is None:
             return
 
-        model = self._problem().model
         for res, resdict in total_sparsity.items():
-            if res in self._responses and self._responses[res]['alias'] is not None:
-                res = self._responses[res]['source']
             if res in self._objs:  # skip objectives
                 continue
+            if res in self._responses and self._responses[res]['alias'] is not None:
+                res = self._responses[res]['source']
             self._res_subjacs[res] = {}
             for dv, (rows, cols, shape) in resdict.items():
                 rows = np.array(rows, dtype=INT_DTYPE)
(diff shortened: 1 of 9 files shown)


robfalck (approved):

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -467,6 +470,9 @@ def run(self):
                 rels = relevant[path]
                 wrt = [v for v in indep_list if dv_meta[v]['source'] in rels]
 
+            if not wrt:
```
Kenneth-T-Moore, line 473: So, this is a constraint that is not dependent on any design variables, and we just skip it. Seems like we should at least warn about it, since it might mean there is a problem in the model.

Kenneth-T-Moore (changes requested):
Just a question about whether we should warn if a constraint doesn't depend on any desvar due to graph-based sparsity.  Although, I seem to recall that the scaling report also raises this warning when it runs, so maybe this would be redundant.

openmdao/drivers/pyoptsparse_driver.py
naylor-b: added warning

Kenneth-T-Moore: I think the message should be more explicit, like "... does not depend on any design variables, and was not added to the optimization."

naylor-b: added

Kenneth-T-Moore (approved):

PR #2778, opened 2023-01-17
N2: Display connections when in NodeInfo mode

### LINKED ISSUE #2720: Remove generic N2 absPathName references
### Description

There are two references to `TreeNode.absPathName` in `Matrix.js` that need to be updated to `TreeNode.path`.

### Example

```python
Matrix.js:139:        const debugStr = `${node.absPathName}(${nodeId})`
Matrix.js:627:                        'label': offscreenNode.absPathName,
```

### OpenMDAO Version

3.21.1-dev

_No response_

### LINKED ISSUE #2725: Allow user to see N2 connections when not zoomed in.

When viewing a large N2, it's common to see group connections "clustered" into large gray boxes.

Zoomed out N2 showing connections

Currently, the user is required to zoom in until the variables are visible to see what the connected variables are.

Instead, using the "detailed node information tool" to hover or click on one of the off-diagonal gray boxes should show a table of all variables involved in the connection.

Minimally, the names of the variables involved should be shown.
If possible, we should show a two-column table with src in one column and targets in the second column.  If there are multiple targets, each should be listed on a separate line in its column.
Names should be shown as model-level promoted paths.

| Source           | Target(s)        |
| ------------- | ------------- |
| G1.x                | G2.c1.x           |
| G1.y                | G2.c1.foo        |
|                        | G2.c1.bar        |

### Possible issues

It's possible for there to be many connections between two systems.  Can we have the table pop up in a scrollable window?

### Associated POEM

N/A

tadkollar:
### Summary

In the N2 diagram with NodeInfo mode active, hovering over an off-diagonal connection square will now list all connections between the children of the two nodes. When just hovering, the list may be long enough to go off screen, but clicking and making the window persistent will reduce the size and add a scrollbar. The list can then be searched with the browser's built-in Find function.

### Related Issues

- Resolves #2725 
- Resolves #2720 

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/gen/NodeInfo.js
+++ b/openmdao/visualization/n2_viewer/gen/NodeInfo.js
@@ -246,7 +246,10 @@ class PersistentNodeInfo extends WindowResizable {
     }
 }
 
-
+/**
+ * Display all connections represented by the current off-diagonal cell.
+ * @typedef NodeConnectionInfo
+ */
 class NodeConnectionInfo extends NodeInfo {
     /**
      * Build a list of the properties we care about and set up
@@ -257,6 +260,12 @@ class NodeConnectionInfo extends NodeInfo {
         this.propList = [];
     }
 
+    /**
+     * Gather all connections between the two cells that this one represents.
+     * @param {Array} connList The connections discovered so far.
+     * @param {TreeNode} srcNode The source node to search.
+     * @param {TreeNode} tgtNode The target node to search.
+     */
     _getConnections(connList, srcNode, tgtNode) {
         if (srcNode.hasChildren()) {
             for (const child of srcNode.children) {
@@ -273,7 +282,10 @@ class NodeConnectionInfo extends NodeInfo {
         }
     }
 
-    /** Create a new persistent window by copying our contents */
+    /**
+     * Create a new persistent window by copying our contents. Set the
+     * max initial height of the window at 300px.
+     */
     pin() {
         if (this.tbody.html() == '') return; // Was already pinned so is empty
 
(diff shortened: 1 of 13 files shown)


openmdao/visualization/n2_viewer/gen/NodeInfo.js
```diff
@@ -230,3 +245,86 @@ class PersistentNodeInfo extends WindowDraggable {
...
+            newRow.append('td')
+                .attr('scope', 'row')
+                .text(conn.src);
+
+            newRow.append('td').html(' &#x2501;&#x2501;&#x2501;&#x25ba; ');
+
+            newRow.append('td').text(conn.tgt);
+        }
+
+        this.title(`Connections from ${cell.srcObj.path} to ${cell.tgtObj.path}`)
```
robfalck, line 325: Is there a way to tell when srcObj and tgtObj are collapsed or not?
When things are collapsed this displays the source system and the target system.
Connection info when collapsed

But when the connection is not collapsed, this expands all the way out to the variable name:
Connection info when not collapsed

If it could always just show `Connections from {system} to {system}` that would be preferred.

robfalck: Also, since connections are made with promoted names, I think the variable names in each row should be promoted names instead of absolute.

Really enjoying using this!

robfalck (changes requested):
Overall I'm really happy with the change, just a few minor issues with the info displayed in the window.

robfalck (approved):

swryan (approved):

PR #2802, opened 2023-02-06
removed broken checking for dynamic overrides of matrix free methods

### LINKED ISSUE #2735: Bug in detection of 'matrix_free' behavior
### Description

Automatic detection of 'matrix_free' behavior can break if compute_jacvec_product or apply_linear are wrapped with a decorator.

### Example

```python
Decorating either of these methods will cause the `matrix_free` attribute on the component to be set to True, even when the component is not matrix free.  In general this won't break the component, but will cause unnecessary computations and could degrade component performance.
```

### OpenMDAO Version

3.21.1-dev

_No response_

naylor-b:
### Summary

Checking for dynamic overrides of matrix free methods was broken in cases where decorators were applied to those methods.  That checking was removed, so now only 'static' checking is done, i.e., if the class overrides the method we will automatically detect it as before, but if the method is added dynamically, it's now the responsibility of the component author to set the `matrix_free` attribute for that component to True manually because the framework will not attempt to detect it.

The same also applies to the `_has_solve_nl` flag in ImplicitComponent which suffered from the same problem as the `matrix_free` flag.

### Related Issues

- Resolves #2735

In certain cases where `compute_jacvec_product`, `apply_linear`, or `solve_nonlinear` methods were added dynamically to a component, it's possible that the framework used to detect them correctly (in the absence of decorators) but now will not.  The fix for this is to set the component's `matrix_free` attribute to True manually.

### New Dependencies

None

### DIFF
--- a/openmdao/components/implicit_func_comp.py
+++ b/openmdao/components/implicit_func_comp.py
@@ -82,6 +82,7 @@ class ImplicitFuncComp(ImplicitComponent):
 
         if solve_nonlinear:
             self.solve_nonlinear = self._user_solve_nonlinear
+            self._has_solve_nl = True
         if linearize:
             self.linearize = self._user_linearize
         if solve_linear:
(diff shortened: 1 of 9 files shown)


swryan (commented):
Would this be considered an API change, and does itrequire documentation?

Kenneth-T-Moore (approved):

naylor-b:
OK I updated the docstring for System (where matrix_free is documented) and also added a little to the docstring for `overrides_method`, which we use to detect the 'typical' case.

swryan (approved):

PR #2806, opened 2023-02-08
Add GitHub workflow to test with latest/pre-release dependencies

### LINKED ISSUE #2772: Add a non-blocking workflow to look for deprecations in dependencies.
### Task Description

The test workflow used on github for CI has a matrix entry to test against the latest versions of dependencies.
When dependencies introduce a breaking change and OpenMDAO tests do not pass, this holds up our development process or forces us to merge changes despite the failures.
A big reason for this is that we're not actively tracking deprecations that we need to handle, such as was the case for numpy 1.24 where a few deprecations were removed and resulted in a break.
A change commonly used package like numpy can take some time to ripple through the python ecosystem.

We need to add tests to catch these pending deprecations so we can keep an eye on them and prioritize handling them.
1. Add a new action that is run outside of our standard test workflow, similar to the security vulnerabilities scan. This should test against the latest beta or release candidates of dependencies like numpy, scipy, networkx, matplotlib, pyoptsparse, etc.
2. Add the ability to easily see deprecations that we need to handle in our development. Can we generate a log of all DeprecationWarnings encountered during a testflo run?
3. Failures in these tests aren't something that should necessarily block our development - we're running using pre-release versions of dependencies that may have broken behaviors - but it should give us a heads up on upcoming deprecations we need to handle.

swryan:
### Summary

Add a GitHub workflow that tests against latest/pre-release versions of dependencies.  This new workflow is scheduled to run daily.

Also:
- uses a new feature of `testflo` to generate a deprecations report in the Latest workflow
- updated `ExternalCodeComp` to resolve `distutils` deprecation
- cleaned up the dependencies in `setup.py` to remove pins to outdated versions of `jupyter-book` and related packages
- updated api translation docs to use current method of creating tabbed panes
- fixed some broken links in the docs
- updated baseline versions in test workflow to use Python 3.11
- rescheduled the audit workflow to run an hour later

### Related Issues

- Resolves #2772 

None

### New Dependencies

None

### DIFF
--- a/.github/workflows/openmdao_audit.yml
+++ b/.github/workflows/openmdao_audit.yml
@@ -3,9 +3,9 @@ name: OpenMDAO Audit
 
 on:
 
-  # Run the workflow daily a 0200 UTC
+  # Run the workflow daily a 0400 UTC
   schedule:
-    - cron: '0 2 * * *'
+    - cron: '0 3 * * *'
 
   # Allow running the workflow manually from the Actions tab
   workflow_dispatch:
--- a/.github/workflows/openmdao_audit.yml
+++ b/.github/workflows/openmdao_audit.yml
@@ -3,7 +3,7 @@ name: OpenMDAO Audit
 
 on:
 
-  # Run the workflow daily a 0400 UTC
+  # Run the workflow daily a 0300 UTC
   schedule:
     - cron: '0 3 * * *'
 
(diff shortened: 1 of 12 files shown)


robfalck (approved):

.github/workflows/openmdao_audit.yml
```diff
@@ -3,9 +3,9 @@ name: OpenMDAO Audit
 
 on:
 
-  # Run the workflow daily a 0200 UTC
+  # Run the workflow daily a 0300 UTC
```
Kenneth-T-Moore, line 6: a -> at

robfalck (approved):

PR #2811, opened 2023-02-09
Change DOEDriver to record unscaled derivatives

tuckerbabcock:
### Summary

This PR changes the behavior of the DOEDriver when recording derivatives. As is, the DOEDriver records derivatives with driver scaling applied to them. When plotting the derivatives against the recorded function values, it appears as if the derivative values are incorrect since the function values are not scaled. This changes the behavior to record unscaled derivatives so that it is consistent with the unscaled function outputs. I've added a test that ensures that the recorded derivatives are the same regardless of driver scaling.

### Related Issues

None

Changes DOEDriver to record unscaled derivatives instead of scaled derivatives.

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/tests/test_doe_driver.py
+++ b/openmdao/drivers/tests/test_doe_driver.py
@@ -1500,6 +1500,42 @@ class TestDOEDriver(unittest.TestCase):
             for dv in ('x', 'y'):
                 self.assertEqual(derivs['f_xy', dv], expected_deriv['f_xy', dv])
 
+    def test_derivative_scaled_recording(self):
+        prob = om.Problem()
+        model = prob.model
+
+        model.add_subsystem('comp', Paraboloid(), promotes=['x', 'y', 'f_xy'])
+        model.set_input_defaults('x', 0.0)
+        model.set_input_defaults('y', 0.0)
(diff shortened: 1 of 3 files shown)


Kenneth-T-Moore (approved):

openmdao/drivers/tests/test_doe_driver.py
```diff
@@ -1500,6 +1500,42 @@ def test_derivative_recording(self):
             for dv in ('x', 'y'):
                 self.assertEqual(derivs['f_xy', dv], expected_deriv['f_xy', dv])
 
+    def test_derivative_scaled_recording(self):
```
swryan, line 1503: Need the following decorator here, since this test requires `pyDOE2`  which is not always installed:
```
    @unittest.skipIf(pyDOE2, "only runs if 'pyDOE2' is not installed")
```

tuckerbabcock: Thanks for pointing that out. Should I also have added that to the other tests I added last time, `test_derivative_recording` and `test_derivative_no_recording`? They also use the `DOEDriver`, but I didn't put that decorator there.

tuckerbabcock: I see. Thank you!

swryan (approved):

PR #2812, opened 2023-02-14
made second line in out_buffer be an f string

### LINKED ISSUE #2749: _assemble_derivative_data() shows FD but should be CS
### Description

_assemble_derivative_data() reports Finite Difference and Complex Step derivatives as:

```
Raw Analytic Derivative (Jfor)
[[3.66286949e-05 0.00000000e+00]]

Raw FD Derivative (Jfd)
[[3.66286949e-05 0.00000000e+00]] 
```

However, even when using CS properly, text displayed is "Raw FD Derivative".
This can be confusing to users because it brings into question if they properly executed the CS derivative check.

### Example

```python
prob.set_val(Aircraft.Engine.SCALED_SLS_THRUST, [28000, 24000], units='lbf')
prob.run_model()
prob.check_partials(compact_print=False, method="cs")
```
```


### OpenMDAO Version

commit 96846427100f14f0b5d33d5e4afef4b97df91a47

_No response_

nsteffen:
### Related Issues

- Resolves #2749 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -2685,10 +2685,10 @@ def _assemble_derivative_data(derivative_data, rel_error_tol, abs_error_tol, out
 
                     if directional:
                         out_buffer.write(f"    Directional {fd_opts['method'].upper()}"
-                                         " Derivative (J{fd_opts['method']})\n{fd}\n")
+                                         f" Derivative (J{fd_opts['method']})\n{fd}\n")
                     else:
                         out_buffer.write(f"    Raw {fd_opts['method'].upper()}"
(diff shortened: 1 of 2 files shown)


naylor-b (approved):

Kenneth-T-Moore (approved):
Approve to fix the feature regression, but the fact that this made it through means our testing is inadequate. I think a good follow-on PR would be to add a test that checks the fd value in the long form output. There are already some tests that check other things in that text, so you an piggyback on one of those.

Kenneth-T-Moore (approved):

PR #2813, opened 2023-02-17
POEM 78: Add is_indep_var and is_design_var filters to list_inputs/list_outputs

### LINKED ISSUE #2804: Implement POEM 78

System.list_inputs should be updated so that only inputs the user is responsible for setting will be displayed.  See POEM 78 for details.

### Associated POEM

78

swryan:
### Summary

[POEM78 ](https://github.com/OpenMDAO/POEMs/pull/158) Implementation.  Adds `is_indep_var` and `is_design_var` filters to the `list_inputs()` and `list_outputs()` methods on `System` and `Case`.  This gives users a way of knowing which inputs they are ultimately responsible for setting. 

### Related Issues

- Resolves #2804

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -4068,7 +4068,8 @@ class System(object):
             self._apply_nonlinear()
 
     def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
-                        includes=None, excludes=None, tags=(), get_remote=False, rank=None,
+                        includes=None, excludes=None, is_indep_var=None, is_design_var=None,
+                        tags=(), get_remote=False, rank=None,
                         return_rel_names=True):
         """
         Retrieve metadata for a filtered list of variables.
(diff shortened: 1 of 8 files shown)


openmdao/core/system.py
```diff
@@ -4230,6 +4239,36 @@ def get_io_metadata(self, iotypes=('input', 'output'), metadata_keys=None,
...
+
+                        src_tags = out_meta['tags'] if 'tags' in out_meta else {}
+                        if is_indep_var is True and 'openmdao:indep_var' not in src_tags:
+                            continue
+                        elif is_indep_var is False and 'openmdao:indep_var' in src_tags:
+                            continue
+
+                    # handle is_design_var
+                    if is_design_var is not None:
+                        des_vars = self.get_design_vars(use_prom_ivc=False)
```
naylor-b, line 4261: We should probably move this outside of the loop.  Also, since we don't need sizing info, you could pass `get_sizes=False` into the call as well.

naylor-b (changes requested):

openmdao/core/system.py
swryan: oops.. I thought I had done this, but apparently only did it in Case.   wilco on get_sizes.

robfalck (approved):

naylor-b (approved):

PR #2817, opened 2023-02-21
POEM 81: Implement Submodel Component

### LINKED ISSUE #2816: Implement POEM 81

Add a subproblem component to allow evaluation of OpenMDAO systems within a problem model.

### Associated POEM

081

nsteffen:
### Summary

[POEM 81](https://github.com/OpenMDAO/POEMs/pull/172) implementation. Adds `SubmodelComp` as a new component allowing for users to evaluate OpenMDAO systems within a component.

### Related Issues

- Resolves #2816 

None

### New Dependencies

None


openmdao/components/tests/test_subproblem_comp.py
```diff
@@ -0,0 +1,37 @@
...
+import unittest
+
+from openmdao.components.subproblem_comp import SubproblemComp
+
+class TestSubproblemComp(unittest.TestCase):
+    def test_subproblem_comp(self):
+        prob = om.Problem()
+
+        model = om.ExecComp('z = x**2 + y')
+        submodel1 = om.ExecComp('x = r*cos(theta)')
```
naylor-b, line 14: There have been discussions recently about deprecating the ability to specify a component as the problem model, so you may want to avoid having to go back and update this test in the future by just making the subproblem models be Groups instead of ExecComps.  Using Groups, and potentially nested ones, in the subproblem models is probably a good idea anyway, in order to test that the name mapping/promotion stuff is working correctly.

naylor-b (changes requested):

Kenneth-T-Moore (approved):
I think this is good, simple, and concise. The only thing missing is a feature doc, which you could do as a follow-on.

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,284 @@
...
+        ----------
+        inputs : Vector
+            Unscaled, dimensional input variables read via inputs[key].
+        outputs : Vector
+            Unscaled, dimensional output variables read via outputs[key].
+        """
+        p = self._subprob
+
+        # switch subproblem to use complex IO if in complex step mode
+        if self.under_complex_step != self._prev_complex_step:
```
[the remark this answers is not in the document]
naylor-b: This comment was marked as resolved with no explanation.  In general it's helpful to respond to all review comments in some way, especially if you disagree with a recommendation.  That way the reviewer at least knows why you didn't make a recommended change.

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,284 @@
...
+
+        elif isinstance(var, str):
+            # check if variable already exists in var_dict[varType]
+            # i.e. no repeated variable names
+            if var in var_dict[varType]:
+                raise Exception(f'Variable {var} already exists. Rename variable'
+                                ' or delete copy of variable.')
+
+            # make dict with given var name as key and meta data from model_vars
+            tmp_dict = {var: meta for _, meta in model_vars.items()
```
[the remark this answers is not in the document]
naylor-b: Not sure why this one was resolved without comment or code change...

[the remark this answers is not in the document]
nsteffen: I changed it from 
```python
tmp_dict = {var: meta for _, meta in model_vars
                        if name.endswith(var)}
```
to
```python
tmp_dict = {var: meta for name, meta in model_vars
                        if name.endswith('.' + var)}
```
so that it catches what you mentioned with `prom_name`. Along with that, I made sure `get_model_io` was passed the absolute names rather than the promoted names so that there will always be a '.' in the name since sometimes `prom_name` won't include the '.' syntax.

openmdao/components/subproblem_comp.py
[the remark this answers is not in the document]
nsteffen: I included the function you gave above, but I still thought it was necessary to keep the `if self.under_complex_step != self._prev_complex_step` conditional block because that way there won't be as many calls to `_set_complex_step_mode` whereas if you replace that block with `self._set_complex_step_mode(self._under_complex_step)` then it will be calling it every time `compute` is called.

I guess I thought that would cause it to perform slower, but I could be wrong. Do you think that would be possible?

Also, I wasn't aware of the responding to reviews rule of thumb. I'll be sure to do that in the future.

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,284 @@
...
+
+        self.prob_args = {'driver': driver,
+                          'comm': comm,
+                          'name': name,
+                          'reports': reports}
+
+        self.prob_args.update(prob_options)
+
+        self.model = model
+        self.list_inputs = inputs
```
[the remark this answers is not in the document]
nsteffen: Changed attribute names

openmdao/components/subproblem_comp.py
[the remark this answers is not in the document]
naylor-b: I think that if you did the override of the _set_complex_step_mode method then you wouldn't have to do the check at all in compute, because the component will be in whatever mode is appropriate.  Basically overriding that method would just keep the subproblem in sync with whatever mode the component is in based on whatever mode is currently active.

openmdao/components/subproblem_comp.py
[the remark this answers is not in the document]
naylor-b: If you'd rather work with promoted names, you could change the check to `if name==var or name.endswith('.' + var)`.  You do have to be a little careful here, making sure that you're comparing the same type of names, i.e. you can compare promoted to promoted, but not absolute to promoted because there's not necessarily any correspondence between a promoted name and the suffix of an absolute name since promoted names are allowed to be aliased.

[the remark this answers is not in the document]
naylor-b: I may have missed it, but I didn't see any promotion aliases in your test case.  You should probably have a test that includes that.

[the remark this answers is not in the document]
nsteffen: For that check, would adding `meta['prom_name'] == var` to the check you gave before suffice since that adds the option that the `prom_name` is checked along with the absolute name?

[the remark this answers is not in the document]
nsteffen: fixed name check and ensured model.list_inputs/outputs uses absolute names for keys.

openmdao/docs/openmdao_book/features/building_blocks/components/subproblem_comp.ipynb
```diff
@@ -0,0 +1,235 @@
+{
```
robfalck, line 1: <p><code>SubproblemComp</code> provides a way to evaluate an OpenMDAO system within an OpenMDAO system. Unlike a typical sub-system, using a subproblem exposes only a limited number of inputs and outputs of the underlying model to the parent system. This may be beneficial from a performance standpoint when there are many inputs and outputs of the internal system that aren't needed in the external system.</p><p> </p><p>Note that this means that nonlinear and linear solvers are needed on the subproblem's model if implicit behavior is present, since the internal variables are invisible to solvers in the parent system.</p><p> </p><p>At this time, OpenMDAO does not evaluate gradients through the optimization process, so a gradient-based driver at the outer level cannot access derivatives across the optimization on the inner level.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMDAO/OpenMDAO/pull/2817/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

nsteffen: Added this under 'Using SubproblemComp' section

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,310 @@
...
+        p = self._subprob
+
+        # switch subproblem to use complex IO if in complex step mode
+        self._set_complex_step_mode(self.under_complex_step)
+
+        # setup input values
+        for inp in self._input_names:
+            p.set_val(self.options['inputs'][inp]['prom_name'], inputs[inp])
+
+        if not isinstance(p.driver, Driver):
```
relf, line 274: I am interested by this PR and looking at the code this test puzzled me. Should the condition not be the opposite?

nsteffen: If the problem's driver is not declared, it is made to be a 'run once' driver from the base class `Driver`. Because of that, we can check if p.driver is declared by seeing if it is an instance of the base class.

However, since it is a 'run once' driver, it essentially just calls `run_model`, so that test is probably unnecessary. I am still new to OpenMDAO development, so I thought there might be a difference between `run_model` and `run_driver` for this case, but it doesn't appear so. I'll likely change that in my next commit, thanks for pointing it out!

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def setup(self):
+        """
+        Perform some final setup and checks.
+        """
+        p = self._subprob = Problem(**self.prob_args)
+        p.model.add_subsystem('subsys', self.model, promotes=['*'])
+
+        p.setup(force_alloc_complex=self._problem_meta['force_alloc_complex'])
+        p.final_setup()
```
robfalck, line 205: Now that I'm experimenting with this in dymos, I think that the first 4 lines of code of `setup` should be moved to `__init__`.  I have a use-case where I want to be able to interrogate the subproblem for its outputs and unconnected inputs during my setup process, which requires that the subproblem's setup process be done already.

Because I don't necessarily know what inputs and outputs are to be transferred to/from the subproblem, we should also implement `add_input` and `add_output` methods specific to this component that do something along the lines of:

```
def add_subproblem_input(self, path, name=None):
    self.model_input_names.append((path, name))
   
def add_subproblem_output(self, path, name=None):
    self.model_output_names.append((path, name)) 
```

robfalck: @naylor-b I would appreciate your thoughts on this.

robfalck (changes requested):

openmdao/components/subproblem_comp.py
relf: My surprise was that if it is **not** an instance of driver then you run the driver! 😉 But I agree in any case you can remove it.

nsteffen: Of course! I appreciate the feedback!

openmdao/components/subproblem_comp.py
naylor-b: We should probably add some error checking with these as well so that add_input/add_output can't be called after a certain point in setup.  Otherwise the calls will basically be ignored.  We should also add a test if we don't have one already that calls setup multiple times when a subprob comp is present just to make sure we don't have any weird setup issues.

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,344 @@
...
+        # boundary inputs are any inputs that externally come into `SubproblemComp`
+        boundary_inputs = p.model.list_inputs(out_stream=None, prom_name=True,
+                                              units=True, shape=True, desc=True, is_indep_var=True)
+        # want all outputs from the `SubproblemComp`
+        all_outputs = p.model.list_outputs(out_stream=None, prom_name=True,
+                                           units=True, shape=True, desc=True)
+
+        # declaring all inputs as design vars and all outputs as constraints allows for coloring
+        # to be computed
+        for _, meta in boundary_inputs:
```
naylor-b, line 217: Isn't setting all inputs as DVs and all outputs as constraints going to mess us up later when the driver is an actual optimizer, i.e. it might only be using some subset of the variables as dv/obj/constraints, so it seems like we need to keep track of what dv/obj/constraints exist as part of the model and be able to save/restore them after we have the coloring, or (maybe better) we update the coloring stuff to work with a caller supplied list of 'of' and 'wrt' vars.

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,344 @@
...
+        ----------
+        inputs : Vector
+            Unscaled, dimensional input variables read via inputs[key].
+        outputs : Vector
+            Unscaled, dimensional output variables read via outputs[key].
+        """
+        p = self._subprob
+
+        # switch subproblem to use complex IO if in complex step mode
+        self._set_complex_step_mode(self.under_complex_step)
```
naylor-b, line 301: I'm not clear on why this call to `_set_complex_step_mode` is necessary.  Wouldn't we already be in the right state since we have overridden the `_set_complex_step_mode` method?

openmdao/components/subproblem_comp.py
```diff
@@ -0,0 +1,344 @@
...
+            self.add_output(var, **meta)
+            meta['prom_name'] = prom_name
+            self._output_names.append(var)
+
+        if self.coloring is None:
+            self.declare_partials(of='*', wrt='*')
+        else:
+            for of, wrt, nzrows, nzcols, _, _, _, _ in self.coloring._subjac_sparsity_iter():
+                if of not in self._output_names or wrt not in self._input_names:
+                    continue
```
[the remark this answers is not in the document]
nsteffen: got rid of that check

openmdao/components/subproblem_comp.py
nsteffen: Made methods for `add_design_vars`, `add_constraints`, and `add_objectives` and made `_setup_var_data` to do coloring after we get design vars, constraints, and objectives

openmdao/components/subproblem_comp.py
nsteffen: Implemented error checking in `add_input` and `add_output` to make sure they don't occur after configure. Also implemented test `test_multiple_setups` to ensure SubproblemComp doesn't have any weird setup issues.

naylor-b (changes requested):

openmdao/components/tests/test_subproblem_comp.py
```diff
@@ -0,0 +1,414 @@
...
+        subprob2 = om.SubproblemComp(model=submodel2)
+
+        subprob1.add_input('psi')
+
+        p.model.add_subsystem('sub1', subprob1)
+        p.model.add_subsystem('sub2', subprob2)
+        p.model.add_subsystem('supModel', model, promotes_inputs=['x','y'],
+                              promotes_outputs=['z'])
+
+        with self.assertRaises(Exception) as ctx:
```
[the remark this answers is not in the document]
nsteffen: added message check with `assertRaises`

openmdao/components/tests/test_submodel_comp.py
```diff
@@ -0,0 +1,372 @@
...
+        p.model.add_subsystem('sub1', subprob1)
+        p.model.add_subsystem('sub2', subprob2)
+        p.model.add_subsystem('supModel', model, promotes_inputs=['x','y'],
+                              promotes_outputs=['z'])
+
+        with self.assertRaises(Exception) as ctx:
+            p.setup(force_alloc_complex=True)
+            
+        msg = 'Variable psi does not exist in model.'
+        self.assertTrue(str(ctx.exception), msg)
```
naylor-b, line 340: You need an `assertEqual` here to check that the exception matches the message.

naylor-b (changes requested):

openmdao/components/tests/test_submodel_comp.py
nsteffen: changed to `assertEqual`

naylor-b (approved):

openmdao/components/submodel_comp.py
```diff
@@ -0,0 +1,447 @@
...
+from openmdao.core.constants import _SetupStatus, INF_BOUND
+
+
+class SubmodelComp(ExplicitComponent):
+    """
+    System level container for systems.
+
+    Parameters
+    ----------
+    model : <System>
```
naylor-b, line 14: model has to be a Group now, so should replace <System> with <Group>.

openmdao/components/submodel_comp.py
```diff
@@ -0,0 +1,447 @@
...
+        The system-level <System>.
+    problem : object
```
naylor-b, line 16: replace `object` with `<Problem>`

openmdao/components/submodel_comp.py
```diff
@@ -0,0 +1,447 @@
...
+        [e.g. (path.to.var, desired_name)].
+    reports : bool
+        Determines if reports should be include in subproblem. Default is False because
+        submodelcomp runs faster without reports.
+    **kwargs : named args
+        All remaining named args that become options for `SubmodelComp`.
+
+    Attributes
+    ----------
+    _subprob : object
```
naylor-b, line 38: object --> `<Problem>`

openmdao/components/submodel_comp.py
```diff
@@ -0,0 +1,447 @@
...
+    """
+
+    def __init__(self, problem, inputs=None, outputs=None, reports=False, **kwargs):
+        """
+        Initialize all attributes.
+        """
+        super().__init__(**kwargs)
+
+        problem._reports = [] if not reports else problem._reports
+        # promote ever var in subsystems automatically
```
naylor-b, line 58: typo:  ever --> every

openmdao/components/submodel_comp.py
```diff
@@ -0,0 +1,447 @@
...
+        for subsys in problem.model._static_subsystems_allprocs.keys():
+            problem.model.promotes(subsys, any=['*'])
```
naylor-b, line 60: It's not clear to me why we're promoting everything here.  It could also cause errors due to collisions between outputs that have the same name.

openmdao/components/submodel_comp.py
nsteffen: changed `<System>` to `<Group>`

openmdao/components/submodel_comp.py
nsteffen: changed to `<Problem>`

openmdao/components/submodel_comp.py
nsteffen: fixed typo

openmdao/components/submodel_comp.py
nsteffen: changed to `<Problem>`

naylor-b (changes requested):

openmdao/components/submodel_comp.py
nsteffen: My original thinking was that we want everything to be promoted so that it's easier to refer to the submodel's variables inside. But you're right, that would cause collisions with outputs that have the same name and if the user wants to promote variables, they can do that using `problem.model.promotes` anyway. I'll go ahead and change that.

nsteffen: got rid of automatic promoting

openmdao/core/total_jac.py
```diff
@@ -271,15 +271,24 @@ def __init__(self, problem, of, wrt, use_abs_names, return_format, approx=False,
...
-                           (driver_wrt, wrt, driver_of, of))
-                    issue_warning(msg, category=DerivativesWarning)
-                    self.simul_coloring = None
+                if self.simul_coloring is not None:
+                    ok = True
+                    if prom_wrt != driver_wrt:
+                        ok = False
+                    # TODO: there's some weirdness here where sometimes the
+                    # of vars are absolute and sometimes they're not...
+                    for pof, aof, dof in zip(prom_of, of, driver_of):
```
Kenneth-T-Moore, line 280: This loop could be placed under an else: branch. When prom_wrt is not equal to driver_wrt, then ok is already false, so no need to check the ofs.

nsteffen: placed loop under else branch

relf:
Just to let you know this one is missing in the release notes.

Issue #2769, opened 2023-01-11
Implementation needed for POEM 75


naylor-b:
### Desired capability or behavior.

POEM 75 changes the convention OpenMDAO uses when computing derivatives involving distributed and non-distributed variables.  The underlying principle is that serial variables should have values and derivative values that are consistent across all ranks where those variables exist.

### Associated POEM

75

PR #2751, opened 2023-01-10
POEM 75 impl branch. 

naylor-b:
### Summary

This changes the convention OpenMDAO uses when transferring data between distributed and non-distributed variables.
The underlying principle is that serial variables and their derivatives must have consistent values across all ranks where those variables exist.

### Related Issues

- Resolves #2769

### Backwards incompatibilities

Some matrix-free components with a mix of distributed and non-distributed variables could break.

### New Dependencies

None

PR #2838, opened 2023-03-08
Fixing Out of Bounds Error in Serial Derivative Consistency Check

bernardopacini:
### Summary

For specific scenarios, we found that OpenMDAO was throwing an out of bounds error in the new serial derivative consistency check. This came up for components that had several serial and distributed inputs and outputs. When running `check_totals()` in parallel we would run into errors such as:

```
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/system.py", line 2967, in _call_user_function
    self.gen.throw(typ, value, traceback)
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/system.py", line 2967, in _call_user_function
    raise err_type(
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/system.py", line 2961, in _call_user_function
    raise err_type(
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/system.py", line 2961, in _call_user_function
    yield
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/implicitcomponent.py", line 265, in _apply_linear
    yield
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/implicitcomponent.py", line 265, in _apply_linear
    self._apply_linear_wrapper(self._inputs, self._outputs,
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/implicitcomponent.py", line 217, in _apply_linear_wrapper
    self._apply_linear_wrapper(self._inputs, self._outputs,
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/implicitcomponent.py", line 217, in _apply_linear_wrapper
    self._check_consistent_serial_dinputs(nzdresids)
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/component.py", line 1723, in _check_consistent_serial_dinputs
    self._check_consistent_serial_dinputs(nzdresids)
  File "/home/bpacini/MDOLabCodes/repos/OpenMDAO/openmdao/core/component.py", line 1723, in _check_consistent_serial_dinputs
    result = inconsistent_across_procs(self.comm, v.asarray()[inds])
IndexError: 'cruise.coupling.solver' <class DAFoamSolver>: Error calling apply_linear(), index 16808 is out of bounds for axis 0 with size 5357
    result = inconsistent_across_procs(self.comm, v.asarray()[inds])
IndexError: 'cruise.coupling.solver' <class DAFoamSolver>: Error calling apply_linear(), index 16367 is out of bounds for axis 0 with size 5027
Finalising parallel run
Finalising parallel run
```

Digging into this function it seems that the search for indices is done on a vector that includes both the inputs AND outputs, but the check is later done on vectors of either the inputs OR the outputs. In our case, our input was at the end of the combined input and output vector so it would be significantly out of bounds of the input-only vector.

I am not sure if this fix is the best approach or if it is the intended behavior, but it fixes our issues and the serial derivative consistency check passes.

### Related Issues

This does not resolve any issue on the repo, though it is related to POEM 75 (#2769) which that implemented in PR #2751.

Edit:
- Resolves: #2840

As far as I can tell, none.

### New Dependencies

None

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1702,7 +1702,7 @@ class Component(System):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
-            output_len = 0 if self.is_explicit() else len(self._outputs) 
+            output_len = 0 if self.is_explicit() else len(self._outputs)
             for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
                 if dist_sizes is None:  # not distributed
                     if offset != end:
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1702,10 +1702,14 @@ class Component(System):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
+            globaloffset = 0
             for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
-                if dist_sizes is None:  # not distributed
+                if name.split(".")[-1] in self._var_rel_names["output"]:
+                    globaloffset += (end - offset)
+                elif dist_sizes is None:  # not distributed
                     if offset != end:
-                        ranges[vec].append(range(offset, end))
+                        ranges[vec].append(range(offset - globaloffset, end - globaloffset))
+
 
             self._serial_idxs = []
             for vec, rlist in ranges.items():
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1702,13 +1702,18 @@ class Component(System):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
-            globaloffset = 0
+            output_len = 0
             for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
-                if name.split(".")[-1] in self._var_rel_names["output"]:
-                    globaloffset += (end - offset)
+                if vec is self._outputs:
+                    output_len = len(self._outputs)
+                    break
+            for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
                 if dist_sizes is None:  # not distributed
                     if offset != end:
-                        ranges[vec].append(range(offset - globaloffset, end - globaloffset))
+                        if vec is self._outputs:
+                            ranges[vec].append(range(offset, end))
+                        else:
+                            ranges[vec].append(range(offset - output_len, end - output_len))
 
             self._serial_idxs = []
             for vec, rlist in ranges.items():
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1706,10 +1706,9 @@ class Component(System):
             for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
                 if name.split(".")[-1] in self._var_rel_names["output"]:
                     globaloffset += (end - offset)
-                elif dist_sizes is None:  # not distributed
+                if dist_sizes is None:  # not distributed
                     if offset != end:
                         ranges[vec].append(range(offset - globaloffset, end - globaloffset))
-
 
             self._serial_idxs = []
             for vec, rlist in ranges.items():
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -1702,11 +1702,7 @@ class Component(System):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
-            output_len = 0
-            for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
-                if vec is self._outputs:
-                    output_len = len(self._outputs)
-                    break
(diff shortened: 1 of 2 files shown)


bernardopacini:
I added a test that shows the behavior without the fix. In doing this I noticed that the problem only arises for implicit components when both inputs and outputs are included.

openmdao/core/component.py
```diff
@@ -1702,10 +1702,13 @@ def _check_consistent_serial_dinputs(self, nz_dist_outputs):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
+            globaloffset = 0
             for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
+                if name.split(".")[-1] in self._var_rel_names["output"]:
```
naylor-b, line 1707: could replace this with `if vec is self._outputs:` here.

naylor-b (approved):

openmdao/core/component.py
bernardopacini: Is this the correct approach overall? I noticed below that it selects either the outputs or the inputs, so would this fix conflict with a case in which the outputs are checked instead of the inputs?

naylor-b: Actually, you're right.  Maybe a better way to do this would be:
```
output_len = len(self._outputs)
for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
    if dist_sizes is None:  # not distributed
        if offset != end:
            if vec is self._outputs:
                ranges[vec].append(range(offset, end))
            else:
                ranges[vec].append(range(offset - output_len, end - output_len))
```

I didn't test it, but I think this should work because `_jac_wrt_iter` yields all of the outputs first, then the inputs (for an implicit component).

bernardopacini: I tried this out but it fails for explicit components since the `_jac_wrt_iter()` object does not include the outputs in that case. I added a loop to check if there are outputs before setting the offset. My fix still feels like a bit of a hack, but it seems to work.

bernardopacini:
I swapped my longer test with a modified version of the one @kejacobson put together in Issue #2840, as it is simpler and cleaner.

openmdao/core/component.py
```diff
@@ -1702,10 +1702,18 @@ def _check_consistent_serial_dinputs(self, nz_dist_outputs):
 
         if self._serial_idxs is None:
             ranges = defaultdict(list)
+            output_len = 0
+            for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
+                if vec is self._outputs:
+                    output_len = len(self._outputs)
+                    break
```
naylor-b, line 1709: Ugh, sorry my first recommended fix was wrong.  Since an explicit comp will have all inputs and an implicit comp will have all outputs followed by all inputs, how about this:

```
            output_len = len(self._outputs) if self.is_explicit() else 0
            for name, offset, end, vec, slc, dist_sizes in self._jac_wrt_iter():
                if dist_sizes is None:  # not distributed
                    if offset != end:
                        if vec is self._outputs:
                            ranges[vec].append(range(offset, end))
                        else:
                            ranges[vec].append(range(offset - output_len, end - output_len))

```

bernardopacini:
I didn't catch that formatting issue but I just fixed it. Hopefully now this is good to go.

bernardopacini:
I am not sure how to go about fixing this issue with the fake comm and still maintain correct derivatives. Is there a trick for this sort of missing feature?

naylor-b:
Yeah just put in a check to see if MPI is active and skip the test if it's not.  We typically use something like

```
from openmdao.utils.mpi import MPI   # MPI will be None here if MPI is not active
try:
    from openmdao.vectors.petsc_vector import PETScVector
except ImportError:
    PETScVector = None

...


@unittest.skipUnless(MPI and PETScVector, "MPI and PETSc are required.")
class MyTestCase(unittest.TestCase):
    def test_mycase(self):
           pass
```

bernardopacini:
> Yeah just put in a check to see if MPI is active and skip the test if it's not. We typically use something like
> 
> ```
> from openmdao.utils.mpi import MPI   # MPI will be None here if MPI is not active
> try:
>     from openmdao.vectors.petsc_vector import PETScVector
> except ImportError:
>     PETScVector = None
> 
> ...
> 
> 
> @unittest.skipUnless(MPI and PETScVector, "MPI and PETSc are required.")
> class MyTestCase(unittest.TestCase):
>     def test_mycase(self):
>            pass
> ```

I just added those lines. Hopefully now good to go!

robfalck (approved):

PR #2839, opened 2023-03-08
Refactor for complex step around Newton.

Kenneth-T-Moore:
### Summary

Previously, when we performed complex step around a model with a Newton solver in it, we needed to "nudge" the outputs vector by some small quantity in order to force it to reconverge, and thereby propagate the imaginary part throughout the vector. This worked in many cases, but sometimes the size of the nudge could be too small or large and this would affect the accuracy of the approximated derivatives.

This PR implements a new method, where Newton is forced to compute a single iteration (because a complex step of 1e-40 is too small to rise above the solver's tolerances). We also disable any special sub iteration, such as subsystem solves or linesearches so that we just get a pure Newton step that propagates the complex step into the rest of the vector via the local derivative.

1.  If the newton solver uses an iterative linear solver such as Linear Gauss-Seidel, the approximated totals are not that great. Note that these derivatives weren't very good with the old method either, but our test was comparing against some wrong values.
2. There may be some issues with ExecComp's "cs"-computed derivatives, but that is unrelated to this PR and there will be a bug story soon.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_check_derivs.py
+++ b/openmdao/core/tests/test_check_derivs.py
@@ -3552,7 +3552,7 @@ class TestProblemCheckTotals(unittest.TestCase):
         prob.model.linear_solver = om.DirectSolver(assemble_jac=True)
         prob.run_model()
 
-        totals = prob.check_totals(of=['obj', 'con1'], wrt=['x', 'z'], method='cs', out_stream=None)
+        totals = prob.check_totals(of=['obj', 'con1'], wrt=['x', 'z'], method='cs')
 
         for key, val in totals.items():
             assert_near_equal(val['rel error'][0], 0.0, 3e-8)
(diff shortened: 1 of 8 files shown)


robfalck (approved):

openmdao/core/tests/test_check_derivs.py
```diff
@@ -3552,7 +3598,7 @@ def test_cs_around_newton_top_sparse(self):
         prob.model.linear_solver = om.DirectSolver(assemble_jac=True)
         prob.run_model()
 
-        totals = prob.check_totals(of=['obj', 'con1'], wrt=['x', 'z'], method='cs', out_stream=None)
+        totals = prob.check_totals(of=['obj', 'con1'], wrt=['x', 'z'], method='cs')
```
naylor-b, line 3601: Shouldn't out_stream=None be there for a test?

naylor-b (approved):

openmdao/core/tests/test_check_derivs.py
Kenneth-T-Moore: Fixed

PR #2860, opened 2023-03-22
Changed handling of _no_check_partials

### LINKED ISSUE #2858: Different `check_partials` behavior locally vs. on GitHub Actions

When running unit tests for our own package, we use `check_partials` to ensure that our derivative implementations are correct. We run unit tests automatically with GitHub Actions.

**The problem:** Our GitHub Actions run was raising an error coming from a `check_partials` call. I tried to reproduce it locally, but was unable. After looking in the `check_partials` source I realized why: it looks for an environment variable automatically set by GitHub Actions to determine whether or not to run `check_partials` on OpenMDAO's built-in components (or anything else with `_no_check_partials == True`). See the check here:

https://github.com/OpenMDAO/OpenMDAO/blob/cb27f67159c5908fb24e58733207320ef752e6fd/openmdao/core/problem.py#L1124

This makes sense for OpenMDAO's own testing framework, but can cause problems for other packages that use OpenMDAO and are running automated testing. In my case, I was complex-stepping a model that contained a vectorized `MetaModelUnstructuredComp`, which does not work (more on this later). Tests passed locally because it was not checking the partials of this metamodel. However, it *was* checking the metamodel's partials on our GitHub Actions. I believe the intended behavior is to check partials of built-in OpenMDAO components only on OpenMDAO's automated testing, correct me if I'm wrong.

**The solution:** instead of using GitHub Actions' default `CI` environment variable, I would propose that OpenMDAO's automated testing set it's own environment variable, such as `OM_CI`. This would then be the variable `check_partials` would look for when evaluating the `under_CI` variable in the line of code above.

**The other problem and its solution:** OpenMDAO's `MetaModelUnstructuredComp` is not complex safe when the `vec_size` option is greater than 1. This results in complex-step derivatives of zero when checking partials using complex-step. To fix this I believe the only thing necessary is to change this line:

https://github.com/OpenMDAO/OpenMDAO/blob/cb27f67159c5908fb24e58733207320ef752e6fd/openmdao/components/meta_model_unstructured_comp.py#L387

to instead be `predicted = np.zeros(output_shape, dtype=flat_inputs.dtype)`. I'm happy to open a PR with this fix if that would be helpful.

Try this bit of code to see the `check_partials` failure with a vectorized `MetaModelUnstructuredComp`:
```python
# create a MetaModelUnStructuredComp, specifying surrogates for the outputs
import numpy as np
import openmdao.api as om

size = 3

# create a vectorized MetaModelUnStructuredComp for sine and cosine
trig = om.MetaModelUnStructuredComp(vec_size=size, default_surrogate=om.KrigingSurrogate())
trig.add_input('x', np.zeros(size))
trig.add_output('y', np.zeros((size, 2)))

# add it to a Problem
prob = om.Problem()
prob.model.add_subsystem('trig', trig)

# provide training data
trig.options['train_x'] = np.linspace(0, 10, 20)
trig.options['train_y'] = np.column_stack((
    .5*np.sin(trig.options['train_x']),
    .5*np.cos(trig.options['train_x'])
))

# train the surrogate and check predicted value
prob.set_val('trig.x', np.array([2.1, 3.2, 4.3]))
prob.run_model()
prob.check_partials(method="cs")
```

### Associated POEM

_No response_

naylor-b:
### Summary

- Environment variable being checked is now `OPENMDAO_CI` instead of `CI`
- Set `OPENMDAO_CI` in our workflow yaml file
- created a testing function called `force_check_partials` that will always check partials regardless of the value of `OPENMDAO_CI` so that tests designed to check partials for a particular component will always check those partials.
- In a number of tests, replaced manually setting _no_check_partials to False with calling `force_check_partials` instead.
- Added calls to `force_check_partials` in a few places where checking of partials wasn't being forced but should have been.
- Fixed the bug in MetaModelUnStructuredComp where the derivatives were wrong when vec_size > 1 and using complex step.

### Related Issues

- Resolves #2858

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/testing_utils.py
+++ b/openmdao/utils/testing_utils.py
@@ -248,7 +248,7 @@ def set_env_vars_context(**kwargs):
                 os.environ[k] = v
 
 
-@set_env_vars(OPENMDAO_CI="1")
+@set_env_vars(OPENMDAO_CHECK_ALL_PARTIALS="1")
 def force_check_partials(prob, *args, **kwargs):
     r"""
     Force the checking of partials even for components with _no_check_partials set.
--- a/openmdao/utils/testing_utils.py
+++ b/openmdao/utils/testing_utils.py
@@ -197,6 +197,7 @@ class set_env_vars(object):
         fnc : function
             The function being wrapped.
         """
+        @functools.wraps(fnc)
         def wrap(*args, **kwargs):
             saved = {}
             try:
@@ -245,6 +246,28 @@ def set_env_vars_context(**kwargs):
                 del os.environ[k]
             else:
                 os.environ[k] = v
+
+
+@set_env_vars(OPENMDAO_CI="1")
+def force_check_partials(prob, *args, **kwargs):
+    r"""
+    Force the checking of partials even for components with _no_check_partials set.
+
+    Parameters
+    ----------
+    prob : Problem
+        The Problem being checked.
+    *args : list
+        Positional args.
+    **kwargs : dict
+        Keyword args.
+
+    Returns
+    -------
+    dict
+        Total derivative comparison data.
+    """
+    return prob.check_partials(*args, **kwargs)
 
 
 class _ModelViewerDataTreeEncoder(json.JSONEncoder):
--- a/.github/workflows/openmdao_test_workflow.yml
+++ b/.github/workflows/openmdao_test_workflow.yml
@@ -288,6 +288,8 @@ jobs:
 
       - name: Run tests
         if: matrix.TESTS
+        env:
+          OPENMDAO_CI: true
         run: |
           echo "============================================================="
           echo "Run tests with coverage (from directory other than repo root)"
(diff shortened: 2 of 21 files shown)


openmdao/utils/testing_utils.py
```diff
@@ -247,6 +248,28 @@ def set_env_vars_context(**kwargs):
                 os.environ[k] = v
 
 
+@set_env_vars(OPENMDAO_CI="1")
+def force_check_partials(prob, *args, **kwargs):
+    r"""
+    Force the checking of partials even for components with _no_check_partials set.
+
```
swryan, line 255: It occurs to me that maybe the env var should be more specific, e.g. `FORCE_CHECK_PARTIALS` since that's the one thing it's used for..  What do you think?

robfalck: I agree with @swryan here.  Undoubtedly in the future we'll do something else that needs a CI flag but a more descriptive name would be useful here.

swryan (changes requested):
see comments re: changing the var name

openmdao/utils/testing_utils.py
naylor-b: fixed

swryan (approved):

PR #2865, opened 2023-03-26
OptionsDictionary update: Added context manager for temporary options values and a set method for setting multiple values at once.

robfalck:
### Summary

When testing, I find myself wanting to change options values temporarily, run a test, and then reset the option back to its default value. This usually looks something like:

```
cache = my_options['name']
my_options['name'] = temp_val

run_tests()

my_options['name'] = cache
```

This seems similar to numpy's printoptions, and so it seemed like the right use for a context manager.
With this change you can now temporarily set and option value within the context.

```
with my_options.temporary(name=temp_val):
    run_tests()
```

I also often prefer to write a setter method to set multiple options in one line rather than using several lines and accessing the various options names with the `__setitem__` interface. This PR adds a `set` method that allows multiple options in the dictionary to be set simultaneously.

```
my_options.set(name='foo', units='kg')
```

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/options_dictionary.py
+++ b/openmdao/utils/options_dictionary.py
@@ -335,11 +335,15 @@ class OptionsDictionary(object):
             and the associated values are the temporary values for those options.
         """
         for option, val in kwargs.items():
-            self._context_cache[option] = self[option]
+            if option not in self._context_cache:
+                self._context_cache[option] = []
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore (approved):
Thanks for adding nesting.

PR #2885, opened 2023-04-12
Fix for a bug found when computing semi-total derivatives.

### LINKED ISSUE #2874: In subsystems with approx_totals, derivative with respect to inputs that connect to multiple subsystems is wrong.
### Description

The fd perturbation is only carried out on one of the inputs, instead of all of them simultaneously. Something must be wrong with the setter when used under approx_totals.

### Example

```python
import numpy as np
import openmdao.api as om

step = 1e-6

class Mult(om.ExplicitComponent):

    def setup(self):

        self.add_input('x', 1.0)
        self.add_input('y', 1.0)

        self.add_output('z', 1.0)

        outputs['z'] = inputs['x'] * inputs['y']
        print(self.pathname, inputs['x'], inputs['y'])


class GeometryAndAero(om.Group):

    def setup(self):

        self.add_subsystem("comp1", Mult(), promotes_inputs=[('y', 'omega')])
        self.add_subsystem("comp2", Mult(), promotes_inputs=[('y', 'omega')])
        self.add_subsystem("comp3", Mult(), promotes_inputs=[('y', 'omega')], promotes_outputs=[('z', 'thrust')])

        self.connect('comp1.z', 'comp2.x')
        self.connect('comp2.z', 'comp3.x')
        
        self.approx_totals(step=step, step_calc="abs", method="fd", form="forward")


prob = om.Problem()

comp = GeometryAndAero()
prob.model.add_subsystem('sub', comp, promotes=['*'])

prob.model.add_design_var("omega")
prob.model.add_objective("thrust")

prob.setup(force_alloc_complex=False, check=False)
prob.set_val("omega", 5.0)

prob.run_model()

# Deriv should be 75. Analytic is wrong.
prob.check_totals(method="fd", form="forward", step=step, step_calc="abs")
```
```


### OpenMDAO Version

3.25.1-dev

_No response_

naylor-b:
### Summary

When computing the approx subjac keys it was looping over promoted input names and only taking the first absolute input name in the list, causing perturbations not to propagate correctly through implicitly connected inputs.

Also fixed a logic error when weeding out candidate wrt's that didn't connect to IndepVarComp outputs.  An absolute component name was being computed from the absolute var name and used in a call to `_get_subsystem`, but `_get_subsystem` requires a relative name, so IndepVarComps were actually being missed.  After this was fixed, a test failed because `_jac_wrt_iter` was now yielding the same name twice.  The duplication was removed and now all tests pass.

### Related Issues

- Resolves #2874

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -2840,32 +2840,46 @@ class Group(System):
         if self._owns_approx_wrt and not self.pathname:
             candidate_wrt = self._owns_approx_wrt
         else:
-            candidate_wrt = self._var_allprocs_abs2meta['input']
+            pro2abs = self._var_allprocs_prom2abs_list
+            candidate_wrt = []
+            for abs_inps in pro2abs['input'].values():
+                for inp in abs_inps:
+                    candidate_wrt.append(inp)
+                    # If connection is inside of this Group, perturbation of all implicitly
+                    # connected inputs will be handled properly via internal transfers. Otherwise,
+                    # we need to add all implicitly connected inputs separately.
+                    if inp in self._conn_abs_in2out:
+                        break
 
         wrt = set()
         ivc = set()
         if self.pathname:  # get rid of any old stuff in here
             self._owns_approx_of = self._owns_approx_wrt = None
+            totals = False
+        else:
+            totals = True
 
-        all_abs2meta_out = self._var_abs2meta['output']
+        all_abs2meta_out = self._var_allprocs_abs2meta['output']
 
+        # When computing totals, weed out inputs connected to anything inside our system unless
+        # the source is an indepvarcomp.
+        prefix = self.pathname + '.' if self.pathname else ''
         for var in candidate_wrt:
-
-            # Weed out inputs connected to anything inside our system unless the source is an
-            # indepvarcomp.
             if var in self._conn_abs_in2out:
-                src = self._conn_abs_in2out[var]
-                meta = all_abs2meta_out[src]
-                if 'openmdao:indep_var' in meta['tags']:
-                    wrt.add(src)
-                    ivc.add(src)
+                if totals:
+                    src = self._conn_abs_in2out[var]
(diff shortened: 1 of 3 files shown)


Kenneth-T-Moore (changes requested):
Awaiting change to _var_allprocs_abs2meta in line 2850. Looks good otherwise.

robfalck (approved):

dingraha:
This appear to fix the problem I was having with using `approx_totals`, but it also appears to require many more `compute` calls than I expected. I've attached an example that is similar to the actual application I'm working on. When I use this PR, running `run_model` and then `compute_totals` requires 25 `compute` calls, judging from the print statements I added to one of the components. When I try with OpenMDAO `3.25.0` it only requires 5 `compute` calls. I'm asking for derivatives with respect to two inputs from the `Group` with `approx_totals`, so I would have expected the `compute_totals` to require 3 compute calls (2 calls to perturb each input, 1 for the unperturbed inputs, 1 for a `compute` call that I imagine `compute_totals` does just to be safe, and then 1 more for the `run_model` that I have in the script just before `compute_totals`). Maybe I'm misunderstanding how `approx_totals` is supposed to work?

Here is the script (actually a Python script, just added the `txt` extension for GitHub):

[mwe2.py.txt](https://github.com/OpenMDAO/OpenMDAO/files/11232378/mwe2.py.txt)

naylor-b:
@dingraha I pushed up a fix.  If you get a chance, please try it out and let me know if your issue is fixed.

dingraha:
@naylor-b Working on it, thanks! Seems to work for the MWE, checking it on the "real" test case now...

dingraha:
@naylor-b This seems to fix my issue. :-)

How hard would it be to test how many times `compute` is called when using `approx_totals` (if that's not done already)?

Kenneth-T-Moore (approved):

PR #2889, opened 2023-04-18
removed promotion warning if either old or new is '*'

### LINKED ISSUE #2856: Propose: move the re-promotion warning out of the default checks.

If you promote a variable using a wildcard, then at a higher subsystem level, you change the promote to a different alias, you get a warning of the form:

    "input variable '...' promoted using '*' was already promoted using 'aircraft:*'

If you have a lot of variables, this can be cluttersome. I woud like to propose moving it out of the default checks. You will still be able to see the warnings if you set check=True (or include the string for this specific warning.)

### Associated POEM

_No response_

naylor-b:
### Related Issues

- Resolves #2856

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2677,7 +2677,7 @@ class System(object):
                                            f"'{tup[0]}' because '{name}' has already been promoted "
                                            f"as '{old_key}'.")
 
-                if old_key != '*':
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore (approved):
This will clean up a lot of warnings that we are suppressing in aviary with a warning filter context.

swryan (approved):

PR #2906, opened 2023-04-28
Made `list_indep_vars` return only one var per source

### LINKED ISSUE #2905: make `list_indep_vars` track source names rather than prom names
### Description

When the user wants the independent variables in a problem, `list_indep_vars` tracks the promoted names of the indep vars rather than the source names. This means there can be repeated variables in `list_indep_vars`, since multiple variables can be connected to a source.

It would be better for `list_indep_vars` to return the variable that other variables are connected to, since that is the only indep var that will truly change.

In the following case, `xComp.r1` and `xComp.theta1` are connected to `indep_r.r` and `indep_theta.theta`, but rather than `indep_r.r` or `indep_theta.theta` showing up in `list_indep_vars`, their connections are there instead.

This is necessary for `SubmodelComp` because its purpose is to hide some variables to the outer model and it uses `list_indep_vars` to track the necessary inputs. A lot of unnecessary variables are then added to the submodel because `list_indep_vars` returns variables that have the same source.

### Example

```python
import openmdao.api as om


p = om.Problem()

p.model.add_subsystem('indep_theta', om.IndepVarComp('theta'))
p.model.add_subsystem('indep_r', om.IndepVarComp('r'))

p.model.add_subsystem('xComp', om.ExecComp('x = r1*cos(theta1)'))
p.model.add_subsystem('yComp', om.ExecComp('y = r2*sin(theta2)'))

p.model.connect('indep_theta.theta', ['xComp.theta1', 'yComp.theta2'])
p.model.connect('indep_r.r', ['xComp.r1', 'yComp.r2'])

p.setup()
p.final_setup()

p.list_indep_vars()

```

```
-----------------------------------------
Problem sample_case Independent Variables
-----------------------------------------
name         units val  
------------ ----- ----
xComp.r1     None  [1.] 
xComp.theta1 None  [1.] 
yComp.r2     None  [1.] 
yComp.theta2 None  [1.] 

```

### OpenMDAO Version

3.26-dev

_No response_

nsteffen:
### Summary

`list_indep_vars` would previously track prom names and return vars whose sources were tagged 'openmdao:indep_var'. However, this can mean there are repeated variables returned  since multiple vars can be connected to the same source. This update makes `list_indep_vars` return the sources tagged 'openmdao:indep_var' if the names aren't auto ivcs. If they are auto ivcs, then it returns the associated prom name.

### Related Issues

- Resolves #2905 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -2264,30 +2264,32 @@ class Problem(object):
             raise RuntimeError("list_indep_vars requires that final_setup has been "
                                "run for the Problem.")
 
-        connections = model._conn_global_abs_in2out
         desvar_prom_names = model.get_design_vars(recurse=True,
                                                   use_prom_ivc=True,
                                                   get_sizes=False).keys()
+
         problem_indep_vars = []
         indep_var_names = set()
 
         default_col_names = ['name', 'units', 'val']
         col_names = default_col_names + ([] if options is None else options)
 
-        for target, meta in model._var_allprocs_abs2meta['input'].items():
-            src = connections[target]
-            smeta = model._var_allprocs_abs2meta['output'][src]
-            src_is_ivc = 'openmdao:indep_var' in smeta['tags']
-            input_name = model._var_allprocs_abs2prom['input'][target]
+        abs2meta = model._var_allprocs_abs2meta['output']
+        prom2abs = self.model._var_allprocs_prom2abs_list['input']
 
-            smeta = {key: val for key, val in smeta.items() if key in col_names}
-            smeta['val'] = self.get_val(input_name)
+        prom2src = {prom: model.get_source(prom) for prom in prom2abs.keys()
+                    if 'openmdao:indep_var' in abs2meta[model.get_source(prom)]['tags']}
 
-            if src_is_ivc \
-                    and (include_design_vars or input_name not in desvar_prom_names) \
-                    and input_name not in indep_var_names:
-                problem_indep_vars.append((input_name, smeta))
-                indep_var_names.add(input_name)
+        for prom, src in prom2src.items():
+            name = prom if '_auto_ivc' in src else src
+            meta = abs2meta[src]
+            meta = {key: val for key, val in meta.items() if key in col_names}
+            meta['val'] = self.get_val(prom)
+
+            if (include_design_vars or name not in desvar_prom_names) \
+                    and name not in indep_var_names:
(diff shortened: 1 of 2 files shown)


openmdao/core/problem.py
```diff
@@ -2264,30 +2264,32 @@ def list_indep_vars(self, include_design_vars=True, options=None,
...
+        abs2meta = model._var_allprocs_abs2meta['output']
+        prom2abs = self.model._var_allprocs_prom2abs_list['input']
+
+        prom2src = {prom: model.get_source(prom) for prom in prom2abs.keys()
+                    if 'openmdao:indep_var' in abs2meta[model.get_source(prom)]['tags']}
 
-            smeta = {key: val for key, val in smeta.items() if key in col_names}
-            smeta['val'] = self.get_val(input_name)
+        for prom, src in prom2src.items():
+            name = prom if '_auto_ivc' in src else src
```
naylor-b, line 2284: `name = prom if src.startswith('_auto_ivc.') else src` would work better in the highly unlikely event that src contained the string '_auto_ivc' but wasn't actually the path of a variable contained in the _auto_ivc component.

openmdao/core/tests/test_problem.py
```diff
@@ -2264,6 +2264,25 @@ def test_list_indep_vars_duplicate_inputs(self):
...
+        prob.model.add_subsystem('yComp', om.ExecComp('y = r2*sin(theta2)'))
+
+        prob.model.connect('indep_theta.theta', ['xComp.theta1', 'yComp.theta2'])
+        prob.model.connect('indep_r.r', ['xComp.r1', 'yComp.r2'])
+
+        prob.setup()
+        prob.final_setup()
+
+        indep_vars = prob.list_indep_vars()
+        self.assertEqual(len(indep_vars), 2)
```
naylor-b, line 2283: I would recommend testing for the actual indep var names you're expecting instead of just testing the length of indep_vars.

naylor-b (changes requested):

openmdao/core/problem.py
nsteffen: added `src.startswith('_auto_ivc.')` to the check

openmdao/core/tests/test_problem.py
nsteffen: changed the asserts to look for the indep var names

robfalck (approved):

naylor-b (approved):

PR #2915, opened 2023-05-15
Split top level subsystems into pre, optimization, and post lists.

### LINKED ISSUE #2908: Allow automated pre/post analysis when running a driver.

Users sometimes request the ability to execute a system in a model only once before or after the optimization has run.
The submodel capability will allow this to be done manually, but that's a sometimes laborious process and we don't necessarily want the user to have to change their model to accommodate this capability.

We can determine those components of the model that need to be executed during optimization and omit those that are determined to be purely pre or post optimization steps. OpenMDAO should allow the user to disable the execution of these systems during optimization if necessary.

We will add a flag to problem, name TBD, that enables this capability. The current functionality will remain the default for now.

When enabled, the optimization driver will not execute those parts of the system that have been identified as purely pre or post processes.

The behavior of recording will need to be worked out when this feature is in use. Are those systems simply not recorded if they are not executed, or do we record them with stale values in place?

### Associated POEM

089

naylor-b:
### Summary

Added a problem option called 'group_by_pre_opt_post` (default False) that, if True, will split the top level subsystems into pre-optimization, optimization, and post-optimization lists, and only the optimization list will be iterated over during the optimization.  If the top level driver is not an optimizer, or if the top level nonlinear solver is not NonlinearRunOnce, the option will be ignored.

### Related Issues

- Resolves #2908

None

### New Dependencies

None


openmdao/core/tests/test_coloring.py
```diff
@@ -658,6 +659,38 @@ def test_dynamic_rev_simul_coloring_pyoptsparse_slsqp(self):
...
+
+        p.model._solve_count = 0
+        p_color.model._solve_count = 0
+
+        J = p.driver._compute_totals()
+        J_color = p_color.driver._compute_totals()
+
+        # coloring saves 16 solves per driver iter  (5 vs 21)
+        self.assertEqual(p.model._solve_count, 21)
+        self.assertEqual(p_color.model._solve_count, 5)
```
swryan, line 692: this test is for an unrelated bug fix?  should probably mention in the PR summary

swryan (approved):

robfalck (approved):

openmdao/core/driver.py
```diff
@@ -10,6 +10,7 @@
```
Kenneth-T-Moore, line 10: I'm kind of impressed that this didn't require any changes to Driver.

openmdao/core/group.py
```diff
@@ -4333,3 +4366,161 @@ def _ordered_comp_name_iter(self):
                 yield from s._ordered_comp_name_iter()
             else:
                 yield s.pathname
+
+    def _solver_subsystem_iter(self, local_only=False):
```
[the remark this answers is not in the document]
naylor-b: No this is actually an iterator for solvers.  The driver doesn't specify which systems to iterate over.  It just tells the model to run solve_nonlinear and then the nonlinear solver actually does the iterations.

openmdao/core/driver.py
naylor-b: Yeah it turns out that Driver doesn't manage the iteration of the model subsystems at all. It just tells the model to do a nonlinear solve and lets the nonlinear solver do the iteration.

openmdao/core/group.py
```diff
@@ -4333,3 +4366,161 @@ def _ordered_comp_name_iter(self):
...
+            # skip any system separation and just do it the old way
+            self._iterated_subsystems = [s for s, _ in self._subsystems_allprocs.values()]
+            if do_separation:
+                issue_warning("The 'group_by_pre_opt_post' option is True, but the top level "
+                              "nonlinear solver is not NonlinearRunOnce, so the option will be "
+                              "ignored.  If an iterative solver is needed, move the iterated "
+                              "subsystems into a subgroup and set the nonlinear solver on that.")
+            return
+
+        graph = self.compute_sys_graph()
```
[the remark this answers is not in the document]
naylor-b: yes, good catch.

naylor-b:
Closing this temporarily in order to make it more useful for Aviary.

PR #2937, opened 2023-06-01
Implemented POEM_088: Enabled custom `load_case` method for a System

### LINKED ISSUE #2936: POEM_088: Allow Systems to provide custom `load_case` logic

The current `load_case` function in the `Problem` class should be updated to allow for the model or any of it's subsystems to have a custom `load_case` function.

The new `load_case` function on `Problem` will perform similarly as it does now by populating the input and output values in the model _except_ for values that belong to a System that provides it's own `load_case` method.  For those values, `load_case` will be delegated to the System function.

The base `System` class should get a default `load_case` method which a user must override in a subclass (Group or Component) to provide custom logic.

### Associated POEM

088

swryan:
### Summary

Updated the `load_case` method in the Problem class to allow for the model or any of it's subsystems to have a custom `load_case` function.

The new `load_case` function on Problem performs similarly to how it does now by populating the input and output values in the model except for values that belong to a System that provides it's own `load_case` method. Loading of those values is delegated to the System function.

### Related Issues

- Resolves #2936 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -10,7 +10,6 @@ import pathlib
 import textwrap
 
 from collections import defaultdict, namedtuple
-from fnmatch import fnmatchcase
 from itertools import product
 
 from io import StringIO, TextIOBase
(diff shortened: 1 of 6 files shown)


openmdao/core/problem.py
```diff
@@ -2103,39 +2103,121 @@ def load_case(self, case):
...
+        # find all subsystems that override the load_case method
+        system_overrides = {}
+        for subsys in model.system_iter(include_self=False, recurse=True):
+            if overrides_method('load_case', subsys, System):
+                system_overrides[subsys.pathname] = subsys
+
+        def set_later(var_name):
+            # determine if variable should be set later via an overridden load_case method
+            for pathname in system_overrides:
+                if var_name.startswith(pathname):
```
naylor-b, line 2128: Need to put a '.' after pathname to be certain that a variable lives under that system.

robfalck (approved):

PR #2950, opened 2023-06-23
Removed hard coded `set_solver_print` in `SubmodelComp`

### LINKED ISSUE #2949: SubmodelComp does not let you change the solver print level
### Description

In SubmodelComp's setup method, set_solver_print is hard coded in as -1. Sometimes a user wants to change that for debugging, but they aren't able to do so because if they try to and setup is called again, that overrides their initial set_solver_print.

### Example

```python
p = om.Problem()
subprob = om.Problem()
subprob.model.add_subsystem('comp', om.ExecComp('x = r*cos(theta)'), promotes=['*'])
submodel = om.SubmodelComp(problem=subprob)

submodel.add_input('r', name='new_r', val=20)
submodel.add_input('theta', name='new_theta', val=0.5)
submodel.add_output('x', name='new_x', val=100)

p.model.add_subsystem('submodel', submodel, promotes=['*'])

p.model.submodel._subprob.set_solver_print(level=1)

p.setup()
p.final_setup()
```

### OpenMDAO Version

3.26.1-dev

_No response_

nsteffen:
### Summary

Sometimes the user wants to have the submodel's internal problem print its solver output. Before, the internal problem's solver output was disabled via hard coding `set_solver_print(-1)` in SubmodelComp's setup method. Getting rid of this now allows the user to set the solver print by doing `comp._subprob.set_solver_print()`.

### Related Issues

- Resolves #2949 

None

### New Dependencies

None

### DIFF
--- a/openmdao/components/submodel_comp.py
+++ b/openmdao/components/submodel_comp.py
@@ -29,8 +29,6 @@ class SubmodelComp(ExplicitComponent):
     reports : bool
         Determines if reports should be include in subproblem. Default is False because
         submodelcomp runs faster without reports.
-    solver_print_options : dict
-        Dict for any options the user would want for the internal problems solver print.
     **kwargs : named args
         All remaining named args that become options for `SubmodelComp`.
 
@@ -38,8 +36,6 @@ class SubmodelComp(ExplicitComponent):
     ----------
     _subprob : <Problem>
         Instantiated problem used to run the model.
-    _subprob_solver_print_options: dict
-        Arguments for the subproblems solver print.
     submodel_inputs : list of tuple
         List of inputs requested by user to be used as inputs in the
         subproblem's system.
@@ -48,8 +44,7 @@ class SubmodelComp(ExplicitComponent):
         subproblem's system.
     """
 
-    def __init__(self, problem, inputs=None, outputs=None, reports=False,
-                 solver_print_options={'level': 2, 'depth': 1e99, 'type_': 'all'}, **kwargs):
+    def __init__(self, problem, inputs=None, outputs=None, reports=False, **kwargs):
         """
         Initialize all attributes.
         """
@@ -58,8 +53,6 @@ class SubmodelComp(ExplicitComponent):
         if not reports:
             clear_reports(problem)
         self._subprob = problem
-
-        self._subprob_solver_print_options = solver_print_options
 
         self.submodel_inputs = {}
         self.submodel_outputs = {}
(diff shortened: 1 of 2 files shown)


openmdao/components/submodel_comp.py
```diff
@@ -75,6 +82,27 @@ def __init__(self, problem, inputs=None, outputs=None, reports=False, **kwargs):
                 else:
                     raise Exception(f'Expected output of type str or tuple, got {type(out)}.')
 
+    def set_subprob_solver_print(self, level=2, depth=1e99, type_='all'):
```
robfalck, line 85: So I think it would be preferable for the user to interact with the subprob in this case rather than saving these settings.

Let's not introduce public functions if they're not necessary.  The user already has to instantiate this component with a problem instance and they therefore have it available at their disposal.  Let them call `set_solver_print` on that problem, and by removing the hard-coded `subprob.set_solver_print(-1)` below, anything the user does there should be respected.

openmdao/components/submodel_comp.py
```diff
@@ -382,7 +410,7 @@ def _setup_var_data(self):
             p.model.add_constraint(prom_name)
 
         # setup again to compute coloring
-        p.set_solver_print(-1)
+        p.set_solver_print(**self._subprob_solver_print_options)
```
robfalck, line 413: Don't bother with this line, just remove the previous line and let whatever settings the user had previously applied remain.

openmdao/components/tests/test_submodel_comp.py
```diff
@@ -440,3 +440,22 @@ def test_add_io_meta_override(self):
...
+        p = om.Problem()
+        subprob = om.Problem()
+        subprob.model.add_subsystem('comp', om.ExecComp('x = r*cos(theta)'), promotes=['*'])
+        submodel = om.SubmodelComp(problem=subprob)
+
+        submodel.add_input('r', name='new_r', val=20)
+        submodel.add_input('theta', name='new_theta', val=0.5)
+        submodel.add_output('x', name='new_x', val=100)
+
+        submodel.set_subprob_solver_print(level=3, depth=20, type_='NL')
```
robfalck, line 454: We already have an instance of subprob to work with here, just use `subprob.set_solver_print(level=3, depth=20, type_='NL')` in this test.

openmdao/components/tests/test_submodel_comp.py
```diff
@@ -440,3 +440,22 @@ def test_add_io_meta_override(self):
...
+
+        p.model.add_subsystem('submodel', submodel, promotes=['*'])
+
+        p.setup()
+        p.final_setup()
+        print()
```
robfalck, line 460: superfluous for a test

robfalck (changes requested):

openmdao/components/submodel_comp.py
nsteffen: got rid of public function

robfalck (approved):

Kenneth-T-Moore (approved):

PR #3013, opened 2023-09-01
Updated the coloring report to use bokeh for visualization.

### LINKED ISSUE #3005: Convert coloring report to bokeh

The total coloring report will be converted to a single html file generated using bokeh.
It will include a heatmap similar to the current coloring report as well as the coloring statistics (total rows, cols, number of nonzeros, number of forward solves, number of reverse solves).

As part of this, mappings from desvar and response names to promoted names will be included in the pickled coloring file so that variables can be labeled with names that the user understands (as opposed to absolute paths or auto_ivc names).

If bokeh is not present this report will contain the sparsity pattern as a text pattern.

Bokeh will not be a hard requirement to install OpenMDAO but will be recommended as we'll be using it for more visualization features.

### Associated POEM

_No response_

robfalck:
### Summary

The total coloring report is now created as an HTML file via bokeh. This will make the behavior of this report consistent regardless of whether it is generated as a report during setup or invoked from the command lines `view_coloring` option.

If bokeh is unavailable, then the total coloring report will be created using plain text via `Coloring.display_txt`.

- Added function `display_coloring` to `openmdao.api` that serves as a general interface to both `Coloring.display_bokeh` and `Coloring.display_txt`.
- Updated documentation to include image from the new viewer as well an explanation of `display_coloring`.
- `Coloring.summary` now accepts an output stream argument.

- Minimum bokeh version updated from 1.3.4 to 2.4.0 (Sept 2021)

### Related Issues

- Resolves #3005 

None

### New Dependencies

None


openmdao/utils/coloring.py
```diff
@@ -876,6 +965,232 @@ def on_resize(event):
...
+            data['fwd_color_idx'] = np.full(nrows * ncols, '', dtype=object)
+            data['rev_color_idx'] = np.full(nrows * ncols, '', dtype=object)
+
+            # Add the color group information to the data source
+            fwd_map = {}
+            if coloring._fwd is not None:
+                for idx_fwd, (_cols, _nz_rows) in enumerate(coloring.color_nonzero_iter('fwd')):
+                    for _row_idx, _col_idx in zip(_nz_rows, _cols):
+                        fwd_grp_idxs = itertools.product(_row_idx, [_col_idx])
+                        fwd_map.update({(i, j): idx_fwd for i, j in fwd_grp_idxs})
```
naylor-b, line 1051: Since _col_idx is a single value, couldn't you just get rid of the product call and do `fwd_map.update({(i, _col_idx): idx_fwd for i in _row_idx})`?

naylor-b (approved):

swryan (approved):

PR #3015, opened 2023-09-06
fixed doc error describing 'openmdao timing'

### LINKED ISSUE #3014: Docs describing command line needed for `openmdao timing` are incorrect
### Task Description

The example command line leaves out the necessary beginning of `mpirun -n <nprocs>`.  Without that it's not possible to generate timing of MPI procs.

naylor-b:
### Related Issues

- Resolves #3014

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/debugging/profiling/timing.ipynb
+++ b/openmdao/docs/openmdao_book/features/debugging/profiling/timing.ipynb
@@ -31,7 +31,7 @@
     "\n",
     "The simplest way to use it is via the command line using the `openmdao timing` command. For example:\n",
     "``` bash\n",
-    "    openmdao timing <your_python_script_here>\n",
+    "    mpirun -n <nprocs> openmdao timing <your_python_script_here>\n",
     "```\n",
     "\n",
     "This will collect the timing data and generate a text report that could look something like this:\n",
--- a/openmdao/docs/openmdao_book/features/debugging/profiling/timing.ipynb
+++ b/openmdao/docs/openmdao_book/features/debugging/profiling/timing.ipynb
@@ -24,9 +24,8 @@
    "source": [
     "# Timing Systems under MPI\n",
     "\n",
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore (approved):

robfalck (approved):

openmdao/docs/openmdao_book/features/debugging/profiling/timing.ipynb
```diff
@@ -31,7 +31,7 @@
     "\n",
     "The simplest way to use it is via the command line using the `openmdao timing` command. For example:\n",
     "``` bash\n",
-    "    openmdao timing <your_python_script_here>\n",
+    "    mpirun -n <nprocs> openmdao timing <your_python_script_here>\n",
```
swryan, line 34: I think the paragraph above is still somewhat ambiguous about whether this tool is applicable when NOT running under MPI:
```
    "There is another way to view timings for methods called on specific group and component instances in \n",
    "an OpenMDAO model.  Not only does it have lower overhead than the previously discussed instance-based \n",
    "profiler, but it also displays, when running your model under MPI, how that model spends its time \n",
    "across MPI processes.\n",```
```
Is this still useful for non-MPI use cases, with the advantage of lower overhead?

The change to the om_command page implies that this is really intended _only_ for MPI use cases.
If so, I think that moving the qualifying clause to the first sentence might be more consistent with that intent:
```
    "There is another way to view timings for methods called on specific group and component instances in \n",
    "an OpenMDAO model when running under MPI.",
    "Not only does it have lower overhead than the previously discussed instance-based \n",
    "profiler, but it also displays how that model spends its time \n",
    "across MPI processes.\n",```
```

naylor-b: I reworded the paragraph to make it clear that it's intended to use only under MPI.

PR #3025, opened 2023-09-13
networkx version update.

carlos-pereyra:
### Summary

Update networkx dependency to later version.

### Related Issues

Security.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -187,7 +187,7 @@ setup(
     },
     python_requires=">=3.7",
     install_requires=[
-        'networkx>=2.0',
+        'networkx>=3.0',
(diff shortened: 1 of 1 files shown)


swryan:
Can you be more specific about what the security issue is?  I'm not seeing anything on OSV:

https://osv.dev/list?ecosystem=PyPI&q=networkx

carlos-pereyra:
This ticket can be closed. I realized the issue I see has more to do with Python's xml library https://docs.python.org/3/library/xml.html

As long as a sufficiently new version of Python >= 3.7 is used then expat doesn't become an issue and likewise for networkx.

PR #3028, opened 2023-09-15
Fix for issue where the desvars and responses are sorted in a different order if the number of procs is less than the number of subsystems in a par group.

### LINKED ISSUE #3027: In problems with a parallelgroup that has more components than procs, the constraints and design variables are in a different order than they are in serial or when there are adequate procs.
### Description

In problems with a parallelgroup that has more components than procs, the constraints and design variables are in a different order than they are in serial or when there are adequate procs. This happens because the subsystems are farmed out to the procs sequentially, so for (climb, cruise, descent) under 2 procs, rank 0 gets climb and descent and rank 1 gets cruise.

One way to fix this would be to preserve the base order when assembling the desvar and response data structures.

### Example

Any aviary benchmark will show this.

### OpenMDAO Version

3.28-1.dev

All envs.

Kenneth-T-Moore:
### Summary

Fix an issue where the desvars and responses are sorted in a different order if the number of procs is less than the number of subsystems in a parallel group. This order difference was causing the results to be different on some sensitive problems, and it made debug_print and the pyoptsparse output more difficult to diff.

### Related Issues

- Resolves #3027

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -4813,6 +4813,23 @@ class Group(System):
         else:
             yield from self._subsystems_myproc
 
+    def _sorted_sys_iter_all_procs(self):
+        """
+        Yield subsystem names in sorted order if Problem option allow_post_setup_reorder is True.
+
+        Otherwise, yield subsystem names in the order they were added to their parent group.
+
+        Yields
+        ------
+        System
+            A subsystem.
+        """
+        if self._problem_meta['allow_post_setup_reorder']:
+            for s in sorted(self._subsystems_allprocs):
+                yield s
+        else:
+            yield from self._subsystems_myproc
+
     def _solver_subsystem_iter(self, local_only=False):
         """
         Iterate over subsystems that are being optimized.
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -4828,7 +4828,7 @@ class Group(System):
             for s in sorted(self._subsystems_allprocs):
                 yield s
         else:
-            yield from self._subsystems_myproc
+            yield from self._subsystems_allprocs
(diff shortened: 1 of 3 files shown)


openmdao/core/system.py
```diff
@@ -3523,32 +3523,59 @@ def get_design_vars(self, recurse=True, get_sizes=True, use_prom_ivc=True):
...
-                    out.update(dvs)
-
             if (self.comm.size > 1 and self._subsystems_allprocs and
                     self._mpi_proc_allocator.parallel):
-                my_out = out
-                out = {}
-                for all_out in self.comm.allgather(my_out):
-                    for name, meta in all_out.items():
+
+                # For parallel groups, we need to make sure that the design varaible dictionary is
```
naylor-b, line 3529: typo: varaible

openmdao/core/group.py
```diff
@@ -4813,6 +4813,23 @@ def _sorted_sys_iter(self):
...
+
+        Otherwise, yield subsystem names in the order they were added to their parent group.
+
+        Yields
+        ------
+        System
+            A subsystem.
+        """
+        if self._problem_meta['allow_post_setup_reorder']:
+            for s in sorted(self._subsystems_allprocs):
```
naylor-b, line 4828: Did it break something if you didn't sort these?  It doesn't seem like it should matter.

Kenneth-T-Moore: Yeah, they weren't in the right order if I didn't sort them. The serial order also sorts them like this.

naylor-b (approved):

swryan (approved):

PR #3029, opened 2023-09-18
Fixed the `openmdao n2` command to work properly with subproblems

### LINKED ISSUE #3023: The `openmdao n2` command shows the N2 for a sub-problem instead of the top-level Problem
### Description

Because a sub-Problem completes it's setup() first, it will be the first to encounter the hook set by the `openmdao n2` command, causing the N2 diagram to be displayed for that sub-Problem followed immediately by a `sys.exit()`.

The `openmdao n2` command should, by default, show the N2 diagram for the top-level problem.

A new `--problem` argument should be added to the `openmdao n2` command so that the user can specify the pathname of a sub-problem if they want to see the N2 for that instead of for the top-level Problem.

### Example

Currently, invoking `openmdao n2` on the following script will display the N2 diagram for the sub-Problem:
```
if __name__ == '__main__':

        import openmdao.api as om

        # create sub problem
        submodel = om.Group()
        submodel.add_subsystem('subcomp', om.ExecComp('y = x1**2 + x2**2 + x3**2'), promotes=['*'])

        subprob = om.Problem(submodel)

        # create top problem
        p = om.Problem()
        p.model.add_subsystem('submodelcomp',
                              om.SubmodelComp(problem=subprob, inputs=['*'], outputs=['*']),
                              promotes=['*'])
        p.model.add_subsystem('supercomp',
                              om.ExecComp('z = 3 * y'),
                              promotes=['*'])

        p.setup()
        p.final_setup()

        # verify model runs correctly
        p.set_val('x1', 1)
        p.set_val('x2', 2)
        p.set_val('x3', 3)

        p.run_model()
```

### OpenMDAO Version

3.28.1-dev

_No response_

swryan:
### Summary

- added the `--problem` argument to the N2 command to specify the problem or subproblem to display
  - by default, the top-level problem will be shown, whereas previously it would display a subproblem
- disabled the report system when running the n2 or scaling report command
  - the report system was setting hooks that were interfering with generating the n2 for the desired problem
 
Also:
- set a maximum object size which will exclude very large values (including option values) from the viewer data
  - this prevents a situation where JSON data cannot be encoded/decoded for recording or viewing in the N2
- removed the value key entirely when values are excluded from N2 data, rather than setting values to None

### Related Issues

- Resolves #3023

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/n2_viewer.py
+++ b/openmdao/visualization/n2_viewer/n2_viewer.py
@@ -33,7 +33,7 @@ from openmdao.visualization.htmlpp import HtmlPreprocessor
 from openmdao import __version__ as openmdao_version
 
 _MAX_ARRAY_SIZE_FOR_REPR_VAL = 1000  # If var has more elements than this do not pass to N2
-_MAX_OBJECT_SIZE = 1000000           # If object is bigger than this do not pass to N2
+_MAX_OBJECT_SIZE = int(1e7)          # If option value is bigger than this do not pass to N2
 
 _default_n2_filename = 'n2.html'
 
--- a/openmdao/visualization/n2_viewer/n2_viewer.py
+++ b/openmdao/visualization/n2_viewer/n2_viewer.py
@@ -164,11 +164,8 @@ def _get_var_dict(system, typ, name, is_parallel, is_implicit, values):
                     # which could be remote under MPI
                     _get_array_info(system, vec, name, prom, var_dict, from_src=False)
 
-            else:
-                var_dict['val'] = None
         except Exception as err:
             issue_warning(str(err))
-            var_dict['val'] = None
     else:  # discrete
         meta = system._var_discrete[typ][name]
         val = meta['val']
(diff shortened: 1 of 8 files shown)


openmdao/visualization/n2_viewer/n2_viewer.py
```diff
@@ -32,6 +33,7 @@
 from openmdao import __version__ as openmdao_version
 
 _MAX_ARRAY_SIZE_FOR_REPR_VAL = 1000  # If var has more elements than this do not pass to N2
+_MAX_OPTION_SIZE = int(1e7)          # If option value is bigger than this do not pass to N2
```
Kenneth-T-Moore, line 36: I wonder if it should be even smaller, since our max array size is only 1000 elements.  I guess this is bytes though; I am not sure how many bytes are in a 1000 element array in numpy.

Kenneth-T-Moore (approved):

openmdao/visualization/n2_viewer/n2_viewer.py
swryan: yeah.. I set it to the absolute max order of magnitude before it would cause an encoding error...  we could definitely set it to a more practical value, since large values are pretty useless visually in the n2 anyway...

```
>>> k.shape
(1000,)
>>> sys.getsizeof(k)
8112
>>> j.shape
(1000, 1000)
>>> sys.getsizeof(j)
8000128
```

swryan: I changed it to `1e4`..  because larger values in the N2 don't really make sense, per above, and because a user may inadvertently and unnecessarily incur the performance overhead of encoding the larger value for no good reason.

robfalck (approved):

naylor-b (approved):

PR #3039, opened 2023-09-29
Added a warning when an excludes or includes recording option has no matches

### LINKED ISSUE #3018: recording_options `includes` and `excludes` should warn when no matches are found for an entry
# Raise warning for missing comma's in includes/excludes list

I think it could be useful to raise a warning when the user forgets to include comma's in the 'includes' or 'excludes' list. It's a silly mistake but if most auto-formatters will break the list into several lines and it is thus easy to forget to place comma's in between. An example given below:

### Wrong
Doesn't raise a warning but also doesn't record OPENAEROSTRUCT.AS_point_0.wing_perf.Cl
```
prob.driver.recording_options['includes'] = ["OPENAEROSTRUCT.wing.geometry.twist",
                                                 "OPENAEROSTRUCT.wing.geometry.chord"
                                                 "OPENAEROSTRUCT.AS_point_0.wing_perf.Cl",
                                                 "OPENAEROSTRUCT.AS_point_0.wing_perf.CDi",
                                                 'OPENAEROSTRUCT.AS_point_0.total_perf.L',
                                                 'OPENAEROSTRUCT.AS_point_0.total_perf.D',
                                                 'RETHORST.velocity_distribution']
```

### Correct
```
prob.driver.recording_options['includes'] = ["OPENAEROSTRUCT.wing.geometry.twist",
                                                 "OPENAEROSTRUCT.wing.geometry.chord",
                                                 "OPENAEROSTRUCT.AS_point_0.wing_perf.Cl",
                                                 "OPENAEROSTRUCT.AS_point_0.wing_perf.CDi",
                                                 'OPENAEROSTRUCT.AS_point_0.total_perf.L',
                                                 'OPENAEROSTRUCT.AS_point_0.total_perf.D',
                                                 'RETHORST.velocity_distribution']
```

Perhaps I'm the only one making this mistake but it has happened several times.

https://github.com/OpenMDAO/OpenMDAO/blob/671375fd394bd5cb17ec460a2f8dd9984bca0b1b/openmdao/core/problem.py#L345

swryan:
### Summary

Added a check that all exclude/include globs in `recording_options` for Problem, Driver or System have at least one matching output or input name in the set of inputs, outputs and/or residuals to be recorded.

### Related Issues

- Resolves #3018 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1980,7 +1980,7 @@ class System(object):
                     myresiduals = myoutputs
 
             elif options['record_residuals']:
-                match_names = match_names | self._residuals.keys()
+                match_names = match_names | set(self._residuals.keys())
                 myresiduals = [n for n in self._residuals._abs_iter()
                                if check_path(abs2prom_output[n], incl, excl)]
 
@@ -1988,11 +1988,11 @@ class System(object):
             for pattern in excl:
                 if not has_match(pattern, match_names):
                     issue_warning(f"{self.msginfo}: No matches for pattern '{pattern}' in "
-                                "recording_options['excludes'].")
+                                  "recording_options['excludes'].")
             for pattern in incl:
                 if not has_match(pattern, match_names):
                     issue_warning(f"{self.msginfo}: No matches for pattern '{pattern}' in "
-                                "recording_options['includes'].")
+                                  "recording_options['includes'].")
 
             self._filtered_vars_to_record = {
                 'input': myinputs,
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -1960,14 +1960,14 @@ class System(object):
             # includes and excludes for inputs are specified using _absolute_ names
             # vectors are keyed on absolute name, discretes on relative/promoted name
             if options['record_inputs']:
-                match_names = match_names | set(abs2prom_inputs.keys())
+                match_names.update(abs2prom_inputs.keys())
                 myinputs = sorted([n for n in abs2prom_inputs
                                    if check_path(n, incl, excl)])
 
             # includes and excludes for outputs are specified using _promoted_ names
             # vectors are keyed on absolute name, discretes on relative/promoted name
             if options['record_outputs']:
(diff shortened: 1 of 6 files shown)


Kenneth-T-Moore (approved):

openmdao/core/system.py
```diff
@@ -1948,16 +1948,27 @@ def _setup_recording(self):
...
+
+            # set of promoted output names and absolute input and residual names
+            # used for matching includes/excludes
+            match_names = set()
+
             # includes and excludes for inputs are specified using _absolute_ names
             # vectors are keyed on absolute name, discretes on relative/promoted name
             if options['record_inputs']:
-                myinputs = sorted([n for n in self._var_abs2prom['input']
+                match_names = match_names | set(abs2prom_inputs.keys())
```
naylor-b, line 1963: should just use `match_names.update(abs2prom_inputs)` here so as not to create any unnecessary temporary sets.

openmdao/core/system.py
```diff
@@ -1969,9 +1980,19 @@ def _setup_recording(self):
...
             elif options['record_residuals']:
-                abs2prom = self._var_abs2prom['output']
+                match_names = match_names | set(self._residuals.keys())
                 myresiduals = [n for n in self._residuals._abs_iter()
-                               if check_path(abs2prom[n], incl, excl)]
+                               if check_path(abs2prom_output[n], incl, excl)]
+
+            # check that all exclude/include globs have at least one matching output or input name
+            for pattern in excl:
+                if not has_match(pattern, match_names):
```
naylor-b, line 1989: It's kind of a bummer that we're doing the glob pattern match a second time for every variable just to check for a bad pattern. Could we improve things a little by only checking the `excl` patterns against the vars that we did match and only checking the 'incl' patterns against vars that we didn't match? I guess that could miss cases where you had multiple patterns where an earlier one matched and a later one didn't, but would the user even care in that case since the intended set of vars was matched even though certain patterns didn't match anything?

naylor-b (approved):

PR #3055, opened 2023-10-23
Changed memleak test to be skipped if coverage is running

### LINKED ISSUE #3054: `test_setup_memleak` fails when run with coverage>=6.0
### Task Description

`testflo` has been updated to use `coverage>=6.0` but the tracing done in the new version of coverage interacts with the trace done by the test (openmdao/core/tests/test_setup_memleak.py), causing the test to fail under most circumstances.

The test should either disable coverage or be skipped if coverage is running.

swryan:
### Summary

Changed `test_setup_memleak` test to be skipped if coverage is running.

Also:
- removed some unnecessary snapshot saving in the test
- disable a common warning coming from the new version of coverage

### Related Issues

- Resolves #3054

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_setup_memleak.py
+++ b/openmdao/core/tests/test_setup_memleak.py
@@ -42,7 +42,6 @@ class TestSetupMemLeak(unittest.TestCase):
 
         for memtest in ITERS:
             prob.model = SellarMDA()
-            snapshots = []
 
             for i in range(memtest):
                 prob.setup(check=False) # called here causes memory leak
--- a/openmdao/core/tests/test_setup_memleak.py
+++ b/openmdao/core/tests/test_setup_memleak.py
@@ -1,6 +1,7 @@
 import unittest
 import tracemalloc
 import gc
+import os
 import os.path
 
 import openmdao.api as om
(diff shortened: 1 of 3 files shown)


robfalck:
Should we be updating the testflo version to avoid the speed in python 3.11/3.12

swryan:
> Should we be updating the testflo version to avoid the speed in python 3.11/3.12

Good catch, I forgot that bit.

I removed the restriction so it will pull the latest version.

swryan:
> > Should we be updating the testflo version to avoid the speed in python 3.11/3.12
> 
> Good catch, I forgot that bit.
> 
> I removed the restriction so it will pull the latest version.

Actually that was a mistake.. I jumped to quickly.  The restriction was a _minimum_ version...   we could update the minimum to be the (upcoming) new version, but I think that's an unnecessary restriction since the old version still works, just not optimally.

robfalck (approved):

openmdao/core/tests/test_setup_memleak.py
```diff
@@ -41,9 +43,11 @@ def test_setup_memleak(self):
                 # Force garbage collection now instead of waiting for an
                 # optimal/arbitrary point since we're tracking memory usage
                 gc.collect()
-                snapshots.append(tracemalloc.take_snapshot())
+                if i == 0:
```
naylor-b, line 46: Shouldn't we be taking the first snapshot *before* prob.setup?

swryan: The way I read the existing code was that it checks that the memory footprint does not change with multiple setup calls, rather than between pre-setup and setup (times n)

naylor-b (approved):

PR #3073, opened 2023-11-21
Put optimizer generated output files into the reports directory

### LINKED ISSUE #3060: Modify the reports system to put all reports in one directory

Change the behavior of the reports system in two ways. The goal is to have one directory that provides a complete, accurate documentation of the run.

1. Before any reports are written during a script run, delete any existing files in the directory. This prevents confusion when making multiple runs and requesting different reports for the runs
2. Have any kind of reports files go to the reports subdirectory (e.g. ./reports/name_of_script). This includes the coloring files and also any files written by the optimizers (e.g. IPOPT.out, SNOPT_summary.out, SNOPT_print.out, pyopt_solution.txt )

### Associated POEM

_No response_

hschilling:
### Summary

To have all the output reports in a single location, the optimizer reports for pyoptsparse will now, by default, go into the reports directory. 

The user can change that. If the output_dir is set to None, then those files will go into the current working directory. If set to a string, that will be used as the location. 

Also, some minor fixes were made.

The code does not handle all the pyoptsparse optimizers. NLPQLP and ParOpt were not included because they don't work. 

### Related Issues

- Resolves #3060 

The default location of the optimizer generated reports is now the reports directory. Previously they were in the current working directory.

### New Dependencies

None


naylor-b (changes requested):

openmdao/core/constants.py
```diff
@@ -128,3 +128,11 @@ def __deepcopy__(self, memo):
...
 # out_stream argument. We run into problems running testflo if we use a default of sys.stdout.
 _DEFAULT_OUT_STREAM = _ReprClass("DEFAULT_OUT_STREAM")
+
+# Used in pyOptSparseDriver and Problem for coloring dir. The default is the reports directory
+#  which includes the directory
+#  named after the Problem name. But when the declare method in that class is called, the driver
+#  does not have a reference to the Problem so can't get the name. This serves as a flag that
+#  the default directory in the reports directory is what is wanted. Then in the run method,
+#  the actual default directory is used
+_DEFAULT_REPORTS_DIR = _ReprClass("DEFAULT_PYOPT_SPARSE_OUTPUT_DIR")
```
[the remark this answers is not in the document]
hschilling: VERY weird. Sorry, that was left over from when this was just going to be used for pyoptsparse optimizer output files. Fixed.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -374,6 +379,39 @@ def run(self):
...
         optimizer = self.options['optimizer']
+
+        # Need to tell optimizer where to put its .out files
+        if self.options['output_dir'] is None:
+            self.options['output_dir'] = "."
+        elif self.options['output_dir'] == _DEFAULT_REPORTS_DIR:
+            problem = self._problem()
+            default_output_dir = pathlib.Path(get_reports_dir()).joinpath(problem._name)
+            pathlib.Path(default_output_dir).mkdir(parents=True, exist_ok=True)
+            self.options['output_dir'] = str(default_output_dir)
```
swryan, line 390: I'm not sure it's a good idea to be changing the value of the user facing 'output_dir' option here, since that is normally under control of the user.  Couldn't you just set the local output_dir variable, which is all that your using in subsequent code anyway?

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -374,6 +379,39 @@ def run(self):
...
+        output_dir = self.options['output_dir']
+
+        if optimizer == 'ALPSO':  # Actually, this is the root of two files generated
```
swryan, line 393: I think you could just use a dictionary like you do in the test and just pull the setting for the optimizer you need rather than have this extended if/then/else construct.

You might also consider moving this code down to around line 572 where it would be collocated with other opt_setting logic

swryan (changes requested):

openmdao/drivers/pyoptsparse_driver.py
hschilling: Correct. Fixed

swryan (approved):

PR #3078, opened 2023-11-29
Allow creation of N2 diagrams for problems that don't exist on global…

### LINKED ISSUE #3077: Allow creation of N2 diagrams for problems that don't exist on global root proc

I am currently running code in which I have multiple MPI communicators, each of which has their own OpenMDAO problem. I noticed that despite calling the `n2` method for the problems on all communicators, only the n2 for the problem on the first communicator would be created.

This is happening because the `n2` method only runs on the global root proc:
https://github.com/OpenMDAO/OpenMDAO/blob/fee4510212d6c91cd1499b4dfa49ffdff4b969e4/openmdao/visualization/n2_viewer/n2_viewer.py#L644-L646

Hence, for any problem whose communicator doesn't include the global root proc, the function will do nothing.

Would it be possible to modify the logic statement above to use the problem's comm in place of `MPI.COMM_WORLD` when the data source passed to the `n2` method is an OpenMDAO problem? 

### Proposed solution

```python
if MPI:
    try:
        comm = data_source.comm
    Except AttributeError:
        comm = MPI.COMM_WORLD
    if comm.rank != 0:
        return
```

### Associated POEM

_No response_

A-CGray:
### Summary

Allows the n2 diagram creator to be run for problems that do not exist on the global root proc by running on the root of the problem's comm rather than the root of the global comm.

### Related Issues

- Resolves #3077 

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/n2_viewer.py
+++ b/openmdao/visualization/n2_viewer/n2_viewer.py
@@ -641,9 +641,16 @@ def n2(data_source, outfile=_default_n2_filename, path=None, values=_UNDEFINED, 
         err_msg = str(err)
         issue_warning(err_msg)
 
-    # if MPI is active only display one copy of the viewer
-    if MPI and MPI.COMM_WORLD.rank != 0:
-        return
+    # If MPI is active only display one copy of the viewer.
+    # If the data_source is a Problem, only run on the root proc of its comm.
+    # Otherwise, only run on the global root proc.
+    if MPI:
+        try:
+            comm = data_source.comm
+        except AttributeError:
+            comm = MPI.COMM_WORLD
+        if comm.rank != 0:
+            return
 
     options = {}
     model_data['options'] = options


naylor-b (approved):

robfalck (approved):

swryan:
would you be able to put together a small unit test to demonstrate your use case and the correct functionality per your fix?

A-CGray:
> would you be able to put together a small unit test to demonstrate your use case and the correct functionality per your fix?

Sure, I added a test that creates 2 problems on two comms and checks that they both write out files. I also added an assert to check the output file exists to the existing `N2ParallelTestCase` because it seemed like all that test was doing was checking that the call to `n2` doesn't hang. Feel free to remove if you want.

swryan:
Note: The 'Ubuntu Minimal` will fail, just ignore that.  we have a fix on another PR.

PR #3092, opened 2024-01-03
In the Custom report example, make sure the entry point doesn't collide with our default names.

Kenneth-T-Moore:
### Summary

In the Custom report example, make sure the entry point doesn't collide with our default names.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
+++ b/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
@@ -269,7 +269,10 @@
     "        }\n",
     "    ```\n",
     "\n",
-    "    As shown above, report entry points are part of the `openmdao_report` entry point group.\n"
+    "    As shown above, report entry points are part of the `openmdao_report` entry point group.\n",
+    "    \n",
+    "    Note that the name for your entry point needs to be unique. Here, we use the name \"summary_report\" because\n",
+    "    an entry point named \"summary\" already exists for one of the built-in reports.\n"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


swryan:
I would suggest that the doc probably needs to tell the user how to avoid name collisions as well...   I think we could tell them to do something like this to determine existing names:
```
from importlib.metadata import entry_points
eps = entry_points(group='openmdao_reports')
print([ep.name for ep in eps])
```

Kenneth-T-Moore:
> I would suggest that the doc probably needs to tell the user how to avoid name collisions as well... I think we could tell them to do something like this to determine existing names:
> 
> ```
> from importlib.metadata import entry_points
> eps = entry_points(group='openmdao_reports')
> print([ep.name for ep in eps])
> ```

This feels like something that might be useful to put into the command line interface.

swryan:
> > I would suggest that the doc probably needs to tell the user how to avoid name collisions as well... I think we could tell them to do something like this to determine existing names:
> > ```
> > from importlib.metadata import entry_points
> > eps = entry_points(group='openmdao_reports')
> > print([ep.name for ep in eps])
> > ```
> 
> This feels like something that might be useful to put into the command line interface.

I had that thought as well, although this is more of a developer-user thing than a user-user thing, if that makes sense ¯\_(ツ)_/¯

Kenneth-T-Moore:
I pushed up a sentence about making sure the names are unique. Sadly, `entry_points` doesn't have a "group" argument in python 3.8, and I just get an empty list in 3.10.8, so I didn't include any way to do it.

swryan (approved):

PR #3112, opened 2024-02-05
Added option `stall_tol_type` to Nonlinear solvers

### LINKED ISSUE #3100: Allow user to specify stall tolerance type for Newton and Broyden solvers.

Currently Newton and Broyden solvers accept `stall_limit` and `stall_tol` for situations where the residual magnitude isn't decreasing after several iterations, but they use the relative tolerance as the metric for this.

In situations where the system starts out converged or nearly so, the relative tolerance can be a poor choice for stall detection.

## Propsed Solution

Add option `stall_tol_type` to the Newton and Broyden solvers, with a default value of `'rel'` (the current behavior), but also allow it to accept `'abs'`.

### Associated POEM

N/A

robfalck:
### Summary

The nonlinear solvers now have an option `stall_tol_type` which can be `'abs'` or `'rel'`.
This option determines whether the `stall_tol` specified by the user is the absolute or relative
residual norm.

### Related Issues

- Resolves #3100

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/building_blocks/solvers/broyden.ipynb
+++ b/openmdao/docs/openmdao_book/features/building_blocks/solvers/broyden.ipynb
@@ -419,7 +419,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "**stall_limit and stall_tol**\n",
+    "**stall_limit, stall_tol, and stall_tol_type **\n",
     "\n",
     "  In some cases, nonlinear solvers can stall out where the norm of the residual stops changing at all. This\n",
     "  can happen for a couple of reasons. You can hit numerical noise problems and just be wandering around in\n",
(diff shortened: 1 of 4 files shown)


naylor-b (approved):

openmdao/visualization/n2_viewer/tests/sellar_initial_values.json
```diff
@@ -253,17 +254,17 @@
                             "desc": "",
```
Kenneth-T-Moore, line 254: Why did all of the initial values for Sellar get zeroed out in this json?

Kenneth-T-Moore: oh, it is probably because you are using save_viewer_data now.

Kenneth-T-Moore (approved):

swryan (approved):

PR #3117, opened 2024-02-08
Skip any relevance/singular_jac checks when `approx_totals` is used

### LINKED ISSUE #3026: Error for constraints not dependent on design vars does not work well for `approx_totals` models
### Description

For Issue #2980, [this change](https://github.com/OpenMDAO/OpenMDAO/pull/2997/files#diff-0f5666af4457f169f61dab7eef161117cefc6c8424bca49eb51b06f4c5dad9c7) was made in PR #2997 .  However, for models that rely on `approx_totals` to compute total derivatives, there often isn't a `self.declare_partials` line in the individual components or groups.  Ideally, the matrix mapping design variables and constraints is also built for models that use `approx_totals`.  Secondly, I request that [this new error](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/driver.py#L951-L952) be pushed inside of an if-statement [similar to as is done here](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/core/group.py#L863-L864).  That way, there is at least an option that can be passed to the driver to suppress it.

In practice, many of our constraints are long vectors along a physical components (blade / wing / structure / tower).  Some of those vector indices may be locked for other reasons and cannot be impacted by the design variables.  Which indices are locked may be use-case dependent and not something we want to hard code into the constraint declaration.  In general, I'm not in favor of this recent change to default to "error" instead of "warn" in these instances, but I acknowledge I may be in the minority there.

### Example

I have hacked a current test file to replicate the issue:
```
import unittest

import numpy as np
import openmdao.api as om

step = 1e-6
size = 3


class FakeGeomComp(om.ExplicitComponent):

    def initialize(self):
        self.options.declare("n", types=int)

    def setup(self):
        n = self.options["n"]
        self.add_input("x0", val=np.zeros(n), units="m")
        self.add_input("feather", val=0.0, units="deg")
        self.add_output("x", val=np.zeros(n), units="m")

        #self.declare_partials("*", "*", method="fd")

        self._counter = 0

        self._counter += 1
        feather_rad = inputs["feather"][0]*np.pi/180
        x0 = inputs["x0"]

        outputs["x"][:] = 3*np.sin(feather_rad + 0.2)*x0 + 3*feather_rad**2


class FakeAeroComp(om.ExplicitComponent):

    def initialize(self):
        self.options.declare("n", types=int)

    def setup(self):
        n = self.options["n"]

        self.add_input("x", val=np.arange(n), units="m")
        self.add_input("omega", val=7000*2*np.pi/60, units="rad/s")

        self.add_output("CT", val=0.5)
        self.add_output("CP", val=0.5)

        #self.declare_partials("*", "*", method="fd")

        self._counter = 0

        self._counter += 1
        omega = inputs["omega"][0]
        x = inputs["x"]

        outputs["CT"][0] = 0.8*omega**2 + np.sum(x)
        outputs["CP"][0] = 0.1*omega**3 + np.sum(x**2)


class GeometryAndAero(om.Group):

    def initialize(self):
        self.options.declare("n", types=int)
        self.options.declare("rho", types=float)
        self.options.declare("vinf", types=float)

    def setup(self):
        n = self.options["n"]

        comp = self.add_subsystem("init_geom", om.IndepVarComp(), promotes_outputs=["x0"])
        comp.add_output("x0", val=np.arange(n) + 1.0, units="m")

        self.add_subsystem("geom", FakeGeomComp(n=n), promotes_inputs=["x0", "feather"], promotes_outputs=["x"])
        self.add_subsystem("aero", FakeAeroComp(n=n), promotes_inputs=["x", "omega"], promotes_outputs=["CT", "CP"])

    def reset_count(self):
        self.geom._counter = 0
        self.aero._counter = 0


class TestSemiTotalsNumCalls(unittest.TestCase):

    def test_call_counts(self):
        size = 10
        rho = 1.17573
        minf = 0.111078231621482
        speedofsound = 344.5760217432
        vinf = minf*speedofsound

        prob = om.Problem()
        prob.driver = om.ScipyOptimizeDriver(singular_jac_behavior="warn")
        prob.driver.options["optimizer"] = "SLSQP"
        prob.driver.options["singular_jac_behavior"] = "warn"

        omega = 7199.759242*2*np.pi/60
        ivc = prob.model.add_subsystem("ivc", om.IndepVarComp(), promotes_outputs=["*"])
        ivc.add_output("feather", val=0.0, units="deg")
        ivc.add_output("omega", val=omega, units="rad/s")

        geom_and_aero = prob.model.add_subsystem('geom_and_aero',
                                                GeometryAndAero(n=size, rho=rho, vinf=vinf),
                                                promotes_inputs=["feather", "omega"],
                                                promotes_outputs=["CT", "CP"])
        geom_and_aero.approx_totals(step=step, step_calc="abs", method="fd", form="forward")

        prob.model.add_design_var("feather", lower=-5.0, upper=25.0, units="deg", ref=1.0)
        prob.model.add_design_var("omega", lower=3000*2*np.pi/60, upper=7500*2*np.pi/60, units="rad/s", ref=1.0)
        prob.model.add_objective("CP", ref=1e0)
        prob.model.add_constraint("CT", lower=0.0)

        prob.setup(force_alloc_complex=False)

        omega = (6245.096992023524*2*np.pi/60) + step
        feather = 0.6362159381168669
        prob.set_val("omega", omega, units="rad/s")
        prob.set_val("feather", feather, units="deg")
        prob.run_driver()
        self.assertTrue(True)

if __name__ == "__main__":
    unittest.main()
```

### OpenMDAO Version

3.28.1-dev

Using OpenMDAO main-branch

robfalck:
### Summary

When a group in OpenMDAO uses approx_totals, the user is not required to explicitly declare partials in their systems.
This renders the relevance_graph in OpenMDAO useless, because it doesn't capture this dependency information.
Checks that use this relevance graph incorrectly report singular rows or columns in the jacobian, and the driver `check_relevance` method raises an exception unnecessarily.

This PR fixes this behavior by skipping these checks when any group in the system tree uses `approx_totals`.

### Related Issues

- Resolves #3026

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -864,17 +864,13 @@ class Group(System):
                     missing_responses.add(output)
 
         if missing_responses:
-            for subsys in self.system_iter(include_self=True, recurse=True, typ=Group):
-                if subsys._has_approx:
-                    break
+            msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
+                    " be impacted by the design variables of the problem because no partials "
+                    "were defined for them in their parent component(s).")
+            if self._problem_meta['singular_jac_behavior'] == 'error':
+                raise RuntimeError(msg)
             else:
-                msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
-                       " be impacted by the design variables of the problem because no partials "
-                       "were defined for them in their parent component(s).")
-                if self._problem_meta['singular_jac_behavior'] == 'error':
-                    raise RuntimeError(msg)
-                else:
-                    issue_warning(msg, category=DerivativesWarning)
+                issue_warning(msg, category=DerivativesWarning)
 
         self._relevance_graph = graph
         return graph
@@ -4037,6 +4033,8 @@ class Group(System):
         missing : dict
             Dictionary containing list of missing derivatives keyed by system pathname.
         """
+        if self._has_approx:
+            return
         for subsys in self._subsystems_myproc:
             subsys._get_missing_partials(missing)
 
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -865,8 +865,8 @@ class Group(System):
 
         if missing_responses:
             msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
-                    " be impacted by the design variables of the problem because no partials "
-                    "were defined for them in their parent component(s).")
+                   " be impacted by the design variables of the problem because no partials "
+                   "were defined for them in their parent component(s).")
             if self._problem_meta['singular_jac_behavior'] == 'error':
                 raise RuntimeError(msg)
             else:
@@ -4026,7 +4026,7 @@ class Group(System):
 
     def _get_missing_partials(self, missing):
         """
-        Return a list of (of, wrt) tuples for which derivatives have not been declared.
+        Provide (of, wrt) tuples for which derivatives have not been declared in the system.
 
         Parameters
         ----------
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -869,8 +869,8 @@ class Group(System):
                     break
             else:
                 msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
-                    " be impacted by the design variables of the problem because no partials "
-                    "were defined for them in their parent component(s).")
+                       " be impacted by the design variables of the problem because no partials "
+                       "were defined for them in their parent component(s).")
                 if self._problem_meta['singular_jac_behavior'] == 'error':
                     raise RuntimeError(msg)
                 else:
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -864,13 +864,17 @@ class Group(System):
                     missing_responses.add(output)
 
         if missing_responses:
-            msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
-                   " be impacted by the design variables of the problem because no partials "
-                   "were defined for them in their parent component(s).")
-            if self._problem_meta['singular_jac_behavior'] == 'error':
-                raise RuntimeError(msg)
+            for subsys in self.system_iter(include_self=True, recurse=True, typ=Group):
+                if subsys._has_approx:
+                    break
             else:
-                issue_warning(msg, category=DerivativesWarning)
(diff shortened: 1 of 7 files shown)


openmdao/core/group.py
```diff
@@ -864,13 +864,17 @@ def get_relevance_graph(self, desvars, responses):
                     missing_responses.add(output)
 
         if missing_responses:
-            msg = (f"Constraints or objectives [{', '.join(sorted(missing_responses))}] cannot"
-                   " be impacted by the design variables of the problem because no partials "
-                   "were defined for them in their parent component(s).")
-            if self._problem_meta['singular_jac_behavior'] == 'error':
-                raise RuntimeError(msg)
+            for subsys in self.system_iter(include_self=True, recurse=True, typ=Group):
```
naylor-b, line 867: Rather than skipping the entire check, we could just modify the `_get_missing_partials` method in Group and Component to just not report any missing partials within that system (or below in the case of Group).  That way, the dependency graph will still leave those system nodes in the graph (so they'll be treated as 'dense', i.e. all outputs depend on all inputs) for those nodes.

openmdao/core/tests/test_semitotals.py
```diff
@@ -310,3 +318,45 @@ def test_call_counts(self):
         assert_check_totals(data, atol=1e-6, rtol=1e-6)
 
         self.assertEqual(geom_and_aero.geom._counter, 4)
+
+    def test_check_relevlance_approx_totals(self):
```
swryan, line 322: ;)

swryan (approved):

openmdao/core/group.py
```diff
@@ -4030,13 +4026,15 @@ def _declared_partials_iter(self):
...
         """
-        Return a list of (of, wrt) tuples for which derivatives have not been declared.
+        Provide (of, wrt) tuples for which derivatives have not been declared in the system.
 
         Parameters
         ----------
         missing : dict
             Dictionary containing list of missing derivatives keyed by system pathname.
         """
+        if self._has_approx:
```
naylor-b, line 4036: Shouldn't we also do this same sort of thing in Component._get_missing_partials?

naylor-b (changes requested):

openmdao/core/group.py
robfalck: For components, I think _has_approx is too broad of a brush, since it is true if only some partials are approximated. The get_missing_partials method should account for this.

naylor-b (approved):

PR #3137, opened 2024-02-27
Added `set_reports_dir` to openmdao.api.

### LINKED ISSUE #3134: Add `set_reports_dir` to `openmdao.api`

OpenMDAO allows the user to set a top-level reports dir using an environment variable `OPENMDAO_REPORTS_DIR`, and the docs discuss it, but this environment variable is read at import time and so the user doesn't have an obvious way to set the reports directory.

The module `openmdao.utils.reports_system` does have a `set_reports_dir` function to change this, but it is not in an obvious place (should be in `openmdao.api`) and it is not discussed in the docs.

`set_reports_dir` should be added to `openmdao.api`, and discussed as the correct programmatic way to change the directory.

We should also explain to the user that the reports directory is a global setting, so they can't set a different one for different problems executing in parallel under MPI.

### Associated POEM

_No response_

robfalck:
### Summary

The `set_reports_dir` function was in reports_dir but should be part of `openmdao.api`.
This PR adds it to the public facing API and explains its use in the documentation.

### Related Issues

- Resolves #3134 

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
+++ b/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
@@ -94,12 +94,10 @@
     "\n",
     "If the user wishes to have the reports directory placed into a different directory other than the default current working directory, the user has two options for doing so.\n",
     "\n",
-    "First, this option can be set programmatically using the `set_reports_dir` function from `openmdao.api`.\n",
+    "1. This option can be set programmatically using the `set_reports_dir` function from `openmdao.api`.\n",
     "Note that this is a global OpenMDAO setting and therefore it cannot take different values for different processors if running multiple problems simultaneously under MPI.\n",
     "\n",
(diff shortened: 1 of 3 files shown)


naylor-b (approved):

openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
```diff
@@ -92,8 +92,20 @@
...
+    "If the user wishes to have the reports directory placed into a different directory other than the default current working directory, the user has two options for doing so.\n",
     "\n",
+    "First, this option can be set programmatically using the `set_reports_dir` function from `openmdao.api`.\n",
+    "Note that this is a global OpenMDAO setting and therefore it cannot take different values for different processors if running multiple problems simultaneously under MPI.\n",
+    "\n",
+    "The default value may be set using the environment variable `OPENMDAO_REPORTS_DIR` to the name of that directory. The directory does not have to exist beforehand. It will be created by the reporting system. As an example, if `OPENMDAO_REPORTS_DIR` is set to \"my_reports\", then, if the `Problem` name is `problem1`, then the reports will be written to the \"my_reports/problem1\" directory.\n",
```
swryan, line 100: I think you should either bulletize the two options, or start the second sentence with "Alternatively" or "The second option" etc to more clearly demarcate the two options..

PR #3113, opened 2024-02-05
Refactored relevance and removed driver dependency for coloring.

naylor-b:
### Summary

Relevance was refactored, the the currently active relevance is determined by the current set of design variables and responses coming out of the model, or by 'of' and 'wrt' variables passed into Problem.compute_totals which don't have to be declared design variables or responses.

Design variables and responses are also no longer necessary to compute a total coloring.  This will lead to simplification of the SubModelComp (that work will be done in a future PR).

Relevance is now active while computing total derivatives using FD/CS.
Relevance is now active while computing objectives when using pyOptSparseDriver, but not ScipyOptimizeDriver, because activating relevance there caused several dymos test failures. This will be investigated and hopefully fixed in a later PR.

Relevance is not active under DirectSolvers or NewtonSolvers.

### Related Issues

- Resolves #2894
- Resolves #3037

### Backwards incompatibilities

The keys used in dictionaries returned by `compute_totals` will now contain 'user' specified names (either promoted names or aliases) instead of absolute names.  So they now match the names specified in add_design_var/add_constraint/add_objective, subject to promotion rules if they were declared below the top level.

The `rel_systems` arg was removed from some internal methods like `_solve_linear` and `_apply_linear`.  It was left in the `solve` method for linear solvers since that's public, but the arg is deprecated and does nothing now.

### New Dependencies

None

PR #3138, opened 2024-03-06
Added release notes for 3.31.0

robfalck:
### Summary

Added release notes for upcoming 3.31.0 release.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -27,6 +27,10 @@ any changes that break your models, please submit an issue so we can resolve the
 
 ## Backwards Incompatible API Changes
 
+Tthe jacobian as returned by compute_totals is now keyed with the 'user facing' names given to the design variables and responses, rather than the absolute names that were used previously. OpenMDAO's general rule of thumb is that users see "promoted names" or aliases for variables. [#3113](https://github.com/OpenMDAO/OpenMDAO/pull/3113)
+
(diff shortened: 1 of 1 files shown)


release_notes.md
```diff
@@ -1,3 +1,69 @@
...
+no effect after OpenMDAO was imported. The `set_reports_dir` has now been added to `openmdao.api`.
+
+Given some of the significant internal changes in this release, we tested it as thoroughly as we could. If you find
+any changes that break your models, please submit an issue so we can resolve them quickly.
+
+## New Deprecations
+
+- Renamed `list_problem_vars` to `list_driver_vars` [#3132](https://github.com/OpenMDAO/OpenMDAO/pull/3132)
+
+## Backwards Incompatible API Changes
```
swryan, line 27: Per Bret, after PR #3113, the jacobian as returned by `compute_totals`  is now keyed with the 'user facing' names given to the design variabless and responses, rather than the absolute names that were used previously.

This should be noted as a backwards incompatible API change.

swryan (changes requested):
see comment regarding PR #3113

release_notes.md
```diff
@@ -1,3 +1,75 @@
...
+
```
naylor-b, line 29: The automatic determination of derivative 'mode' is now done separately for the driver vs. a direct call to compute_totals or check_totals with a set of 'of' and 'wrt' variables that differ from the driver responses and design variables.  This can modify the keys in the dict return value of check_totals, where for example the dict used to contain a 'J_fwd' key because it ran in fwd mode, it might now contain a 'J_rev' key instead because the best derivative mode was determined to be 'rev' based on the relative sizes of the 'of' and 'wrt' variables passed to check_totals.

swryan (approved):

PR #3142, opened 2024-03-07
Moved pre/opt/post logic into the Relevance class.

naylor-b:
### Summary

All relevance operations, both linear and nonlinear, are now handled by the Relevance class.

Made some improvements to the SubmodelComp class.

The `mode` arg was removed from some private methods, but that shouldn't impact users.

### New Dependencies

None


swryan (approved):

timryanb:
I may have missed it in the PR @naylor-b , but was there a test added to catch the parallel derivative bug in the future?

naylor-b:
Good catch @timryanb .  I added a test.

PR #3166, opened 2024-03-28
added zero_tol parameter for check_partials method

### LINKED ISSUE #3164: Add Zero Tolerance to Check Partials for Relative Error Checks

In some cases, derivatives are very close to 0. This can cause relative error checks to fail. It may be worth to adding a parameter `zero_tol` to the `check_partials` call that rounds the calc mag to zero if it is bellow this tolerance before comparing it to the check mag.

```
-------------------------------------------------------------------------------------------------------
Component: Comp 'comp'
-------------------------------------------------------------------------------------------------------
'<output>'       wrt '<variable>'        | calc mag.  | check mag. | a(cal-chk) | r(cal-chk)
--------------------------------------------------------------------------------------------

'a' wrt 'b'  | 1.4410e-17 | 0.0000e+00 | 1.4410e-17 | 1.0000e+00 >REL_TOL
```

### Associated POEM

_No response_

kikutabryan:
### Summary

- Added parameter to check_partials method called zero_tol.
- If calculated magnitude of partials is below zero_tol, they are set equal to zero to avoid situations where the check magnitude is zero and the calculated magnitude has some slight error resulting in a relative error of 1.

Before:

image

After:

image

### Related Issues

- Resolves #3164 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/problem.py
+++ b/openmdao/core/problem.py
@@ -1109,7 +1109,7 @@ class Problem(object):
             self.check_config(logger, checks=checks)
 
     def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes=None,
-                       compact_print=False, abs_err_tol=1e-6, rel_err_tol=1e-6,
+                       compact_print=False, abs_err_tol=1e-6, rel_err_tol=1e-6, zero_tol=0.0,
                        method='fd', step=None, form='forward', step_calc='abs',
                        minimum_step=1e-12, force_dense=True, show_only_incorrect=False):
         """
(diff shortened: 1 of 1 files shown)


robfalck:
Thanks for your contribution!  Would you be able to add a test for our CI that verifies that this works?

It can be as simple as capturing the text from your example above and making sure it matches the expected value.
If you dont think you have the time to do this let us know.

robfalck:
After testing this implementation a bit, I have a few suggestions.

I think `zero_tol` should specify the magnitude of the `calc_mag` below which the relative error is not calculated.
A tol on the derivatives themselves overlaps with `abs_err_tol` and `rel_err_tol` and is a bit confusing.

Interested to hear thoughts on this.

kikutabryan:
> After testing this implementation a bit, I have a few suggestions.
> 
> I think `zero_tol` should specify the magnitude of the `calc_mag` below which the relative error is not calculated. A tol on the derivatives themselves overlaps with `abs_err_tol` and `rel_err_tol` and is a bit confusing.
> 
> Interested to hear thoughts on this.

Not entirely sure if I understood correctly, but I just updated the to code use the norm of the Jacobian and if the norm is greater than the specified threshold (renamed zero_tol to rel_err_norm_thresh), the relative error will be computed, if it is below, it will not be computed (as you suggested).

Let me know if this sounds better.

andrewellis55:
@robfalck, kikutabryan was an intern on my team who has now finished up. If there's any changes you'd like to this I can make them so this can get closed out (if the PR is accepted)

robfalck:
The tolerance checks in check_partials as of 3.37.0 have superseded this, as of PR #3442

PR #3185, opened 2024-04-10
Implementation branch for POEM93 (Linear solution caching)

naylor-b:
### Summary

This adds an 'rhs_checking' option to DirectSolver, PETScKrylov, and ScipyKrylov linear solvers.  This functionality is inactive by default.  The value of 'rhs_checking' can be True, False, or a dict that overrides any of the default entries: {'check_zero': False, 'rtol': 3e-16, 'atol': 3e-16, 'max_cache_entries': 3, 'collect_stats': False, 'auto': False, 'verbose': False}

Also added a method to Relevance that detects when RHS caching may be beneficial.

See also: [POEM 093](https://github.com/OpenMDAO/POEMs/pull/190/files)

None

### New Dependencies

None


openmdao/utils/numba.py
```diff
@@ -0,0 +1,25 @@
...
+try:
+    import numba
+    from numba import *
+except ImportError:
+    numba = None
+    prange = range
+
+    # If numba is not available, just write a dummy jit wrapper.
+    # Code will still run at a significant performance hit.
+    def jit(f, *args, **kwargs):
```
robfalck, line 16: I think this will confuse the autodoc tools. They'll pick up the documentation for jit rather than the wrapped function.  The solution is to use `functools.wraps` like here:

https://github.com/OpenMDAO/dymos/blob/8cb3c2666b363287eb23d2b79001e8ce3ebcd2f0/dymos/transcriptions/explicit_shooting/barycentric_control_interp_comp.py#L12

robfalck (changes requested):

openmdao/utils/numba.py
naylor-b: I thought functools.wraps was only necessary if the function was actually being wrapped, and it's not in this case.  The original function is just returned by the decorator, and that should be happening at class load time which I would think would be before the autodoc stuff even runs.

robfalck (approved):

swryan (approved):

PR #3197, opened 2024-04-17
Added capability for IVCs to be instantiated with a list of variables

johnjasa:
### Summary

There's an error message for IndepVarComp that says you can use an iterable at instantiation, but I don't think the code was there anymore.
I added the option to do that here and two tests accordingly.

Shout-out to @crecine for trying this and saying "hey should this work?"

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/indepvarcomp.py
+++ b/openmdao/core/indepvarcomp.py
@@ -39,6 +39,16 @@ class IndepVarComp(ExplicitComponent):
         if isinstance(name, str):
             super().add_output(name, val, **kwargs)
 
+        elif isinstance(name, (list, tuple)):
+            for tup in name:
+                if len(tup) == 2:
+                    super().add_output(tup[0], tup[1], **kwargs)
+                elif len(tup) == 3:
+                    super().add_output(tup[0], tup[1], **tup[2])
+                else:
+                    raise ValueError("Each entry in the list of tuples must be of the form "
+                                     "(name, value) or (name, value, keyword_dict).")
+
         elif name is None:
             pass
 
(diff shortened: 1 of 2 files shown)


openmdao/core/indepvarcomp.py
```diff
@@ -39,6 +39,16 @@ def __init__(self, name=None, val=1.0, **kwargs):
         if isinstance(name, str):
             super().add_output(name, val, **kwargs)
 
+        elif isinstance(name, (list, tuple)):
+            for tup in name:
+                if len(tup) == 2:
```
naylor-b, line 44: You might want to verify here that tup is actually a tuple, or at least not a string. Otherwise in cases where the user gives you a list/tuple of strings, you'll get weird error messages.  I could see a user just providing a list of strings and assuming those variables would all have default values.

johnjasa: Good call, I've modified this and pushed up!

swryan: I think the docstring for `name` needs to be updated for these new possibilities...

Also maybe some kind of error checking on the `tup` values (i.e. tup[0] must be a string, tup[2] must be a dict?)?

swryan (changes requested):

openmdao/core/indepvarcomp.py
johnjasa: Thanks Steve, I've updated the docstring and added type checking!

swryan (approved):

Issue #2659, opened 2022-10-06
Making multiple calls to declare_coloring on the same component doesn't work.


naylor-b:
### Description

It's possible that one might want to make multiple calls to declare_coloring on a component in order to use different metadata (approx method, etc.) for different sets of inputs.  In the example below, this results in an IndexError.

Note that in order for this to work properly, a separate coloring would have to be computed for each part of the jacobian corresponding to the inputs mapping to each declare_coloring call.  Otherwise columns with differing approx method, etc. could be grouped together in the same color.

### Example

```python
import openmdao.api as om
import numpy as np

size = 5
om.ExecComp.register('unsafe', lambda x: x**2, complex_safe=False)
om.ExecComp.register('safe', lambda x: x**2, complex_safe=True)
p = om.Problem()
comp = p.model.add_subsystem('comp', om.ExecComp(['out1 = unsafe(x) * z',
                                                  'out2 = safe(y) + z'], shape=size))
comp.declare_coloring(wrt=('x', 'z'), method='fd')
comp.declare_coloring(wrt='y', method='cs')

p.setup()
xx = np.arange(1, size + 1, dtype=float)
p['comp.x'] = x = xx * 3.
p['comp.y'] = y = xx * 4.
p['comp.z'] = z = xx * 5.
p.run_model()

J = p.compute_totals(of=['comp.out1', 'comp.out2'], wrt=['comp.x', 'comp.y', 'comp.z'])
```

### OpenMDAO Version

3.21.1-dev

PR #3201, opened 2024-04-20
declare_coloring index error 

leenahk:
### Summary
Modified ‘declare_coloring’ implementation

def declare_coloring(self, wrt, method='fd', **kwargs):
    key = f'{method}:{wrt}' if isinstance(wrt, tuple) else f'{method}:{wrt}'
    if key in self._colorings:
        self._colorings[key].update(method=method, **kwargs)
    else:
        self._colorings[key] = ColoringConfig(wrt=wrt, method=method, **kwargs)

### Issue
Addressing issue #2659. This issue involves where making multiple calls to declare_coloring on the same component does not work as expected, leading to an IndexError.

### Example (from user naylor-b)
import openmdao.api as om
import numpy as np

size = 5
om.ExecComp.register('unsafe', lambda x: x**2, complex_safe=False)
om.ExecComp.register('safe', lambda x: x**2, complex_safe=True)
p = om.Problem()
comp = p.model.add_subsystem('comp', om.ExecComp(['out1 = unsafe(x) * z',
                                                  'out2 = safe(y) + z'], shape=size))
comp.declare_coloring(wrt=('x', 'z'), method='fd')
comp.declare_coloring(wrt='y', method='cs')

p.setup()
xx = np.arange(1, size + 1, dtype=float)
p['comp.x'] = x = xx * 3.
p['comp.y'] = y = xx * 4.
p['comp.z'] = z = xx * 5.
p.run_model()

J = p.compute_totals(of=['comp.out1', 'comp.out2'], wrt=['comp.x', 'comp.y', 'comp.z'])


robfalck:
Hello!  Do you have an example demonstrating the bug you're trying to address?

robfalck (changes requested):
This PR needs a lot of cleanup and needs to explain the actual problem being addressed

robfalck:
This PR commits a lot of files from the documentation _build directory, which should not be committed to the codebase. There also don't seem to be any meaningful changes to actual OpenMDAO code, other than a whitespace change in Driver.

PR #3202, opened 2024-04-22
Fix pyOptSparse `hotstart_file` `@property` method

eytanadler:
### Summary

The pyOptSparse hot start and history file attributes were switched to options. To maintain backward compatibility, `@property` methods were added that reference the option. However, the `hotstart_file` property method (presumably accidentally) references the `hist_file` option instead of the `hotstart_file` option. This PR fixes that.

None

### New Dependencies

None


swryan (changes requested):
Good catch..   this was a test fail as well...  we should update the test:
https://github.com/eytanadler/OpenMDAO/pull/1

swryan (approved):

PR #3214, opened 2024-05-01
Changed return of `prob.run_driver()` from a bool to an object containing information about the Driver execution.

### LINKED ISSUE #3206: POEM 094 Part I: DriverResults

This issue is the first of multiple parts in the implementation of [POEM 094](https://github.com/OpenMDAO/POEMs/pull/191)

### Related Issues

- OpenMDAO/Aviary#250

In this part we will change the return value of `run_driver` to return a `DriverResults` object, rather than a boolean value that is `False` if the run succeeded.

### Associated POEM

094

robfalck:
### Summary

This is the first of several implementation changes in support of [POEM 094](https://github.com/OpenMDAO/POEMs/pull/191).

Users were finding the fact that run_driver returned `True` if something had failed to be confusing.

To address this, a new `DriverResult` object has been added, containing the information previously stored in `driver.opt_result`.
This DriverResult object exists for all drivers, including non-optimization ones, so it is now stored in `Driver.result`.
Previously, ScipyOptimizeDriver stored scipy's OptimizeResult object as `driver.result`, but this is now changed to `scipy_optimize_result.`

DOEDriver now successfully populates the attributes of `DriverResult`.
Hooks have been added, though not yet implemented, to allow drivers to add driver/optimizer-specific attributes to the DriverResults object.

DriverResult currently overrides `__bool__` with a deprecation warning to retain existing behavior.

### Related Issues

- Resolves #3206 

`ScipyOptimizeDriver.result` is now a DriverResults object. Scipy-specific OptimizeResult objects are now stored in `ScipyOptimizeDriver._scipy_optimize_result`, though some of this information will be added to the DriverResult object.

### New Dependencies

None


openmdao/core/driver.py
```diff
@@ -22,7 +22,80 @@
...
+
+    def __init__(self):
+        """
+        Initialize the DriverResult object.
+        """
+        self.runtime = 0.0
+        self.iter_count = 0
+        self.obj_calls = 0
+        self.deriv_calls = 0
+        self.exit_status = 'NOT_RUN'
```
Kenneth-T-Moore, line 57: In Aviary and MBSE, we have been using python enums for things like this.  It makes it easer for someone coding a new driver, so that they can just import the enum and select one of the exit statuses instead of accidentally making up a new string. (Of course, if they really do have a new kind of status, they can add that to the enum.)

Kenneth-T-Moore (approved):
Approve, but make sure to look at my comment. That can be a future consideration if you want.

openmdao/core/driver.py
robfalck: Hmm, this isn't a new addition though. This is someothing that's alread existent that the optimization report relies upon.

openmdao/core/driver.py
```diff
@@ -22,7 +22,80 @@
...
+        A boolean that dictates whether or not the driver was successful.
+    """
+
+    def __init__(self):
+        """
+        Initialize the DriverResult object.
+        """
+        self.runtime = 0.0
+        self.iter_count = 0
+        self.obj_calls = 0
```
naylor-b, line 55: What do you think about maybe keeping track of runtime for obj_calls and deriv_calls separately?  Might be nice to know where the time is being spent.

openmdao/core/driver.py
```diff
@@ -97,6 +117,47 @@ def __bool__(self):
                       category=OMDeprecationWarning)
         return not self.success
 
+    @staticmethod
+    def track_stats(kind):
+        """
+        Decorate methods to tracks either the objective the deriv time and call count for a Driver.
```
naylor-b, line 123: should be 'to track' instead of 'to tracks'

openmdao/core/driver.py
```diff
@@ -97,6 +117,47 @@ def __bool__(self):
...
+
+        This decorator should be applied to the _objfunc or _gradfunc (or equivalent) methods
+        of drivers. It will either accumulated the elapsed time in driver.result.obj_time or
```
naylor-b, line 126: acumulated --> acumulate

openmdao/core/driver.py
```diff
@@ -22,7 +23,140 @@
...
+        driver.result.deriv_time, based on the value of time_type.
+
+        Parameters
+        ----------
+        kind : str
+            One of 'obj' or 'deriv', specifying which statistics should be accumulated.
```
Kenneth-T-Moore, line 132: maybe pedantic, but i don't know about this being called 'obj', since you don't always have an objective (and, when you do want an objective, you want the constraints too, which also come along for free.) I know that I can't just say this without offerig some alternatives ... 

"compute" , "compute_totals"
"function", "gradient"  (note, this allows for our future "hessian")
"eval", "deriv_eval"

naylor-b (approved):

openmdao/drivers/doe_driver.py
```diff
@@ -73,6 +77,8 @@ def __init__(self, generator=None, **kwargs):
         self._name = ''
         self._problem_comm = None
         self._color = None
+        self._num_model_evals = 0
+        self._num_deriv_evals = 0
```
swryan, line 81: I don't see where these attributes are consumed, they seem redundant with the decorator.. am I missing something?

robfalck: Bah you're right.

openmdao/core/driver.py
```diff
@@ -723,7 +877,7 @@ def get_driver_objective_calls(self):
         int
             Number of objective evaluations made during a driver run.
         """
-        return 0
+        return self.result.model_evals
```
swryan, line 880: The change in terminology makes me wonder if these method names should be deprecated in favor of `get_driver_model_evals` and `get_driver_deriv_evals`.

But then I'm not sure why these public methods need to exist...  they were only used to populate the `opt_result`, which is no longer the case.   They are not documented anywhere for the user...

robfalck: Good point.  Changed them to being deprecated and removed their override in DifferentialEvolutionDriver.

PR #3216, opened 2024-05-01
Added ability to set print_results to "minimal"

crecine:
### Summary

Instead of just True/False, minimal will only print variables and constraints that have a non-empty status (for example ones that violate a bound)
This extends the acceptable values for print_result and automates setting the minimal_print value that is added here https://github.com/mdolab/pyoptsparse/pull/395
Even if the version of pyoptsparse a user has doesn't include this feature, it won't cause problems.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/pyoptsparse_driver.py
+++ b/openmdao/drivers/pyoptsparse_driver.py
@@ -253,7 +253,7 @@ class pyOptSparseDriver(Driver):
                              desc='Title of this optimization run')
         self.options.declare('print_opt_prob', types=bool, default=False,
                              desc='Print the opt problem summary before running the optimization')
-        self.options.declare('print_results', types=bool, default=True,
+        self.options.declare('print_results', types=(bool,str), default=True,
                              desc='Print pyOpt results if True')
         self.options.declare('gradient_method', default='openmdao',
                              values={'openmdao', 'pyopt_fd', 'snopt_fd'},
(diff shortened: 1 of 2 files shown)


robfalck:
Wait for pyoptsparse merge:  https://github.com/mdolab/pyoptsparse/pull/395

swryan:
also needs a test

crecine:
> also will need a test
I'm not seeing any tests in test_pyoptsparse_driver.py that set print_results to True, so I'm not sure what a test would look like. Could you elaborate?

swryan:
> > also will need a test
> 
> I'm not seeing any tests in test_pyoptsparse_driver.py that set print_results to True, so I'm not sure what a test would look like. Could you elaborate?

Hmmm. yeah that's unfortunate.. there probably should be a test that sets it to True and captures stdout to make sure it works.

A minimal test (no pun intended) for this feature I guess would be to just set the option to 'minimal' and make sure it was accepted as valid...  A more complete test would capture stdout and check for the expected output, but since the content of that is outside of OM control, it's probably a step too far...

crecine:
> > > also will need a test
> >
> > I'm not seeing any tests in test_pyoptsparse_driver.py that set print_results to True, so I'm not sure what a test would look like. Could you elaborate?
> 
> Hmmm. yeah that's unfortunate.. there probably should be a test that sets it to True and captures stdout to make sure it works.
> 
> A minimal test (no pun intended) for this feature I guess would be to just set the option to 'minimal' and make sure it was accepted as valid... A more complete test would capture stdout and check for the expected output, but since the content of that is outside of OM control, it's probably a step too far... just setting the value to `minimal` would get coverage on the if statement.

I added a test that makes sure that both cases run.
If you want to try it to see the difference in the results, you can run it with:
```
  test = TestPyoptSparse()
  test.test_minimal_print()
```
Let me know if this is satisfactory, or if you are looking for something else.

swryan:
excellent.. thanks 👍🏼

swryan (approved):

PR #3228, opened 2024-05-08
Added ability to handle reading case recorder files with class instances when the associated class cannot be imported

### LINKED ISSUE #3205: Case recorder files can contain Python objects whose classes are not importable
### Description

If a user makes use of some external packages like VSP wrappers, the case recorder file will have pickled objects from that package. If that case recorder file is opened in a Python environment which does not have that package installed, there will be an exception since the class of that object will not be available to be imported. 

Need to handle the reading of case recorder files more intelligently so that the reading doesn't completely fail. Should read in what is doable and set to empty what is not. Give the user a warning about what happened. 

### Example

The example that brought out this issue was a user doing an Aviary run using OpenVSP and then sending the case recorder file to another user who did not have OpenVSP installed. 

The test created as part of the PR that fixes this bug will give an example of the bug. 

### OpenMDAO Version

3.31.2-dev

NA

hschilling:
### Summary

Added ability to handle reading case recorder files with class instances when the associated class cannot be imported.

Designed to handle the following example scenario. 

A user does case recording on a computer where a module MyModule exists. An instance of that class is recorded in the component options metadata for a component in the model. 

This case recorder file is shared with another user that does not have access to that module. Currently if that second user reads the case recorder file, the reading of the file fails completely. Not data can be read from it. 

This PR solves this problem by skipping reading of that one class instance and reading everything else. 

### Related Issues

- Resolves #3205 

None

### New Dependencies

None

### DIFF
--- a/openmdao/recorders/sqlite_reader.py
+++ b/openmdao/recorders/sqlite_reader.py
@@ -29,7 +29,7 @@ from json import loads as json_loads
 from io import TextIOBase
 
 
-class _InstanceWithNoClassDefinition:
+class UknownType:
     """
     A class used by _RestrictedUnpickler.
 
@@ -64,7 +64,7 @@ class _RestrictedUnpicklerForCaseReader(pickle.Unpickler):
         # Returning this acts as a kind of flag to indicate that
         # the unpickler can't generate instances of classes whose class definition
         # is not available
-        return _InstanceWithNoClassDefinition
+        return UknownType
 
     def loads_and_return_errors(self):
         unpickled_contents = self.load()
@@ -88,7 +88,7 @@ def _loads_and_return_errors(s):
     for key in keys:
         try:
             isInstanceWithNoClassDefinition = isinstance(dictionary[key],
-                                                         _InstanceWithNoClassDefinition)
+                                                         UknownType)
             if isInstanceWithNoClassDefinition:
                 dictionary.undeclare(key)
         except RuntimeError:  # OptionDictionary item might not be set yet
--- a/openmdao/recorders/sqlite_reader.py
+++ b/openmdao/recorders/sqlite_reader.py
@@ -29,7 +29,7 @@ from json import loads as json_loads
 from io import TextIOBase
 
 
-class UknownType:
+class UnknownType:
     """
     A class used by _RestrictedUnpickler.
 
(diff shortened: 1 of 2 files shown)


openmdao/recorders/tests/test_sqlite_reader.py
```diff
@@ -3055,6 +3056,89 @@ def test_constraints_with_aliases(self):
...
+            prob.record('final')
+            
+            prob.cleanup()
+            
+            # Really remove the module!
+            del sys.modules['mymodule']
+            os.remove(module_path)
+
+
+        create_case_recording_file()
```
naylor-b, line 3127: Instead of adding current dir to sys.path inside of `create_case_recording_file`, you might consider doing it here and have a try: finally block around the function call so you can restore sys.path back to its original form.

openmdao/recorders/sqlite_reader.py
```diff
@@ -26,6 +29,74 @@
...
+
+    # remove any InstanceWithNoClassDefinition classes
+    # Only want to return legit values in the OptionsDictionary
+    keys = [key for key, value in dictionary.items()]
+    for key in keys:
+        try:
+            isInstanceWithNoClassDefinition = isinstance(dictionary[key],
+                                                         _InstanceWithNoClassDefinition)
+            if isInstanceWithNoClassDefinition:
+                dictionary.undeclare(key)
```
swryan, line 93: Do we really want to remove the option completely?  

Seems like we would want it to show up as `Unknown Type` or something like that...  
Do things blow up if you just leave the `_InstanceWithNoClassDefinition` in there?

hschilling: At first I had it that way. Things don't blow up if you just leave the _InstanceWithNoClassDefinition in there. 

But I thought users would be confused by seeing an instance of some weird class that they didn't know anything about. On the other hand, seeing it might make them realize that something is not quite right. We do issue a warning but they might miss that. 

I do like the name of `UnknownType`. That might be better than `_InstanceWithNoClassDefinition`

I will go with that.

openmdao/recorders/tests/test_sqlite_reader.py
hschilling: Good idea. Changed

swryan (approved):

naylor-b (approved):

PR #3237, opened 2024-05-15
Added `usedforsecurity=False` flag to hashlib.md5 uses for FIPS-enabled systems.

### LINKED ISSUE #3236: Exception thrown using recording manager on Linux computer with FIPS
### Description

The OpenMDAO system._generate_md5_hash() method uses the hashlib.md5 method, which throws an exception on Linux systems with FIPS (Federal Information Processing Standard) enabled, as described here: https://til.simonwillison.net/python/md5-fips

This method is used in the recording manager and the n2 viewer, preventing users on Linux systems with FIPS from using these features.

In Python 3.9+, you can avoid this exception by changing a setting in the md5 method: usedforsecurity=False

Suggested solution for the system._generate_md5_hash() is to have a try-catch to attempt to run the hashlib.md5 method with the the usedforsecurity set to false:

```
try:
    return hashlib.md5(s.encode("utf8"), usedforsecurity=False).hexdigest()
except TypeError:
    # For Python 3.8 which does not support usedforsecurity=False
    return hashlib.md5(s.encode("utf8")).hexdigest()
```

This would enable users on Linux systems with FIPS enabled to access the recording manager and n2 viewer with newer Python versions.

### Example

Exception message thrown for a user when trying to use the recording manager on a RHEL9 machine with FIPS enabled:

> Python.Runtime.PythonException: [digital envelope routines] unsupported
>   File "[path to python]/python-3.11.8/lib/python3.11/site-packages/openmdao/core/system.py", line 5832, in _generate_md5_hash
>     return hashlib.md5(str(data).encode()).hexdigest()  # nosec: content not sensitive
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

### OpenMDAO Version

3.27.0

_No response_

robfalck:
### Summary

Users of Linux systems with FIPS (Federal Information Processing Standard) enabled were unable to use OpenMDAO features which called hashlib.md5.  The argument `usedforsecurity=False` allows this code to operate, but is only available in Python >= 3.9.

This change adds the `usedforsecurity=False` to `hashlib.md5` calls.
WIth Python 3.8 this will fall back to the previous method without the argument, so FIPS systems will need to use Python 3.9 or later.

### Related Issues

- Resolves #3236

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -6104,7 +6104,13 @@ class System(object):
         for key in sorted(self._conn_global_abs_in2out):
             data.append(self._conn_global_abs_in2out[key])
 
-        return hashlib.md5(str(data).encode()).hexdigest()  # nosec: content not sensitive
+        try:
+            hash = hashlib.md5(str(data).encode(),
+                               usedforsecurity=False).hexdigest() # nosec: content not sensitive
+        except TypeError:
+            hash = hashlib.md5(str(data).encode()).hexdigest() # nosec: content not sensitive
+
+        return hash
 
     def _get_full_dist_shape(self, abs_name, local_shape):
         """
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -6106,9 +6106,9 @@ class System(object):
 
         try:
             hash = hashlib.md5(str(data).encode(),
-                               usedforsecurity=False).hexdigest() # nosec: content not sensitive
+                               usedforsecurity=False).hexdigest()  # nosec: content not sensitive
         except TypeError:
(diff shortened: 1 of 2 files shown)


naylor-b (approved):

andrewellis55:
@robfalck if it's only one try except I would very much appreciate keeping the support for python 3.8 at least until the end of life. My team has an action to move towards a higher version by the end of the year but for the moment we havn't quite moved yet.

andrewellis55:
Just scheduling. It's very firmly on the todo list and should be complete by the official 3.8 End of Life. There are a few (minor) extra steps we need to take when performing upgrades because of corporate IT policies.

PR #3241, opened 2024-05-20
Updates for NumPy 2.0 compatibility and testing

### LINKED ISSUE #3240: NumPy 2.0 pre-release testing should be done separately from our usual "latest" test
### Task Description

Our `Latest` test workflow currently ends up downgrading NumPy from the 2.0 pre-release due to compatibility issues with some dependencies (PETSc in particular, which is installed via Conda and thus does not have a NumPy 2.0rc build).  We also require `numpy<2` now for OpenMDAO, pending the availability of the release version for testing with compatible dependencies.

We should split the workflow to test with the NumPy 2.0 release candidates, excluding dependencies that do not yet work with it, and also with the latest `numpy<2` including all dependencies.

swryan:
### Summary

- Updated the `Latest` workflow to test NumPy 2.0 pre-release separately (excluding PETSc and pyOptSparse)
- Added NumPy 2.0 compatibility check to the `Tests` workflow
- Fixed several NumPy 2.0 compatibility issues on The OpenMDAO codebase
  - Replaced `NaN` with `nan`
  - Updated `ComplexWarning` imports to use new location
  - Changed `array(copy=False)` to `asarray()`, which will not copy if the object is already a numpy array
  - Updated tests to handle change in the default print format
  - Updated the `cs_safe.abs()` method to preserve current numpy<2 logic for the `sign` operation

### Related Issues

- Resolves #3240

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/cs_safe.py
+++ b/openmdao/utils/cs_safe.py
@@ -20,7 +20,20 @@ def abs(x):
         Absolute value.
     """
     if isinstance(x, np.ndarray):
-        return x * np.sign(x)
+        # Changed in NumPy 2.0: Definition of complex sign changed to follow the Array API standard.
+        # 1.x: For complex inputs, the sign function returns sign(x.real) + 0j if x.real != 0
+        #      else sign(x.imag) + 0j.
+        # 2.0: For complex inputs, the sign function returns x / abs(x), and 0 if x==0.
+        if np.__version__[0] == '2' and np.any(np.iscomplex(x)):
+            # replicate NumPy 1.x behavior for complex arrays
+            z0_idx = x.real == 0
+            nz_idx = x.real != 0
+            signs = np.zeros(x.shape, dtype=complex)
+            signs[nz_idx] = np.sign(x[nz_idx]).real + 0j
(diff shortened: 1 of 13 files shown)


openmdao/utils/cs_safe.py
```diff
@@ -20,7 +20,20 @@ def abs(x):
         Absolute value.
     """
     if isinstance(x, np.ndarray):
-        return x * np.sign(x)
+        # Changed in NumPy 2.0: Definition of complex sign changed to follow the Array API standard.
+        # 1.x: For complex inputs, the sign function returns sign(x.real) + 0j if x.real != 0
+        #      else sign(x.imag) + 0j.
+        # 2.0: For complex inputs, the sign function returns x / abs(x), and 0 if x==0.
+        if np.__version__[0] == '2' and np.any(np.iscomplex(x)):
```
robfalck, line 27: Rather than do string matching here, do it at import and save a bool.

robfalck (changes requested):

robfalck (approved):

Kenneth-T-Moore (approved):

PR #3248, opened 2024-05-28
fix for group set_val when using set_input_defaults

### LINKED ISSUE #3246: Unit ambiguity issue when using set_val
### Description

Using `set_val()` at the subsystem level doesn't register `set_input_defaults()` being called. If two subsystems promote a variable to the same variable name but have different units at the subsystems, this can cause an issue depending on what level `set_val()` is called. In the example below, using `p.set_val('comp.x', 2.0, units='m')` works but using `comp.set_val('x', 2.0, units='m')` causes the error 

`The following inputs, ['comp.ec1.x', 'comp.ec2.x'], promoted to 'x', are connected but their metadata entries ['units'] differ. Call <group>.set_input_defaults('x', units=?), where <group> is the Group named 'comp' to remove the ambiguity.`

### Example

`class Comp(om.Group):`
`    def setup(self):`
`        self.add_subsystem('ec1', om.ExecComp('y=x', x={'units': 'm'}), promotes_inputs=['x'])`
`        self.add_subsystem('ec2', om.ExecComp('z=x**2', x={'units': 'ft'}), promotes_inputs=['x'])`
`        self.set_input_defaults('x', val=1.0, units='m')`

`if __name__ == '__main__':`
`    p = om.Problem(om.Group())`
`    comp = p.model.add_subsystem('comp', Comp())`
`    p.setup()`
`    # p.set_val('comp.x', 2.0, units='m')`
`    comp.set_val('x', 2.0, units='m')`
`    p.run_model()`

### OpenMDAO Version

3.32.1-dev

_No response_

naylor-b:
### Summary

When calling set_val on a group, it was accessing the group input defaults for the model instead of the specified group.

### Related Issues

- Resolves #3246

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -5369,7 +5369,7 @@ class System(object):
         try:
             ginputs = self._group_inputs
         except AttributeError:
-            ginputs = {}  # could happen if top level system is not a Group
+            ginputs = {}  # could happen if this system is not a Group
 
         if post_setup:
             abs_names = name2abs_names(self, name)
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -5367,7 +5367,7 @@ class System(object):
         n_proms = 0  # if nonzero, name given was promoted input name w/o a matching prom output
 
         try:
-            ginputs = model._group_inputs
+            ginputs = self._group_inputs
         except AttributeError:
             ginputs = {}  # could happen if top level system is not a Group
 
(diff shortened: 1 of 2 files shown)


openmdao/core/system.py
```diff
@@ -5367,7 +5367,7 @@ def set_val(self, name, val, units=None, indices=None):
         n_proms = 0  # if nonzero, name given was promoted input name w/o a matching prom output
 
         try:
-            ginputs = model._group_inputs
+            ginputs = self._group_inputs
```
swryan, line 5370: should probably update/remove the comment in the `except` block....  it implies that the previous behavior was intended (i.e. only looking at the top level)

naylor-b: done

robfalck (approved):

swryan (approved):

PR #3265, opened 2024-06-05
Fix relevance check for empty groups

andrewellis55:
### Summary

Relevancy checks currently fail if any empty groups are present. This is a bit of an edge case, but we sometimes have a few empty groups when various user set constraints\configurations aren't applied.

### Example
This code will fail in 3.32.1-dev

```python
import openmdao.api as om

prob = om.Problem()
model = prob.model

grp1: om.Group = model.add_subsystem('empy_group', om.Group(), promotes=['*'])
grp2: om.Group = model.add_subsystem('non_empty_group', om.Group(), promotes=['*'])
grp2.add_subsystem('idv', om.IndepVarComp('x', val=1), promotes=['*'])
grp2.add_subsystem('comp', om.ExecComp('y=2*x**2'), promotes=['*'])
model.add_design_var('x')
model.add_objective('y')

prob.setup()
prob.run_driver()
```

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/tests/test_relevance.py
+++ b/openmdao/utils/tests/test_relevance.py
@@ -61,3 +61,5 @@ class TestRelevanceEmptyGroups(unittest.TestCase):
 
         prob.setup()
         prob.run_driver()
+
+        assert_check_totals(prob.check_totals(method='cs', out_stream=None))
--- a/openmdao/utils/tests/test_relevance.py
+++ b/openmdao/utils/tests/test_relevance.py
@@ -59,7 +59,7 @@ class TestRelevanceEmptyGroups(unittest.TestCase):
 
         prob.driver = om.ScipyOptimizeDriver()
 
-        prob.setup()
+        prob.setup(force_alloc_complex=True)
         prob.run_driver()
 
         assert_check_totals(prob.check_totals(method='cs', out_stream=None))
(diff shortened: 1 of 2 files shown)


openmdao/utils/tests/test_relevance.py
```diff
@@ -43,3 +43,21 @@ def compute_jacvec_product(self, inputs, d_inputs, d_outputs, mode):
...
+        grp2: om.Group = model.add_subsystem('non_empty_group', om.Group(), promotes=['*'])
+        grp2.add_subsystem('idv', om.IndepVarComp('x', val=1), promotes=['*'])
+        grp2.add_subsystem('comp', om.ExecComp('y=2*x**2'), promotes=['*'])
+        model.add_design_var('x')
+        model.add_objective('y')
+
+        prob.driver = om.ScipyOptimizeDriver()
+
+        prob.setup()
+        prob.run_driver()
```
naylor-b, line 63: Can you add the following to the bottom of this test, to verify that the totals are correct?
```
        assert_check_totals(prob.check_totals(method='cs', out_stream=None))
```

naylor-b (changes requested):

openmdao/utils/tests/test_relevance.py
andrewellis55: done!

naylor-b (approved):

robfalck (approved):

PR #3267, opened 2024-06-06
making it easier to use jax to compute component derivatives

naylor-b:
### Summary

Made updates to ExplicitComponent to allow an option to be set to specify how derivs should be computed.  Current choices are 'jax', 'cs', and 'fd'.  Created a JaxExplicitComponent class that requires the user to do nothing but define a compute_primal method.  

If the compute_primal outputs depend on component options or some other 'static' variables then the component author also has to define a `get_static_arg` method that returns a tuple containing all of the static variables.

Also added jax capability for Groups, but this is pretty experimental, works only for Groups that are fully explicit (no internal cycles or implicit comps anywhere).  The primary reason why the majority of the jax functionality is included in ExplicitComp rather than in JaxExplicitComp is so that when the user specifies that a Group should use jax to compute derivatives, all of the components contained in it will have their compute methods converted to compute_primal methods on the fly.

None

### New Dependencies

None


openmdao/docs/openmdao_book/features/experimental/jax_explicitcomp_api.ipynb
```diff
@@ -0,0 +1,303 @@
+{
```
robfalck, line 1: <pre class="ql-syntax" spellcheck="false">Line #12.            self.add_input('z', val=np.zeros(2))
</pre><p>Is it necessary to use both <code>np</code> and <code>jnp</code> in this component definition? Users might find that confusing, or say "well I'm importing jnp anyway so I'll just use <code>jnp.zeros</code>."</p><p> <span class="ql-cursor">﻿</span></p><p>Could we override add_input and add_output here to allow jax arrays?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMDAO/OpenMDAO/pull/3267/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

naylor-b: It works either way.  I changed it to jnp.

naylor-b:
Closing this because I have another branch where I added jax for ImplicitComponent.

PR #3275, opened 2024-06-13
Optimizers now keep track of design vars that impact linear constraints and nonlinear responses separately.

### LINKED ISSUE #3221: OpenMDAO issues an errant DerivativesWarning if design variables do not impact nonlinear constraints.
### Description

In `_TotalJacInfo`, method `check_total_jac` only considers the nonlinear jacobian.  This is the jacobian used for coloring, as we don't want to consider linear responses for coloring.

However, this check will issue a warning if a design variable only impacts linear constraints, because it appears to have a column of zeros in the corresponding nonlinear total jacobian.

### Example

Executing the following test in test_pyoptsparse_driver demonstrates the issue.

```
    def test_total_jac_warning_linear_cons(self):

        prob = om.Problem()
        model = prob.model

        model.add_subsystem('p1', om.IndepVarComp('x', 50.0), promotes=['*'])
        model.add_subsystem('p2', om.IndepVarComp('y', 50.0), promotes=['*'])
        model.add_subsystem('con', om.ExecComp('c = x - y + k'), promotes=['*'])

        prob.driver = pyOptSparseDriver()
        prob.driver.options['optimizer'] = OPTIMIZER
        if OPTIMIZER == 'SLSQP':
            prob.driver.opt_settings['ACC'] = 1e-9
        prob.driver.options['print_results'] = False

        model.add_design_var('x', lower=-50.0, upper=50.0)
        model.add_design_var('y', lower=-50.0, upper=50.0)
        model.add_design_var('k', lower=-1, upper=1)
        model.add_objective('f_xy')
        model.add_constraint('c', lower=15.0, linear=True)

        prob.setup()

        with assert_no_warning(category=om.DerivativesWarning):
            prob.run_driver()
```

### OpenMDAO Version

3.31.2-dev

_No response_

naylor-b:
### Summary

The warning seen in issue #3221 no longer appears if the design variable in question is one that only impacts linear constraints.

Also, in pyOptSparseDriver, the linear and nonlinear total jacobians now only contain the appropriate design variables.  This was not done in ScipyOptimizeDriver because, unlike pyoptsparse, it doesn't keep track of different jacobians internally.

The scaling viewer was also updated so that the linear jacobian now only contains design variables that impact the linear constraints.

### Related Issues

- Resolves #3221

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -14,7 +14,6 @@ from openmdao.core.total_jac import _TotalJacInfo
 from openmdao.core.constants import INT_DTYPE, _SetupStatus
 from openmdao.recorders.recording_manager import RecordingManager
 from openmdao.recorders.recording_iteration_stack import Recording
-from openmdao.utils.hooks import _setup_hooks
 from openmdao.utils.record_util import create_local_meta, check_path, has_match
 from openmdao.utils.general_utils import _src_name_iter
 from openmdao.utils.mpi import MPI
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -444,6 +444,36 @@ class Driver(object):
             msg += '.'.join(self._designvars_discrete)
             raise RuntimeError(msg)
 
+        # Determine which design vars are relevant to linear constraints and which are relevant to
+        # nonlinear constraints or objectives.
+        lin_dvs = set()
+        nl_dvs = set()
+        lin_cons = {name: meta for name, meta in self._cons.items() if meta['linear']}
+        nl_resps = {name: meta for name, meta in self._cons.items() if not meta['linear']}
+        nl_resps.update(self._objs)
+
+        relevance = model._relevance
+        revsrcs = relevance._single_seed2relvars['rev']
+        with relevance.seeds_active():
+            for dv, dvmeta in self._designvars.items():
+                dvarr = relevance._single_seed2relvars['fwd'][dvmeta['source']]
+                for linmeta in lin_cons.values():
+                    linarr = revsrcs[linmeta['source']]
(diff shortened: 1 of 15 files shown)


openmdao/core/driver.py
```diff
@@ -1744,3 +1780,62 @@ def record_iteration(requester, prob, case_name):
...
+                yield tup
+    else:
+        for tup in metadict_items:
+            if tup[1][key]:
+                yield tup
+
+
+def filter_by_meta(metadict_items, key, chk_none=False):
+    """
+    Yield only items where meta[key] is not True.
```
Kenneth-T-Moore, line 1817: These functions are practically the same.  Seems like they could be generalized into one function, maybe with includes and excludes.  Just some thoughts.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -710,7 +706,7 @@ def _objfunc(self, dv_dict):
             signal.signal(sigusr, self._signal_handler)
 
         try:
-            for name in self._indep_list:
+            for name in self._nl_dvs:
```
Kenneth-T-Moore, line 709: I think co-pilot forgot to add `_nl_dvs` to the attributes list.

Kenneth-T-Moore (approved):
Approved, but see notes.

openmdao/drivers/pyoptsparse_driver.py
naylor-b: fixed

openmdao/core/driver.py
naylor-b: done

swryan (approved):

PR #3276, opened 2024-06-17
update to hook system to allow args and/or return value to be passed to hook functions

naylor-b:
### Summary

Herb wanted to create a Report where the hook function needed to know the return value of the hooked function.

None

### New Dependencies

None

### DIFF
--- a/openmdao/utils/hooks.py
+++ b/openmdao/utils/hooks.py
@@ -32,8 +32,43 @@ def _reset_all_hooks():
     _hooks = {}
 
 
+def _hook_meta_factory(pre, pass_args, pass_return):
+    """
+    Return classes for hook functions that accept different arguments.
+
+    Parameters
+    ----------
+    pre : bool
+        If True, create a pre-hook class.
+    pass_args : bool
+        If True, create a class that passes args to the hook function.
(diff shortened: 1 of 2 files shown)


hschilling (approved):

openmdao/utils/hooks.py
```diff
@@ -113,7 +175,7 @@ def add_hooks(self, hooks):
         self.pre_hooks = sorted(self.pre_hooks, key=lambda x: x._stamp)
         self.post_hooks = sorted(self.post_hooks, key=lambda x: x._stamp)
 
-    def _run_hooks(self, hooks):
+    def _run_hooks(self, hooks, args, kwargs, ret=None):
```
swryan, line 178: should update the docstring for the new args...

also should probably add a docstring to `_HookMeta`

swryan (changes requested):
see previous comment.. one docstring is out of date and the _HookMeta constructor takes a lot of arguments that should probably be documented there as well

openmdao/utils/hooks.py
naylor-b: done

swryan (approved):

PR #3292, opened 2024-07-12
Branch for POEM098 - Automatically associating solvers with cycles

naylor-b:
### Summary

This adds the `add_subsolvers` method to Group, which creates an invisible CycleGroup for an associated sub-cycle within the Group.

Also adds a tool to view a graph of cycles in the model (`openmdao graph --type=cycle <fname>`), and a tool to list Groups that contain one or more sub-cycles, meaning cycles in the graph of direct children of the Group that do not contain all children of the Group.  These are the situations where it might be beneficial to associate solvers with the sub-cycle rather than with the entire Group.

### Related Issues

- Resolves #

- deprecates the `allow_post_setup_reorder` option.  Order can be set before or during config but not after.

### New Dependencies

None


naylor-b:
Closing this, at least for now.  I still haven't come up with a great API for this, and the implementation where CycleGroup doesn't have its own pathname has introduced a lot of complexity and broken some of our debugging tools.

PR #3293, opened 2024-07-15
Added the 'list_vars' method to the Case  object

### LINKED ISSUE #3272: `_check_discrete_dependance` can become very slow for models with many discrete variable
### Description

My team has recently move up to openmdao 3.33.0 (our project was previously pinned at 3.17.0, which was getting a little dated). Upon upgrading, it seems that many of the optimizations that used to sit in the 30 seconds to 2 minute runtime zone saw a notable increase in run time. After some debugging, it seems like a large source of the runtime difference is `_TotalJacInfo._check_discrete_dependence()`.

The models we use tend to be quite discrete variable heavy. In version 3.17.0 the first call to `compute_totals` is usually 0.5-2 seconds, whereas in 3.33.0 this has jumped to 15-30 seconds for the first call.

We're going to look at swapping a good portion of these discrete outputs to become options to get things back up to the speeds they were at before, but I figured I'd make an issue just to make a note that this can be an area where performance is affected for discrete variable heavy models.

In the meantime though, is there any way to make openmdao skip this check?

Thanks

### Example

Unfortunately I can't provide a copy of the models

### OpenMDAO Version

3.33.0

_No response_

swryan:
### Summary

Added the list_vars() function to the Case object

Also:
- changed the `tags` argument to `list_inputs`, `list_outputs` and `list_vars` to accept wildcards
- added a `print_tags` argument to `list_inputs`, `list_outputs` and `list_vars` to enable showing or not showing tags independent of filtering
- moved the rectangle example components/model to the test suite
- changed the three `list_*` functions in Case to use a common `get_io_metadata` method similar to System
- added test to exercise args and compare output from Case method to that of System method

### Related Issues

- Resolves #3272

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -4153,62 +4158,41 @@ def list_vars(self,
...
+        for var_name in var_list:
+            if var_name in outputs:
+                var_dict[var_name] = outputs[var_name]
+                var_dict[var_name]['io'] = 'output'
+            else:
+                var_dict[var_name] = inputs[var_name]
+                var_dict[var_name]['io'] = 'input'
 
         if all_procs or self.comm.rank == 0:
             write_var_table(self.pathname, var_list, 'all', var_dict,
```
naylor-b, line 4192: Since python dict is now ordered, we could probably get rid of the var_list arg to write_var_table and just use var_dict.

swryan: oh good call.. forgot about that... wilco

openmdao/core/system.py
```diff
@@ -4648,26 +4636,35 @@ def _get_vars_exec_order(self, inputs=False, outputs=False, variables=None):
...
-            for io in in_or_out:
-                for var_name in chain(real_vars[io], disc_vars[io]):
+                for var_name in chain(*var_dicts):
                     if not variables or var_name in variables:
-                        var_list.append(var_name)
+                        if var_name.startswith(prefix):
+                            var_list.append(var_name)
+        else:
+            for var_name in chain(*var_dicts):
+                if not variables or var_name in variables:
```
naylor-b, line 4666: 'not variables' should really be 'variables is None' here since passing an empty 'variables' shouldn't match anything.

openmdao/recorders/case.py
```diff
@@ -392,15 +395,371 @@ def get_responses(self, scaled=True, use_indices=True):
...
+                    del meta[key]
+                except KeyError:
+                    pass
+
+        if var_dict:
+            # variables should be grouped by component in the order those components are executed
+            if 'execution_order' in self._var_info:
+                var_list = []
+                var_order = self._var_info['execution_order']
+                # execution order will
```
naylor-b, line 734: Is this comment missing something?

openmdao/recorders/case.py
```diff
@@ -392,15 +395,371 @@ def get_responses(self, scaled=True, use_indices=True):
...
+                cmp_names = [var_name.rsplit('.') for var_name in var_order if var_name in var_dict]
+                cmp_order = []
+                for cmp_name in cmp_names:
+                    if cmp_name not in cmp_order:
+                        cmp_order.append(cmp_name)
```
naylor-b, line 739: It doesn't look like `cmp_order` gets used.

naylor-b (approved):

openmdao/recorders/tests/test_list_vars.py
```diff
@@ -0,0 +1,299 @@
...
+        prob.run_model()
+        prob.cleanup()
+
+        case = om.CaseReader('list_vars.db').get_case(0)
+
+        for arg in (
+            'val',
+            'prom_name',
+            'residuals',
+            # 'residuals_tol=None,
```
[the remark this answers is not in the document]
swryan: I left them there on purpose, as documentation that those args were not being tested here because they are not boolean.

I can remove them or add a comment to clarify, whichever...

robfalck (approved):

openmdao/recorders/tests/test_list_vars.py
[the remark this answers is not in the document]
swryan: I updated the test to use a  more robust method of identifying boolean args and testing with the non-default value

PR #3295, opened 2024-07-15
Added InputResidsComp to the list of components. Added `setup_residuals` method for compatibility with dynamic shaping.

### LINKED ISSUE #3294: InputResidsComp and dynamic-IO-compatible `add_residual`.
### Description

The `add_residual` method is sometimes useful for adding residuals associated with input variables. However, if those variables use dynamic shaping, the current implementation makes it impossible to automatically determine the correct size of the residual.

Since the sizes are all known during `final_setup`, I propose adding a new `setup_residuals` method, similar to `setup_partials`, that is executed during `final_setup` after all IO shapes are known.

We should be encouraging more users to move to automatic shaping, and this implementation would remove one roadblock to doing so.

### Example

The InputResidsComp definition follows.  Note the use of both `setup_partials` and `setup_residuals` to make it compatible with dynamic shaping.

```
import numpy as np

import openmdao.api as om


class InputResidsComp(om.ImplicitComponent):
    """Class definition for the InputResidsComp.

    Uses all inputs as residuals while allowing individual outputs that are not necessarily
    associated with a specific residual.

    Parameters
    ----------
    **kwargs : dict
        Dictionary of optional arguments.

    """
    def __init__(self, **kwargs):
        self._refs = {}
        super().__init__(**kwargs)

    def add_input(self, name, val=1.0, shape=None, units=None, desc='', tags=None,
                  shape_by_conn=False, copy_shape=None, compute_shape=None, distributed=None,
                  ref=None):
        """Add an input to be used as a residual.

        Parameters
        ----------

        name : str
            Name of the variable in this component's namespace.
        val : float or list or tuple or ndarray or Iterable
            The initial value of the variable being added in user-defined units.
            Default is 1.0.
        shape : int or tuple or list or None
            Shape of this variable, only required if val is not an array. Default is None.
        units : str or None
            Units in which this input variable will be provided to the component
            during execution. Default is None, which means it is unitless.
        desc : str
            Description of the variable.
        tags : str or list of strs
            User defined tags that can be used to filter what gets listed when calling
            list_inputs and list_outputs.
        shape_by_conn : bool
            If True, shape this input to match its connected output.
        copy_shape : str or None
            If a str, that str is the name of a variable. Shape this input to match that of
            the named variable.
        compute_shape : function
            A function taking a dict arg containing names and shapes of this component's outputs
            and returning the shape of this input.
        distributed : bool
            If True, this variable is a distributed variable, so it can have different sizes/values
            across MPI processes.
        ref : float or ndarray or None
            Scaling parameter. The value in the user-defined units of this residual
            when the scaled value is 1. Default is 1.

        """
        self._refs[name] = ref
        super().add_input(name, val=val, shape=shape, units=units, desc=desc, tags=tags,
                          shape_by_conn=shape_by_conn, copy_shape=copy_shape,
                          compute_shape=compute_shape, distributed=distributed)

    def setup_residuals(self):
        for name in self._var_rel_names['input']:
            meta = self._var_rel2meta[name]
            resid_name = f'resid_{name}'
            self.add_residual(resid_name, shape=meta['shape'], units=meta['units'],
                              desc=meta['desc'], ref=self._refs[name])

    def setup_partials(self):
        size = self._get_var_meta('x', 'size')
        self.mat = np.eye(size) * 3.
        rng = np.arange(size)
        self.declare_partials('y', 'x', rows=rng, cols=rng, val=3.0)
        for name in self._var_rel_names['input']:
            resid_name = 'resid_' + name
            size = self._get_var_meta(name, 'size')
            ar = np.arange(size, dtype=int)
            self.declare_partials(of=resid_name, wrt=name, rows=ar, cols=ar, val=1.0)

    def apply_nonlinear(self, inputs, outputs, residuals):
        """Compute residuals given inputs and outputs.

        The model is assumed to be in an unscaled state.

        Parameters
        ----------
        inputs : Vector
            Unscaled, dimensional input variables read via inputs[key].
        outputs : Vector
            Unscaled, dimensional output variables read via outputs[key].
        residuals : Vector
            Unscaled, dimensional residuals written to via residuals[key].

        """
        residuals.set_val(inputs.asarray())
```

### OpenMDAO Version

3.33.1-dev

_No response_

robfalck:
### Summary

Adds `InputResidsComp` to the list of OpenMDAO components.
`InputResidsComp` reduces boilerplate necessary to implement a simple implicit system by treating the added inputs as residual values.

This takes advantage of renamed residuals to allow the `InputResidsComp` to have a different number of inputs (residuals) and implicit outputs (states) so long as the total size of the inputs and outputs is equal.

To make `InputResidsComp` compatible with dynamic shaping, a new `setup_residuals` method was added that allows for the addition of residuals during final setup, once the sizes of inputs and outputs has been determined. Overriding `setup_residuals` in an explicit component will result in an error.

### Related Issues

- Resolves #3294

None

### New Dependencies

None


openmdao/components/input_resids_comp.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def setup_partials(self):
+        """
+        Delay calls to declare_partials for the component.
+
+        This method is used because input/residual sizes
+        may not be known until final setup.
+        """
+        size = self._get_var_meta('x', 'size')
+        self.mat = np.eye(size) * 3.
```
naylor-b, line 102: self.mat doesn't appear to be used anywhere

openmdao/components/input_resids_comp.py
```diff
@@ -0,0 +1,126 @@
...
+        rng = np.arange(size)
+        self.declare_partials('y', 'x', rows=rng, cols=rng, val=3.0)
+        for name in self._var_rel_names['input']:
+            resid_name = 'resid_' + name
+            size = self._get_var_meta(name, 'size')
```
naylor-b, line 107: Since you already have the relative name, it would be quicker to just get size from self._var_rel2meta[name]['size']

naylor-b (changes requested):

swryan (approved):

naylor-b (approved):

PR #3296, opened 2024-07-16
Scaling report was causing an extra call to compute totals if there were linear constraints

### LINKED ISSUE #3231: Scaling report can cause an extra call to compute_totals when there are linear constraints
### Description

Bernardo Pacini at U of Michigan was seeing an extra call to compute_totals which took 90 minutes for his bigger models.  The cause is that the hook function that kicks off the scaling report is being triggered after the computation of the linear constraint derivatives, which does not set the '_total_jac' attribute on the Driver.  That `_total_jac` attribute is used by the scaling report to determine if it should call `_compute_totals` on the Driver.  

As a general solution that may be useful for other hook functions in the future, we could update the hook machinery to allow a boolean function to be passed to `register_hook` that will be evaluated prior to triggering the hook, and the hook would only be triggered if the boolean function returned True.

### Example

No example code provided.

### OpenMDAO Version

3.32.1

_No response_

naylor-b:
### Summary

Added a 'predicate' option to hooks to allow the post _compute_totals hook skip firing on the first call if that call was to compute the linear constraint jacobian.

### Related Issues

- Resolves #3231

None

### New Dependencies

None


openmdao/utils/hooks.py
```diff
@@ -80,8 +80,12 @@ class _HookMeta(object):
         If True, pass the hooked function's positional and keyword arguments to the hook function.
     pass_return : bool
         If True, pass the return value to the hook function.  Only valid for post hooks.
-    **kwargs : dict of keyword arguments
-        Keyword arguments that will be passed to the hook function.
+    predicate : function or None
+        If not None, a function that will be called to determine if the hook should run. The
+        function should take the instance as its only argument, returning True if the hook should
```
swryan, line 85: `take the instance as its only argument` does not seem to be True:
```
return self.predicate(inst, **self.reg_kwargs)
```

swryan (approved):

robfalck (approved):

openmdao/utils/hooks.py
swryan: no problem, very minor

PR #3287, opened 2024-06-26
POEM 097 Implementation, Part 1

robfalck:
### Summary

This POEM begins the reorganization of problem output files for [POEM 097](https://github.com/OpenMDAO/POEMs/pull/198).

### Related Issues

- Resolves #

### Backwards incompatibilities

The global `_reports_dir` variable is removed, along with associated functions (set_reports_dir, reset_reports_dir).

Reports may not be generated pre-setup, as the path to the reports directory is not known until setup time.

problem_setup_stack is removed in favor of a parent/child hierarchy for problems.

### New Dependencies

None

PR #3298, opened 2024-07-17
POEM 097 Implementation - Part 2

robfalck:
### Summary

This Pull request builds upon #3287 to complete the implementation of POEM 097.

Output coloring files are now placed in `f'{prob_name}_out/coloring_files'`, regardless of where fixed coloring files come from.

Problem `coloring_dir` directory is now _DEFAULT_COLORING_DIR by default. When set to this value, the coloring directory used for _input_ (loading fixed colorings) will be set to the `coloring_dir` subdirectory in the problem output directory.

Using named problems will make loading this existing coloring more simple.

Added `InvalidColoringError` class to be raised when the coloring check finds an issue with a coloring. The next step is to trigger an automatic dynamic coloring when the found static coloring is invalid.

### Related Issues

- Resolves #

None

### New Dependencies

None


swryan (approved):

openmdao/core/problem.py
```diff
@@ -2543,6 +2544,33 @@ def get_outputs_dir(self, *subdirs, mkdir=True):
...
+        mkdir : bool
+            If True, attempt to create this directory if it does not exist.
+
+        Returns
+        -------
+        pathlib.Path
+            The path to the directory where reports should be written.
+        """
+        if mode == 'input':
+            return pathlib.Path(self.options['coloring_dir'])
```
Kenneth-T-Moore, line 2567: It doesn't looks like the 'input' mode is used in the code or covered in the tests.

openmdao/core/system.py
```diff
@@ -1843,11 +1859,14 @@ def _get_static_coloring(self):
...
             if static is _STD_COLORING_FNAME:
-                fname = self.get_coloring_fname()
+                fname = self.get_coloring_fname(mode='input')
             else:
                 fname = static
-            print("%s: loading coloring from file %s" % (self.msginfo, fname))
+            print(f"{self.msginfo}: loading coloring from file {fname}")
             info.coloring = coloring = Coloring.load(fname)
+
+            self._save_coloring(coloring)
```
Kenneth-T-Moore, line 1868: Why didn't we save a coloring here before?

naylor-b (approved):

openmdao/core/system.py
robfalck: Previously coloring files were in `$PWD/coloring_files`.  Now that we have an output directory, my intent was to save all the coloring files associated with an execution to the output directory, even if they came from another location (i.e. "input" coloring files).  This basically just says "save off the coloring being used."

Kenneth-T-Moore (approved):

PR #3307, opened 2024-07-24
fixed issue with ScipyOptimizeDriver introduced in the scaling report PR

naylor-b:
### Summary

Disabled splitting of DVs into linear and nonlinear for ScipyOptimizeDriver because it doesn't split them internally like pyOptSparseDriver does.  Added a supports['linear_only_designvars"] option to Driver to indicate to the framework whether DVs should be split or not.

Was not able to reproduce the dymos tests failures using an OpenMDAO-only test case.

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/pyoptsparse_driver.py
+++ b/openmdao/drivers/pyoptsparse_driver.py
@@ -207,6 +207,7 @@ class pyOptSparseDriver(Driver):
         self.supports['multiple_objectives'] = True
         self.supports['two_sided_constraints'] = True
         self.supports['linear_constraints'] = True
+        self.supports['linear_only_designvars'] = True
         self.supports['simultaneous_derivatives'] = True
         self.supports['total_jac_sparsity'] = True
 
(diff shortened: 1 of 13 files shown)


openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -207,6 +207,7 @@ def __init__(self, **kwargs):
         self.supports['multiple_objectives'] = True
         self.supports['two_sided_constraints'] = True
         self.supports['linear_constraints'] = True
+        self.supports['linear_only_designvars'] = True
```
swryan, line 210: could I suggest adding a 'desc' to this option?  I know the others don't have it, but this one seems a bit subtle, having to do with the driver implementation ("doesn't split them internally "), and it may not be clear to users what it means...

swryan (approved):

robfalck (approved):

swryan:
The significant part of this PR was incorporated into #3308.  There were a couple of changes that could not be included in that PR and are residual here after merging out.

PR #3318, opened 2024-08-02
Added `openmdao clean` command line utility and `openmdao.api.clean_outputs` function.

### LINKED ISSUE #3317: Add `openmdao clean` command line utility.

As part of POEM 097 we create unique output directories so that multiple OpenMDAO scripts in the same directory do not clobber each others output files.

One side-effect of this is that OpenMDAO tends to generate a lot of output directories now, especially in testing and doc building.

To mitigate this, we're adding a the ability to automatically find OpenMDAO output files and remove them.

This will be accomplished with:
1. `openmdao clean` will recurse through the given path and remove any OpenMDAO output directories by default.
2. `openmdao.api.clean_outputs` will clean output directories from within python scripts.
3. OpenMDAO problem output directories will contain an empty file `.openmdao_out` so that we're not locating output directories on naming patters alone.

### Associated POEM

097

robfalck:
### Summary

Added the ability to recursively find and remove OpenMDAO output directories to keep users filesystems from getting cluttered.

### Related Issues

- Resolves #3317 

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/other_useful_docs/om_command.ipynb
+++ b/openmdao/docs/openmdao_book/other_useful_docs/om_command.ipynb
@@ -185,6 +185,32 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "(om-command-clean)=\n",
+    "### openmdao clean\n",
+    "\n",
+    "OpenMDAO creates an output directory for each problem it runs. This directory contains recording files, optimizer output files, derivative coloring information (if generated), and reports (if generated).\n",
+    "The `openmdao clean` command recurse through the given path and remove any OpenMDAO output directories found.\n",
+    "\n",
+    "OpenMDAO output directories are identified by the presence of an `.openmdao_out` file, which is automatically added upon their creation."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "remove-input"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "!openmdao clean --help"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "(om-command-n2)=\n",
     "### openmdao n2\n",
     "\n",
--- a/openmdao/docs/openmdao_book/other_useful_docs/om_command.ipynb
+++ b/openmdao/docs/openmdao_book/other_useful_docs/om_command.ipynb
@@ -189,7 +189,7 @@
     "### openmdao clean\n",
     "\n",
     "OpenMDAO creates an output directory for each problem it runs. This directory contains recording files, optimizer output files, derivative coloring information (if generated), and reports (if generated).\n",
-    "The `openmdao clean` command recurse through the given path and remove any OpenMDAO output directories found.\n",
+    "The `openmdao clean` command recurses through the given path and remove any OpenMDAO output directories found.\n",
     "\n",
     "OpenMDAO output directories are identified by the presence of an `.openmdao_out` file, which is automatically added upon their creation."
    ]
--- a/openmdao/utils/file_utils.py
+++ b/openmdao/utils/file_utils.py
@@ -1,7 +1,8 @@
 """
 Utilities for working with files.
 """
-
+from collections.abc import Iterable
+from itertools import chain
 from multiprocessing import Value
 import sys
 import os
@@ -10,6 +11,10 @@ import types
 from fnmatch import fnmatch
 from os.path import join, basename, dirname, isfile, split, splitext, abspath
 import pathlib
+import shutil
+
+import openmdao
+from openmdao.utils.om_warnings import issue_warning
 
 
 def get_module_path(fpath):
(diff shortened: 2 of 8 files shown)


openmdao/utils/file_utils.py
```diff
@@ -481,3 +489,117 @@ def _get_outputs_dir(obj=None, *subdirs, mkdir=True):
...
+    directory : str or Path
+        The directory to check.
+
+    Returns
+    -------
+    bool
+        True if the directory is an OpenMDAO output directory, False otherwise.
+    """
+    directory = pathlib.Path(directory)
+    return directory.is_dir() and directory.name.endswith('_out') \
```
swryan, line 509: it seems like the directory_name check is unnecessary? the presence of the dot file is conclusive...
(maybe the is_dir check is redundant as well)

swryan (approved):

openmdao/utils/file_utils.py
robfalck: It probably is, but when it comes to automatically removing things from a user's system I wanted to do belt and suspenders.

openmdao/docs/openmdao_book/other_useful_docs/om_command.ipynb
```diff
@@ -181,6 +181,32 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "(om-command-clean)=\n",
+    "### openmdao clean\n",
+    "\n",
+    "OpenMDAO creates an output directory for each problem it runs. This directory contains recording files, optimizer output files, derivative coloring information (if generated), and reports (if generated).\n",
+    "The `openmdao clean` command recurse through the given path and remove any OpenMDAO output directories found.\n",
```
naylor-b, line 192: recurse -> recurses

naylor-b (approved):

openmdao/utils/file_utils.py
```diff
@@ -481,3 +489,117 @@ def _get_outputs_dir(obj=None, *subdirs, mkdir=True):
...
+    Parameters
+    ----------
+    obj : Problem or System or Solver or str or Path
+        The problem or system or solver whose output file should be removed.
+    recurse : bool
+        If True, and if obj is a string or Path, recurse into it
+        finding and removing OpenMDAO output directories along the way.
+        This option is ignored if obj is a Problem, System, or Solver.
+    prompt : bool
+        If True, prompt the user to confirm directories to be removed.
```
Kenneth-T-Moore, line 570: With the "prompt" option, my expectation would be that it would prompt me to delete each one individually instead of all at once. I think that is the only reason i would use this.

openmdao/utils/om.py
```diff
@@ -280,6 +281,40 @@ def _tree_setup_parser(parser):
...
+        The parser we're adding options to.
+    """
+    parser.add_argument('path', nargs='*', default='.',
+                        help='Path(s) from which OpenMDAO output directories should be remove.')
+    parser.add_argument('-f', '--force', action='store_true', dest='force',
+                        help='Remove output directories without confirmation.')
+    parser.add_argument('-n', '--norecurse', action='store_true', dest='no_recurse',
+                        help='Do not recurse into subdirectories to find directories to remove.')
+    parser.add_argument('-d', '--dryrun', action='store_true', dest='dryrun',
+                        help='Highlight directories to be removed but do not actually remove them.')
```
Kenneth-T-Moore, line 300: I think maybe prompt should be in the parser args.  It probably should also default to True.

I am a little worried about accidentally running `openmdao clean` at the root directory and accidentally deleting years worth of runs without realizing it.

openmdao/utils/file_utils.py
```diff
@@ -439,6 +444,10 @@ def _get_outputs_dir(obj=None, *subdirs, mkdir=True):
     The resulting outputs directory will be nested where each problem's output directory
     contains its own output files and subdirectories as well as any subproblems.
 
+    This directory also will include a .openmdao_outputs hidden file that
```
Kenneth-T-Moore, line 447: This is a useful identifier, though I wonder if we should warn people not to use that filename for anything else.

Kenneth-T-Moore (changes requested):
Holding for my comments since we already have 2 approvals.

openmdao/utils/file_utils.py
robfalck: In the command line usage, prompt is on by default.  The `--force` option will do the deletion without prompting, but otherwise the user will get an "are you sure you want to delete these?" prompt.

openmdao/utils/om.py
robfalck: Right now `prompt` is on by default, and the `--force` argument is the same as `do not prompt`.  I'll change it to `--noprompt` for consistency.

openmdao/utils/file_utils.py
robfalck: We should, but I'm going to write more thorough documentation of the output directory as part of part 3 of the POEM, since POEM 097 will be fully implemented at that point.

Kenneth-T-Moore (approved):

PR #3321, opened 2024-08-06
added better error msg when a distributed variable is part of an invalid connection or is unconnected

### LINKED ISSUE #3312: Unclear error message when improperly connecting to distributed inputs
### Description

A common issue when setting up complicated MPhys models is missing or incorrect connections. The error message from OpenMDAO can be unclear in these cases when dealing with distributed variables. For the example below, I've intentionally misnamed the IVC's output that should be connected to distributed input of the adder. The resulting error message is:

```
    raise RuntimeError(f'Distributed component input "{tgt}" requires an IndepVarComp.')
RuntimeError: Distributed component input "adder.x" requires an IndepVarComp.
```
This could be more helpful in a couple of ways: 1) the inputs to adder.x doesn't not necessarily require an IVC; it could be the output of another implicit or explicit component, so the reported error is incorrect information. 2) The source of the error is that the input variable/component that is the source in the connection does not exist. This information could be relayed to the user rather than saying a new component is needed.

### Example

```python
import numpy as np
import openmdao.api as om
import openmdao
print(openmdao.__version__)

class Adder(om.ExplicitComponent):
    def setup(self):
        self.add_input('x', shape_by_conn=True, distributed=True)
        self.add_output('x_sum', shape=1)

        outputs['x_sum'] = np.sum(inputs['x'])

prob = om.Problem()
ivc.add_output('x', val = np.ones(10), distributed=True)

prob.model.add_subsystem('adder', Adder())

# error here: either a component that doesn't exist or a bad variable name from a valid component gives the same error
prob.model.connect('dvs.x','adder.x')
#prob.model.connect('ivc.x0','adder.x')

prob.setup()
prob.run_model()
```

### OpenMDAO Version

3.34.3-dev

_No response_

naylor-b:
### Summary

The error message no longer erroneously says that the distributed input must be connected to an IndepVarComp output, and it distinguishes between an unconnected input and an input that has been connected to a nonexistent output.
 
### Related Issues

- Resolves #3312

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/group.py
+++ b/openmdao/core/group.py
@@ -200,6 +200,8 @@ class Group(System):
         within this group, keyed by active response.  These determine if contributions
         from all ranks will be added together to get the correct input values when derivatives
         in the larger model are being solved using reverse mode.
+    _bad_conn_vars : set
+        Set of variables involved in invalid connections.
     """
 
     def __init__(self, **kwargs):
@@ -229,6 +231,7 @@ class Group(System):
         self._post_components = None
         self._iterated_components = None
         self._fd_rev_xfer_correction_dist = {}
+        self._bad_conn_vars = None
 
         # TODO: we cannot set the solvers with property setters at the moment
         # because our lint check thinks that we are defining new attributes
@@ -1385,7 +1388,7 @@ class Group(System):
 
         self._resolve_src_indices()
 
-        if self.comm.size > 1:
+        if self.comm.size > 1 and not self._bad_conn_vars:
             abs2idx = self._var_allprocs_abs2idx
             all_abs2meta = self._var_allprocs_abs2meta
             all_abs2meta_in = all_abs2meta['input']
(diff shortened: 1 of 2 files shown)


openmdao/core/group.py
```diff
@@ -200,6 +200,8 @@ class Group(System):
         within this group, keyed by active response.  These determine if contributions
         from all ranks will be added together to get the correct input values when derivatives
         in the larger model are being solved using reverse mode.
+    _bad_conn_vars : set
+        Set of variables involved in invalid connections.
```
swryan, line 204: Set of tuples (input_var, output_var) representing an invalid connection

swryan (approved):

openmdao/core/group.py
naylor-b: It's just a set of variable names, not tuples.  I just pass a tuple containing both the input and output name to the `update` method to avoid having to make 2 calls to the `add` method.

robfalck (approved):

PR #3345, opened 2024-09-10
Changed the N2 to display "Jacobian" when hovering over an input-to-output "connection" within a component, instead of "Connections"

### LINKED ISSUE #3342: N2 Diagram: Blocks in the jacobian shouldn't be labeled "connection" when moused-over.

The blocks within components show something like the following when moused-over and info turned on.

Screenshot 2024-09-06 at 8 35 11 AM

These blocks indicate defined partials not actual connections (in OpenMDAO terms) and some users may find that confusing.

Change the connection label to 'partial' or 'jacobian.'

### Associated POEM

_No response_

robfalck:
### Summary

N2 Diagram now labels the popup window "Jacobian" rather than "Connections" when the user has info enabled and mouses-over a jacobian element within a component..

### Related Issues

- Resolves #3342 

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/gen/NodeInfo.js
+++ b/openmdao/visualization/n2_viewer/gen/NodeInfo.js
@@ -336,7 +336,8 @@ class NodeConnectionInfo extends NodeInfo {
             newRow.append('td').html(conn.tgt);
         }
 
-        let title = 'Connections';
+        let title = (color == "black") ? "Jacobian" : "Connections";
+
         if ( ! (cell.srcObj.isLeaf() || cell.tgtObj.isLeaf()) ) {
             title += ` from ${cell.srcObj.path} to ${cell.tgtObj.path}`;
         }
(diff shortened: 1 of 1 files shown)


Kenneth-T-Moore (approved):

openmdao/visualization/n2_viewer/gen/NodeInfo.js
```diff
@@ -336,7 +336,8 @@ class NodeConnectionInfo extends NodeInfo {
             newRow.append('td').html(conn.tgt);
         }
 
-        let title = 'Connections';
+        let title = (color == "black") ? "Jacobian" : "Connections";
```
robfalck, line 339: Right now this is hard coded to detect if the cell under the cursor is black. Ideally this would use `Style.colors.declaredPartial`, but I haven't been able to get that to work.

swryan, line 339: suggest updating the docstring in line 314 to explain connection vs jacobian

PR #3353, opened 2024-09-18
Added a `require_connection` argument to `add_input`

### LINKED ISSUE #3280: Feature Suggestion: add_input `require_connection` parameter

Is there a way a feature could be added that raises an error if a value is unconnected? When a value if left unconnected, it gets a default value of `1` which is often a perfectly viable value and thus users can often not realize that a value was left unconnected and therefor get unexpected results. 

The approach that's been taken in my project is that we default many things to np.nan that way there's no way it will run unless all required connections have been made, but this seems like a feature that others might benefit from in a more formal way at the API level.

### Associated POEM

_No response_

swryan:
### Summary

Added a `require_connection` argument to `add_input`.   If the input is not connected at `setup` time and is not a design variable, an Exception is raised.

### Related Issues

- Resolves #3280

None

### New Dependencies

None


openmdao/core/tests/test_group.py
```diff
@@ -489,6 +489,46 @@ def test_reused_output_promoted_names(self):
         with self.assertRaisesRegex(Exception, msg):
             prob.setup()
 
+    def test_unconnected_input_required_connection(self):
```
naylor-b, line 492: I didn't see it but might be good to have a test for when the variable isn't connected but is declared as a design variable.

openmdao/core/component.py
```diff
@@ -519,6 +520,8 @@ def add_input(self, name, val=1.0, shape=None, units=None, desc='', tags=None,
         compute_shape : function
             A function taking a dict arg containing names and shapes of this component's outputs
             and returning the shape of this input.
+        require_connection : bool
+            If True, this input must be connected to an output.
```
naylor-b, line 524: I would modify the description to include being declared as a design variable.

naylor-b (changes requested):

openmdao/core/group.py
```diff
@@ -783,6 +783,29 @@ def _setup(self, comm, prob_meta):
         # determine which connections are managed by which group, and check validity of connections
         self._setup_connections()
```
robfalck, line 784: OpenMDAO [allows connections to be issued in configure](https://openmdao.org/newdocs/versions/latest/theory_manual/setup_stack.html#usage-of-setup-vs-configure), so this check should happen at the end of configure.

robfalck (changes requested):

robfalck (approved):

PR #3359, opened 2024-09-20
Added JaxExplicitComponent and JaxImplicitComponent

naylor-b:
### Summary

- Updated the Component API to allow any Component to define a `compute_primal` method instead of a `compute` method (explicit) or an `apply_nonlinear` method (implicit).  Doing this should probably be encouraged because not only does it make it easier to do AD on a component, but it also has less overhead than all of the dictionary access that's required to pull values out of the Vector arguments that are passed to `compute` or `apply_linear`.
- made a command line tool `openmdao to_compute_primal`, to do automatic conversion from compute/apply_nonlinear to compute_primal. It doesn't work in all cases, but for typical components it saves some effort.
- Created JaxExplicitComponent and JaxImplicitComponent.  If a new component inherits from one of these, the component author only needs to define `compute_primal` and jax will be used automatically.  If the component uses static variables from 'self' in compute_primal, then another method, `get_self_statics` must be defined for that component.
- Setting the `matrix_free` attribute of a jax component will cause the component to use `jax.jvp` or `jax.vjp` to compute derivatives without creating a jacobian.

### New Dependencies

jax is required to use either JaxExplicitComponent or JaxImplicitComponent, but otherwise it's optional.


openmdao/docs/openmdao_book/features/features.ipynb
```diff
@@ -77,7 +77,9 @@
     "\n",
     "- [Determining Variable Shapes at Runtime](experimental/dyn_shapes.ipynb)\n",
     "- [Simultaneous Coloring of Approximated Derivatives](experimental/approx_coloring.ipynb)\n",
-    "- [Working with Plugins](experimental/plugins.ipynb)\n"
+    "- [Working with Plugins](experimental/plugins.ipynb)\n",
+    "- [Jax for ExplicitComponents using the JaxExplicitComponent class](experimental/jax_explicitcomp_api.ipynb\n",
```
swryan, line 81: missing closing paren here, extra in the next line

openmdao/docs/openmdao_book/features/experimental/jax_explicitcomp_api.ipynb
```diff
@@ -0,0 +1,295 @@
...
+    "        self.add_output('y1', val=1.0, lower=0.1, upper=1000., ref=0.1)\n",
+    "\n",
+    "    def setup_partials(self):\n",
+    "        self.declare_partials('*', '*')\n",
+    "\n",
+    "    # because our compute primal output depends on static variables, in this case self.staticvar\n",
+    "    # and self.options['static_opt'], we must define a get_self_statics method. This method must\n",
+    "    # return a tuple of all static variables. Their order in the tuple doesn't matter.  If your\n",
+    "    # component happens to have discrete inputs, do NOT return them here. Discrete inputs are passed\n",
+    "    # into the compute_primal function individually, after the continuous variables.\n",
```
swryan, line 139: this says discrete inputs come "after the continuous variables", but in line 96 above it says "prior to the continuous input variables".

openmdao/core/explicitcomponent.py
```diff
@@ -120,6 +135,21 @@ def _setup_partials(self):
         """
         Call setup_partials in components.
         """
+        # if self.options['derivs_method'] in ('cs', 'fd'):
```
Kenneth-T-Moore, line 138: Is this block of coded waiting for a future PR?

naylor-b: No I moved that block into Component.  It's cleaned up now.

openmdao/docs/openmdao_book/features/features.ipynb
naylor-b: fixed

openmdao/components/jax_implicit_comp.py
```diff
@@ -0,0 +1,36 @@
...
+An ImplicitComponent that uses JAX for derivatives.
+"""
+
+import sys
+from openmdao.core.implicitcomponent import ImplicitComponent
+from openmdao.utils.jax_utils import jax
+from openmdao.utils.om_warnings import issue_warning
+
+
+class JaxImplicitComponent(ImplicitComponent):
```
Kenneth-T-Moore, line 11: There really isn't much in the jax imp and exp components, aside from option values. I thought there might be something to prevent the user from defining compute or compute_partials.  Do we really need the components?

openmdao/core/explicitcomponent.py
```diff
@@ -129,19 +144,13 @@ def _setup_partials(self):
         # call the super version of setup_partials. This is still in the final setup.
         for out_abs, meta in self._var_abs2meta['output'].items():
 
-            # No need to FD outputs wrt other outputs
```
Kenneth-T-Moore, line 132: Why don't we need to delete output-output keys anymore?

openmdao/core/explicitcomponent.py
```diff
@@ -500,18 +509,36 @@ def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
...
         outputs : Vector
             Unscaled, dimensional output variables read via outputs[key].
-        discrete_inputs : dict or None
-            If not None, dict containing discrete input values.
-        discrete_outputs : dict or None
-            If not None, dict containing discrete output values.
+        discrete_inputs : dict like or None
+            If not None, dict like object containing discrete input values.
+        discrete_outputs : dict like or None
+            If not None, dict like object containing discrete output values.
```
Kenneth-T-Moore, line 524: probably "dict-like" is more clear

openmdao/core/group.py
```diff
@@ -689,7 +718,7 @@ def _list_states(self):
         for subsys in self._sorted_sys_iter():
             states.extend(subsys._list_states())
 
-        return sorted(states)
+        return states
```
Kenneth-T-Moore, line 721: Was this `sorted` originally added for MPI? Just making sure, because those bugs are hard to find.

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -712,7 +712,7 @@ def _objfunc(self, dv_dict):
                 func_dict = self.get_objective_values()
                 func_dict.update(self.get_constraint_values(lintype='nonlinear'))
                 # convert func_dict to use promoted names
-                func_dict = model._prom_names_dict(func_dict)
+                # func_dict = model._prom_names_dict(func_dict)
```
Kenneth-T-Moore, line 715: Why are we sending absolute names to pyoptsparse now?

openmdao/components/jax_implicit_comp.py
naylor-b: Most of the machinery for jax lives in ExplicitComponent and ImplicitComponent, because my original intention was to allow for an entire Group to be 'jaxxed', which required converting normal openmdao components to jax at runtime.  It's not clear at this point whether the Group functionality will be useable or not, so for now everything stays where it is, but if we decide in the future that the Group stuff isn't viable we can move all of that functionality into JaxExplicitComponent and JaxImplicitComponent.

openmdao/core/explicitcomponent.py
naylor-b: That logic didn't really make sense because any changes it made would have just been overwritten at the end of the loop when the [out_abs, out_abs] entry is set to a completely new dict.

openmdao/core/group.py
naylor-b: list_states returns only local states so I don't think it'll cause any MPI bugs.  The states are sorted anyway because of _sorted_sys_iter.  There is a separate method called _list_states_allprocs that currently returns a set, which is definitely not sorted, but it's currently only used in SqliteRecorder and it isn't iterated over so that shouldn't be a problem either.

openmdao/core/explicitcomponent.py
naylor-b: fixed

openmdao/drivers/pyoptsparse_driver.py
naylor-b: No the keys stored in the internal dicts are now promoted names or aliases.

Kenneth-T-Moore (approved):

robfalck (approved):

PR #3360, opened 2024-09-21
POEM_96 Implementation: Minimization of Constraint Violation (SciPy Driver Only)

andrewellis55:
### Summary

I've created a possible implementation for POEM_96 (https://github.com/OpenMDAO/POEMs/pull/194). I tried to make minimal changes to the overall codebase, so this approach adds a driver level option `minimize_constraint_violation` to activate this functionality. Currently this is only implemented for the SciPy driver as it's most relevant to SLSQP. Those running pyOptSparse with SNOPT or IPOPT already get a form of this build into those optimizers.

This proposal modified the `ScipyOptimizeDriver._objfunc` and `ScipyOptimizeDriver._gradfunc` functions directly to achieve this.

I'll add tests and updates to the docs if this approach is deemed reasonable. I've tested this in my teams main tool that runs OpenMDAO and found it works well and is quite useful. I can share some examples of using the feature in our tool over video call if you would like.

### Example Code
Here's a breif example to run the feature. When toggled on, the constraint violation is printed as the `Current function value` in the SciPy printout
```python
import openmdao.api as om

# Define a simple model
prob = om.Problem()
model = prob.model

# Design Vars
ivc: om.IndepVarComp = prob.model.add_subsystem('ivc', om.IndepVarComp(), promotes_outputs=['*'])
ivc.add_output('x', val=5.0)
ivc.add_design_var('x', lower=0.0, upper=10.0)

# Constraints
con = model.add_subsystem('comp', om.ExecComp('y = (x)**2'), promotes=['*'])
con.add_constraint('y', lower=1.25, upper=4.0)

con2 = model.add_subsystem('comp2', om.ExecComp('y2 = x+0.5'), promotes=['*'])
con2.add_constraint('y2', equals=2)

# Objective
model.add_subsystem('obj_comp', om.ExecComp('obj = (x-1)**2'), promotes=['*'])
model.add_objective('obj')

# Driver
prob.driver.options['optimizer'] = 'SLSQP'
prob.driver.options['minimize_constraint_violation'] = True  # Minimize constraint violations

# Setup the problem and run the optimization
prob.setup()
prob.run_driver()

# Print the final design variable and constraint values
print(f"Final design variable 'x': {prob['x']}")
print(f"Final constraint 'y': {prob['y']}")
print(f"Objective 'obj': {prob['obj']}")
```

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/scipy_optimizer.py
+++ b/openmdao/drivers/scipy_optimizer.py
@@ -14,6 +14,7 @@ from openmdao.core.driver import Driver, RecordingDebugging
 from openmdao.core.group import Group
 from openmdao.utils.class_util import WeakMethodWrapper
 from openmdao.utils.mpi import MPI
+from openmdao.utils.om_warnings import issue_warning, DriverWarning
 
 
 # Optimizers in scipy.minimize
@@ -189,6 +190,9 @@ class ScipyOptimizeDriver(Driver):
                              "ignore - don't perform check.")
         self.options.declare('singular_jac_tol', default=1e-16,
                              desc='Tolerance for zero row/column check.')
+        self.options.declare('minimize_constraint_violation', default=False, types=bool,
+                            desc='Set to True to minimize the sum of constraint violations '
+                            'instead of the objective function.')
 
     def _get_name(self):
         """
@@ -261,6 +265,12 @@ class ScipyOptimizeDriver(Driver):
 
         self._check_for_missing_objective()
         self._check_for_invalid_desvar_values()
+
+        if self.options['minimize_constraint_violation']:
+            issue_warning('As `minimize_constraint_violation` is set to `True`, the '
+                          + 'objective function will be ignored and the sum of '
+                          + 'the constraint violation will instad be minimized.',
+                          category=DriverWarning)
 
         # Initial Run
         with RecordingDebugging(self._get_name(), self.iter_count, self):
@@ -463,6 +473,11 @@ class ScipyOptimizeDriver(Driver):
         # compute dynamic simul deriv coloring if option is set
         prob.get_total_coloring(self._coloring_info, run_model=False)
 
+        if self.options['minimize_constraint_violation']:
+            constraints_to_pass = []
+        else:
+            constraints_to_pass = constraints
+
         # optimize
         try:
             if opt in _optimizers:
@@ -476,7 +491,7 @@ class ScipyOptimizeDriver(Driver):
                                   hess=hess,
                                   # hessp=None,
                                   bounds=bounds,
-                                  constraints=constraints,
+                                  constraints=constraints_to_pass,
                                   tol=self.options['tol'],
                                   # callback=None,
                                   options=self.opt_settings)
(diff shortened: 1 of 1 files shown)


robfalck:
Thanks for the contribution!

I like the implementation.  This seems like something we should be able to extend to the other optimizers.

I'm in the process of reworking the DOEDriver and following that I'm going to try to generalize an OptimizationDriver base class, and I think I'll be able to implement this there.

Have you tested any other way of adding the constraint violation to the pseudo objective?  I do wonder if the non-differentiable min/max functions will present any challenges in larger problems.

andrewellis55:
I've now added an option for L1 or L2 norm minimization. L2 should have continuous derivatives as a violation crosses the 0 point.

Here's my script I was playing around with to test L1 and L2

```python
import openmdao.api as om

# Define a simple model
prob = om.Problem()
model = prob.model

# Design Vars
ivc: om.IndepVarComp = prob.model.add_subsystem('ivc', om.IndepVarComp(), promotes_outputs=['*'])
ivc.add_output('x', val=5)
ivc.add_design_var('x', lower=-50, upper=10.0)
ivc.add_output('xx', val=0.5)
ivc.add_design_var('xx', lower=0, upper=5)
ivc.add_output('xxx', val=0.5)
ivc.add_design_var('xxx', lower=0.6, upper=5)
ivc.add_output('xxxx', val=0.5)
ivc.add_design_var('xxxx')

# Constraints
con = model.add_subsystem('comp', om.ExecComp('y = (x+0.10)**2'), promotes=['*'])
con.add_constraint('y', lower=1, upper=4.0)

con2 = model.add_subsystem('comp2', om.ExecComp('y2 = xxx+0.5'), promotes=['*'])
con2.add_constraint('y2', lower=2, upper=2.2)

con3 = model.add_subsystem('comp3', om.ExecComp('y3 = log(xx-0.3)'), promotes=['*'])
con3.add_constraint('y3', lower=0.42, upper=0.5)

con3 = model.add_subsystem('comp4', om.ExecComp('y4 = log(xxxx)'), promotes=['*'])
con3.add_constraint('y4', equals=3)

# Objective
model.add_subsystem('obj_comp', om.ExecComp('obj = (x-1)**2'), promotes=['*'])
model.add_objective('obj')

# Driver
prob.driver.options['optimizer'] = 'L-BFGS-B'
prob.driver.options['minimize_constraint_violation'] = 'L2'  # Minimize constraint violations

# Setup the problem and run the optimization
prob.setup()
prob.run_driver()

prob.list_driver_vars(
    desvar_opts=['lower', 'min', 'upper'],
    cons_opts=['lower', 'min', 'upper', 'equals'],
)
```

andrewellis55:
Based off some experimentation, the L2 seems to work best. 

Additionally, I've been playing around with adding scalers. I think there should be some manner for a user to manually set one. I'll see about adding an option for this tool as they based off some playing around, they seem to be able to help.

robfalck:
@andrewellis55 

My apologies, it's been some time but I've reworked this into a slighly different implementation.

See my branch here at https://github.com/robfalck/OpenMDAO/tree/poem_096_find_feasible

The base Driver class now has a `find_feasible` method.

This will find the minimum constraint violation (in a least squares sense, with some additional options, see below).

Instead of tying it to whatever particular optimizer is in use, and redefining the objective, scipy.least_squares is used.  This obeys bounds on the design variables and allows the L2 norm to be used (default), in addition to a soft L1 approximation as well as others (available through the `loss` argument passed to `scipy.optimize.least_squares`).

For a tool like dymos, I also wanted the ability to minimize the collocation method defects given a control history (find the trajectory that most closely follows the given control history without modifying that control history itself).  There is now an `excludes` keyword that allows one or more design variables to be removed from the feasibility search.

Implementing it this way allows us to pretty easily find a feasible starting point and then trigger the actual optimization.
If you're still using this, let me know what you think.

PR #3365, opened 2024-09-26
AnalysisDriver

robfalck:
### Summary

Adds AnalysisDriver to the list of available drivers.

AnalysisDriver is a more functional form of DOEDriver.
DOEDriver automatically samples points from the design variables and records the outputs.
AnalysisDriver allows the user to explicitly specify inputs via a sample. 
These inputs do not need to be design variables in the model, but the model does check that they are either unconnected inputs (provided by IndepVarComps) or that they are implicit outputs (in which case the initial value provides a guess for the nonlinear solvers).

To easily generate cases, the `AnalysisGenerator` serves as a base class for (currently) two generators, `ProductGenerator` and `ZipGenerator`.  ProductGenerator uses `itertools.product` to provide every possible combination of given inputs to be executed. ZipGenerator uses `zip` to provide pairs from various lists of sample values.

For instance, given a sequence of values for `x` as `[0, 1, 2]` and values for `y` as `[8, 9, 10]`, the `ProductGenerator` would produce 9 cases: `[[0, 8], [0, 9], [0, 10], [1, 8], [1, 9], [1, 10], [2, 8], [2, 9], [2, 10]` while `ZipGenerator` would produce 3 cases: `[[0, 8], [0, 9], [0, 10]]`. These are lazy, pythonic generators that do not realize all possible combinations in memory, so in theory extremely large combinations of variables will be possible now.

For now, `add_response` has been added to `AnalysisDriver` to allow the user to specify an output of interest that may not already be an objective or constraint in the model.

** Note: This is a draft PR provided for review/comment and under active development **

### Related Issues

- Resolves #

None

### New Dependencies

None


openmdao/docs/openmdao_book/features/building_blocks/drivers/analysis_driver.ipynb
```diff
@@ -0,0 +1,580 @@
...
+    "\n",
+    "ProductGenerator, like ZipGenerator, takes a dictionary that provides values and optionally units and indices for\n",
+    "each variable to be sampled.\n",
+    "It uses Python's `itertools.product` to produce every possible combination of the values of sampled variables.\n",
+    "\n",
+    "For instance\n",
+    "\n",
+    "```\n",
+    "ProductGenerator({'x': {'vals': [0.0, 1.0, 2.0], units=None},\n",
+    "                  'y': {'vals': [4.0, 5.0, 6.0], units=None})\n",
```
naylor-b, line 310: I know you mention it later that variables can have different numbers of values, but might be even better to actually have x and y have different numbers of vals in this example.

openmdao/docs/openmdao_book/tests/test_jupyter_output_linting.py
```diff
@@ -59,7 +59,8 @@ def test_output(self):
         for file in _get_files():
             with self.subTest(file):
                 with open(file) as f:
-                    json_data = json.load(f)
+                    data = f.read()
+                    json_data = json.loads(data)
```
naylor-b, line 63: Just curious, why load the file into a string then convert string to json vs. just loading json from the file?

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+        bool
+            Failure flag; True if failed to converge, False is successful.
+        """
+        comm = self._problem_comm
+        model = self._problem().model
+        self.result.reset()
+        self.iter_count = 0
+
+        # Variables allowed samples are the inputs or implicit outputs in the model.
+        # Non-model-inputs would just have their value overridden when evaluating the model.
```
naylor-b, line 253: Non-model-inputs is somewhat confusing.  Aren't we really talking about inputs connected to IVCs?

openmdao/drivers/analysis_generator.py
```diff
@@ -0,0 +1,243 @@
...
+
+class AnalysisGenerator(Iterator):
+    """
+    Provide a generator which provides case data for AnalysisDriver.
+
+    Parameters
+    ----------
+    var_dict : dict
+        A dictionary whose keys are promoted paths of variables to be set, and whose
+        keys are the arguments to `set_val`.
```
swryan, line 22: and whose *_values_* are the arguments to `set_val`

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+
+            while not samples_complete:
+                if comm.rank == 0:
+                    job_queues = [deque() for _ in range(n_procs)]
+                    # Rank 0 pushes batch_size jobs to the ranks in job_queues
+                    for i, sample in enumerate(self._samples):
+                        # Ranks of the same color get the same samples
+                        color_idx = next(color_cycler)
+                        for rank_idx in color_to_rank_map[color_idx]:
+                            job_queues[rank_idx].appendleft((sample_num, sample))
```
naylor-b, line 283: In a later version we might want to revisit how we're distributing the jobs to avoid losing the 'laziness' of some of the generators.

openmdao/drivers/tests/test_analysis_driver.py
```diff
@@ -0,0 +1,556 @@
...
+            self.assertSetEqual(case_nums, {i for i in range(rank, 100, 2)})
+
+            last_case = cr.get_cases(source='driver')[-1]
+            print(last_case)
+            print('inputs')
+            print(last_case.list_inputs())
+            print('outputs')
+            print(last_case.list_outputs())
+
+    # def test_beam_np4(self):
```
swryan, line 365: are you keeping this around for some future work?  should explain or delete...

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+    .. code-block:: python
+
+        prob.driver.add_response('z', units='m')
+
+    In systems with a lot of outputs, this would be a very tedious process,
+    in which case we can use the add_responses method.
+
+    .. code-block:: python
+
+        prob.driver.add_responses(['foo', 'bar', 'baz'])
```
swryan, line 50: `add_reponse` is a `System` method, not `Driver` method...

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+
+            if comm.rank == 0:
+                color_to_rank_map = {num: [i for i, x in enumerate(colors)
+                                     if x == num] for num in set(colors)}
+
+            while not samples_complete:
+                if comm.rank == 0:
+                    job_queues = [deque() for _ in range(n_procs)]
+                    # Rank 0 pushes batch_size jobs to the ranks in job_queues
+                    for i, sample in enumerate(self._samples):
```
naylor-b, line 279: I was under the impression that the self._samples would act like a normal iterable, but if that's the case then this while loop would keep resetting the iterator after each batch since a normal iterable resets at the beginning of a for loop.  Maybe the AnalysisGenerators handle this properly but we also allow lists and tuples as self._samples so it seems like this could be a problem.

swryan (dismissed):

naylor-b (changes requested):

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+
+    Alternatively, samples can be provided as an instance of AnalysisGenerator,
+    which will provide each sample in a lazily-evaluated way.
+
+    Responses are the outputs of the model to be recorded. These can be added
+    using the standard driver interface.
+
+    .. code-block:: python
+
+        prob.driver.add_response('z', units='m')
```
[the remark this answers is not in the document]
Kenneth-T-Moore: I kind of thought we tended to avoid putting code blocks in the OpenMDAO docstrings, since they are difficult to test.

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+            Failure flag; True if failed to converge, False is successful.
+        """
+        comm = self._problem_comm
+        model = self._problem().model
+        self.result.reset()
+        self.iter_count = 0
+
+        # Variables allowed samples are the inputs or implicit outputs in the model.
+        # Non-model-inputs would just have their value overridden when evaluating the model.
+        # Implicit outputs can override the value given in the case, but it might be a
```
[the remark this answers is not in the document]
robfalck: Right. They're allowed because you'd want to provide a guess, but the value coming from the analysis driver will be overridden when the model executes.

openmdao/drivers/analysis_driver.py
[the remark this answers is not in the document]
robfalck: Good point, will remove.

openmdao/docs/openmdao_book/tests/test_jupyter_output_linting.py
robfalck: Just an oversight on my part, will fix.

openmdao/drivers/analysis_driver.py
```diff
@@ -0,0 +1,451 @@
...
+        The number of total MPI colors for the run.
+    _prev_sample_vars : set
+        The set of variables seen in the previous iteration of the driver on this rank.
+    """
+
+    def __init__(self, samples=None, **kwargs):
+        """
+        Construct an AnalysisDriver.
+        """
+        if isinstance(samples, (list, tuple, AnalysisGenerator)):
```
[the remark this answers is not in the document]
robfalck: I was thinking I didn't want to worry about these being recorded, but drivers don't record their options anyway.  I'll change it.

openmdao/drivers/analysis_driver.py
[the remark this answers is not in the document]
robfalck: @swryan regarding this being a system method, for the purpose of the AnalysisDriver it seems like its more natural to interact with the driver and tell it what should be recorded, rather than the system.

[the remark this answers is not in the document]
swryan:
```
    Responses are the outputs of the model to be recorded. These can be added
    using the standard driver interface.
```
This is not True, am I right?    `add_response` is not part of the standard driver interface...  Do you intend to add it to the base class?  or is it a special feature of this driver?

naylor-b (approved):

PR #3369, opened 2024-09-27
Removed unnecessary shape correction of jacobian when doing bidirectional coloring and fixed bidirectional coloring bug

naylor-b:
### Summary

User was asking why he didn't get 3 colors for his bidirectional coloring case.  Removing the shape correction resulted in 3 colors, but during testing after the change, an actual bug in bidirectional coloring was found and then fixed.

The bug was during the formation of the column adjacency matrix for a partition of the jacobian.  Connections in the adjacency matrix were being missed in cases where a column from a different partition  and a column from the current partition both impacted the same row.

- also added current row and column to the tooltip displayed while using the jacobian coloring viewer, e.g., `openmdao total_coloring --view <myfile.py>`.

- added an environment variable, OPENMDAO_NO_RELEVANCE, to allow turning off all relevance while debugging.

### Related Issues

- issue was a StackOverflow post

In some rare cases, due to a change in the jacobian coloring,  the values in the computed jacobian could change slightly.

### New Dependencies

None


andrewellis55:
I was the one that posted the question on Stack Overflow. I've done some testing with your branch and gotten some interesting behavior. This may be due to the heuristics algorithm, but there seems to be some variability in the results based on the size of the jacobian.

I tested your new branch with the same problem I posted on SO and got the same results as before. However your test problem (which has essentially the same structure as mine, just smaller) passes. I did some tests to increase the number of rows on the matrix and as the number of rows grows, the coloring algorithm starts to fail.

Here are my test cases
[Book2.xlsx](https://github.com/user-attachments/files/17179457/Book2.xlsx)

**Large Case**
```python
coloring2 = MNCO_bidir(coo_matrix(pd.read_excel('C:\Book2.xlsx', sheet_name='Large').to_numpy()))
coloring2.summary()
```
```
Jacobian shape: (621, 8)  (25.12% nonzero)
FWD solves: 8   REV solves: 0
Total colors vs. total size: 8 vs 8  (0.00% improvement)
```

**Medium Case (slightly larger than your test)**
```python
coloring2 = MNCO_bidir(coo_matrix(pd.read_excel('C:\Book2.xlsx', sheet_name='Medium').to_numpy()))
coloring2.summary()
```
```
Jacobian shape: (41, 8)  (26.83% nonzero)
FWD solves: 8   REV solves: 0
Total colors vs. total size: 8 vs 8  (0.00% improvement)
```

**Small case**
```python
coloring2 = MNCO_bidir(coo_matrix(pd.read_excel('C:\Book2.xlsx', sheet_name='Small').to_numpy()))
coloring2.summary()
```
```
Jacobian shape: (21, 8)  (28.57% nonzero)
FWD solves: 2   REV solves: 1
Total colors vs. total size: 3 vs 8  (62.50% improvement)
```

All problems are the same structure
**Small**  
image

**Medium**
image

naylor-b:
Thanks Andrew, I'll take a look at your test cases.

naylor-b:
@andrewellis55 I found an issue with the bidirectional coloring algorithm and now I'm getting 3 colors for all of your test cases.  It is important to keep in mind though that the algorithm doesn't guarantee that the minimal coloring will be found.  For example, when playing around with your test cases, I tried a matrix like yours except that I removed the last column.  The best coloring should still have been 3 in that case, but the algorithm got 5 colors.

andrewellis55:
@naylor-b thanks for this! I just tested it out on a number of more complex cases with this (general) structure and it seems to work well.

image
image

andrewellis55:
Just another note on this. After now testing coloring on a few large problems, there isn't actually much different in the gradient evaluation times (in fact there's an increase if anything). Here's an example where an increase was seen of a taking 5 iteration of a test problem where I would have expected significant improvements:

**Meta**  
Driver: ScipyOptimize + SLSQP  
Problem Structure: Upper triangular, no solvers\implicit components.

**Problem Summary**  
```
============== Problem Summary ============
Groups:             460
Components:        1954
Max tree depth:       7

Design variables:            0   Total size:        0

Nonlinear Constraints:       0   Total size:        0
    equality:                0                      0
    inequality:              0                      0

Linear Constraints:          0   Total size:        0
    equality:                0                      0
    inequality:              0                      0

Objectives:                  0   Total size:        0

Input variables:          4269   Total size:   178663
Output variables:         2525   Total size:   114553

Total connections: 4269   Total transfer data size: 178663

Driver type: Driver
Linear Solvers: [LinearRunOnce x 460]
Nonlinear Solvers: [NonlinearRunOnce x 460]
```

**Timing without Coloring**
Looking at the timing, we can see that the `_solve_linear` is by far the dominant aspect of the problem, which to me indicates that it's a good candidate for an improvement via coloring
image


**Coloring**   
image

**Run with 5 iteration (No Coloring) OpenMDAO 3.33**  
```
Iteration limit reached    (Exit mode 9)
            Current function value: 30705.180863195863
            Iterations: 5
            Function evaluations: 5
            Gradient evaluations: 5
Optimization FAILED.
Iteration limit reached
-----------------------------------
  Driver Overhead Time:  14.62  s
  Model Exec. Time    :  5.51  s
  Grad. Eval. Time    :  84.01  s
-----------------------------------
```
**Run with 5 iteration (With Coloring) coloring_ok branch**
```
Sparsity computed using tolerance: 1e-25
Time to compute sparsity:  64.6408 sec
Time to compute coloring:  14.5963 sec
Memory to compute coloring:  47.6094 MB
Coloring created on: 2024-10-12 17:02:50
-----------------------------------
Iteration limit reached    (Exit mode 9)
            Current function value: 30705.179990420656
            Iterations: 5
            Function evaluations: 5
            Gradient evaluations: 5
Optimization FAILED.
Iteration limit reached
-----------------------------------
  Driver Overhead Time:  112.23  s
  Model Exec. Time    :  7.33  s
  Grad. Eval. Time    :  104.23  s
-----------------------------------
```

**Notes**
driver overhead is computed as `driveroverhead = driver_res.runtime - driver_res.model_time - driver_res.deriv_time`. I'm assuming this jump is from the coloring computation.

naylor-b:
I did find something in the coloring algorithm itself that was quite slow.  I've pushed up that change if you'd like to try your benchmark again.

robfalck (approved):

andrewellis55:
@naylor-b, I've re-run this test again with the latest version of the branch. The same results were seen that the optimization was slower with the coloring, despite the 75% reduction in number of colors compared to each openmdao 3.33 and your coloring_ok branch with coloring off. This could be some form of artifact of the problem structure. I can take a look at some other problems we have in our test set, but the latest change does not seem to have had a notable effect on runtime of the optimization and specifically the gradient compute time.

naylor-b:
@andrewellis55 can you post the sparsity data for the 7601 x 232 jacobian you were testing with?  I'd like to create a test case using that and do some profiling to see if I can zero in on the slowdown.

naylor-b:
@andrewellis55 I tried various test cases with large Jacobians and didn't see slowdowns after the coloring was computed.  For very large Jacobians, it's expected that the sparsity computation (which computes typically 3 full dense jacobians) will take a bit of time, and the coloring computation also, although not as much as the sparsity computation.  After that is done though, the later calls to compute the gradient should be sped up based on the number of colors vs. the number of rows or cols in the full jacobian.

robfalck (changes requested):
We should add the new environment variable to the OpenMDAO environment variables documentation:  https://github.com/OpenMDAO/OpenMDAO/blob/878c3a043cec751d3fe250450e063dfcc0966bd7/openmdao/docs/openmdao_book/other_useful_docs/environment_vars.ipynb#L4

andrewellis55:
@naylor-b I just re-tested this. I think the time to compute the coloring was slipping into my runtimes. I re-ran the tests where for the run with the coloring, I loaded from a static file rather than computing it. The new results now show that use coloring doesn't add any, but it also doesn't really provide any speed up. I'm still a little surprised by this, but perhaps it could just be a factor of the structure of the problem. 

There coloring file is here
[total_coloring.zip](https://github.com/user-attachments/files/17787447/total_coloring.zip)

The results I'm observing are:

**Coloring on**
```
loading total coloring from file C:\...\coloring_files\total_coloring.pkl
Iteration limit reached    (Exit mode 9)
            Current function value: 51942.72369333709
            Iterations: 30
            Function evaluations: 30
            Gradient evaluations: 30
Optimization FAILED.
Iteration limit reached
-----------------------------------
  Setup Time  :  8.6  s
  Final Setup Time    :  17.38  s
  Model Exec. Time    :  18.56  s
  Grad. Eval. Time    :  335.26  s
  Driver Overhead Time:  215.1  s (driver_res.runtime - driver_res.model_time - driver_res.deriv_time)
-----------------------------------
  Driver Eval. Pace   :  6.94  s/itr
  Model Eval. Pace     :  0.6  s/model eval.
  Grad. Eval. Pace     :  11.18  s/grad. eval.
-----------------------------------
Num. Design Variables  :  232
Num. Constraints       :  12404
  Non-Linear InEq. Cons:  12404
  Non-Linear Eq. Cons  :  0
  Linear InEq. Cons    :  0
  Linear Eq. Cons      :  0
Total Jac. Size        :  2877728
-----------------------------------
Case Recording Time:  0.26
```

**Coloring off**
```
Iteration limit reached    (Exit mode 9)
            Current function value: 51942.73637931126
            Iterations: 30
            Function evaluations: 30
            Gradient evaluations: 30
Optimization FAILED.
Iteration limit reached
-----------------------------------
  Setup Time  :  8.76  s
  Final Setup    :  17.73  s
  Model Exec. Time    :  20.45  s
  Grad. Eval. Time    :  336.2  s
  Driver Overhead Time:  225.15  s  (driver_res.runtime - driver_res.model_time - driver_res.deriv_time)
-----------------------------------
  Driver Eval. Pace   :  7.26  s/itr
  Model Eval. Pace     :  0.66  s/model eval.
  Grad. Eval. Pace     :  11.21  s/grad. eval.
-----------------------------------
Num. Design Variables  :  232
Num. Constraints       :  12404
  Non-Linear InEq. Cons:  12404
  Non-Linear Eq. Cons  :  0
  Linear InEq. Cons    :  0
  Linear Eq. Cons      :  0
Total Jac. Size        :  2877728
-----------------------------------
```

Kenneth-T-Moore (approved):

robfalck (approved):

PR #3387, opened 2024-10-21
Added additional documentation for the `set_input_defaults` function

### LINKED ISSUE #2711: Users are confused by `set_input_defaults`
### Task Description

New users often think that `set_input_defaults` will set all values for systems, including inputs and outputs. This is not the current or expected behavior and `set_val` should generally be used in that case.

I propose that we update the docs to better explain and reflect this difference, specifically:

- Use `set_input_defaults` at the group level to resolve inconsistencies between auto-IVC values
- Use `set_val` at the run script level to propagate values or units throughout the system
- Have examples in a doc landing page that shows when to use which method and explain how the info is passed into the system
- Detail _when_ the values that are given to `set_input_defaults` are used. For example, if I call `set_input_defaults`, then right after do a `prob.model.list_inputs`, will it print those default values? Or do I need to call `setup` in between? What's the stack like?

Currently if you [search for `set_input_defaults`](http://openmdao.org/twodocs/versions/latest/search.html?q=set_input_defaults) you only see it being used, not really being explained. Hopefully this doc enhancement will help users understand it better.

swryan:
### Summary

Added additional documentation for the `set_input_defaults` function

Also:
- Enabled the use of `set_input_defaults` for discrete variables

### Related Issues

- Resolves #2711

None

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/core_features/running_your_models/set_input_defaults.ipynb
+++ b/openmdao/docs/openmdao_book/features/core_features/running_your_models/set_input_defaults.ipynb
@@ -105,9 +105,8 @@
     "# Note that the we have specified the default metadata for the AutoIVC output that will supply values\n",
     "# for the promoted 'x' (_auto_ivc.v0) but it has not been applied yet. It will be applied during setup.\n",
     "\n",
-    "# Calling list_inputs at this point will show the default values from the component definitions\n",
-    "model.list_inputs(units=True)\n",
-    "model.list_outputs(units=True, list_autoivcs=True);"
+    "# Calling list_vars (or list_inputs/list_outputs) at this point will show the default values from the component definitions\n",
+    "model.list_vars(units=True, list_autoivcs=True);"
    ]
   },
   {
@@ -119,8 +118,7 @@
     "p1.setup()\n",
     "\n",
     "# now the default value and units for the AutoIVC have been applied\n",
-    "model.list_inputs(units=True)\n",
-    "model.list_outputs(units=True, list_autoivcs=True);"
+    "model.list_vars(units=True, list_autoivcs=True);"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


robfalck (approved):

openmdao/docs/openmdao_book/features/core_features/running_your_models/set_input_defaults.ipynb
```diff
@@ -0,0 +1,475 @@
...
+    "\n",
+    "-  *Scope*: \n",
+    "   `set_input_defaults` is used at the group level to define default metadata for promoted inputs, while `set_val` is used at the problem level to set specific values for variables.\n",
+    "\n",
+    "- *Purpose*: \n",
+    "  `set_input_defaults` resolves ambiguities when multiple inputs are promoted to the same name, while `set_val` is used to assign values to variables.\n",
+    "\n",
+    "- *Timing*: \n",
+    "  `set_input_defaults` is typically called during the model setup phase, while `set_val` can be called before or during a run of the model."
+   ]
```
Kenneth-T-Moore, line 449: "- `set_val` is used at the problem level to set the actual value of a variable, including inputs, outputs, and independent variables."

We might want to also mention that initial values for implicit states should be set with `set_val`. Not sure what independent variables are here, and how they differ from inputs, so maybe we just say "inputs and implicit states".

openmdao/docs/openmdao_book/features/core_features/running_your_models/set_input_defaults.ipynb
```diff
@@ -0,0 +1,475 @@
...
+   "outputs": [],
+   "source": [
+    "model.set_input_defaults('x', val=1., units='m')\n",
+    "\n",
+    "# Note that the we have specified the default metadata for the AutoIVC output that will supply values\n",
+    "# for the promoted 'x' (_auto_ivc.v0) but it has not been applied yet. It will be applied during setup.\n",
+    "\n",
+    "# Calling list_inputs at this point will show the default values from the component definitions\n",
+    "model.list_inputs(units=True)\n",
+    "model.list_outputs(units=True, list_autoivcs=True);"
```
Kenneth-T-Moore, line 110: I wonder if using `list_vars` instead of `list_inputs` and `list_outputs` is more convenient for the user.

PR #3410, opened 2024-12-05
Fix bug that occurred when recording inputs by absolute name.

### LINKED ISSUE #3385: Recording an input raises an exception when running in parallel.
### Description

When adding a variable to `recording_options['includes']`, the variable must be given as an absolute path if it is an input.
However, the code in `System._retrieve_data_of_kind` will raise an exception when this is done.

The block that currently encounters this exception is showing up as uncovered on coveralls.

The offending block is

```python
    if variables:
            ...
            if comm.size == 1:
                ...
            elif parallel:
                get = self._abs_get_val
                vdict = {}
                if discrete_vec:
                    ...
                else:       
                    for name in variables:
                        if vec._contains_abs(name):
                            vdict[name] = get(name, get_remote=True, rank=0,
                                              vec_name=vec_name, kind=kind)
                        else:
                            ivc_path = conns[prom2abs_in[name][0]]  # name IS NOT IN prom2abs_in
                            vdict[name] = get(ivc_path, get_remote=True, rank=0,
                                              vec_name=vec_name, kind='output')
```

In this case `name` is the absolute name of the variable, but we're assuming its a key in `prom2abs_in`, hence the exception.
This routine doesn't seem to make any distinction for whether `kind == 'input'`, which seems like it would be necessary here.

### Example

Running the following code as a test with testflo, NPROC=4, will deomonstrate the issue.

```python3
        import openmdao.api as om
        from openmdao.test_suite.groups.parallel_groups import FanInGrouped
        
        # run cases in parallel with 2 procs per model
        # (cases will be split between the 2 parallel model instances)
        run_parallel = True
        procs_per_model = 2

        prob = om.Problem(FanInGrouped())
        model = prob.model

        model.add_design_var('x1', lower=0.0, upper=1.0)

        model.add_objective('c3.y')

        samples = [
            [('x1', 0.)],
            [('x1', .1)],
            [('x1', .2)],
            [('x1', 0.3)],
            [('x1', 0.4)],
            [('x1', 0.5)],
            [('x1', 0.6)],
            [('x1', 0.7)],
            [('x1', 0.8)],
            [('x1', 0.9)],
        ]

        prob.driver = om.DOEDriver(samples)
        prob.driver.add_recorder(om.SqliteRecorder("cases.sql"))
        # prob.driver.recording_options['includes'].append('x2')  # Doesn't work, must be absolute because its an input. Fails silently
        prob.driver.recording_options['includes'].append('sub.c2.x')    # Doesn't work, but in system.py.

        prob.driver.options['run_parallel'] = run_parallel
        prob.driver.options['procs_per_model'] = procs_per_model

        prob.setup()
        prob.final_setup()
        # Manually record the iteration to trigger the exception.
        prob.driver.record_iteration()

        prob.run_driver()
        prob.cleanup()
        if prob.comm.rank < 2:
            rec_file = prob.get_outputs_dir() / f'cases.sql_{prob.comm.rank}'
            cr = om.CaseReader(rec_file)
            last_case = cr.get_case(-1)
            last_case.list_inputs()
            last_case.list_outputs()
```

### OpenMDAO Version

3.35.1-dev

_No response_

naylor-b:
### Summary

- Fixed adding absolute input name to recorder includes
- Fixed an mpi hang due to a collective MPI call being made only on ranks that were recording data

### Related Issues

- Resolves #3385

- The 'inputs' dict of a Case object now contains inputs by absolute name only.  Promoted input names now refer to the connected output and should be accessed via the 'outputs' dict of the Case object.

### New Dependencies

None


robfalck (approved):

openmdao/docs/openmdao_book/features/debugging/profiling/index.ipynb
```diff
@@ -62,7 +62,32 @@
     "    # do some other stuff that I don't want to profile/trace...\n",
     "```\n",
     "\n",
-    "* [Timing Systems under MPI](timing.ipynb)\n"
+    "* [Timing Systems under MPI](timing.ipynb)\n",
+    "\n",
+    "## Dumping debug info to file based on MPI rank and/or PID\n",
+    "\n",
+    "The `OPENMDAO_DUMP` environment variable can be set to a number of values to control what information\n",
+    "is dumped and which file it is written to. `OPENMDAO_DUMP` can have values like `stdout`, `stderr`, \n",
```
swryan, line 70: this should probably say the value ca be a comma separated list of keywords (as is implied by the examples )

swryan (approved):

openmdao/recorders/sqlite_reader.py
```diff
@@ -324,26 +320,6 @@ def _collect_metadata(self, cur):
                 if 'upper' in meta and meta['upper'] is not None:
                     meta['upper'] = np.resize(np.array(meta['upper']), meta['shape'])
 
-            # Map source names to input display text.
```
Kenneth-T-Moore, line 327: Are there any backwards compatibility concerns for older sql files?

Kenneth-T-Moore (approved):

openmdao/docs/openmdao_book/features/debugging/profiling/index.ipynb
naylor-b: fixed

PR #3424, opened 2025-01-06
Modified assert_check_partials to use numpy.assert_allclose

### LINKED ISSUE #3413: Change request: Use more robust error checking in derivative testing functions
## Problem statement

OpenMDAO's `assert_check_partials` and `assert_check_totals` are very useful for testing. However, using standard absolute and relative error tolerances, it is very hard to specify a single absolute tolerance that is valid for test values that span multiple orders of magnitude, and relative errors can be very high even for accurate derivatives when the test values are very small. The result of this is that developers often need to make their test tolerances several orders of magnitude looser in order to avoid false positive test failures. This then increases the chance of missing real test failures.

A related issue is that, when computing relative errors, OpenMDAO will divide by the test value if the reference value is zero, this will always lead to a test failure because the relative error is then guaranteed to be 100%. For example, if your reference (FD/complex-step) value is zero, and your test (fwd/rev) value is 1e-200, OpenMDAO will compute the relative error as `rel_error = (test - ref)/test = 1e-200/1e-200 = 1`. See https://github.com/mdolab/pygeo/pull/258 for an example of this kind of failure.

A good way to get around these issues is to use numpy's [assert_allclose method](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html), which combines the relative and absolute tolerance into a single check:

`fail = abs( test - ref ) > atol + rtol * abs(ref)`

Where ref is the finite-difference or complex-step value and test is the analytic value

When `abs(ref) >> atol/rtol` this check acts like a relative tolerance check, but for `abs(ref) << atol/rtol` it acts like an absolute tolerance check, thus avoiding the issues associated with relative tolerances of very small values.

In TACS, we switched to using this kind of check and were able to tighten our unit test tolerances by 1000x (see https://github.com/smdogroup/tacs/pull/216).

## Requested changes

Switch to using assert_allclose in `assert_check_totals`, `assert_check_partials` and similar testing methods

### Associated POEM

_No response_

naylor-b:
### Summary

Also made `check_partials` callable directly on a Component.

### Related Issues

- Resolves #3413

Could see slightly different output from check_partials/assert_check_partials

### New Dependencies

None

### DIFF
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -2417,9 +2417,7 @@ class Component(System):
 
         # add pathname to the partials dict to make it compatible with the return value
         # from Problem.check_partials and passable to assert_check_partials.
-        data = {self.pathname: partials_data}, worst
-
-        return data
+        return {self.pathname: partials_data}, worst
 
 
 class _DictValues(object):
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -2288,7 +2288,7 @@ class Component(System):
             self._inputs.set_val(input_cache)
             self._outputs.set_val(output_cache)
 
-            self.run_apply_nonlinear()
+            # self.run_apply_nonlinear()
 
         all_fd_options = {}
 
@@ -2306,6 +2306,7 @@ class Component(System):
         alloc_complex = self._outputs._alloc_complex
 
         for step in steps:
+            self.run_apply_nonlinear()
             approximations = {'fd': FiniteDifference(), 'cs': ComplexStep()}
 
             added_wrts = set()
--- a/openmdao/core/component.py
+++ b/openmdao/core/component.py
@@ -12,7 +12,7 @@ from numpy import ndarray, isscalar, ndim, atleast_1d, atleast_2d, promote_types
 from scipy.sparse import issparse, coo_matrix
 
 from openmdao.core.system import System, _supported_methods, _DEFAULT_COLORING_META, \
-    global_meta_names, collect_errors
+    global_meta_names, collect_errors, _iter_derivs
 from openmdao.core.constants import INT_DTYPE, _DEFAULT_OUT_STREAM
 from openmdao.jacobians.dictionary_jacobian import DictionaryJacobian, _CheckingJacobian
 from openmdao.utils.array_utils import shape_to_len, submat_sparsity_iter
@@ -2007,7 +2007,7 @@ class Component(System):
                         issue_warning(msg, prefix=self.msginfo,
                                       category=OMInvalidCheckDerivativesOptionsWarning)
 
-    def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes=None,
+    def check_partials(self, out_stream=_DEFAULT_OUT_STREAM,
                        compact_print=False, abs_err_tol=1e-6, rel_err_tol=1e-6,
                        method='fd', step=None, form='forward', step_calc='abs',
                        minimum_step=1e-12, force_dense=True, show_only_incorrect=False):
@@ -2019,12 +2019,6 @@ class Component(System):
         out_stream : file-like object
             Where to send human readable output. By default it goes to stdout.
             Set to None to suppress.
-        includes : None or list_like
-            List of glob patterns for pathnames to include in the check. Default is None, which
-            includes all components in the model.
-        excludes : None or list_like
-            List of glob patterns for pathnames to exclude from the check. Default is None, which
-            excludes nothing.
         compact_print : bool
             Set to True to just print the essentials, one line per input-output pair.
         abs_err_tol : float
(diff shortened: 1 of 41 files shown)


openmdao/core/component.py
```diff
@@ -1934,14 +1937,488 @@ def col_size_iter():
...
+                        issue_warning(msg, prefix=self.msginfo,
+                                      category=OMInvalidCheckDerivativesOptionsWarning)
+
+    def check_partials(self, out_stream=_DEFAULT_OUT_STREAM,
+                       compact_print=False, abs_err_tol=1e-6, rel_err_tol=1e-6,
+                       method='fd', step=None, form='forward', step_calc='abs',
+                       minimum_step=1e-12, force_dense=True, show_only_incorrect=False,
+                       show_worst=True):
+        """
+        Check partial derivatives comprehensively for all components in your model.
```
Kenneth-T-Moore, line 2016: copy/paste error -- should just check this component.

openmdao/core/problem.py
```diff
@@ -1221,25 +1210,38 @@ def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes
...
-        if self._metadata['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
-            self.final_setup()
-
         model = self.model
 
+        # on master, we were doing this:
+        # if self._metadata['setup_status'] < _SetupStatus.POST_FINAL_SETUP:
+        #     self.final_setup()
+
+        # model.run_apply_nonlinear()
```
Kenneth-T-Moore, line 1219: It is interesting that we were only doing a run_apply, and I wonder if this caused some funny behavior. As a matter of habit, I always call run_model before check_partials, so I don't encounter problems, but that could be a learned behavior caused by this.

openmdao/core/system.py
```diff
@@ -6796,3 +6797,717 @@ def _allprocs_exec_order(self):
             if path not in seen:
                 seen.add(path)
                 yield path
+
+    def _deriv_display(self, err_iter, derivatives, rel_error_tol, abs_error_tol, out_stream,
```
Kenneth-T-Moore, line 6801: Kind of wonder why these live in system when they are only called in component.

openmdao/core/tests/test_check_partials.py
```diff
@@ -1445,14 +1416,14 @@ def test_directional_derivative_option(self):
...
-        # 5~8  - FD wrt bb, non-directional
-        # 9    - FD wrt x1, directional
-        # 10   - FD wrt x2, directional
-        self.assertEqual(mycomp.exec_count, 10)
+        # 2    - Called apply_nonlinear at the start of analytic deriv calculations
+        # 3    - Called apply_nonlinear to clean up before starting FD
+        # 4-7  - FD wrt bb, non-directional
+        # 8    - FD wrt x1, directional
+        # 9    - FD wrt x2, directional
+        self.assertEqual(mycomp.exec_count, 9)
```
Kenneth-T-Moore, line 1426: I guess this means check partials is 10% faster now.

Kenneth-T-Moore (approved):
Pre-emptive approve, but just one comment to fix.

openmdao/core/component.py
naylor-b: fixed

openmdao/core/problem.py
```diff
@@ -1256,426 +1258,50 @@ def check_partials(self, out_stream=_DEFAULT_OUT_STREAM, includes=None, excludes
...
-            msg += "\nTo enable complex step, specify 'force_alloc_complex=True' when calling " + \
-                   "setup on the problem, e.g. 'problem.setup(force_alloc_complex=True)'"
-            issue_warning(msg, category=DerivativesWarning)
-
-        _assemble_derivative_data(partials_data, rel_err_tol, abs_err_tol, out_stream,
-                                  compact_print, comps, all_fd_options, self.comm,
-                                  indep_key=indep_key, print_reverse=print_reverse,
-                                  show_only_incorrect=show_only_incorrect)
-
-        if not do_steps:
```
robfalck, line 1670: That's some great simplification there.

openmdao/utils/testing_utils.py
```diff
@@ -299,6 +299,86 @@ def force_check_partials(prob, *args, **kwargs):
...
+             'directional_fd_rev']
+
+    for name in names:
+        if name in data:
+            data[name] = data[name][0]
+    if 'steps' in data:
+        del data['steps']
+
+
+def compare_prob_vs_comp_check_partials(probdata, compdata, comp):
```
robfalck, line 321: I'm not finding any usages of this function.  Is this only necessary for debugging of this PR?

robfalck (approved):

openmdao/utils/testing_utils.py
naylor-b: it's used in some tests

PR #3433, opened 2025-01-21
Schur complement Solver

ArshSaja:
### Summary

This PR contains the novel Schur complement solvers for carrying out saddle-point optimizations. We can consider a problem which has 2 components. The 1st component has the main model such as a CFD model. The 2nd component has some balance residuals which we want to solve at the model level at each design iteration. For instance, we generally enforce the target lift value in aerodynamic shape optimization as constraints. But if we wanted to solve the target lift value at each design iteration `(r_2(alpha)=CL-CL*)` by changing the angle of attack (alpha) as a coupling variable, this would create a saddle-point Jacobian where the `dr_2/du_2= dr_2/dalpha=0`. This coupling problem cannot be solved with BGS because BGS require each diagonal in the coupled Jacobian to be invertible.  So, we use Schur complement method to develop new solvers which can handle these type of problems. This is very helpful if we want to obtain feasible designs at each design iteration

These solvers works in a specific way. We define the problem with 2 components. The 1st component can be solved by the 1st component's specialized nonlinear and linear solvers and the 2nd component is solved by Schur complement solvers (See the figure below). 

Screenshot

Currently, we input the component names to the solver so that they know which one is 1st component and which one is 2nd component. It can be automated I think. These solvers may need more cleaning and work in order to be consistent with OM.

I am tagging @robfalck and @anilyil here who are familiar with the work.

### Related Issues

None

None

### New Dependencies

None


JustinSGray:
This is a HUGE advancement in the state of the art. Awesome

setup.py
swryan, line 1: we don't use setuptools anymore..   see `pyproject.toml`

swryan (changes requested):

robfalck (changes requested):
This is great!  Aside from the setup.py/pyproject.toml issue, we're going to need some simple tests and documentation. If you have the basics of a simple script that uses it and could send that our way, we can handle the documentation/testing aspect, if you like.

setup.py
ArshSaja: I removed the setup file and added schur solver to pyproject.toml. Please check

ArshSaja:
> This is great! Aside from the setup.py/pyproject.toml issue, we're going to need some simple tests and documentation. If you have the basics of a simple script that uses it and could send that our way, we can handle the documentation/testing aspect, if you like.

@robfalck I have attached a simple script to run the solver. It may be used for the tests. Please check
[simple_problem.zip](https://github.com/user-attachments/files/18526165/simple_problem.zip)

robfalck:
Thanks for the example of usage. I'm in the process of constructing tests and documentation from that, your paper on the solver, and the other references.

As I make changes, I'm going to close this PR and issue one from my branch, if that's ok with you.

Changes I'm in the process of making:
1. Nonlinear and Linear Solvers can be applied to components (Systems in general, not just Groups). So `groupNames` becomes `sys_names`.
2. If the system only has two subsystems, we should be able to make specification of `sys_names` optional (None by default) in which case subsystem1 and subsystem2 will just be the two subsystems, in order.
3. In LinearSchur's solve there are some unused variables that i suspect are copy-paste leftovers from the Nonlinear version.

```
        inpu_cache = system._dinputs.asarray(copy=True)
        outp_cache = system._doutputs.asarray(copy=True)
        resd_cache = system._dresiduals.asarray(copy=True)

        r_data = rvec.asarray(copy=True)
        o_data = ovec.asarray(copy=True)
        i_data = ivec.asarray(copy=True)
```

I don't think there's any TODO's related to these so I plan on removing them.

ArshSaja:
> Thanks for the example of usage. I'm in the process of constructing tests and documentation from that, your paper on the solver, and the other references.
> 
> As I make changes, I'm going to close this PR and issue one from my branch, if that's ok with you.
> 
> Changes I'm in the process of making:
> 
> 1. Nonlinear and Linear Solvers can be applied to components (Systems in general, not just Groups). So `groupNames` becomes `sys_names`.
> 2. If the system only has two subsystems, we should be able to make specification of `sys_names` optional (None by default) in which case subsystem1 and subsystem2 will just be the two subsystems, in order.
> 3. In LinearSchur's solve there are some unused variables that i suspect are copy-paste leftovers from the Nonlinear version.
> 
> ```
>         inpu_cache = system._dinputs.asarray(copy=True)
>         outp_cache = system._doutputs.asarray(copy=True)
>         resd_cache = system._dresiduals.asarray(copy=True)
> 
>         r_data = rvec.asarray(copy=True)
>         o_data = ovec.asarray(copy=True)
>         i_data = ivec.asarray(copy=True)
> ```
> 
> I don't think there's any TODO's related to these so I plan on removing them. 4. NonlinearSchur's `_linearize_children` method is never used. It contains a reference to variable `system` is never defined. An exhaustive test of every combination of problem and solvers in your example never hit this error, so I believe we can safely remove `_linearize_children`.

@robfalck Thanks for making the tests and documentation! The journal paper is in the minor revision. In case you need more reference, please let me know.

Yes, please close this PR and open an issue in your branch. If you could keep the commit history, that would be great to track later in the future work.

I understand. yeah the changes looks good.

Yes, I forgot to remove that copy-paste leftovers, Please remove them.

I see, yeah. We will not need linear solvers for Nonlinear Schur. So I think we can remove `_linearize_children`.

ArshSaja:
@robfalck A quick follow up question on the last comment. When we linearize the Jacobian, we would need `_linearize_children` . But is this only relevant to the Newton solver because it solves the linear system with the Direct solver? 

        do_sub_ln = self.linear_solver._linearize_children()

        # Disable local fd
        approx_status = system._owns_approx_jac
        system._owns_approx_jac = False

        system._vectors["residual"]["linear"].set_vec(system._residuals)
        system._vectors["residual"]["linear"] *= -1.0
        my_asm_jac = self.linear_solver._assembled_jac

        system._linearize(my_asm_jac, sub_do_ln=do_sub_ln)
        if my_asm_jac is not None and system.linear_solver._assembled_jac is not my_asm_jac:
            my_asm_jac._update(system)

robfalck:
Searching the codebase, we only ever call _linearize_children on linear solvers, never on nonlinear solvers.  Should that method be moved to LinearSchur?

ArshSaja:
I see. I had that part of code in order to follow the Newton solver in OM. I think it would useful to have `_linearize_children` in the linear Schur in case if a user wants to use the linear Schur for the Newton solver.  What do you think?

robfalck:
Closing...superceded by #3443

PR #3443, opened 2025-01-29
Schur Complement linear and nonlinear solvers

robfalck:
### Summary

This is a cleaned up version of the Schur Complement solver work done by ArshSaja.  Once documentation is added it will be moved out of draft status.

This PR contains the novel Schur complement solvers for carrying out saddle-point optimizations. We can consider a problem which has 2 components. The 1st component has the main model such as a CFD model. The 2nd component has some balance residuals which we want to solve at the model level at each design iteration. For instance, we generally enforce the target lift value in aerodynamic shape optimization as constraints. But if we wanted to solve the target lift value at each design iteration `(r_2(alpha)=CL-CL*)` by changing the angle of attack (alpha) as a coupling variable, this would create a saddle-point Jacobian where the `dr_2/du_2= dr_2/dalpha=0`. This coupling problem cannot be solved with BGS because BGS require each diagonal in the coupled Jacobian to be invertible.  So, we use Schur complement method to develop new solvers which can handle these type of problems. This is very helpful if we want to obtain feasible designs at each design iteration

These solvers works in a specific way. We define the problem with 2 components. The 1st component can be solved by the 1st component's specialized nonlinear and linear solvers and the 2nd component is solved by Schur complement solvers (See the figure below). 

Screenshot

Currently, we input the component names to the solver so that they know which one is 1st component and which one is 2nd component. It can be automated I think. These solvers may need more cleaning and work in order to be consistent with OM.

I am tagging @ArshSaja and @anilyil here who are familiar with the work.

### Related Issues

None

None

### New Dependencies

None


openmdao/solvers/linear/linear_schur.py
```diff
@@ -0,0 +1,381 @@
...
+        resd_to_solve = [*subsys2_residuals.keys()]
+
+        # total size of the jacobian
+        n_vars = 0
+        for var in vars_to_solve:
+            n_vars += subsys2_outputs[var].size
+
+        # initialize the schur complement jacobian for these variables
+        # TODO better way to get the dtype?
+        schur_jac = np.zeros((n_vars, n_vars),
```
naylor-b, line 126: Since you're only using the system._vectors['residual']['linear'] vector array to determine the dtype, you shouldn't set `copy=True`

openmdao/solvers/linear/linear_schur.py
```diff
@@ -0,0 +1,381 @@
...
+            subsys1_rhs = self._rhs_vec[off: off + len(b_vec)].copy()
+            subsys2_rhs = self._rhs_vec[off2: off2 + len(b_vec2)].copy()
+
+            # Beg solve for subsys 2
+            # schur_jacobian
+
+            # Schur_Jac = D - C A^-1 B ##
+
+            # put back the vectors
+            rvec.set_val(np.zeros(len(rvec)))
```
naylor-b, line 155: You don't need to create temporary np.zeros arrays here.  You can just do, e.g., rvec.set_val(0.0) and that will zero out the whole vector.

openmdao/solvers/linear/linear_schur.py
```diff
@@ -0,0 +1,381 @@
...
+                # run the apply linear. we do it on the complete system here
+                # the result is the final jacobian for this using the schur
+                # complement method D[:,{ii}] - C A^-1 B[:,{ii}]
+                scope_out, scope_in = system._get_matvec_scope()
+                scope_out = self._vars_union(self._scope_out, scope_out)
+                scope_in = self._vars_union(self._scope_in, scope_in)
+                system._apply_linear(None, mode, scope_out, scope_in)
+
+                # put this value into the jacobian.
+                schur_jac[:, ii] = subsys2._vectors['residual']['linear'].asarray(copy=True)
```
naylor-b, line 201: Don't need copy=True here

openmdao/solvers/linear/linear_schur.py
```diff
@@ -0,0 +1,381 @@
...
+            # Schur_Jac = D - C A^-1 B ##
+
+            # put back the vectors
+            rvec.set_val(np.zeros(len(rvec)))
+            ovec.set_val(np.zeros(len(ovec)))
+            ivec.set_val(np.zeros(len(ivec)))
+
+            for ii, var in enumerate(vars_to_solve):
+                # set the linear seed of the variable we want to solve for in subsys 2
+                ovec[f'{subsys2.name}.{var}'] = 1.0
```
naylor-b, line 161: Just making sure that the intent here is to set all entries corresponding to the variable to 1.0 rather than just one entry at a time like we normally would when computing a jacobian.

openmdao/solvers/linear/linear_schur.py
```diff
@@ -0,0 +1,381 @@
...
+            scope_out = self._vars_union(self._scope_out, scope_out)
+            scope_in = self._vars_union(self._scope_in, scope_in)
+            subsys2._apply_linear(None, mode, scope_out, scope_in)
+
+            schur_rhs = subsys2._vectors['residual']['linear'].asarray(copy=True)
+
+            # Beg solve for subsys 2
+
+            schur_rhs = subsys2_rhs - schur_rhs
+            iD_schur = np.eye(n_vars,
```
naylor-b, line 230: No copy needed for dtype.  May also want to allocate this outside the loop.

openmdao/solvers/linear/linear_schur.py
robfalck: I suspect this is an error, but not sure how to correct it yet.  Working on an example for testing.

robfalck:
Rather than implement this with a separate nonlinear solver, we're going to address this by allowing groups to optionally be treated as atomic explicit calculations. This should largely reproduce the desired capability.

PR #3450, opened 2025-02-14
_CheckingJacobian no longer raises if the sparsity pattern is bad

### LINKED ISSUE #3449: The _CheckingJacobian's test for invalid sparsity patterns interrupts `check_partials` by raising.
### Description

When running check_partials, the use of `_CheckingJacobian` will raise if there are non-covered nonzeros in the user's sparsity pattern.
This should be changed to a warning so that check partials can complete.

### Example

N/A

### OpenMDAO Version

3.36.1-DEV

_No response_

robfalck:
### Summary

_CheckingJacobian `set_col` was causing an exception to be raised during `check_partials` which was preventing the completion of the partials check. This information is now displayed in the output of check partials.

This condition is triggered by nonzeros in the FD or CS jacobian where there should only be zeros.
In some cases, calculations can result in non-zero values that are in the noise. This condition is now only triggered if their magnitudes exceed 1.0E-16.

### Related Issues

- Resolves #3449 

None

### New Dependencies

None


openmdao/jacobians/dictionary_jacobian.py
```diff
@@ -357,7 +358,7 @@ def set_col(self, system, icol, column):
                     arr[row_inds] = 0.  # zero out the rows that are covered by sparsity
                     nzs = np.nonzero(arr)[0]
                     if nzs.size > 0:
-                        self._errors.append(f"{system.msginfo}: User specified sparsity (rows/cols)"
-                                            f" for subjac '{of}' wrt '{wrt}' is incorrect. There "
-                                            f"are non-covered nonzeros in column {loc_idx} at "
-                                            f"row(s) {nzs}.")
+                        issue_warning(f"{system.msginfo}: User specified sparsity (rows/cols)"
```
naylor-b, line 361: I thought about this a little more after our initial discussion, I think the right way to handle this would be similar to the way we handle inconsistent keys, where it's really an error and we just add it to the data structure so that it can be part of the check_partials report at the end.  I understand why you don't want it to bail out too early, but having an incorrect sparsity really is a significant error, and making it a warning could easily result it it being ignored, leading to wrong answers.

naylor-b (changes requested):

openmdao/jacobians/dictionary_jacobian.py
robfalck: I agree that we should be displaying the information in the output. Maybe along the lines of `Nonzero rows/cols in Jfd not in Jfwd/Jrev`. The text output of the current warning makes it really difficult to visualize where the problem lies. I'll work on this to add it to the check partials data for display.

naylor-b (approved):

swryan (approved):

PR #3455, opened 2025-02-19
Added work_dir as an option to problem,

### LINKED ISSUE #3453: add_recorder in a read-only folder

Hi OpenMDAO team,

Thank you for your continuous hard work.

At [NREL](https://www.nrel.gov) we use OpenMDAO v3.36.0 in our conceptual wind turbine design tool [WISDEM](https://github.com/WISDEM/WISDEM)

WISDEM is being deployed on some HPC system by one of our industrial partners

The industrial partner has created a read-only folder where a WISDEM wrapper is executed. The user launches the script passing keyword arguments pointing to input files and folder directories. The read-only requirement is to make sure that the code, and all its dependencies, runs successfully and cannot be manipulated/broken/changed accidentally.

We recently ran into an issue where we could not turn on the OpenMDAO recorder without running into a write-access error

@johnjasa and I believe that the issue is that when the recorder is turned on, OpenMDAO prints out the recorder in the target output folder, but prints out `.html` reports in the working directory

We've gone through the documentation and learnt that we can fix this by setting the environment variable `OPENMDAO_WORKDIR`

However, we think that it would make more sense to expose the path to python so that the user can set it automatically during the python call without going through an extra terminal command. If the user can set where the .sql recorder file goes, shouldn't we do the same for the reports?

Thank you, and again thanks for the work
@gbarter 

### Associated POEM

_No response_

robfalck:
### Summary

This PR allows the user to control the parent directory of the problem output folder without using an environment variable.

The file utility get_work_dir has been renamed _get_work_dir, since it should no longer be part of the user facing API.  Users should use `problem.options['work_dir']` to get the path of the working directory associated with a problem.

### Related Issues

- Resolves #3453 

openmdao.file_utils.get_work_dir is now `openmdao.file_utils._get_work_dir`. If necessary, users should obtain the top level working directory on a problem-by-problem basis from `problem.options['work_dir']`

### New Dependencies

None

### DIFF
--- a/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
+++ b/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
@@ -105,7 +105,7 @@
     "\n",
     "2. The [openmdao clean](om-command-clean) or `openmdao.api.clean_outputs` function can be used to remove these directories.\n",
     "\n",
-    "In addition to setting `OPENMDAO_WORKDIR` to set the top level output directory for all problems, a user can set `work_dir` as an argument or option of Problem in order to set the top level output directory under which that problems `'{problem_name}_out'` directory will be placed.\n"
+    "As an alternative to setting `OPENMDAO_WORKDIR` to set the top level output directory for all problems, a user can set `work_dir` as an option of Problem in order to set the top level output directory under which that problem's `'{problem_name}_out'` directory will be placed. If provided, the value of the `work_dir` option takes precedence.\n"
    ]
   },
   {
--- a/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
+++ b/openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
@@ -103,7 +103,9 @@
     "\n",
     "1. The user has the option of setting a common output directory for ALL problems using the `OPENMDAO_WORKDIR` environment variable. Thus there is a single directory that contains all OpenMDAO output data.\n",
     "\n",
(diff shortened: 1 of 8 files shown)


openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
```diff
@@ -103,7 +103,9 @@
     "\n",
     "1. The user has the option of setting a common output directory for ALL problems using the `OPENMDAO_WORKDIR` environment variable. Thus there is a single directory that contains all OpenMDAO output data.\n",
     "\n",
-    "2. The [openmdao clean](om-command-clean) or `openmdao.api.clean_outputs` function can be used to remove these directories.\n"
+    "2. The [openmdao clean](om-command-clean) or `openmdao.api.clean_outputs` function can be used to remove these directories.\n",
+    "\n",
+    "In addition to setting `OPENMDAO_WORKDIR` to set the top level output directory for all problems, a user can set `work_dir` as an argument or option of Problem in order to set the top level output directory under which that problems `'{problem_name}_out'` directory will be placed.\n"
```
naylor-b, line 108: typo: "under which that problems"  -> "under which that problem's"

naylor-b (approved):

openmdao/docs/openmdao_book/features/reports/reports_system.ipynb
swryan: "In addition to" should probably say "As an alternative to"...  

There is a question of precedence I guess.. it might be worthwhile to say that the problem option supersedes the env var...

PR #3457, opened 2025-02-20
smooth_round function added to om.jax

ehariton:
### Summary

A smooth round function is useful in a number of contexts and should be included as part of the om.jax library of files for all our users to leverage. Docs updated as well with an example.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/jax/smooth.py
+++ b/openmdao/jax/smooth.py
@@ -30,7 +30,8 @@ def act_tanh(x, mu=1.0E-2, z=0., a=-1., b=1.):
     mu : float
         A shaping parameter which impacts the "abruptness" of
         the activation function. As this value approaches zero
-        the response approaches that of a step function.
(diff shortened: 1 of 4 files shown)


openmdao/jax/smooth.py
```diff
@@ -132,3 +132,30 @@ def smooth_abs(x, mu=1.0E-2):
...
+
+    Parameters
+    ----------
+    x : float or array
+        The argument to round.
+    mu : float
+        A shaping parameter which impacts the tradeoff between the
+        smoothness and accuracy of the function. As this value
+        approaches zero the response approaches that of the true
+        absolute value.
```
naylor-b, line 150: 'absolute value' should be 'round' I think.

swryan (approved):

naylor-b (approved):

PR #3458, opened 2025-02-21
Calling set_val can now set shapes of dynamically shaped variables (shape_by_conn)

naylor-b:
### Summary

It's no longer necessary to create IndepVarComps purely to set the shapes on shape_by_conn variables.  Calling set_val on a shape_by_conn variable can now set its shape.

In testing, some things that use to happen in Problem.setup now happen later, so some tests may have to be updated to produce the same behavior.  We had many tests that tested for setup errors and those errors now occur in final_setup so the tests had to be updated.

### New Dependencies

None


Kenneth-T-Moore (approved):

swryan:
I ran the `dymos` test suite against this PR and it passes cleanly, so I'll go ahead and merge @robfalck

PR #3469, opened 2025-03-04
openmdao commands now correctly use the script name to name the output directory

### LINKED ISSUE #3467: openmdao command running scripts results in incorrect Problem name and name for reports directory
### Description

A Problem created in a script and run with the python command has the correct name, which is the name of the script minus the extension. 

But when the script is run using the openmdao command, the Problem name is `openmdao`. This results in the reports directory being name `openmdao_out`, which is not the desired behavior. It should also be the script name. 

### Example

Here is a script that can be used to show the problem. Assume the name of the script is `model_script.py`

```
import openmdao.api as om
prob = om.Problem()
prob.setup()
prob.final_setup()
print(f'{prob._name=}')

```

Then run the script using python and openmdao to see the difference

```
$ python model_script.py
prob._name='model_script'
$ openmdao model_script.py
prob._name='openmdao'
```

### OpenMDAO Version

3.37.1-dev

_No response_

swryan:
### Summary

openmdao  commands now pass script name via `os.environ`

### Related Issues

- Resolves #3467

None

### New Dependencies

None


naylor-b (approved):

openmdao/utils/file_utils.py
```diff
@@ -190,6 +190,8 @@ def _load_and_exec(script_name, user_args):
     with open(script_name, 'rb') as fp:
         code = compile(fp.read(), script_name, 'exec')
 
+    os.environ['OPENMDAO_SCRIPT_NAME'] = script_name
```
robfalck, line 193: We probably want to wrap this in `with set_env_vars_context(OPENMDAO_SCRIPT_NAME=script_name)` in case there's some unforeseen side effect of the environment variable after this.

robfalck (changes requested):

robfalck (approved):

PR #3475, opened 2025-03-10
Modifying ImplicitComponent to respect run_root_only option for solve_linear

timryanb:
### Summary

Modifying ImplicitComponent to respect `run_root_only` option for `solve_linear`

### Related Issues

- Resolves #

None

### New Dependencies

None


naylor-b (approved):

Kenneth-T-Moore (approved):

Kenneth-T-Moore:
Just a quick question, but would you ever need separate behavior of the root_only functionality for linear and nonlinear solvers?

timryanb:
> Just a quick question, but would you ever need separate behavior of the root_only functionality for linear and nonlinear solvers?

I can't think of a reason why. Usually if you want a component to run only on root, you'd expect every computational call to run only on root

PR #3493, opened 2025-03-28
Added colorized printout to check_totals(compact_print=False)

robfalck:
### Summary

When checking totals of a large vectorized model, it can be difficult to find the elements that are incorrect.
This update makes it so the check_partials and check_totals functions print out in colorized, formatted text if the rich python package is installed.

### Added openmdao.utils.rich_utils module

The rich_utils module includes three functions to make things easier with rich:
1. `rich_wrap(s, tags)` will apply the given rich tags to a string if `rich` is available. If rich is unavailable, the string is returned unmodified. Tags can be a set or list of rich tags such as `'bold'` or `'bright_red'`.
2. `strip_tags(s)` will remove any rich tags from the given string. This is useful in testing or in determining the width of a rich-formatted string as it is printed.
3. `strip_formatting(s)` removes formatting that rich has applied to a string. It converts the tags to some ascii code that starts with `'\xb1'` that is interpreted by the terminal. This is useful for testing the content of a rich-formatted string.

### `check_partials` and `check_totals` now save output to the reports directory.
If rich is available, the output is a formatted html file.  If not, the text of the check is saved in a basic html file.
The results of check_totals are saved to a single file, while `check_partials` results are saved to a file based on the system name.

**It may be time to default `show_only_incorrect` to True.**

An example of the output for a vectorized dymos component is below:

In this case, the nonzeros in the sparsity pattern are entirely along the diagonal. Elements not included in the sparsity pattern are dimmed. I intentionally ran with a larger FD step size to force some errors to occur, which are highlighted in red.

Screenshot 2025-03-28 at 3 19 19 PM

### Related Issues

None

None

### New Dependencies

None


swryan (approved):

robfalck:
> Currently the user has no way to turn the coloring off aside from not installing `rich`, but that could get installed for other reasons in the environment. It would be nice if check_totals and check_partials took an arg that allowed you to turn off the coloring if you needed to. If you don't want to deal with that now I could add it in a later PR.

That shouldn't be a big deal, let's we can hold off on merging until I put that in.

PR #3494, opened 2025-03-31
Added real-time plotting of optimization progress

### LINKED ISSUE #3403: Create a real-time plotting capability for monitoring optimization runs

Create a tool that lets the user plot, in real-time, values from an optimization run. 

### Associated POEM

_No response_

hschilling:
### Summary

This PR adds the ability to create real-time and post-processing plotting of the progress of an optimization run. 

### Related Issues

- Resolves #3403 

None

### New Dependencies

None


robfalck (changes requested):
I think this looks great.  The only thing I would change, as we mentioned before, is to call it `optplot1 or `rtoptplot` or some variation like thiat.

We need to do something like this for the analysis driver where constraints/desvars/objectives aren't relevant and I don't think we should have to pick a different name for the rtplot tool later when we have multiple real-time plotting tools.

robfalck:
I'm rescinding my previous comment.  Let's keep the command name and the tool itself should be able to figure out what kind of plot to make based on whether we're using an AnalysisDriver or an OptimizationDriver.

robfalck (approved):

openmdao/docs/openmdao_book/features/model_visualization/real_time_optimization_progress_plot.ipynb
```diff
@@ -0,0 +1,145 @@
...
+   "source": [
+    "## Running the plot as part of an optimization run\n",
+    "\n",
+    "Likely the most common way users will use this feature is to run the optimization script is using the `opendmao rtplot` subcommand.\n",
+    "\n",
+    "```\n",
+    "openmdao rtplot path_to_script.py\n",
+    "```\n",
+    "\n",
+    "This will run the optimization script and also automatically launch the plotting app in a browser window at the appropriate time during the optimization run. \n",
```
swryan, line 30: It seems to me that, to be consistent, this should be an option rather than a subcommand...  much like the  `openmdao --view_reports script.py`  command...  The `openmdao` command itself will run the script, with the _option_ to get the real time plots.

https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/utils/om.py#L748

Just a thought...

openmdao/docs/openmdao_book/features/model_visualization/real_time_optimization_progress_plot.ipynb
```diff
@@ -0,0 +1,145 @@
...
+   "metadata": {},
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c1860288",
+   "metadata": {},
+   "source": [
+    "```\n",
+    "openmdao realtime_opt_plot optimize_system_out/system_driver_recording.sql\n",
```
swryan, line 87: _realtime_ is really a misnomer for this version of the command.. shouldn't it just be `opt_plot`?  ;)

openmdao/docs/openmdao_book/features/model_visualization/real_time_optimization_progress_plot.ipynb
hschilling: I like that idea but it causes a problem in that there are options to the rtplot subcommand so hard to handle those when the subcommand because an option itself.

Kenneth-T-Moore (approved):
This works really well, but I needed to update Bokeh and panel to the most recent version.

PR #3502, opened 2025-04-03
Fixed an issue with check_partials reports not being generated correctly without rich.

robfalck:
### Summary

When rich isn't available in the environment, the deriv_display_compact method tries to write to
 an html file but it is not opened with write access.

Added a test to verify that the reports are generated correctly with and without rich and with and without compact print.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_check_partials.py
+++ b/openmdao/core/tests/test_check_partials.py
@@ -2805,10 +2805,10 @@ class TestCheckPartialsMultipleSteps(unittest.TestCase):
         return tables
 
     def test_report_generation(self):
-        for use_rich in ('1', '0'):
+        for disable_rich in ('1', '0'):
             for compact in (True, False):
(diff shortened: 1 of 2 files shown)


openmdao/core/tests/test_check_partials.py
```diff
@@ -2803,6 +2804,19 @@ def get_tables(self, content):
                 tlines = []
         return tables
 
+    def test_report_generation(self):
+        for use_rich in ('1', '0'):
+            for compact in (True, False):
+                with self.subTest(f'{use_rich=} {compact=}'):
+                    with set_env_vars_context(OPENMDAO_USE_RICH=use_rich,
+                                            TESTFLOW_RUNNING='0',
```
swryan, line 2812: typo  TESTFLOW -> TESTFLO
(also indentation)

naylor-b (approved):

swryan (approved):

PR #3512, opened 2025-04-14
Some internal refactoring of Vector classes and name resolution functionality in Systems

naylor-b:
### Summary

The goal is to localize the name resolution functionality (handling of absolute and promoted names) in one place and to decrease Vector's dependence on System in order to make the code cleaner and more maintainable.

- Removed access to variables from Problem using 'hybrid' names.  Hybrid names are a combination of a system's pathname with the promoted name of its variables within the scope of that system.  For example, suppose a variable 'x' is promoted up from a component whose pathname is 'G1.G2.comp', and suppose that 'x' is promoted at every level up to the top, so its top level promoted name is 'x'.  Using a 'hybrid' name, it could be accessed by prom['G1.G2.x'] or prom['G1.x'].  These names will no longer work.  Only the top level promoted name or the absolute name can be used now, so prom['x'] would work, and prom['G1.G2.comp.x'] would work.  In order to use promoted names below the top level, you'll now need to access the system at the desired scope and use get_val with the promoted name from that scope.  For example, you could access prob.model.G1.G2.get_val('x').

### New Dependencies

None


openmdao/core/group.py
```diff
@@ -1843,27 +1845,20 @@ def _setup_var_data(self):
...
-                                    "multiple outputs: {}.".format(self.msginfo, prom_name,
-                                                                   sorted(abs_list)))
-
-        for io in ('input', 'output'):
-            a2p = self._var_allprocs_abs2prom[io]
-            for prom, abslist in self._var_allprocs_prom2abs_list[io].items():
-                for abs_name in abslist:
-                    a2p[abs_name] = prom
+        # for prom_name, abs_list in allprocs_prom2abs_list['output'].items():
+        #     if len(abs_list) > 1:
```
Kenneth-T-Moore, line 1849: This check is commented out, but what does it mean? multiple outputs with the same path means two outputs promoted to the same name, but this message doesn't look like the one we get for that.

openmdao/core/tests/test_src_indices.py
```diff
@@ -106,14 +106,11 @@ def test_multiple_inputs_different_src_indices(self):
         assert_near_equal(p.model.get_val('a'), 99.)
         assert_near_equal(p.model.get_val('b'), 2.)
         assert_near_equal(p['g1.y'], 101.)
-        assert_near_equal(p['g1.a'], 99.)
```
Kenneth-T-Moore, line 109: Wow, we actually had a test with the hybrid names intentionally added to it.

openmdao/jacobians/jacobian.py
```diff
@@ -63,10 +62,20 @@ def _get_abs_key(self, key):
         try:
             return self._abs_keys[key]
         except KeyError:
-            abskey = key2abs_key(self._system(), key)
-            if abskey is not None:
+            resolver = self._system()._resolver
+            of = resolver.any2abs(key[0], 'output')
+            wrt = resolver.rel2abs(key[1], check=True)
```
Kenneth-T-Moore, line 67: This step feels a lot more complicated now.

openmdao/core/group.py
naylor-b: The check was just moved over into NameResolver.  There's a test for the error message.  Do you recall what the different message looked like?

openmdao/core/system.py
```diff
@@ -2305,16 +2225,14 @@ def _setup_var_data(self):
...
-        self._var_abs2prom = {'input': {}, 'output': {}}
-        self._var_allprocs_abs2prom = {'input': {}, 'output': {}}
         self._var_allprocs_abs2meta = {'input': {}, 'output': {}}
         self._var_abs2meta = {'input': {}, 'output': {}}
         self._var_allprocs_discrete = {'input': {}, 'output': {}}
         self._var_allprocs_abs2idx = {}
         self._owning_rank = defaultdict(int)
         self._var_sizes = {}
         self._owned_sizes = None
+        self._resolver = NameResolver(self.pathname, self.msginfo)
```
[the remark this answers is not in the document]
naylor-b: Good catch.  It's really only needed in init and System._setup_var_data.

openmdao/core/tests/test_src_indices.py
naylor-b: Yeah I think we actually had multiple tests for hybrid names.

openmdao/jacobians/jacobian.py
naylor-b: You're definitely right about that.  It's similar now to the previous version.

openmdao/core/group.py
Kenneth-T-Moore: I don't remember, but I feel like it is a clearer message than this.

Kenneth-T-Moore (approved):

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -629,7 +630,8 @@ def run(self):
         # framework is left in the right final state
         dv_dict = sol.getDVs()
         for name in self._designvars:
-            self.set_design_var(name, dv_dict[model._get_prom_name(name)])
+            # self.set_design_var(name, dv_dict[model._get_prom_name(name)])
```
swryan, line 633: could clean up some old commented code in this file

openmdao/vectors/default_vector.py
```diff
@@ -592,32 +525,32 @@ def idxs2nameloc(self, idxs):
...
+
+    #     Returns
+    #     -------
+    #     dict
+    #         state minus system member.
+    #     """
+    #     state = self.__dict__.copy()
+    #     if '_system' in state:
+    #         del state['_system']
+    #     return state
```
swryan, line 553: could clean up old commented code

swryan (approved):

openmdao/vectors/default_vector.py
naylor-b: done

openmdao/drivers/pyoptsparse_driver.py
naylor-b: done

PR #3519, opened 2025-04-28
Added a simple method to output a list of options on a system.

### LINKED ISSUE #3511: New Feature on System: list_options()

In the spirit of `list_vars` and `list_driver_vars`, a new method `list_options` would be very useful for documenting and comparing component options.

1. When called on a single component, it should print and return a list of all options on the component, their default/current values, and other defined metadata keys (desc, types).
2. When called on a group, it should print and return a the above information for all components using the same order and style as list_vars.
3. Options that are not serializable shouldn't display a value in the print. A string with class type is fine for those. You may want them in the return value, but that should be optional.

### Associated POEM

_No response_

Kenneth-T-Moore:
### Summary

Added a simple method to output a list of options on a system.

### Related Issues

- Resolves #3511

None

### New Dependencies

None


openmdao/core/system.py
```diff
@@ -4954,6 +4958,95 @@ def _get_vars_exec_order(self, inputs=False, outputs=False, variables=None, loca
...
+            When True, include the built-in openmdao system options. Default is True.
+        include_solvers : bool
+            When True, include options from nonlinear_solver and linear_solver.
+        out_stream : file-like
+            Where to send human readable output. Default is sys.stdout.
+            Set to None to suppress.
+        return_format : str
+            Indicates the desired format of the return value. Can have value of 'list' or 'dict'.
+            If 'list', the return value is a list of (name, metadata) tuples.
+            if 'dict', the return value is a dictionary mapping {name: metadata}.
```
swryan, line 4981: I think these should say: a list/dict of tuples of the form: (system options, nonlinear solver options, linear solver options)   

above and below

I wonder if it would be more user-friendly if the dict format converted the tuples to dicts as well... e.g.
```
[
  { 'cycle':
       {'options':  { 'opt1': val1, 'opt2': val2},
       {'nonlinear': { 'opt1': val1, 'opt2': val2},
       {'linear': { 'opt1': val1, 'opt2': val2}
    },
    {'objcmp':
       {'options':  { 'opt1': val1, 'opt2': val2},
       {'nonlinear': { 'opt1': val1, 'opt2': val2},
       {'linear': { 'opt1': val1, 'opt2': val2}
    },
    etc..
 }    
```

openmdao/utils/variable_table.py
```diff
@@ -308,3 +308,124 @@ def _write_variable(out_stream, row, column_names, var_dict, print_arrays):
             indented_lines = [(left_column_width + indent_inc) * ' ' +
                               s for s in out_str.splitlines()]
             out_stream.write('\n'.join(indented_lines) + '\n')
+
+
+def write_options(pathname, opt_list, out_stream=_DEFAULT_OUT_STREAM):
```
swryan, line 313: a nit, but I would name this `write_options_table` to be consistent with the other functions in this file.

openmdao/core/system.py
Kenneth-T-Moore: These are all good suggestions. I've pushed up a new version with them incorporated.

swryan (approved):

naylor-b (approved):

PR #3530, opened 2025-05-14
Added `compute_lagrange_multipliers` method to Driver.

robfalck:
### Summary

This is an interim commit on the way to providing full post-optimality sensitivities.
Adds `compute_lagrange_multipliers` and some other related methods to Driver.
This method will solve for the Lagrange multipliers of the active constraints, where the user is responsible for setting the tolerance of what determines if a constraint is active.

Assuming that the KKT conditions have been satisfied by the optimizer, then the Lagrangian is

```math
\nabla_{\bar\theta} f(\bar\theta, \bar{p}) + \lambda \nabla_{\bar\theta} \bar{g}(\bar\theta, \bar{p}) = 0
```

where $\bar{\theta}$ are the design variables of the problem and $\bar{p}$ are any input parameters of the system.

We assume that $\bar{g}$ consists of all active equality and inequality constraints, and active bounds.

The returned values of `compute_lagrange_multipliers` are two dictionaries: one for active design variable bounds, and one for active constraints. These dictionaries contain the multipliers, the active indices of the dv/constraint, and the active bound of the dvs and constraints.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/driver.py
+++ b/openmdao/core/driver.py
@@ -8,6 +8,7 @@ import os
 import weakref
 
 import numpy as np
+import scipy.sparse as sp
 
 from openmdao.core.group import Group
 from openmdao.core.total_jac import _TotalJacInfo
@@ -1732,8 +1733,6 @@ class Driver(object, metaclass=DriverMetaclass):
         for constraint, con_options in constraints.items():
             constraint_value = con_vals[constraint]
             con_size = con_options['size']
-            adder = 0.0 if con_options['total_adder'] is None else con_options['total_adder']
-            scaler = 1.0 if con_options['total_scaler'] is None else con_options['total_scaler']
 
             if con_options.get('equals', None) is not None:
                 # Equality constraint, all indices active
@@ -1741,8 +1740,8 @@ class Driver(object, metaclass=DriverMetaclass):
                                            'active_bounds': np.zeros(con_size, dtype=int)}
             else:
                 # Inequality constraint, determine active indices and bounds
-                constraint_upper = (con_options.get("upper", np.inf) + adder) * scaler
-                constraint_lower = (con_options.get("lower", -np.inf) + adder) * scaler
+                constraint_upper = con_options.get("upper", np.inf)
+                constraint_lower = con_options.get("lower", -np.inf)
 
                 if np.all(np.isinf(constraint_upper)):
                     upper_idxs = np.empty()
@@ -1769,10 +1768,8 @@ class Driver(object, metaclass=DriverMetaclass):
         for des_var, des_var_options in des_vars.items():
             des_var_value = dv_vals[des_var]
             des_var_size = des_var_options['size']
-            adder = 0.0 if con_options['total_adder'] is None else con_options['total_adder']
-            scaler = 1.0 if con_options['total_scaler'] is None else con_options['total_scaler']
-            des_var_upper = (np.ravel(des_var_options.get("upper", np.inf)) + adder) * scaler
-            des_var_lower = (np.ravel(des_var_options.get("lower", -np.inf)) + adder) * scaler
+            des_var_upper = np.ravel(des_var_options.get("upper", np.inf))
+            des_var_lower = np.ravel(des_var_options.get("lower", -np.inf))
 
             if assume_dvs_active:
                 active_dvs[des_var] = {'indices': np.arange(des_var_size, dtype=int),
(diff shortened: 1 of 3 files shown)


naylor-b (approved):

openmdao/core/driver.py
```diff
@@ -1687,6 +1688,289 @@ def _update_result(self, result):
...
+        feas_atol : float
+            Feasibility absolute tolerance
+        feas_rtol : float
+            Feasibility relative tolerance
+        assume_dvs_active : bool
+            Override to force design variables to be treated as active.
+
+        Returns
+        -------
+        active_cons : dict[str: dict
```
Kenneth-T-Moore, line 1709: missing closing bracket

swryan:
Should there be a doc page for this, explaining how to use it.. or is it too early for that?

openmdao/core/driver.py
```diff
@@ -1687,6 +1688,289 @@ def _update_result(self, result):
...
+                                               'active_bounds': active_bounds}
+
+        for des_var, des_var_options in des_vars.items():
+            des_var_value = dv_vals[des_var]
+            des_var_size = des_var_options['size']
+            des_var_upper = np.ravel(des_var_options.get("upper", np.inf))
+            des_var_lower = np.ravel(des_var_options.get("lower", -np.inf))
+
+            if assume_dvs_active:
+                active_dvs[des_var] = {'indices': np.arange(des_var_size, dtype=int),
```
Kenneth-T-Moore, line 1775: There is a weird corner case where both upper and lower can be active. It looks like this is supported by this overall method. There is one weird byproduct when assume_dvs_active is True, where the active bounds are all "0", which means equality constraint instead of containing both lower and upper. I am not sure if this matters or has any repercussions.

openmdao/core/driver.py
```diff
@@ -1687,6 +1688,289 @@ def _update_result(self, result):
...
+
+        grad_f_vec = np.zeros((n))
+        offset = 0
+        for inp in grad_f.keys():
+            inp_size = grad_f[inp].size
+            grad_f_vec[offset:offset + inp_size] = grad_f[inp]
+            offset += inp_size
+
+        actives = active_cons | active_dvs
+        n_active = np.sum(np.fromiter((len(c['indices']) for c in actives.values()), dtype=int))
```
Kenneth-T-Moore, line 1912: n_active isn't used for anything other than to bail if we don't have any active cons or desvars. Isn't it easier just to check if active_cons and active_dvs are empty dicts?

openmdao/core/driver.py
```diff
@@ -1687,6 +1688,289 @@ def _update_result(self, result):
...
+        multipliers = dict()
+        offset = 0
+
+        opts = ['indices', 'ref0', 'ref', 'units']
+        driver_vars = prob.list_driver_vars(out_stream=None,
+                                            desvar_opts=opts + ['flat_indices'],
+                                            cons_opts=opts + ['flat_indices'],
+                                            objs_opts=opts,
+                                            return_format='dict')
+        driver_vars = driver_vars['design_vars'] | driver_vars['constraints']
```
Kenneth-T-Moore, line 1956: is driver_vars used somewhere?

openmdao/core/driver.py
```diff
@@ -1687,6 +1688,289 @@ def _update_result(self, result):
...
+            n += grad_f[inp].size
+
+        grad_f_vec = np.zeros((n))
+        offset = 0
+        for inp in grad_f.keys():
+            inp_size = grad_f[inp].size
+            grad_f_vec[offset:offset + inp_size] = grad_f[inp]
+            offset += inp_size
+
+        actives = active_cons | active_dvs
```
Kenneth-T-Moore, line 1911: Does this cause problems if we have a desvar and a constraint on the same variable (maybe on different source indices)?

```
>> {'a': 3} | {'b' : 4, 'a': 5}
{'a': 5, 'b': 4}
```

openmdao/drivers/tests/test_compute_lagrange_multipliers.py
```diff
@@ -0,0 +1,201 @@
...
+                # if we decrease y bound by 0.01 units, we should see the objective increase by lambda[y] * 0.01 units.
+                # we use a loose tolerance when testing this due to FD across the optimization
+                multipliers, active_info = prob.driver.compute_lagrange_multipliers(driver_scaling=False)
+                lambda_y = multipliers['y']
+                f_nom = prob.get_val('f_xy')
+
+                prob = self._make_problem(driver=driver, y_lower=-7.01)
+                result =  prob.run_driver()
+                self.assertEqual(result.exit_status, 'SUCCESS',
+                                 msg=f'Failed to converge optimization with {driver_name}.')
```
Kenneth-T-Moore, line 194: `use_sparse_solve` seems to be untested.

openmdao/core/driver.py
robfalck: ah good catch

openmdao/core/driver.py
robfalck: Vestige of the way I was doing things previously, fixing.

openmdao/core/driver.py
robfalck: It will take me a bit of time to come up with a test and the logic to do this correctly.
I think if we find that an active desvar is also an active constraint, we merge the two sets of active information.
In that case, lower, upper, and equals will have to be amended to contain the appropriate values depending on the index.  The lower bound of the desvar at certain indices may be different that the lower bound on a constraint at other indices.

Kenneth-T-Moore (approved):

PR #3538, opened 2025-05-27
Update minimum required python and scipy versions (and test workflow)

### LINKED ISSUE #3045: Test with NumPy 2.0 when release candidate becomes available
### Task Description

https://github.com/numpy/numpy/issues/24300

swryan:
### Summary

Updated minimum supported version of Python to 3.10 and SciPy to 1.12

Updated tests workflow:
- Test with NumPy 2.2 on Ubuntu and MacOS
- Test `pyoptsparse` from `conda-forge`  as well as using `build_pyoptsparse` script
- Test on MacOS 15 (previously MacOS 14)
- Removed `NPY201` compatibility check since we are now testing with NumPy 2
- Added skip conditions to some tests

*Note:*  there is some kind of issue with the `coverage` package (in conjunction with `testflo`?) when using Py313/NP2 on Ubuntu that dumps the stack of `_collectors`, resulting in a lot of extraneous output after running the tests.  It's not a failure, as coverage is still collected, but is very annoying.  Partially for this reason, only the ubuntu `minimal` job uses this configuration and the more important ubuntu full build still uses Python 3.12 and NumPy 1.26

*Note:* different options for building/installing `pyoptsparse` are tested, with the `build_pyoptsparse` script as before or from `conda-forge` using a separately built `SNOPT` library.

### Related Issues

- Resolves #3045

Updated minimum supported version of Python to 3.10 and SciPy to 1.12

### New Dependencies

None

### DIFF
--- a/.github/workflows/openmdao_docs_workflow.yml
+++ b/.github/workflows/openmdao_docs_workflow.yml
@@ -32,12 +32,12 @@ jobs:
 
           # build docs (baseline versions)
           - NAME: Build Docs
-            OS: ubuntu-latest
-            PY: '3.11'
+            OS: ubuntu-24.04
+            PY: '3.12'
             NUMPY: '1.26'
-            SCIPY: '1.13'
-            PETSc: '3.19'
-            PYOPTSPARSE: 'v2.13.0'
+            SCIPY: '1.15'
+            PETSc: '3.20'
+            PYOPTSPARSE: 'v2.13.1'
             SNOPT: '7.7'
             OPTIONAL: '[all]'
             PUBLISH: true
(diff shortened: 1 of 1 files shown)


.github/workflows/openmdao_test_workflow.yml
```diff
@@ -237,35 +217,43 @@ jobs:
...
         run: |
           echo "============================================================="
-          echo "Install compilers for PETSc"
+          echo "Install PETSc"
           echo "============================================================="
-          if [[ "${{ matrix.OPENMPI }}" ]]; then
-            COMPILERS="cython compilers openmpi-mpicc=${{ matrix.OPENMPI }}"
+
+          if [[ "${{ matrix.BREW }}" ]]; then
+            brew install open-mpi
```
Kenneth-T-Moore, line 246: I brew pretty much a requirement now for getting things working on the mac?

Kenneth-T-Moore (approved):

.github/workflows/openmdao_test_workflow.yml
Kenneth-T-Moore: I don't think we have to change anything. Mostly just curious.

PR #3543, opened 2025-06-02
Fixed a bug where connections within the same group were being flagged as invalid if issued from above.

### LINKED ISSUE #3542: `connect` incorrectly marks a connection invalid based on promoted names.
### Description

In the example below, connect fails because both input and output appear to belong to group2, but in reality they are inputs and outputs of subsystems. This connection is valid if issued on `group2`, or even if promoted to the top level.

We just need to delay the check until setup and base it upon the absolute path of the input and output, rather than the promoted path.

### Example

```
import openmdao.api as om
import pytest

# Generate the components used in the tests
class Comp1(om.ExplicitComponent):

    def setup(self):
        self.add_input('x', val=1.0)
        self.add_output('y', val=1.0)

        outputs['y'] = 3 * inputs['x']


class Comp2(om.ExplicitComponent):

    def setup(self):
        self.add_input('w', val=1.0)
        self.add_output('z', val=1.0)

        outputs['z'] = 4 * inputs['w']


def test_subgroup_promoted_connection():
    """
    This example demonstrates a problem which fails with an error when trying
    to connect the variables promoted up to the same sub-group.
    """
    class Group2(om.Group):

        def setup(self):
            self.add_subsystem('comp1', Comp1(), promotes=['*'])
            self.add_subsystem('comp2', Comp2(), promotes=['*'])

    class Group1(om.Group):
        def setup(self):
            self.add_subsystem('group2', Group2())
            self.connect('group2.y', 'group2.w')

    prob = om.Problem()
    prob.model = Group1()
    prob.model.add_subsystem('group1', Group1())
    partial_msg = 'Output and input are in the same System*'
    prob.setup()


```

### OpenMDAO Version

3.39.1-dev

_No response_

robfalck:
### Summary

Wait until setup to flag connections within the same system.
This needs to be done using absolute names of the source and target, which are not available at the time the connection is issued.

Fortunately the correct logic already existed in the setup logic, but connect was flagging this issue early.

### Related Issues

- Resolves #3542 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_group.py
+++ b/openmdao/core/tests/test_group.py
@@ -2526,9 +2526,11 @@ class TestConnect(unittest.TestCase):
         msg = "Output and input are in the same System for connection " + \
               "from 'tgt.y' to 'tgt.x'."
 
+        p.model.sub.connect('tgt.y', 'tgt.x', src_indices=[1])
+
         with set_env_vars_context(OPENMDAO_FAIL_FAST='1'):
             with self.assertRaisesRegex(Exception, msg):
-                p.model.sub.connect('tgt.y', 'tgt.x', src_indices=[1])
+                p.setup()
 
     def test_connect_within_system_with_promotes(self):
         prob = om.Problem(name='connect_within_system_with_promotes')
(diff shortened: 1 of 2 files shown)


openmdao/core/tests/test_group.py
```diff
@@ -2547,6 +2549,53 @@ def test_connect_within_system_with_promotes(self):
 
         self.assertEqual(str(ctx.exception), msg)
 
+    def test_subgroup_promoted_connection(self):
+        """
+        This example demonstrates a problem which fails with an error when trying
```
rob-hetterich, line 2554: Only thing I would really suggest changing is to remove this comment or change the working. Since you made the fix, this example is no longer meant to fail which means the description is wrong

rob-hetterich (changes requested):
One small comment

naylor-b (approved):

rob-hetterich (approved):

PR #3546, opened 2025-06-10
Added print_mean to variable list methods

johnjasa:
### Summary

We use OpenMDAO's built-in `list_inputs` and `list_outputs` methods in our [H2Integrate tool](https://github.com/NREL/H2Integrate).
We often work with arrays of length 8760 (hours in a year) when modeling energy systems.
The norm of these arrays isn't necessarily useful, but seeing the mean would be helpful.
This PR adds the option to print the mean alongside the existing min and max printouts.

Originally suggested by @jmartin4nrel here: https://github.com/NREL/H2Integrate/issues/155
I first went to implement Jonathan's changes (optionally replace the norm with the mean), but saw that `print_min` and `print_max` already existed and liked the parallelism for the mean.

Feel free to suggest other related changes and let me know if I missed anywhere else in the codebase where we'd need to update accordingly.

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -4243,6 +4243,7 @@ class System(object, metaclass=SystemMetaclass):
                   out_stream=_DEFAULT_OUT_STREAM,
                   print_min=False,
                   print_max=False,
+                  print_mean=False,
                   return_format='list'):
         """
         Write a list of inputs and outputs sorted by component in execution order.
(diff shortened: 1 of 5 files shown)


robfalck (approved):

Kenneth-T-Moore (approved):
Mean is much more useful than the norm we are printing. Also, I think our norm is missing a 1/N in the calculation.

PR #3552, opened 2025-06-26
Refactor of Jacobian class internals

naylor-b:
### Summary

General cleanup of all code related to partial jacobians.  A later PR will attempt to do the same for total jacobians and to improve code reuse between partial and total jacobians.

The 'diagonal' keyword has been added when calling declare_partials, making it unnecessary to specify the 'rows' and 'cols' arguments in that case.

The 'csc' value for the 'assembled_jac_type' option in Group and ImplicitComponent has been deprecated.  The preferred values are now either 'dense' or 'sparse'.  If 'sparse' is chosen, the linear solver will determine the specific sparse format used internally.  A warning will be issued if the user specified value, for example 'csc', differs from the preferred format for the linear solver but the user specified format will still be used.

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_check_totals.py
+++ b/openmdao/core/tests/test_check_totals.py
@@ -1374,9 +1374,9 @@ class TestProblemCheckTotals(unittest.TestCase):
         data = prob.check_totals(method='cs', out_stream=stream, directional=True)
         content = strip_formatting(stream.getvalue())
 
-        self.assertEqual(content.count('rev value:'), 0)
-        self.assertEqual(content.count('fwd value:'), 1)
-        self.assertEqual(content.count('fd value:'), 1)
+        self.assertEqual(content.count('rev value @ max viol:'), 0)
+        self.assertEqual(content.count('fwd value @ max viol:'), 1)
+        self.assertEqual(content.count('fd value @ max viol:'), 1)
         self.assertEqual(content.count('Directional Derivative (Jfwd)'), 1)
         self.assertEqual(content.count('Directional CS Derivative (Jfd)'), 1)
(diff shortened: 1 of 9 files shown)


openmdao/core/tests/test_check_totals.py
```diff
@@ -937,14 +937,14 @@ def setup(self):
                 obj = sub.add_subsystem('obj_cmp', om.ExecComp('obj = x**2 + z[1] + y1 + exp(-y2)', obj=0.0,
                                                          x=0.0, z=np.array([0.0, 0.0]), y1=0.0, y2=0.0),
                                   promotes=['obj', 'x', 'z', 'y1', 'y2'])
-                obj.declare_partials(of='*', wrt='*', method='cs')
+                obj.declare_partials(of='*', wrt='*', method='fd')
```
Kenneth-T-Moore, line 940: why did this test switch from cs to fd?

Kenneth-T-Moore (approved):
Tested against some aviary benches.

robfalck (approved):

openmdao/core/tests/test_check_totals.py
naylor-b: With cs it was failing with errors around 1e-9.  I just went back and took a look at this and it turns out that there was a bug on master in the code that switches over to FD to avoid nested CS.  A single variable name was being passed into `add_approximation` instead of a key tuple, so the name for the wrt variable inside of `add_approxmation` was wrong.  After fixing that bug on master, both branches now get identical total derivatives.  I updated the checking of the totals to use 2e-9 as the tolerance and now the test passes using cs.

PR #3553, opened 2025-06-30
Added real time plot for analysis driver runs

### LINKED ISSUE #3505: Create a real time plot capability for AnalysisDriver.

Real-time plotting should work with AnalysisDriver, though the format of the plot should be different.

When invoked, the realtime plotting tool should determine the perturbed variables and collected responses.
Sampled variables are available in `AnalysisDriver._get_sampled_vars` or in the 'includes' option of the recorded variables. Responses are added to the model using `add_response` so should be available in the standard way.

## Kinds of Plots

When the real time plotting tool is plotting the results of an AnalysisDriver, two types of plots should be available, with some sort of way to change them in the GUI.  First, a scatterplot matrix should contain an NxN array of scatterplots when there are N sampled variables. Each point plotted should have its value displayed on a color axis. The user should be able to choose which response variable is plotted on the color axis.

A second kind of plot is a [Parallel coordinate plot](https://en.wikipedia.org/wiki/Parallel_coordinates). Here, each sampled variable gets its own axis, and within a single run we connect all of the values of the various sampled variables.  Again, the color of the line corresponds to the value of a selected response variable.

### Associated POEM

_No response_

hschilling:
### Summary

This PR includes code for doing real-time plots of a run using AnalysisDriver. Also does refactoring to enable using some common code between the AnalysisDriver plot and the optimizer plot. 

### Related Issues

- Resolves #3505 

None

### New Dependencies

None


openmdao/utils/shell_proc.py
```diff
@@ -357,3 +359,35 @@ def check_call(args, stdin=None, stdout=None, stderr=None, env=None,
...
+            return True
+        else:
+            # If OpenProcess fails, check if it's due to access being denied.
+            ERROR_ACCESS_DENIED = 5
+            if ctypes.windll.kernel32.GetLastError() == ERROR_ACCESS_DENIED:
+                return True
+            return False
+    else:
+        try:
+            os.kill(pid, 0)
```
naylor-b, line 384: Maybe just put a comment here indicating that calling kill with 0 doesn't actually send a kill signal to the process.

Kenneth-T-Moore (commented):
I've been playing with this using an NPSS model in an external code comp optimizing 6 design variables, and it is pretty slick. I will have some possible suggestions, but I think this is ready for people to play around with.

robfalck (approved):

PR #3593, opened 2025-08-12
Added implementation of `find_feasible` for POEM 096

### LINKED ISSUE #3161: Implementation of POEM 096: Minimization of Constraint Violation

POEM 096 proposes the ability to minimize the constraint violation from some starting point of driver.  This would allow users to find a feasible starting point given the current design variable values.

SNOPT supports the option to find a feasible starting point, but this capability isn't available in other optimizers.
Adding it to OpenMDAO would be a valuable addition.

The POEM proposes a capability, but not an implementation, and that needs to be developed.

In this case, we're asking the optimizer to find the feasible starting point, so it seems like it should be associated with the driver.

### Proposal for initial implementation:

Calling `prob.run_driver(find_feasible=True)` will replace the user-defined objective with a new objective that is the norm of the constraint violations.

This new objective is not the output of an OpenMDAO component itself, so we will have to implement the reverse mode derivatives more manually.

### Associated POEM

096

### LINKED ISSUE #3592: Provide a mechanism for minimizing constraint violations (finding a feasible solution if possible)

POEM 096 provides a proposal for finding a feasible solution by minimizing the L1 or L2 norm of the constraint violation.

This can be accomplished using `scipy.optimize.least_squares` and providing it with the magnitude of the violation of each constraint.

### Associated POEM

096

robfalck:
### Summary

Adds a `find_feasible` method to Problem. This method will use `scipy.optimize.least_squares` to attempt to find design variable values which minimize the constraint violations in the least-squares sense.

By default, `least_squares` will minimize the sum of the squares of the constraint violations in driver-scaled units, but the `loss` parameter which is forwarded to least_squares allows this to be changed to  a `soft_l1` approximation as well as a few other options.

The `find_feasible` method supports an `excludes_desvars` argument which allows for the exclusion of design variables from the search via glob patterns.  This allows the user to perform a search that minimizes infeasibilities while holding some design variables fixed.

### Related Issues

- Resolves #3592 
- Resolves #3161 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_driver.py
+++ b/openmdao/core/tests/test_driver.py
@@ -871,8 +871,6 @@ class TestDriver(unittest.TestCase):
             prob.final_setup()
 
     def test_no_desvars(self):
-        import openmdao.api as om
-
         prob = om.Problem()
 
         c1 = om.ExecComp()
--- a/openmdao/core/tests/test_problem.py
+++ b/openmdao/core/tests/test_problem.py
@@ -2464,8 +2464,6 @@ class TestProblem(unittest.TestCase):
                                     assert_near_equal(prob.get_val('x'), 1.5, tolerance=1.0E-8)
 
     def test_find_feasible_no_feasible_solution(self):
-            import openmdao.api as om
-
             prob = om.Problem()
 
             c1 = om.ExecComp()
@@ -2530,8 +2528,6 @@ class TestProblem(unittest.TestCase):
             self.assertTrue(failed)
 
     def test_find_feasible_exclude_desvars(self):
-        import openmdao.api as om
-
         prob = om.Problem()
 
         c1 = om.ExecComp()


openmdao/core/driver.py
```diff
@@ -2001,6 +2033,384 @@ def compute_lagrange_multipliers(self, driver_scaling=False, feas_tol=1.0E-6,
...
+            nl_con_viol_dict = self.get_constraint_values(lintype='nonlinear',
+                                                          driver_scaling=driver_scaling,
+                                                          viol=True)
+
+            return np.concatenate([v.ravel() for v in
+                                   list(lin_con_viol_dict.values()) +
+                                   list(nl_con_viol_dict.values())])
+
+        except Exception as e:
+            raise ValueError('foo') from e
```
Kenneth-T-Moore, line 2113: is "foo" a placeholder exception message?

openmdao/core/driver.py
```diff
@@ -2001,6 +2033,384 @@ def compute_lagrange_multipliers(self, driver_scaling=False, feas_tol=1.0E-6,
...
+        driver_scaling : bool
+            If True, consider the constraint violation in driver-scaled units. Otherwise, it
+            will be computed in the model's units.
+        exclude_desvars : str or Sequence[str] or None
+            If given, a pattern of one or more design variables to be excluded from
+            the least-squares search.  The allows for finding a feasible (or least infeasible)
+            solution when holding one or more design variables to their current values.
+        method : {'trf', 'dogbox', or 'lm'}
+            The method used by scipy.optimize.least_squares. One or 'trf', 'dogbox', or 'lm'.
+        ftol : float or None
```
Kenneth-T-Moore, line 2194: I am not sure None is valid for ftol/xtol/gtol.

Kenneth-T-Moore (approved):

openmdao/core/tests/test_driver.py
```diff
@@ -870,6 +870,32 @@ def test_record_residuals_includes_excludes(self):
         with assert_warnings(expected_warnings):
             prob.final_setup()
 
+    def test_no_desvars(self):
```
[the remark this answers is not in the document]
rob-hetterich: Okay, that's fine. Just double checking if the system level test was sufficient vs some more granular testing of the individual methods.

rob-hetterich (approved):

PR #3604, opened 2025-09-02
Turn on pre-opt-post default. Move warning about redundant adjoint solves into optional checks. Remove warning about overwriting recorder files.

Kenneth-T-Moore:
### Summary

1. Turned on the `group_by_pre_opt_post` option. We have been using it for a while, and it is the way we should run.
2. Moved a warning about redundant adjoint solves with DirectSolver into the check_config section, so that they only run when requested (i.e., check=True).
3. Removed the warning about overwriting case files. By now, everyone expects case files to be overwritten.

### Related Issues

- Resolves #

It is possible that `group_by_pre_opt_post` might cause some unexpected behavior in some complicated models, including those with Mphys running in MPI. We need to check on these.

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -223,6 +223,26 @@ class DirectSolver(LinearSolver):
 
         self.supports['implicit_components'] = True
 
+    def check_config(self, logger):
+        """
+        Perform optional error checks.
+
+        Parameters
+        ----------
+        logger : object
+            The object that manages logging output.
+        """
+        if self.options['rhs_checking'] is False:
+            system = self._system()
+            redundant_adj = system.pathname in system._relevance.get_redundant_adjoint_systems()
+            if redundant_adj:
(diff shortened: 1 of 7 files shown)


openmdao/solvers/linear/direct.py
```diff
@@ -223,6 +223,26 @@ def _declare_options(self):
 
         self.supports['implicit_components'] = True
 
+    def check_config(self, logger):
```
naylor-b, line 226: Previously this check was done for DirectSolver, PETScDirectSolver, and both KSP solvers, so we should probably add this check to the KSP solvers too.

naylor-b (changes requested):

openmdao/solvers/linear/direct.py
Kenneth-T-Moore: Oh yeah, I forgot to check for other usages. Fixed.

PR #3608, opened 2025-09-05
Added ability to set solver debug_print option via set_solver_print.

### LINKED ISSUE #3607: Improved solver debug printing.

Solver debug printing is currently only triggered when the solver fails and raises an exception, as enabled by the option `"err_on_non_converge"`.

There are times when solvers print warnings about "'inf' or 'Nan' after N iterations", and it would be convenient to see  the solver debug information that resulted in this information.

Furthermore, solver debug printing shouldn't be reliant on `err_on_non_converge` also being set.
OpenMDAO should just print the debug information if it is requested.

Setting the debug_print flag on solvers isn't as trivial as using `set_solver_print`, so OpenMDAO should also allow `debug_print` as an argument to `set_solver_print`. Its default value of `None` will leave debug printing unchanged, while values of True or False will enable or disable it.

### Associated POEM

_No response_

robfalck:
### Summary

- `set_solver_print` can now be used to set `debug_print` behavior on solvers in a Problem.
- Changed solver `debug_print` behavior to always be displayed during a NL solver failure, regardless of `err_on_non_converge`.
- Allow solver print `level=None` to allow it to remain unchanged when the user only wants to change `debug_print`. For now, the default level remains 2 to maintain current behavior.

### Related Issues

- Resolves #3607 

None

### New Dependencies

None

### DIFF
--- a/openmdao/visualization/n2_viewer/tests/betz_tree.json
+++ b/openmdao/visualization/n2_viewer/tests/betz_tree.json
@@ -5,7 +5,8 @@
     "expressions": null,
     "nonlinear_solver": "NL: RUNONCE",
     "nonlinear_solver_options": {
-        "iprint": 1
+        "iprint": 1,
+        "debug_print": false
     },
     "linear_solver": "LN: RUNONCE",
     "linear_solver_options": {


openmdao/solvers/tests/test_solver_debug_print.py
```diff
@@ -185,6 +186,7 @@ def test_solver_debug_print_multi_run_scenario(self):
         output = run_model(p)
 
         # Should be empty since solver debugging printing was turned off
+        self.maxDiff = 10000
```
Kenneth-T-Moore, line 189: This looks like it was accidentally included.

PR #3628, opened 2025-09-30
Updated GitHub workflows to properly download SNOPT 

swryan:
### Summary

Updated workflows to properly download SNOPT 

- moved SNOPT download logic into a bash script
- updated `install_pyoptsparse` action to use the new script
- updated the `docs` workflow to build the docs with SNOPT support
- updated the `latest` workflow to use the new script
- updated the `test` workflow to use the new script

> [!NOTE]  
> Since SNOPT 7.2 is no longer used in CI testing, this could all be cleaned up to just use SNOPT 7.7 from SNOPT_LOCATION. This is left as future work (as part of assumed further refactoring).

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/.github/actions/install_pyoptsparse/download_snopt.sh
+++ b/.github/actions/install_pyoptsparse/download_snopt.sh
@@ -1,4 +1,4 @@
-# this script will create a SNOPT firectory and populate it
+# this script will create a SNOPT directory and populate it
 # with the specified version of the SNOPT source code
 # via secure copy over SSH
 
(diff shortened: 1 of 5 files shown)


.github/actions/install_pyoptsparse/download_snopt.sh
```diff
@@ -0,0 +1,39 @@
+# this script will create a SNOPT firectory and populate it
```
naylor-b, line 1: typo 'firectory'

PR #3644, opened 2025-11-25
add_unit can now include prefixed-units.

robfalck:
### Summary

Previously `add_unit` would fail if the new unit was defined relative to a non-base unit.
For instance,

```
add_unit('DU', '1.495978707E8*km')
```

This would result in a very uninformative message. This fix applies `_find_unit` to any units detected in the unit definition string, which adds those units to the table of known units, if they are not already there.

### Related Issues

None

None

### New Dependencies

None

### DIFF
--- a/openmdao/solvers/linear/direct.py
+++ b/openmdao/solvers/linear/direct.py
@@ -528,7 +528,7 @@ class DirectSolver(LinearSolver):
             x_vec[:] = sol_array = scipy.linalg.lu_solve(self._lup, b_vec, trans=trans_lu)
 
         if not system.under_complex_step and self._lin_rhs_checker is not None and mode == 'rev':
-            self._lin_rhs_checker.add_solution(b_vec, sol_array, copy=True)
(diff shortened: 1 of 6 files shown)


openmdao/utils/units.py
```diff
@@ -654,14 +654,35 @@ def add_unit(name, unit, comment=''):
...
+            # Now try evaluating again
+            try:
+                unit = eval(unit, {'__builtins__': None, 'pi': pi},  # nosec: scope limited
+                            _UNIT_LIB.unit_table)
+            except Exception as eval_error:
+                raise ValueError(
+                    f"Could not evaluate unit definition '{unit}' for new unit '{name}'. "
+                    f'Original error: {eval_error}. '
+                ) from eval_error
+
```
Kenneth-T-Moore: Does this mean it will allow units like kilo-kilometer (kkm)?

robfalck: Interesting point.  I'll give that a try.  I'm not opposed to big kid rules and letting users define what they want.

PR #3674, opened 2026-01-06
Add an output shape checker for compute_primal in Jax components.

### LINKED ISSUE #3673: Returning a scalar from `compute_primal` in a `JaxExplicitComponent` causes JAX VJP to fail
### Description

Returning a scalar from `compute_primal` in a `JaxExplicitComponent` causes a shape mismatch when derivatives are computed in reverse mode. According to Claude, OpenMDAO builds cotangents from declared output shapes (self._outputs.values()). But JAX expects cotangents matching the actual traced output shapes from `compute_primal`.

```
ValueError: unexpected JAX type (e.g. shape/dtype) for argument to VJP function: 
    got float64[1], but expected float64[] because the corresponding output of the 
    differentiated function had JAX type float64[]
```

This can happen unexpectedly. For example I got this error when I returned the contents of an array I'd unpacked. Unpacks with `shape ()` instead of `shape (1,)`).

### Example

import numpy as np
import jax.numpy as jnp
import openmdao.api as om

class ScalarReturningComp(om.JaxExplicitComponent):
    def setup(self):
        self.add_input("x", val=1.0)
        self.add_output("y", val=1.0)
    
    def compute_primal(self, x):
        return x[0] * 2.0

prob = om.Problem()
prob.model.add_subsystem("comp", ScalarReturningComp(), promotes=["*"])
prob.setup()
prob.run_model()
prob.model.comp.compute_sparsity(direction='rev') # error

### OpenMDAO Version

3.41.0

jax==0.8.1
jaxlib==0.8.1
numpy==2.3.5

naylor-b:
### Summary

We now flag when shapes of outputs from compute_primal don't match the shapes that the outputs are declared with.

### Related Issues

- Resolves #3673

None

### New Dependencies

None


openmdao/utils/jax_utils.py
```diff
@@ -1094,6 +1095,50 @@ def _jax_derivs2partials(self, deriv_vals, partials, ofnames, wrtnames):
...
+
+    computed = []
+    for val in retvals:
+        try:
+            computed.append(val.shape)
+        except AttributeError:
+            computed.append(None)
+
+    computed_out_shapes = []
+    for oshape in computed:
```
rob-hetterich: Minor thing, but is the `computed` variable needed? Seems like it's just created in the previous loop to then get looped through here. Could reduce the LOC and the amount of looping that is done by just setting the `computed_out_shapes` in the previous loop of `retvals`, no?

openmdao/components/jax_implicit_comp.py
```diff
@@ -541,3 +564,9 @@ def _compute_output_shape(self, name, input_shapes):
             self._output_shapes = {n: shp for n, shp in zip(self._var_rel_names['output'],
                                                             out_shapes)}
         return self._output_shapes[name]
+    
+    def _solve_nonlinear(self):
```
rob-hetterich: Is this redundant? You already have a `_solve_nonlinear` method earlier in the class

rob-hetterich (changes requested):
Just a few minor things and questions

Kenneth-T-Moore (approved):

PR #3690, opened 2026-02-02
Functional interface for `Problem`

### LINKED ISSUE #3701: Functional interface organization.
### Task Description

Propose functional interface and decompose into assignable tasks.

dingraha:
### Summary

This PR implements the "functional interface" to an OpenMDAO `Problem` as described by [POEM 108](https://github.com/dingraha/POEMs/blob/master/POEM_108.md).

~~TODOs~~ see the POEM linked above.

Closes #3701


openmdao/core/total_jac.py
```diff
@@ -88,7 +90,10 @@ class _TotalJacInfo(object):
 
     def __init__(self, problem, of, wrt, return_format, approx=False,
                  debug_print=False, driver_scaling=True, get_remote=True, directional=False,
-                 coloring_info=None, driver=None):
+                 coloring_info=None, driver=None, of_indices=None, wrt_indices=None,
+                 do_special_functional_api_thing=False, always_include_linear=False,
```
robfalck: While I appreciate the explicitness of `do_special_functional_api_thing`, I'd suggest we shorten this to `functional`. Arguably this should be a "private" argument or put into **kwargs.

openmdao/core/problem.py
```diff
@@ -128,6 +130,475 @@ def _default_prob_name():
     return name.stem
 
 
+class _FunctionalCallback(object):
```
robfalck: this seems like a long enough class definition that it should live in its own file.

`object` is the implicit super type of classes in python >3 so we can just do `class _FunctionalCallback:`

openmdao/core/problem.py
```diff
@@ -128,6 +130,475 @@ def _default_prob_name():
...
+        Return the Problem associated with this callback.
+
+        Returns
+        -------
+        Problem
+            The Problem instance that was passed to the constructor.
+        """
+        return self._problem
+
+    def create_input_vector(self):
```
robfalck: For very large multi-variable problems, it may be a chore for users to know what indices of each element of the flat vector correspond to various inputs or outputs. Having this method return (or optionally return) a mapping of variable name to a slicer object might be useful.

```python
x, map = f.create_input_vector(return_map=True)
x[map['foo']] = [ 4., 3., 2. ]
```

or, alternatively, use a set version of your `get_input_val` and `get_output_val` methods:

```python
x = f.create_input_vector()
f.set_input_val('foo', x, [4., 3., 2.])
```

That would give the user a functional interface but also make setting values of specific inputs a little bit easier.

openmdao/docs/openmdao_book/features/experimental/functional_interface.ipynb
robfalck, line 1: Thorough documentation but maybe include a really basic example of design space exploration (replacing analysis driver) or optimization without using an OpenMDAO driver.

openmdao/core/total_jac.py
```diff
@@ -155,11 +182,64 @@ def __init__(self, problem, of, wrt, return_format, approx=False,
         orig_of = of
         orig_wrt = wrt
 
+        # DJI: This block added by Claude.
```
robfalck: extraneous comment

robfalck (changes requested):

openmdao/core/problem.py
```diff
@@ -2433,6 +2434,103 @@ def get_outputs_dir(self, *subdirs, mkdir=False):
...
+            is required.
+
+        Returns
+        -------
+        _FunctionalCallback
+            A callable object whose ``__call__`` method maps a flat NumPy input vector
+            to the requested outputs and/or Jacobian.
+
+        Examples
+        --------
```
Kenneth-T-Moore: Putting examples in the docstrings is a bit of a gray area since (as far as I know) we don't test them.

openmdao/core/functional_callback.py
```diff
@@ -0,0 +1,525 @@
+from collections import OrderedDict
+from collections.abc import Mapping
+
+import numpy as np
+
+from openmdao.core.total_jac import _TotalJacInfo
+
+
+class _FunctionalCallback:
```
Kenneth-T-Moore, line 9: Should this object be public instead of private, particularly since the user is calling public functions on it?

openmdao/core/group.py
```diff
@@ -3770,6 +3770,68 @@ def get_responses(self, recurse=True, get_sizes=True, use_prom_ivc=False):
 
         return out
 
+    def _get_totals_of_metadata(self, driver=None, of=None):
+        if isinstance(of, str):
+            of = [of]
+
+        if not driver:
+            if of is None:
+                raise RuntimeError("driver must be specified if of variables are not "
```
Kenneth-T-Moore, line 3779: Do these checks represent states that are even possible to reach? This is a private method, so it is only going to be called under the controlled circumstances that we have set up.

If it is possible to reach these as a user, then the error message needs to tell the user what to do rather than describe the function-call mechanics that occurred here.

This applies to the methods below here too.

openmdao/core/total_jac.py
dingraha: Gotcha, I renamed it to `_functional`.

openmdao/core/problem.py
dingraha: I moved the class definition to a new file `core/functional_interface.py`.

openmdao/core/problem.py
dingraha: Sounds good, I added a `return_index_map` argument to `create_input_vector`, `create_output_vector`, and `create_jacobian_matrix`. A few examples in the docs show how it's used.

openmdao/docs/openmdao_book/features/experimental/functional_interface.ipynb
dingraha: Claude and I added a couple examples to the docs: one that re-implements one of the examples from the `AnalysisDriver` docs using the functional interface, and another that minimizes the famous paraboloid example using `scipy.optimize.minimize` directly.

openmdao/core/problem.py
dingraha: Ah, good point. This example is in the docs anyway, so I'll remove it.

openmdao/core/functional_callback.py
dingraha: Hadn't thought about that. I did some googling and couldn't find a clear answer either way. I don't have a strong opinion---I was just trying to communicate that the user isn't intended to construct a `_FunctionalCallback` directly, and instead receive one from `Problem.get_callback`.

openmdao/core/group.py
dingraha: I've added a check in the `_FunctionalInterface` class that will throw a `ValueError` if a user tries to create a functional interface object for a `Problem` that doesn't have a driver *and* doesn't specify either an input or output variable, which should avoid those runtime errors from ever being hit. On the other hand `Group._get_totals_of_metadata` and `Group._get_totals_wrt_metadata` throw those `RuntimeError`s just like `Group._get_totals_metadata` (which this PR doesn't change), so I thought it was fair game to follow the `Group._get_totals_metadata` example. But we could remove the `RuntimeError`s in `_get_totals_of_metadata` and `_get_totals_wrt_metadata` since they're only used for the functional interface stuff.

openmdao/core/total_jac.py
dingraha: Removed.

Kenneth-T-Moore (approved):

dingraha:
I think this is ready to go now. I did have to make a few adjustments to fix the doc build errors.

robfalck (approved):

PR #3718, opened 2026-02-26
Add missing return to `get_subjac_sparsity`

### LINKED ISSUE #3717: The "get_subjac_sparsity" method on the Coloring class doesn't return anything
### Description

The [`get_subjac_sparsity`](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/utils/coloring.py#L1912) method on the `Coloring` class is missing a return. It builds the sparsity structure, but then doesn't do anything with it which means that it will always be returning `None`. This method is only used in the [pyOptSparseDriver driver](https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/drivers/pyoptsparse_driver.py#L932) to extract sparsity information for the optimizer. But since this method will always return `None`, the driver interprets that as there being no total sparsity and none will be used in setting up the nonlinear constraints.

### Example

You should be able to see this for any problem that has sparsity via coloring which runs with the pyoptsparse driver. If you check the `_con_subjacs` attribute on the driver, it will always be empty even if there is sparsity.

```python
p = om.Problem()

exec = om.ExecComp(['y = a*x**2',
                    'z = a + x**2'],
                    a={'shape': (1,)},
                    y={'shape': (101,)},
                    x={'shape': (101,)},
                    z={'shape': (101,)})

p.model.add_subsystem('exec', exec)

p.model.add_design_var('exec.a', lower=-1000, upper=1000)
p.model.add_objective('exec.y', index=50)
p.model.add_constraint('exec.z', indices=[0], equals=25)
p.model.add_constraint('exec.z', indices=[-1], lower=20, alias="ALIAS_TEST")

p.driver = om.pyOptSparseDriver()
p.driver.options['optimizer'] = "SLSQP"

p.driver.declare_coloring()

p.setup(mode='rev')

p.set_val('exec.x', np.linspace(-10, 10, 101))

p.run_model()
p.run_driver()

print(f'\n\n{p.driver._con_subjacs}')
```

### OpenMDAO Version

3.42.1-dev

_No response_

rob-hetterich:
### Summary

The `get_subjac_sparsity` method on the `Coloring` class was missing a return which caused it to always return None. This means that that the `pyOptSparseDriver` always thought that there was no sparsity in the optimization problem.

### Related Issues

- Resolves #3717 

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/tests/test_pyoptsparse_driver.py
+++ b/openmdao/drivers/tests/test_pyoptsparse_driver.py
@@ -2758,6 +2758,80 @@ class TestPyoptSparse(unittest.TestCase):
         assert_near_equal(J[('exec.z', 'exec.a')].flatten(), np.array([1.]))
         assert_near_equal(J[('ALIAS_TEST', 'exec.a')].flatten(), np.array([1.]))
 
+    @require_pyoptsparse('IPOPT')
+    def test_driver_sparsity(self):
+        prob = om.Problem()
+        prob.driver = driver = om.pyOptSparseDriver(optimizer='IPOPT', print_results=False)
+        driver.opt_settings['print_level'] = 0
+        driver.opt_settings['max_iter'] = 1000
+        driver.declare_coloring()
+        model = prob.model
+
+        shape = (1, 5)
+        ivc = model.add_subsystem('ivc', om.IndepVarComp())
+        ivc.add_output('x', np.ones(shape))
+        ivc.add_output('y', np.ones(shape))
+
+        model.add_subsystem(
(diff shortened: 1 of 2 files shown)


robfalck (approved):

Kenneth-T-Moore (changes requested):
Fix looks good. I think we talked about testing this by parsing the pyoptsparse or snopt output to look for the number of jacs defined.

robfalck (approved):

Kenneth-T-Moore (approved):

PR #3719, opened 2026-02-27
Autoscaler implementation

### LINKED ISSUE #3697: Implement AutoScaler class

[POEM 094](https://github.com/OpenMDAO/POEMs/blob/master/POEM_094.md#autoscale-api-) Lays out an API for an Autoscaler class to automatically scale design variables and responses for an optimization problem.

The implementation should include DefaultAutoscaler as defined in the POEM, others (HessianAutoscaler) will follow later.

### Associated POEM

094

robfalck:
### Summary

This implements an Autoscaler class, though the implementation has changed compared to the notional one in [POEM 094](https://github.com/OpenMDAO/POEMs/blob/master/POEM_094.md)

This is a substantial change to the way OpenMDAO scales and unscales data between the optmizer space and the model/physical space.

#### OptimizerVector

To allow for scaling techniques that assume the design variables, constraints, and objectives are vectors, such as applying a linear operator (`x_scaled = M @ x_model`), this PR introduces `OptimizerVector`.  `OptimizerVector` will handle the getting and setting of design variables, constraints, and objectives from and to the model, optionally applying the autoscaling to them.

Important methods and attributes of `OptimizerVector` are
- `asarray()` : returns a view into the contiguous array of variables of interest
-  `_to_dict()` : return a dictionary of VOI names and values, optionally filtering for constraint types, and optionally returning remote or local values.
- `driver_scaled` : True if the underlying data is in a driver/optimizer scaled state.

#### Autoscaler

The `Autoscaler` class is designed to provide an efficient way to transform the variables of interest.

The default `Autoscaler` class performs the same block-level affine scaling that OpenMDAO currently performs via scaler, adder, ref0, and ref.

Important methods for the `Autoscaler` are:

- setup(): Called during driver setup.
- update(): Called during initial driver execution after the model has been evaluated.
- get_bounds_scaling(): Returns OptimizerVectors of the element-wise scaling of the bounds (lower, upper, equals). For design variables, None is returned for equals.
- apply_design_var_unscaling(): Unscale the given design variable vector from the driver space to the model space.
- apply_design_var_scaling(): Scale the given design variable vector from the model space to the optimizer space.
- apply_constraint_scaling(): Scale the given constraint vector from the model space to the optimizer space.
- apply_objective_scaling(): Scale the given objective vector from the model space to the optimizer space.
- apply_mult_unscaling(): Used to unscale the Lagrange multipliers which are naturally computed in the optimizer space.
- apply_jac_scaling(): Used to scale a dictionary jacobian from the model space (via compute totals) to the optimizer space.

**Important**: OpenMDAO uses the sign of the objective function to determine whether the optimizer is minimizing or maximizing. When implementing automatic scaling algorithms, one should make sure they don't accidentally flip the scaling.

### Related Issues

- Resolves #3697 

While the default `Autoscaler` uses VOI keys `total_scaler` and `total_adder` to perform affine scaling, these no longer include the affects of unit scaling/unscaling.  Instead, the driver method `_get_voi_val` now accepts a `driver_units` argument that will first convert model outputs to the units specified in `add_design_var`, `add_constraint`, and `add_objective` (if any).

Methods `get_design_var_values`, `get_constraint_values`, and `get_objective_values` values now always return the given VOI in the units specified in `add_design_var`, `add_constraint`, and `add_objective`.

In case recording, we no longer necessarily have access to the logic of the autoscaler, and to avoid saving both model and driver-scaled values to the recording, **the `Case` methods `get_design_vars`, `get_constraints`, `get_objectives`, and `get_responses` will now return model-scaled values but convert them to the driver-specified units by default.**

### New Dependencies

None


openmdao/drivers/autoscalers/__init__.py
```diff
@@ -0,0 +1 @@
+from .autoscaler import Autoscaler
```
Kenneth-T-Moore: I don't think we usually import anything in our __init__ files in OpenMDAO.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,665 @@
...
+        if con_multipliers:
+            for name, mult in con_multipliers.items():
+                # Get the constraint scaler from cached combined scalers
+                scaler = self._var_meta['constraint'][name]['total_scaler'] or 1.0
+                mult *= scaler / obj_scaler
+
+        return desvar_multipliers, con_multipliers
+
+    def apply_jac_scaling(self, jac_dict):
+        """
```
Kenneth-T-Moore, line 512: We kind of have a precedent that "required" methods will raise an error if called on a base class without being specified in the derived class. Do we want to do that here? If we do, it'll require implementing the default autoscaler as a derived class, and raising the errors in the base.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,665 @@
...
+
+    setup(driver)
+        Initialize the autoscaler with driver metadata.
+        Called once at the start of the driver.run method.
+    
+    setup_requires_run_model()
+        Return True if setup of the autoscaler requires the
+        model to be in an executed state.
+
+    apply_vec_scaling(vec)
```
Kenneth-T-Moore: These methods (apply_vec_scaling, apply_vec_unscaling) seem to be private methods in the code and the tests.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,665 @@
...
+                    else:
+                        # Unknown input, skip scaling this entry
+                        continue
+
+                    # Scale the Jacobian block in-place: J_scaled = J_model * out_scaler / in_scaler
+                    # Use in-place operations to preserve view relationship with underlying array
+                    if out_scaler is not None:
+                        block[...] = (out_scaler * block.T).T
+                    if in_scaler is not None:
+                        block *= 1.0 / in_scaler
```
Kenneth-T-Moore, line 572: Thinking ahead, the linear operator scaler is going to need an inverse linear operator for the `/ in_scaler` counterpart. I wonder if we should provide that even in this simple scaler.

in_scaler_inv = self._var_meta['design_var'][in_name]['inverse_total_scaler']
block *=  in_scaler_inv

openmdao/drivers/pyoptsparse_driver.py
```diff
@@ -711,11 +723,8 @@ def _objfunc(self, dv_dict):
             signal.signal(sigusr, self._signal_handler)
 
         try:
-            for name in self._designvars:
-                self.set_design_var(name, dv_dict[name])
-
-            # print("Setting DV")
-            # print(dv_dict)
+            dv_vec.set_data(dv_dict, driver_scaling=True)
```
Kenneth-T-Moore, line 719: Could this be a one-step process instead of two?

openmdao/drivers/scipy_optimizer.py
```diff
@@ -61,11 +61,13 @@
 # With "old-style" a bound is a tuple, with "new-style" a Bounds instance
 # In principle now everything can work with "old-style"
 # These settings have no effect to the optimizers implemented before SciPy 1.1
-_supports_new_style = {'trust-constr'}
-if Version(scipy_version) >= Version("1.4"):
-    _supports_new_style.add('differential_evolution')
+_supports_new_style = {'trust-constr', 'differential_evolution'}
+# if Version(scipy_version) >= Version("1.4"):
```
Kenneth-T-Moore: Might as well delete the commented scipy 1.4 lines.

openmdao/recorders/tests/test_sqlite_recorder.py
```diff
@@ -2764,13 +2764,8 @@ def test_simple_paraboloid_scaled_desvars(self):
...
         unscaled_y = des_vars['y'][0]
 
-        des_vars = last_case.get_design_vars(scaled=True)
-        scaled_x = des_vars['x'][0]
-        scaled_y = des_vars['y'][0]
-
-        adder, scaler = determine_adder_scaler(ref0, ref, None, None)
-        self.assertAlmostEqual((unscaled_x + adder) * scaler, scaled_x, places=12)
-        self.assertAlmostEqual((unscaled_y + adder) * scaler, scaled_y, places=12)
+        self.assertAlmostEqual(unscaled_x, prob.get_val('x'), places=12)
```
Kenneth-T-Moore, line 2767: Probably rename this test from "...scaled_desvars" to "...unscaled_desvars"

openmdao/core/tests/test_problem.py
```diff
@@ -2249,13 +2250,14 @@ def test_list_driver_vars_driver_scaling_bounds(self):
             sys.stdout = stdout
         output = strout.getvalue().split('\n')
 
-        self.assertTrue('35.'  in output[5])  # design var: x, value
+        # Not outputs are return
+        self.assertTrue('1.6'  in output[5])  # design var: x, value
```
Kenneth-T-Moore, line 2254: This test highlights the big change, where driver_vars are always displayed in the driver units, not the model unints.

openmdao/core/tests/test_problem.py
```diff
@@ -2474,11 +2477,16 @@ def test_find_feasible(self):
...
+                                    if method=='dogbox' and list(term_tol.keys())[0] in {'xtol', 'ftol'}:
+                                        # for dogbox we cant reliably determine which tolerance triggers
+                                        # the stopping condition
+                                        pass
+                                    else:
+                                        self.assertIn(list(term_tol.keys())[0], prob.driver.result.exit_status)
                                 else:
                                     self.assertFalse(failed)
                                     self.assertTrue(prob.driver.result.success)
-                                    assert_near_equal(prob.get_val('x'), 1.5, tolerance=1.0E-8)
```
Kenneth-T-Moore, line 2481: I am curious what is different that causes this change to be needed. This problem only has some refs and no units.  Should probably look into this a little and make sure something isn't going on, maybe due to bounds being scaled differently (standardized max and min INF), or due to something that isn't being reset due to setup only being run once.

openmdao/core/tests/test_system_set_solver_bounds_scaling_options.py
```diff
@@ -305,6 +305,10 @@ def test_set_design_var_options_scaler_adder_ref_ref0(self):
         prob.final_setup()
         des_vars_using_add_design_var = prob.model.get_design_vars()
 
+        from pprint import pprint
```
Kenneth-T-Moore: Debug to remove.

openmdao/core/tests/test_system_set_solver_bounds_scaling_options.py
```diff
@@ -316,6 +320,7 @@ def test_set_design_var_options_scaler_adder_ref_ref0(self):
         prob.final_setup()
         des_vars_using_set_design_var_options = prob.model.get_design_vars()
 
+        self.maxDiff = None
```
Kenneth-T-Moore: This (and the one at line 624) don't seem to be used.

openmdao/core/driver.py
```diff
@@ -265,6 +268,10 @@ class Driver(object, metaclass=DriverMetaclass):
         Variables to record based on recording options.
     _in_find_feasible : bool
         True if the driver is currently executing find_feasible.
+    _vectors : dict
```
Kenneth-T-Moore, line 271: Related to earlier comment, but I wonder if the autoscaler should just own the _vector.

openmdao/core/driver.py
```diff
@@ -591,6 +664,45 @@ def _setup_driver(self, problem):
                     issue_warning("Derivatives are turned off.  Skipping simul deriv coloring.",
                                   category=DerivativesWarning)
 
+        self._autoscaler.setup(driver=self)
```
Kenneth-T-Moore: Does autoscaler.setup need to be called twice? It gets called here for all drivers, but then we also call it during `run` in the base driver, pyoptsparse, and scipyopt.

Kenneth-T-Moore (changes requested):
See comments.

openmdao/drivers/autoscalers/autoscaler.py
robfalck: Yeah I debated whether having AutoscalerBase and Autoscaler (DefaultAutoscaler?) was necessary. I can go back to that separation.

openmdao/core/driver.py
robfalck: Also went back and forth on this. In my mind, the vectors are an intermediate space owned by the driver via which it interacts with the optimizer and the model.

openmdao/core/driver.py
robfalck: I'm working on making it so it can be called once.  Originally I had a two step process because some autoscalers will need to be setup after the model has been executed (if they need information from the execution to setup the scaling).

I think theres a cleaner way around it though.

openmdao/core/tests/test_problem.py
robfalck: I suspect its something like that.  The numerical noise is just a touch higher now.

openmdao/drivers/autoscalers/autoscaler.py
robfalck: For this I'd like to see what exactly we need to implement the NRA work. I'm sure there will be other changes needed.

openmdao/core/driver.py
robfalck: OK its now called once. One small hiccup was that _find_feasible, which almost acts like another separate driver, also needed to call setup.  But I think its cleaner now.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,665 @@
...
+    
+    setup_requires_run_model()
+        Return True if setup of the autoscaler requires the
+        model to be in an executed state.
+
+    apply_vec_scaling(vec)
+        Scale a vector from model space to optimizer space.
+        Modifies vec in-place.
+
+    apply_vec_unscaling(vec, name)
```
Kenneth-T-Moore: These two api functions are wrong (`apply_vec_unscaling`, `apply_vec_scaling`), as they don't seem to exist. I think you meant to put these ones into the class docstring instead: [apply_design_var_scaling, apply_design_var_unscaling, apply_constraint_scaling, apply_constraint_unscaling]

robfalck:
Made a separate AutoscalerBase class, with Autoscaler  class being the default implementation. 
Fixed docstring issues.
Moved autoscaler setup call into driver setup so that it happens in final_setup rather than waiting for run. Future autoscalers that require the model to be in an executed state will have to call run_solve_nonlinear in their setup (so within final_setup).

openmdao/core/driver.py
```diff
@@ -858,6 +957,9 @@ def _get_voi_val(self, name, meta, remote_vois, driver_scaling=True,
             of the value that's on the current process for a distributed variable.
         rank : int or None
             If not None, gather value to this rank only.
+        driver_units : bool
+            If True, return the given variable interest in the units specified for the
```
naylor-b: 'variable interest'  --> 'variable of interest'

openmdao/core/total_jac.py
```diff
@@ -1363,6 +1372,69 @@ def directional_jac_setter(self, inds, mode, meta):
...
+            return
+            
+        is_flat = isinstance(next(iter(jac_dict)), tuple)
+
+        if is_flat:
+            for (out_name, in_name), block in jac_dict.items():
+                # Apply row scaling if the output is unit-active
+                out_scaler = self._resp_unit_scalers.get(out_name)
+                if out_scaler:
+                    block[...] = (out_scaler * block.T).T
```
naylor-b: Aren't in_scaler and out_scaler always just floats, and if so, why are we doing a transpose here?  Couldn't we just combine in_scaler and out_scaler into a single number and do a single 'block *= final_scaler' multiply?

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,599 @@
...
+                val_arr[finite] *= scaler if np.isscalar(scaler) else np.asarray(scaler)[finite]
+
+        # Restore sentinel for unbounded elements (scaling may have perturbed them)
+        val_arr[inf_mask] = -INF_BOUND if is_lower else INF_BOUND
+
+        val_arr = val_arr.ravel()
+
+        return val_arr
+
+    def _compute_scaled_bounds(self, voi_type):
```
naylor-b, line 200: There's a version of this in autoscaler_base.py that looks identical.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,599 @@
...
+    def apply_objective_scaling(self, vec: 'OptimizerVector'):
+        """
+        Scale the objective from the model space to the optimizer space.
+
+        Notes
+        -----
+        Use caution in the definition of this method. OpenMDAO **always** minimizes
+        the objective, and negates the sign of the objective when maximizing (generally
+        by setting scaler or ref to a negative value). If your implementation changes
+        the sign of the objective, you max accidentally change an objective minimization
```
naylor-b: max --> may

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,599 @@
...
+                    # Determine input scaler
+                    if in_name in self._var_meta['design_var']:
+                        in_scaler = self._var_meta['design_var'][in_name]['total_scaler']
+                    else:
+                        # Unknown input, skip scaling this entry
+                        continue
+
+                    # Scale the Jacobian block in-place: J_scaled = J_model * out_scaler / in_scaler
+                    # Use in-place operations to preserve view relationship with underlying array
+                    if out_scaler is not None:
```
naylor-b, line 569: Not sure how common it is, but might be worth a check to see if out_scaler is a scalar and just do an in-place mult.

openmdao/vectors/optimizer_vector.py
```diff
@@ -0,0 +1,642 @@
...
+        Parameters
+        ----------
+        val : ArrayLike or dict[str, ArrayLike]
+            Values to which the entire internal vector should be set. May be given
+            as an array-like value or as a dict (as in pyoptsparse) that maps
+            names to associated values.
+        driver_scaling : bool
+            If True, set the
+        order : str
+            The order in which val is flattened, as accepted by
```
naylor-b: Missing the end of the description for order?

openmdao/vectors/optimizer_vector.py
```diff
@@ -0,0 +1,642 @@
...
+
+        # Check cache
+        cache_key = self._make_filter_key(**kwargs)
+        if cache_key not in self._filters:
+            self._filters[cache_key] = self._compute_filter_indices(kwargs)
+
+        indices = self._filters[cache_key]
+        return self._data[indices]
+
+    def _from_dict(self, var_dict):
```
naylor-b: maybe name this _update_from_dict?

openmdao/vectors/optimizer_vector.py
```diff
@@ -0,0 +1,642 @@
...
+            value = np.asarray(var_dict[name]).ravel()
+            expected_size = meta['end_idx'] - meta['start_idx']
+
+            if value.size != expected_size:
+                raise ValueError(
+                    f"Size mismatch for variable '{name}': expected {expected_size}, "
+                    f"got {value.size}"
+                )
+
+            self._data[meta['start_idx']:meta['end_idx']] = value
```
naylor-b: I would suggest storing a 'slice' entry in meta that contains a slice(start, end) object instead of two separate 'start_idx' and 'end_idx' entries because it'll save you a dict lookup each time you access a slice of the data array.

naylor-b (changes requested):

openmdao/drivers/autoscalers/autoscaler.py
robfalck: For now I'm not going to worry about this, transposes are extremely cheap and i doubt we'll save any considerable time here in most use-cases.

openmdao/core/tests/test_system_set_solver_bounds_scaling_options.py
```diff
@@ -616,6 +616,7 @@ def test_set_constraints_options_scaler_adder_ref_ref0(self):
         prob.final_setup()
         constraints_using_set_constraint_options = prob.model.get_constraints()
 
+        self.maxDiff=None
```
Kenneth-T-Moore, line 619: This doesn't seem to be used.

robfalck: It's used in the assertEqual on the following line, so that you can see any differences rather than it truncating them.

Kenneth-T-Moore: oh, this is `unittest` API attributes.

openmdao/core/system.py
```diff
@@ -1197,109 +1175,99 @@ def set_constraint_options(self, name, ref=_UNDEFINED, ref0=_UNDEFINED,
...
                 scaler = None
-        elif scaler == 1.0:
-            scaler = None
 
-        if isinstance(adder, np.ndarray):
-            if not np.any(adder):
+            if isinstance(adder, np.ndarray):
+                if not np.any(adder):
+                    adder = None
+            elif adder == 0.0 if adder is not None else False:
```
Kenneth-T-Moore, line 1266: `elif adder is not None and adder==0.0` seems cleaner

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def __init__(self):
+        self._var_meta : dict[str, dict[str, dict]] = {}
+        self._has_scaling : bool = False
+        self._scaled_lower = {}
+        self._scaled_upper = {}
+        self._scaled_equals = {}
+        self._driver_ref = None
+
+    def _get_inst_id(self):
```
Kenneth-T-Moore, line 30: I think this is identical to the base class method.

openmdao/vectors/optimizer_vector.py
```diff
@@ -0,0 +1,569 @@
...
+        'slice' and 'size' keys.
+    _filters : dict[tuple, ndarray]
+        Cache for computed filter indices. Keys are tuples of sorted (key, value)
+        filter criteria, values are integer arrays of indices into _data.
+        Used by asarray() to avoid recomputing indices for repeated filter calls.
+    _dist_driver_vars : dict[str, tuple]
+        Mapping of distributed driver variables to (local_indices, sizes, _)
+
+    Examples
+    --------
```
Kenneth-T-Moore, line 49: Bad thing about examples in the docstring is they are untested.

Kenneth-T-Moore (approved):
Approved, though see comments.

openmdao/drivers/autoscalers/autoscaler.py
```diff
@@ -20,32 +20,7 @@ class Autoscaler(AutoscalerBase):
...
-            A string of the form '<driver_inst_id>.autoscaler', or None if setup has not
-            been called yet.
-        """
-        if self._driver_ref is None:
-            return None
-        driver = self._driver_ref()
-        if driver is None:
-            return None
-        return f'{driver._get_inst_id()}.autoscaler'
+        super().__init__()
```
naylor-b: If __init__ does nothing but call super().__init__ then you don't need to define __init__ at all.

naylor-b (approved):

PR #3736, opened 2026-03-19
Fixed a bug in relative step-size calculation for parallel fd.

### LINKED ISSUE #3708: Parallel finite differences with relative element step size results in IndexError
### Error Message
When setting up an optimization problem that requires parallel finite differences, setting the "step_calc" and "step" inputs to the "model.approx_totals" method results in the following IndexError:

```
File "/Users/sandleda/opt/Miniconda/4.12.0/DS_Build/lib/python3.9/site-packages/openmdao/approximation_schemes/finite_difference.py", line 333, in _run_point
    results_array *= current_coeff[idxs - idx_start[0]]
IndexError: index 6 is out of bounds for axis 0 with size 1
```

### Example Code Description
The reproducing code provided below is a lightly modified version of the example script from the [Speeding up Derivative Approximations with Parallel Finite Difference](https://openmdao.org/newdocs/versions/latest/features/core_features/working_with_derivatives/parallel_fd.html) page in the documentation. The changes made are as follows:

1. A second input to the MatMultComp explicit component ("z") has been added.
2. A step size of 1E-3 and a step_calc method of "rel_element" has been added to the declare_partials line in the component.
3. A step size of 1E-3 and a step_calc method of "rel_element" has been added to the approx_totals method applied to the model.
4. The compute_totals call has been modified so that the "wrt" argument includes "comp.z" followed by "comp.x".

### Conditions and Observations
I originally tried to set the relative element step size with just the options in "declare_partials" within the component. However, when this is combined with the defaults for "model.approx_totals" using the following line, the step sizes used seem to be absolute and have a value of 1E-6:
`model.approx_totals(method='fd')`
I'm not sure if there is some other way to set a relative element step size on the component itself that is not overridden.

I then set the step size and step_calc at the top level with:
`model.approx_totals(method='fd',step=1E-3,step_calc='rel_element')`
With a single input (the original problem's "x"), this works. However, adding a second input ("z") results in the error given above. 

This only happens when the order that the inputs are added to the component does not match the order that the inputs are listed in the "wrt" argument of "compute_totals". For example, adding "x" and then "z" as inputs to the MatMultComp component
and requesting totals with respect to ['comp.z', 'comp.x'] will fail. Similarly, computing totals with respect to only ['comp.z'] also fails. Computing totals with respect to ['comp.x', comp.z'] will work however.

This problem occurs regardless of whether the script is run in serial or parallel. It does not seem to occur when the step size and step_calc are omitted from "model.approx_totals", but I am then left with the default absolute step size. Finally, although there is the simple workaround of adding the inputs in the correct order, there does not seem to be a way to avoid the error when running an optimization with run_driver() which is the end goal for my problem.

### Example

import time
import openmdao.api as om
import numpy as np

class MatMultComp(om.ExplicitComponent):
    def __init__(self, mat, approx_method='exact', sleep_time=0.1, **kwargs):
        super().__init__(**kwargs)
        self.mat = mat
        self.approx_method = approx_method
        self.sleep_time = sleep_time

    def setup(self):
        self.add_input('x', val=np.ones(self.mat.shape[1]))
        self.add_input('z', val=1.0)
        self.add_output('y', val=np.zeros(self.mat.shape[0]))
        self.num_computes = 0

    def setup_partials(self):
        self.declare_partials('*', '*', method=self.approx_method, step=1E-3, step_calc='rel_element')

        print(f"x: {inputs['x']}, z: {inputs['z']}")
        outputs['y'] = self.mat.dot(inputs['x'])
        self.num_computes += 1
        time.sleep(self.sleep_time)

mat = np.arange(30, dtype=float).reshape(5, 6)

p = om.Problem(model=om.Group(num_par_fd=7))
model = p.model
model.approx_totals(method='fd',step=1E-3,step_calc='rel_element')
#model.approx_totals(method='fd')
comp = model.add_subsystem('comp', MatMultComp(mat, approx_method='fd'))

model.set_input_defaults('comp.x', val=np.ones(mat.shape[1]))
p.setup(mode='fwd')
p.run_model()

pre_count = comp.num_computes

#J = p.compute_totals(of=['comp.y'], wrt=['comp.x','comp.z'], return_format='array')
J = p.compute_totals(of=['comp.y'], wrt=['comp.z','comp.x'], return_format='array')
#J = p.compute_totals(of=['comp.y'], wrt=['comp.z'], return_format='array')

post_count =  comp.num_computes

# how many computes were used in this proc to compute the total jacobian?
# Each proc should be doing 2 computes.
jac_count = post_count - pre_count

print(jac_count)

### OpenMDAO Version

3.39.0

numpy==1.26.4
scipy==1.13.1

robfalck:
### Summary

The bug occurred in openmdao/approximation_schemes/finite_difference.py when using step_calc='rel_element' with multiple input variables in parallel finite differences.
    
This involved fixes in FiniteDifference._run_point():
- When multiplying results by current_coeff, compute the correct local index by finding the minimum index and subtracting it.
- When multiplying results by coeff in the FD loop, apply the same local index computation.
- When applying perturbations with delta, compute the correct local index and handle both scalar and array cases properly.

### Related Issues

- Resolves #3708 

None

### New Dependencies

None

### DIFF
--- a/openmdao/approximation_schemes/finite_difference.py
+++ b/openmdao/approximation_schemes/finite_difference.py
@@ -6,12 +6,6 @@ import numpy as np
 from openmdao.approximation_schemes.approximation_scheme import ApproximationScheme, _is_group
 
 
-# For rel_element step sizing, we need to track local indices for each idx_info tuple.
-# Local indices are computed once (either per iteration in the generator or upfront for colored)
-# and stored here to avoid repeatedly computing them in _run_point and _run_sub_point.
-VecIdxInfo = namedtuple('IdxInfo', ['vec', 'idxs', 'local_idxs'])
-
-
 DEFAULT_ORDER = {
     'forward': 1,
     'backward': 1,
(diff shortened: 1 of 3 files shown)


openmdao/approximation_schemes/finite_difference.py
```diff
@@ -328,8 +328,14 @@ def _run_point(self, system, idx_info, data, results_array, total, idx_range=ran
...
-
-                        results_array *= current_coeff[idxs - idx_range[0]]
+                        # For rel_element, current_coeff is per-element for the wrt variable.
+                        # idxs contains indices that may be global or relative to the wrt variable.
+                        # We need to compute the correct offset to get local indices [0, 1, 2, ...].
+                        idxs_arr = np.atleast_1d(idxs)
+                        # The starting index for this variable is the minimum index we see
+                        var_start_idx = int(idxs_arr.min())
+                        local_idxs = idxs_arr - var_start_idx
+                        results_array *= current_coeff[local_idxs]
```
Kenneth-T-Moore: Could combine these 2 lines (all 3 instances)

openmdao/approximation_schemes/finite_difference.py
```diff
@@ -328,8 +328,14 @@ def _run_point(self, system, idx_info, data, results_array, total, idx_range=ran
 
                 for vec, idxs in idx_info:
                     if vec is not None and idxs is not None:
-
-                        results_array *= current_coeff[idxs - idx_range[0]]
+                        # For rel_element, current_coeff is per-element for the wrt variable.
+                        # idxs contains indices that may be global or relative to the wrt variable.
+                        # We need to compute the correct offset to get local indices [0, 1, 2, ...].
+                        idxs_arr = np.atleast_1d(idxs)
```
naylor-b: Instead of doing these array operations every time inside of these loops, could we instead compute the local indices once in init_approximations/init_colored_approximations so the 'idxs' that we loop over are the correct ones to begin with?

Kenneth-T-Moore:
Superseded by #3741

PR #3749, opened 2026-04-24
Create pymooDriver

### LINKED ISSUE #3735: Support pymoo for gradient-free optimization.

# Summary 
Currently, OpenMDAO maintains internal implementations for Differential Evolution (DifferentialEvolutionDriver) and Genetic Algorithms (GeneticAlgorithmDriver). While these serve as entry-level heuristic solvers, [pymoo](https://pymoo.org) is essentially the industry standard that provides many more algorithms and algorithms that are more capable.

# Proposal
- Implement a PymooDriver
- Deprecate or remove GeneticAlgoirthmDriver and DifferentialEvolutionDriver

# Notes
- Those pymoo algorithms which support constraints do so with a one-sided model (g <= 0). We'll need to take care in the implementation of equality constraints.
- Pymoo has some support of gradients so we should provide those when necessary.

### Associated POEM

_No response_

rob-hetterich:
### Summary

Wrapped the [pymoo](https://pymoo.org/) package in an OpenMDAO Driver.

### Related Issues

- Resolves #3735 

None

### New Dependencies

pymoo


openmdao/drivers/pymoo_driver.py
```diff
@@ -0,0 +1,1032 @@
...
+
+        # What we don't support
+        self.supports['active_set'] = False
+        self.supports['distributed_design_vars'] = False
+        self.supports['gradients'] = False
+        self.supports._read_only = True
+
+        # The user places algorithm-specific settings in here that are passed
+        # into the algorithm instantiation.
+        self.alg_settings = {}
```
Kenneth-T-Moore, line 598: In the scipy and pyoptsparse drivers, we use a dictionary called `opt_settings` to pass in algorithm specific settings. I wonder if we should use the same name here.

EDIT: I see why you did this, because there is a `run_settings` that gets passed in differently. The user does need to know which option goes where, so I guess they have to be more familiar with the pymoo api.

rob-hetterich: yeah, pymoo has two different points where you pass in settings 1) When you instantiate the algorithm 2) when the algorithm is actually setup and run. I didn't name it `opt_settings` to kind of differentiate it, but I don't mind changing `alg_settings` to `opt_settings` if that is preferred. It does require knowledge of pymoo and which settings go where, but I also try to explain it a bit in  the documentation and link to where you can find some of that info (though it's not all well documented in the pymoo docs)

Kenneth-T-Moore: It looks like the set of run_settings is pretty small, they are potentially optimizer-agnostic, and half of them are taken care of by openmdao, so maybe we should go with opt_settings for the alg_settings.

robfalck (approved):

openmdao/drivers/pymoo_driver.py
robfalck: That's a good point

Kenneth-T-Moore (approved):

openmdao/drivers/pymoo_driver.py
rob-hetterich: Note that we discussed this offline and agreed to leave the settings interface as is due to how pymoo's API is setup.

PR #3751, opened 2026-04-28
Fixed System._call_user_function to handle OutOfBoundsError

### LINKED ISSUE #3750: Calls to compute which hit an OutOfBoundsError results in a secondary error
### Description

OutOfBoundsError is constructed using multiple arguments rather than a single string.

If `System._call_user_function` hits an out of bounds error, it attempts to raise a new one but does so using a single string argument, resulting in a secondary and unuseful error being raised.

The fix is to have `_call_user_function` raise a RuntimeError that embeds the original error message.

### Example

```python
"""
Test that OutOfBoundsError raised inside a user compute() function is reported
clearly rather than being masked by a secondary TypeError.

Background
----------
`System._call_user_function` (openmdao/core/system.py) wraps user-defined
callbacks (compute, apply_nonlinear, etc.) and re-raises any exception as::

    raise err_type(f"{msginfo}: Error calling {fname}(), {err}")

For most exception types this is fine because they accept a single string
argument.  `OutOfBoundsError`, however, requires five positional arguments::

    OutOfBoundsError(message, idx, value, lower, upper)

Attempting to construct it with only a formatted message string raises a
secondary ``TypeError: OutOfBoundsError.__init__() missing 4 required
positional arguments``.  This TypeError then masks the original
OutOfBoundsError and provides no useful diagnostic information to the user.
"""

import unittest

import openmdao.api as om
from openmdao.components.interp_util.outofbounds_error import OutOfBoundsError
from openmdao.utils.assert_utils import assert_warning
from openmdao.utils.testing_utils import use_tempdirs


class RaisesOutOfBoundsComp(om.ExplicitComponent):
    """
    Minimal component that intentionally raises OutOfBoundsError in compute().

    This simulates what an interpolation component does when an input value
    falls outside the tabulated domain.
    """

    def setup(self):
        self.add_input('x', val=0.5)
        self.add_output('y', val=1.0)

        x = float(inputs['x'].flat[0])
        lower, upper = 0.0, 1.0
        if x < lower or x > upper:
            raise OutOfBoundsError(
                f"Input x={x} is outside the interpolation range [{lower}, {upper}].",
                idx=0,
                value=x,
                lower=lower,
                upper=upper,
            )
        outputs['y'] = x ** 2


@use_tempdirs
class TestCallUserFunctionOutOfBounds(unittest.TestCase):
    """
    Verify that OutOfBoundsError raised in a component compute() surfaces as a
    clear RuntimeError (or OutOfBoundsError) rather than a secondary TypeError.

    In OpenMDAO <= 3.43.0, this was raising the following:

        TypeError: OutOfBoundsError.__init__() missing 4 required positional
        arguments: 'idx', 'value', 'lower', and 'upper'
    """

    def test_outofbounds_not_masked_by_typeerror(self):
        """OutOfBoundsError in compute() must not be hidden by a secondary TypeError."""
        p = om.Problem()
        p.model.add_subsystem('comp', RaisesOutOfBoundsComp())
        p.setup()
        p.set_val('comp.x', 2.0)  # outside [0, 1] — will trigger OutOfBoundsError

        # Before the fix this raises:
        #   TypeError: OutOfBoundsError.__init__() missing 4 required positional arguments
        #
        # After the fix it should raise RuntimeError (or OutOfBoundsError) with
        # a message that mentions both the component and the original error.
        with self.assertRaises((RuntimeError, OutOfBoundsError)) as ctx:
            p.run_model()

        # The error message must contain the component path so the user can
        # identify which component is out of bounds.
        self.assertIn('comp', str(ctx.exception))

        # The error message must also carry the original out-of-bounds description.
        self.assertIn('outside the interpolation range', str(ctx.exception))


if __name__ == '__main__':
    unittest.main()
```

### OpenMDAO Version

3.43.1

_No response_

robfalck:
### Summary

When `System._call_user_function` encounters an error, it attempts to reraise it, wrapping the message string.  However, OutOfBoundsError was not able to be instantiated with a single string argument. To make this robust since other errors may also be instantiated in non-standard ways, `System._call_user_function` now reraises the same exception after modifying the string message.

### Related Issues

- Resolves #3750 

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2954,8 +2954,9 @@ class System(object, metaclass=SystemMetaclass):
             if str(err).startswith(self.msginfo):
                 raise
             else:
-                err.add_note(f'{self.msginfo}: Encountered error calling {fname}')
-                raise
+                msg = err.args[0] if err.args else ''
+                err.args = (f"{self.msginfo}: Error calling {fname}(), " + msg,) + err.args[1:]
+                raise 
         finally:
             self._inputs.read_only = False
             self._outputs.read_only = False
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2950,19 +2950,13 @@ class System(object, metaclass=SystemMetaclass):
 
         try:
             yield
-        except Exception:
-            err_type, err, trace = sys.exc_info()
+        except Exception as err:
+            # err_type, err, trace = sys.exc_info()
             if str(err).startswith(self.msginfo):
                 raise
             else:
-                new_msg = f"{self.msginfo}: Error calling {fname}(), {err}"
-                try:
-                    new_err = err_type(new_msg)
-                except TypeError:
-                    # OutOfBoundsError requires idx, value, lower, upper).
-                    # Fall back to RuntimeError so the diagnostic message is not lost.
-                    raise RuntimeError(new_msg).with_traceback(trace) from None
-                raise new_err.with_traceback(trace)
+                err.add_note(f'{self.msginfo}: Encountered error calling {fname}')
+                raise
         finally:
             self._inputs.read_only = False
             self._outputs.read_only = False
--- a/openmdao/core/system.py
+++ b/openmdao/core/system.py
@@ -2951,7 +2951,6 @@ class System(object, metaclass=SystemMetaclass):
         try:
             yield
         except Exception as err:
-            # err_type, err, trace = sys.exc_info()
             if str(err).startswith(self.msginfo):
                 raise
             else:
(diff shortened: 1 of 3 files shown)


robfalck:
Looking at the availability of the `Exception.add_note` method in python >= 3.11, we might want to switch to that functionality here:

```python

try:
    raise ValueError("Invalid input")
except ValueError as e:
    # Adding context without losing the original exception
    e.add_note("Occurred during the user registration process.")
    e.add_note("Attempted with username: 'testuser123'")
    raise
```

robfalck:
It turns out using `add_note` wasn't a great fit here. We were testing a for error messages, and the notes are not part of the error message. Plus the interaction with warnings wasn't really what we wanted.

It was much simpler to just adjust the `args` of the caught exception and re-raise it.

naylor-b (approved):

rob-hetterich (approved):

PR #3760, opened 2026-05-15
Refactor matrices

rob-hetterich:
### Summary

Refactoring some of the matrix object to be more memory efficient and a bit faster. In some of my larger Dyreqt campaigns I found that these refactorings helped to reduce the memory usage by around 600 MB (or around 15%)

### Related Issues

- Resolves #

None

### New Dependencies

None


rob-hetterich:
Just to highlight the main changes in this PR and the reasoning:

1. **COOMatrix**
    - The crux of this was that in the original approach all of the subjacs were compiled into a list then concatenated. As a result, there's a temporary spike in memory usage as the concatenated data and list of subjacs exist at the same time before the `rows` and `cols` variables are written over. 
    - The new approach is more memory efficient as it preallocates the `rows` and `cols` variables, then fills in the subjacs and removes them from memory as it iterates. This way you never have all of the subjacs existing in memory at the same time as the filled in `rows` and `cols` variables.
    - This adjustment was the most minor as it just deals with a short, temporary spike in memory usage rather than persistent memory usage.

2. **CSRMatrix**
    - The crux of this was that in the original approach, the COO and CSR matrices are persistently left in memory at the same time. The class relied on the COO matrix getting updated with the subjacs, then converting the COO into a CSR matrix whenever the jacobian is needed.
    - The new approach is more memory efficient as in the `_build` it does a one-time build of a mapping data structure to go from the subjac data directly to the CSR matrix. This means that once the mapping has been built the COO matrix can be removed from memory and for jacobian calculations the updated subjacs can be directly inserted into the CSR matrix with the mapping.
    - The extra compute time to create the mapping data structure is only a one time cost and I've found it to be very negligible compared to the rest of the problem run time. On the other hand, I've found that this fix reduced the memory usage for some large Dyreqt problems by about 15%.

2. **CSCMatrix**
    - All of the points outlined for `CSRMatrix` also apply to `CSCMatrix`

openmdao/matrices/csc_matrix.py
```diff
@@ -13,31 +15,145 @@ class CSCMatrix(COOMatrix):
...
+        randgen : numpy.random.RandomState
+            Random number generator for complex step.
+        """
+        csc_indices = self._coo_to_csc_map[self._coo_slices[subjac.key]]
+        data = subjac.get_as_coo_data(randgen)
+        if subjac.factor is not None:
+            data = data * subjac.factor
+        if self._has_within_subjac_duplicates[subjac.key]:
+            # Rare case: within-subjac duplicate (row, col) entries require unbuffered add
+            np.add.at(self._matrix.data, csc_indices, data)
```
naylor-b, line 120: This is minor since this situation is rare, but np.bincount is faster than np.add.at when you have non-negative 1D index arrays.

naylor-b (approved):

rob-hetterich:
So when I did this refactor there were no existing tests that were directly for the matrix objects, just indirect tests like `test_jacobian.py`. Should I built a suite of tests that are directly testing the matrix objects? Or are we still fine with the indirect tests?

openmdao/matrices/csc_matrix.py
rob-hetterich: `np.bincount` isn't complex safe though while `np.add.at` is complex safe, so I think it would break when doing a problem level complex step

rob-hetterich:
> So when I did this refactor there were no existing tests that were directly for the matrix objects, just indirect tests like `test_jacobian.py`. Should I built a suite of tests that are directly testing the matrix objects? Or are we still fine with the indirect tests?

I ended up adding the tests for the matrix classes

robfalck (approved):

openmdao/matrices/csc_matrix.py
naylor-b: Turns out also that apparently newer versions of numpy have optimized add.at so that now there's basically no advantage to using bincount anyway.

Issue #3742, opened 2026-04-13
Look into using machine learning models for surrogate models


hschilling:
### Desired capability or behavior.

OpenMDAO should support using machine learning models for surrogate models easily.

PR #3766, opened 2026-05-27
DenseNNComp() & JaxFlaxComp() - neural network component and test

ehariton:
### Summary
Added two components for easy neural network exploration DenseNNComp() and JaxFlaxComp(). 

DenseNNComp() is a basic dense neural network component and test created to enable easier onboarding for machine learning community. Component uses Jax for derivatives and vmap for vectorization.

JaxFlaxComp() is a basic interface to a generic Flax-defined neural network meant to enable drop-in flax neural network capability.

### Related Issues

- Partial progress toward #3742 

None

### New Dependencies

None


openmdao/components/dense_nn_comp.py
```diff
@@ -0,0 +1,160 @@
...
+            values=(None, "sigmoid", "relu", "tanh"),
+        )
+
+    def setup(self):
+        n = self.options["vec_size"]
+
+        self._layer_sizes, self._weights, self._biases = self._load_weights()
+
+        self._num_inputs = self._layer_sizes[0]
+        self._num_outputs = self._layer_sizes[-1]
```
Kenneth-T-Moore: Don't need to save these in `self` since they are only used in setup.

openmdao/components/dense_nn_comp.py
```diff
@@ -0,0 +1,160 @@
...
+        if activation == "sigmoid":
+            return 1.0 / (1.0 + jnp.exp(-z))
+
+        elif activation == "relu":
+            return jnp.maximum(0.0, z)
+
+        elif activation == "tanh":
+            return jnp.tanh(z)
+
+        else:
```
Kenneth-T-Moore, line 140: The OptionsDictionary validation prevents `activation` from being anything else, so this branch isn't necessary.

openmdao/components/dense_nn_comp.py
ehariton: layer_sizes, num_inputs, num_outputs all removed from self. weight and biases are used later in compute primal

openmdao/components/dense_nn_comp.py
ehariton: updates complete

ehariton:
@naylor-b @robfalck I need some recommendations on how you'd like to deal with declaring the inputs for compute_primal() in the generic JaxFlaxComp(). We need to dynamically assign these based on the users inputs.

robfalck:
> @naylor-b @robfalck I need some recommendations on how you'd like to deal with declaring the inputs for compute_primal() in the generic JaxFlaxComp(). We need to dynamically assign these based on the users inputs.

If we're just going to wrap a Flax NN in a JaxExplicitComponent, maybe it would make more sense to do this as a documentation example rather than a generic capability?

We do acknowldge that the compute_primal implementation is a little limited in this generic "variable number of arguments" wrapping and we'll probably just go with a standard `compute(inputs, outputs)` interface in OpenMDAO 4.

naylor-b:
JaxExplicitComponent already overrides compute, so we could probably do something in there to make the actual call to jax a little more flexible, maybe passing an inputs dict and expecting compute_primal to return an outputs dict or something like that, and use an option on the component to control which calling convention we want to use.  It might be worth trying this out in OpenMDAO 3 to see how well it works since we're thinking of doing it in om4.

PR #3775, opened 2026-06-11
CI workflow and doc updates

robfalck:
### Summary

Doc and test scripts were installing from wheel, but this is not possible on forks of the OpenMDAO/OpenMDAO repo.
These changes allow running of the tests on forks while still testing the wheel install on the main repo.

Docs were failing and issues with ipyparallel finally drove me to drop jupyter-book and ipyparallel.

#### Doc Overhaul

- Added `openmdao.devtools.build_docs` command line tool.
- Doc notebooks are now copied from the _source_ openmdao_book directory into the executed using papermill. We could use nbconvert here, but papermill is a bit more capable.
- Sphinx is used to build the docs from the notebooks with execution set to "off" (it just builds docs with the notebooks that papermill executed).
- The doc building script executes all serial notebooks with a parallel pool.
- The MPI-relatated notebooks are detected by having a metadata entry `mpi: True` in their notebook metadata. These notebooks (there are 12) are run in series to make sure each gets the necessary resources.
- Created an `mpi_exec` function that is to be used to run MPI commands in the notebooks. This lets us demonstrate the use of MPI without the need of ipyparallel.
- All of these changes bring the doc build process down to ~5 minutes on a 16 core m4, and about 20 minutes on CI (limiting it to 3 procs).
- Updated documentation of documentation to explain how things work now.
- Added pixi task `install-julia-deps`, since julia is currently necessary to build the docs.

We could potentialy let the docs actually build without mpi or julia, for convenience. We would need to make sure `build_docs` returns a nonzero exit code so CI would fail the doc build in that case. Currently this idea is not implemented though.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):

rob-hetterich:
I'm looking through the built docs artifact in the CI because I wanted to look and see if things were still building as intended with the changes. I'm noticing a few things in the documents that I wanted to call out. Some of these changes we might not care about that much and some we might care more about.

1. It looks like the way it is treating code cells with remove-input and remove-output tags is a bit different. If you look at this "Getting Started" in the docs for example, the inputs and outputs that should be removed are included as drop downs rather than not being shown at all (which is what is intended with the tags). I think the issue might just be with cells that have one of the tags because cells with both remove-input and remove-output seem to be hidden fine.

2. Another thing that I noticed is that when we use glue and grab a class definition, it no longer shows as a drop down, it's just the full block of code with no way to hide it. The SellarDis1 in this example originally was a drop down which defaulted to being closed but is no longer a dropdown in the new build.

3. I don't know if it's related to your changes or something else, but in the implicit_with_balancecomp.html page the N2 at the end of the page displays fine in the current deployed docs but in the artifact the N2 is broken.

4. The "Welcome to OpenMDAO" page lists every single documentation page at the bottom. Also the sidebar does not display the documentation hierarchically, it just displays all of the links at once.
image

robfalck:
> I'm looking through the built docs artifact in the CI because I wanted to look and see if things were still building as intended with the changes. I'm noticing a few things in the documents that I wanted to call out. Some of these changes we might not care about that much and some we might care more about
> 
> It looks like the way it is treating code cells with remove-input and remove-output tags is a bit different. If you look at this "Getting Started" in the docs for example, the inputs and outputs that should be removed are included as drop downs rather than not being shown at all (which is what is intended with the tags). I think the issue might just be with cells that have one of the tags because cells with both remove-input and remove-output seem to be hidden fine: image

This behavior hasn't changed, it looks the same on the latest version of the published docs.  In the case of getting started, I agree that they should be "remove" and not "hide".

> 
> Another thing that I noticed is that when we use glue and grab a class definition, it no longer shows as a drop down, it's just the full block of code with no way to hide it. The SellarDis1 in this example originally was a drop down which defaulted to being closed but is no longer a dropdown in the new build: image
> 
> I don't know if it's related to your changes or something else, but in the implicit_with_balancecomp.html page the N2 at the end of the page displays fine in the current deployed docs but in the artifact the N2 is broken: image

robfalck:
> 3\. I don't know if it's related to your changes or something else, but in the implicit_with_balancecomp.html page the N2 at the end of the page displays fine in the current deployed docs but in the artifact the N2 is broken.

So this is actually a security issue in switching to using Ipython IFrame to view the n2 diagrams. They work when served but don't work when just viewed as local files (file://).  To get around this, I added a `view` subcommand to the build_docs devtool.  So `python -m openmdao.devtools.build_docs view --port 9000` will spin up an http server and show the docs as they are when hosted.

You can also point it at a different directory in case you're viewing the doc artifact from github: ` python -m openmdao.devtools.build_docs view --path ~/Downloads/artifact-clean `

robfalck:
> 4\. The "Welcome to OpenMDAO" page lists every single documentation page at the bottom. Also the sidebar does not display the documentation hierarchically, it just displays all of the links at once.

Toctree issue fixed. Juypterbook took the entire toctree from one user defined file, in sphinx the user is expected to put the locally-relative toctree in each section heading notebook.

robfalck:
> 2\. Another thing that I noticed is that when we use glue and grab a class definition, it no longer shows as a drop down, it's just the full block of code with no way to hide it. The SellarDis1 in this example originally was a drop down which defaulted to being closed but is no longer a dropdown in the new build.

This was another thing that got changed by the slight change in the openmdao theme.  It's been fixed.

rob-hetterich:
I can't quite look through the 15,000 lines of changes, but I did the next best thing and looked over the built, archived docs after your updates and everything looks good. Approving.

rob-hetterich (approved):

PR #3778, opened 2026-06-16
Make Egor optimizer available through the modOptDriver

### LINKED ISSUE #3772: `modOptDriver` makes `modOpt` always handle constraints as double-sided constraints.
### Description

A constraint like `c >= 0` will be define as `con['c'] = {'lower': 0, 'upper': 1e30 }`
while `modOpt` would better work with `con['c'] = {'lower': 0, 'upper': n.inf }` 

The issue comes from the fact that OpenMDAO uses `INF_BOUND = 1e30` as infinite marker while modOpt uses `np.inf`.

### Example

This slightly modified modOptDriver test fails

```python
    def test_simple_paraboloid_lower(self):
        """
        Test modOptDriver with a simple lower bound constraint.

        Verifies that SLSQP correctly optimizes a paraboloid problem with a
        lower bound constraint (con >= lower_bound). The optimal solution
        should satisfy the constraint.
        """
        prob = om.Problem()
        model = prob.model

        model.add_subsystem('p1', om.IndepVarComp('x', 50.0), promotes=['*'])
        model.add_subsystem('p2', om.IndepVarComp('y', 50.0), promotes=['*'])
        model.add_subsystem('con', om.ExecComp('c = x + y'), promotes=['*'])

        prob.driver = modOptDriver(optimizer='SLSQP')
        prob.driver.options['disp'] = False
        prob.driver.options['turn_off_outputs'] = True

        model.add_design_var('x', lower=-50.0, upper=50.0)
        model.add_design_var('y', lower=-50.0, upper=50.0)
        model.add_objective('f_xy')
        model.add_constraint('c', lower=-15.0)

        prob.setup()

        prob.run_driver()

        self.assertEqual(prob.driver._mo_prob.nl_con_bounds["c"]["lower"], -15.0)
        self.assertEqual(prob.driver._mo_prob.nl_con_bounds["c"]["upper"], np.inf)   # ERROR
```

This test can be fixed in [modOptDriver](https://github.com/OpenMDAO/OpenMDAO/blob/6d45412a0ee6ee98d1b0a097b50166588bfe4a64/openmdao/drivers/modopt_driver.py#L952-L956) with something like:`

```
		nl_con_bounds[name] = {
			'lower': lower_con[name] if lower_con[name] > -INF_BOUND else -np.inf,
			'upper': upper_con[name] if upper_con[name] < INF_BOUND else np.inf,
			'size': size
		}
```
to be duplicated for other kinds of constraints

### OpenMDAO Version

3.44

_No response_

### LINKED ISSUE #3780: modOptDriver `disp` option does not work when a boolean is given
### Description

While integrating Egor (PR #3778 ), I noticed the handling of `disp` as a boolean does not work due to erroneous int type detection in [`_setup_verbosity`](https://github.com/OpenMDAO/OpenMDAO/blob/a2cc3cc9d5afcf99456270d6252bcd62ac2a7e50/openmdao/drivers/modopt_driver.py#L666) of the `modOptDriver`

```python 
if isinstance(disp, int):    # always True when disp is a boolean as bool is a subclass of int
  self.opt_settings['verbose'] = disp 
else:
  self.opt_settings['verbose'] = 2 if disp else 0  # For Egor we want info log level when disp is True
```

A fix is to write: `if type(disp) is int:`

### Example

Exemple with Egor setting 
```python
import numpy as np
import openmdao.api as om
from openmdao.test_suite.components.sellar import SellarNoDerivatives

prob = om.Problem()
model = prob.model = SellarNoDerivatives(nonlinear_solver=om.NonlinearBlockGS,
                                         linear_solver=om.ScipyKrylov)

prob.driver = om.modOptDriver(optimizer='Egor')
prob.driver.options["maxiter"] = 30
prob.driver.options["disp"] = True   # NOT WORKING while it works with `= 2`
prob.driver.opt_settings["seed"] = 42

model.add_design_var('z', lower=np.array([-10.0, 0.0]), upper=np.array([10.0, 10.0]))
model.add_design_var('x', lower=0.0, upper=10.0)
model.add_objective('obj')
model.add_constraint('con1', upper=0.0)
model.add_constraint('con2', upper=0.0)

prob.setup(check=False, mode='rev')
prob.run_driver()
```

### OpenMDAO Version

3.44.1-dev

_No response_

relf:
### Summary

As [Egor is now integrated within modOpt](https://github.com/LSDOlab/modopt/pull/8), we can make it available for  OpenMDAO.

### Related Issues

- Resolves #3772 
- Resolves #3780 

None

### New Dependencies

None

### DIFF
--- a/openmdao/drivers/modopt_driver.py
+++ b/openmdao/drivers/modopt_driver.py
@@ -687,7 +687,7 @@ class modOptDriver(Driver):
         # CVXOPT and ConvexQPSolvers are rejected at runtime due to Hessian requirements
         if opt == 'IPOPT':
             if 'print_level' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['print_level'] = min(max(disp, 0), 12)
                 else:
                     self.opt_settings['print_level'] = 5 if disp else 0
@@ -695,7 +695,7 @@ class modOptDriver(Driver):
         elif opt == 'SNOPT':
             if 'major print level' not in opt_keys_lower \
                     and 'minor print level' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     level = min(max(disp, 0), 10)
                     self.opt_settings['Major print level'] = level
                     self.opt_settings['Minor print level'] = level
@@ -705,35 +705,35 @@ class modOptDriver(Driver):
 
         elif opt == 'TrustConstr':
             if 'verbose' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['verbose'] = min(max(disp, 0), 3)
                 else:
                     self.opt_settings['verbose'] = 1 if disp else 0
 
         elif opt == 'PySLSQP':
             if 'iprint' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['iprint'] = disp
                 else:
                     self.opt_settings['iprint'] = 1 if disp else 0
 
         elif opt == 'LBFGSB':
             if 'iprint' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['iprint'] = disp
                 else:
                     self.opt_settings['iprint'] = 0 if disp else -1
 
         elif opt == 'OpenSQP':
             if 'verbosity' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['verbosity'] = disp
                 else:
                     self.opt_settings['verbosity'] = 1 if disp else 0
 
         elif opt == 'Egor':
             if 'verbose' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['verbose'] = disp
                 else:
                     self.opt_settings['verbose'] = 2 if disp else 0
@@ -742,7 +742,7 @@ class modOptDriver(Driver):
             # Most SciPy-based optimizers (SLSQP, COBYLA, COBYQA, BFGS, NelderMead)
             # use 'disp' as boolean
             if 'disp' not in opt_keys_lower:
-                if isinstance(disp, int):
+                if type(disp) is int:
                     self.opt_settings['disp'] = disp > 0
                 else:
                     self.opt_settings['disp'] = disp
--- a/openmdao/drivers/modopt_driver.py
+++ b/openmdao/drivers/modopt_driver.py
@@ -20,13 +20,14 @@ Gradient-Free:
     - COBYLA: Constrained Optimization BY Linear Approximation
     - COBYQA: Constrained Optimization BY Quadratic Approximation
     - NelderMead: Nelder-Mead simplex algorithm (unconstrained)
+    - Egor: Efficient Global Optimization in Rust (bayesian optimization)
 
 Notes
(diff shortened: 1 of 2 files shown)


robfalck:
Thanks for the update!  We had a doc-building issue blocking things. I'll merge this out and put it through CI now.

We have an upcoming change on how we handle INF_BOUND in various optimizers, but we'll accept this first and then build on top of that.

relf:
The cobyla failing test passes on my windows machine (I get c = -7e-5 instead of -1.7e-4 in CI test) with the same numpy and scipy version.

In the last commit, I try to set a huge upper bound 1e29 to see if it makes the test pass (to check if it could relate to the bound handling changes). In any case, I would revert this last commit and I think I would just increase the tolerance to 2e-4. 

@robfalck, @rob-hetterich , what do you think?

robfalck:
I agree. Interesting I saw the same behavior in the inf_bound-refactor that I'm working on. Go with 2.e4 and I'll apply my PR once you're done here.

relf:
Ok, adding upper=1e29 makes the original test pass! 🤦‍♂️ . @robfalck, I've increased the test tolerance.

robfalck (approved):

rob-hetterich:
Does the modopt implementation of egor support discrete variables? That would require some additional updates to the modopt driver since I originally set it up with the assumption that I was only using continuous variables

relf:
The current Egor wrapper in modopt considers only continuous variables. Actually I didn't think about making it work for discrete variables as there was no handling of discrete variables in modopt. 
But I agree we could pass variable types to Egor and enable the handling of discrete variables (without modopt being aware)

robfalck:
I'd debated whether it was worth having an example of its use in the docs or tests, and adding it as a dependency to the pixi environment. I think it's acceptable to say "you can use this but you have to pip install it", but if you would prefer that it worked out-of-the-box we could include those.

relf:
Well... If you ask me, sure it would be great if it could just work out-of-the-box! But I understand you may not want to manage the egobox dependency, so pip install instructions works for me as well.

relf:
As an example, Egor can optimize the classic Sellar problem:

```python
import numpy as np
import openmdao.api as om
from openmdao.test_suite.components.sellar import SellarNoDerivatives

prob = om.Problem()
model = prob.model = SellarNoDerivatives(nonlinear_solver=om.NonlinearBlockGS,
                                         linear_solver=om.ScipyKrylov)

prob.driver = om.modOptDriver(optimizer='Egor')
prob.driver.options["maxiter"] = 30
prob.driver.options["disp"] = 2
prob.driver.opt_settings["seed"] = 42

model.add_design_var('z', lower=np.array([-10.0, 0.0]), upper=np.array([10.0, 10.0]))
model.add_design_var('x', lower=0.0, upper=10.0)
model.add_objective('obj')
model.add_constraint('con1', upper=0.0)
model.add_constraint('con2', upper=0.0)

prob.setup(check=False, mode='rev')
prob.run_driver()
```

relf:
@robfalck, @rob-hetterich, I think I'm all set with this PR. Let me know if there is anything that needs to be changed or added.

robfalck (approved):

rob-hetterich (approved):
Looks fine to me. Confirmed that I am able to run the driver with egor if I update my modopt install. I just added a more descriptive error raise to the modOptDriver for if an optimizer a user is trying to use is not available in their current install environment. Also Rob^2 talked and we'd rather leave egobox out of our direct dependencies since modopt does not include it in its default install. We are just sticking with what modopt includes by default.

robfalck:
Done, thanks!

PR #3781, opened 2026-07-01
Add a bounds-based autoscaler, minor driver clean-up

johnjasa:
### Summary

I've updated this PR to use the autoscaler capability. Here we add an autoscaler based on the bounds; when enabled, design variables (and their bounds) are rescaled to `[0, 1]` using their declared lower/upper bounds before being handed the driver. This lets scalar step-size options such as COBYLA's `rhobeg` act uniformly across design variables with drastically different magnitudes (e.g. a plant size in thousands of kW alongside a 0-1 ratio). Requires finite bounds on all design variables; raises a clear error otherwise. Default behavior is unchanged.

Also cleans up SciPy version checks in scipy_optimizer.py for versions older than the required 1.13.

Also removes the check for discrete variables within the optimization problem when using `approx_totals` at the model level. This is useful to allow us to use Scipy's SLSQP when internal components pass discrete variables.

### Related Issues

- Resolves #

None

### New Dependencies

None


johnjasa:
Tagging @robfalck, let me know if you want to see any changes!

robfalck:
This feels like something that should be implemented with the newish Autoscaler capability, and then could be made available to any optimizer.

https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/drivers/autoscalers/autoscaler.py

johnjasa:
Thanks for the suggestion, @robfalck! It's a good idea. I've moved this capability to a new autoscaler class. Let me know if you want to see something different.

robfalck (approved):

Kenneth-T-Moore (approved):

PR #3787, opened 2026-07-13
Fix for a double unit conversion on design variables with different units than the model variable.

### LINKED ISSUE #3786: When a design variable has a different unit than the model, the value from the optimizer is double converted.
### Description

When a design variable has a different unit than the model source, there is a unit conversion that occurs when the optimizer provides a new value. This is being converted twice due to a logic error.

### Example

This bug was discovered by @tad2216
Example provided in PR

### OpenMDAO Version

3.44.1-dev and a few versions prior

_No response_

Kenneth-T-Moore:
### Summary

Fixes a small logic error that caused a design variable to be double-converted when set by the optimizer whenever its declared units were different than the model units.

### Related Issues

- Resolves #3786

None

### New Dependencies

None

### DIFF
--- a/openmdao/core/tests/test_driver.py
+++ b/openmdao/core/tests/test_driver.py
@@ -798,6 +798,36 @@ class TestDriver(unittest.TestCase):
         msg = "<model> <class Group>: Target for constraint x has no units, but 'ft' units were specified."
         self.assertEqual(str(context.exception), msg)
 
+    def test_units_bug(self):
+        # This tests a bug in the set_design_var method where a design variable was being double
+        # converted when it was set.
+        prob = om.Problem()
+        model = prob.model
+
+        eq = om.ExecComp(
+            ['y=x', 'obj=x'],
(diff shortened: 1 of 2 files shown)


openmdao/core/tests/test_driver.py
```diff
@@ -798,6 +798,36 @@ def test_units_error_messages(self):
...
+        eq = om.ExecComp(
+            ['y=x', 'obj=x'],
+            x={'units': 'm'},
+            y={'units': 'm'},
+            obj={'units': 'm'},
+        )
+
+        model.add_subsystem('comp1', eq, promotes=['*'])
+
+        prob.driver = om.pyOptSparseDriver(optimizer='SLSQP')
```
robfalck: Should test with Scipy optimization driver here

robfalck (changes requested):

naylor-b (approved):

robfalck (approved):

PR #3788, opened 2026-07-14
Fix intermittent MPI test timeouts on macOS runners

robfalck:
### Summary

The N_PROCS=8 tests (e.g., in test_parallel_fd.py) were swamping the available cores on MacOS, which do not yield to other ranks by default. This bogged the tests down and led to timeouts.

This is fixed in two ways:
- Sets OMPI_MCA_mpi_yield_when_idle=1 in the test step so waiting ranks yield the CPU instead of spinning at 100%.
- Caps testflo concurrency at `-n 2` on macOS to prevent stacking too many additional test workers on top of 8-rank MPI jobs.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/.github/actions/run_openmdao_tests_pixi/action.yml
+++ b/.github/actions/run_openmdao_tests_pixi/action.yml
@@ -184,6 +184,7 @@ runs:
           echo "============================================================="
           export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
           export OMPI_MCA_rmaps_base_oversubscribe=1
+          export OMPI_MCA_mpi_yield_when_idle=1
           export OMPI_MCA_btl=^openib
           echo "-----------------------"
           echo "Quick test of mpi4py:"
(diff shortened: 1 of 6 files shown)


naylor-b (approved):

.github/actions/run_openmdao_tests_pixi/action.yml
```diff
@@ -184,6 +184,7 @@ runs:
           echo "============================================================="
           export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
           export OMPI_MCA_rmaps_base_oversubscribe=1
+          export OMPI_MCA_mpi_yield_when_idle=1
```
Kenneth-T-Moore, line 187: Do you think we should restrict this setting to just MacOS?

robfalck: As I understand it, this should be the default behavior anyway, for some reason macos isn't respecting it.  Also, in general we really don't want procs just spinning and not yielding on any platform. Things are still working fine on ubuntu with this change.

Kenneth-T-Moore (approved):

PR #3790, opened 2026-07-14
Fix for a key-error for input-input connections that are created in subgroups.

Kenneth-T-Moore:
### Summary

Fix for a key-error for input-input connections that are created in subgroups. This area of the code is cleaning up after converting input-input connections to input-output connections. The manual connections storage isn't propagated to prob.model, so some connections leave nothing to clean up at the top.

### Related Issues

- Resolves #

None

### New Dependencies

None


openmdao/core/tests/test_conn_graph.py
```diff
@@ -109,6 +109,18 @@ def test_no_promotion(self):
...
+        prob = om.Problem()
+        model = prob.model
+
+        sub = model.add_subsystem('sub', om.Group())
+        sub.add_subsystem('c1', om.ExecComp('y=x'))
+        sub.add_subsystem('c2', om.ExecComp('y=x'))
+        sub.connect('c1.x', 'c2.x')
+
+        # Make sure it doesn't raise a Keyerror.
+        prob.setup()
```
robfalck, line 122: Could you please set a value of `c1.x` and call `run_model`. We should actually verify the connection is behaving as expected.

robfalck (changes requested):

robfalck (approved):

naylor-b (approved):

PR #3805, opened 2026-08-18
Adding jax non-dependent partials logic

johnjasa:
## Summary

Allow `JaxExplicitComponent` to skip JAX derivative computation entirely for outputs that don't need it instead of computing derivatives for every output and throwing away the ones that weren't requested.

## Problem

`JaxExplicitComponent` builds a single `jax.jacfwd`/`jax.jacrev` transform over a function that returns *all* continuous outputs, then discards any `(of, wrt)` entries that weren't declared via `declare_partials`. This was noted in the code itself:

```python
if key not in partials:
    # FIXME: this means that we computed a derivative that we didn't need
    continue
```

## Change

- `declare_partials(of='my_output', wrt='*', dependent=False)` now causes the JAX derivative transform to skip that output entirely, rather than just discarding its (still computed) result.
- If any outputs are excluded, `_update_jac_functs` builds a restricted version of `compute_primal` that returns only the wanted outputs, so `jax.jacrev`'s cost (proportional to output count) actually drops.

## Usage

```python
class MyComp(om.JaxExplicitComponent):
    def setup(self):
        self.add_input('x')
        self.add_output('y')
        self.add_output('debug_metric')  # don't care about derivatives of this

    def setup_partials(self):
        self.declare_partials('debug_metric', '*', dependent=False)

    def compute_primal(self, x):
        y = ...
        debug_metric = ...
        return y, debug_metric
```

## Testing

- Added `test_jax_dependent_false_output_excluded_from_derivs` in `openmdao/jax_funcs/tests/test_jax_explicit.py`, which verifies:
  - `_deriv_output_idxs`/`_deriv_output_names` correctly reflect the excluded output.
  - The excluded output's `(of, wrt)` pairs are absent from `_subjacs_info`.
  - `check_partials` passes for the remaining outputs.

## Files changed

- `openmdao/components/jax_explicit_comp.py`
- `openmdao/utils/jax_utils.py`
- `openmdao/jax_funcs/tests/test_jax_explicit.py`


robfalck (approved):
Approved, but seems like this needs to be merged  after #3806

Kenneth-T-Moore (approved):

Issue #2929, opened 2023-05-30
convergence issues of nonlinear solver when under FD


fzahle:
### Description

When using the `NonlinearBlockGS` class to solve an aero-elastic steady state, we observe that when under FD (approx_totals on the full problem, and yes I know it's a bad idea), the nonlinear solver for some cases doesn't converge. This may be because a given DV doesn't affect the nonlinear solution, and the solver therefore tries to converge something already converged, and fails to reduce the residual by the required orders of magnitude.

Applying the nudge implemented for CS to FD works and greatly reduces the number of failed solves and speeds up the overall workflow.

Two options:

* Add an `fd_reconverge` option
* rename `cs_reconverge` to `approx_reconverge` or something similar.

This would apply to the Broyden, Newton and GS solvers.

### Example

```python
code is part of a bigger, not currently sharable code base
```

### OpenMDAO Version

3.26

### Relevant environment information

N/A

robfalck:
If you could find the time to provide some code for a test case/example of this capability, it would help us to implement it.

PR #3809, opened 2026-08-19
Preserve independent outputs during NonlinearBlockGS complex step reconvergence

### LINKED ISSUE #3800: NonlinearBlockGS cs_reconverge shifts IndepVarComp outputs, so check_totals(method='cs') differentiates at the wrong real point
### Description

In OpenMDAO 3.45.0, `check_totals(method='cs')` can evaluate derivatives at a translated real design when a top-level `NonlinearBlockGS` with `cs_reconverge=True` owns `IndepVarComp` outputs.

`ComplexStep._run_point` adds `i*h` to the selected output and calls `run_solve_nonlinear()`. `NonlinearBlockGS._iter_initialize` then adds `norm(system._outputs) * 1e-10` to the complete output vector. Because `IndepVarComp._solve_nonlinear` is a no-op, that real nudge persists in independent/design outputs throughout the perturbed solve. The returned complex-step total is therefore evaluated at a nearby real point rather than the declared point. The discrepancy is deterministic and independent of `h` to leading order.

This appears related to #2929 and the nudge-accuracy discussion in #2839, but neither covers the `NonlinearBlockGS` / `IndepVarComp` wrong-base-point case.

### Minimal reproducer

```python
import openmdao.api as om


def run(extra):
    prob = om.Problem(reports=False)
    ivc = prob.model.add_subsystem("ivc", om.IndepVarComp(), promotes=["*"])
    ivc.add_output("x", 22.0)
    ivc.add_output("irrelevant", extra)  # deliberately disconnected
    prob.model.add_subsystem("sq", om.ExecComp("y = x**2"), promotes=["*"])
    prob.model.nonlinear_solver = om.NonlinearBlockGS()
    prob.model.nonlinear_solver.options["cs_reconverge"] = True
    prob.model.nonlinear_solver.options["use_aitken"] = False
    prob.run_model()
    data = prob.check_totals(
        of=["y"], wrt=["x"], method="cs", step=1e-16, out_stream=None
    )
    return data["y", "x"]["J_fd"].item()


print(run(0.0))  # 44.000000096899946; expected 44.0
print(run(1e8))  # 44.02; a disconnected output changes the derivative
```

A probe around the perturbed nonlinear solve records `Re(x)` changing from `22.0` to `22.000000048449973`. The derivative error, `9.689994584505257e-08`, is exactly twice the shift, as expected for `y=x**2`. With the disconnected `1e8` output, `Re(x)` changes by `0.01` and the derivative error becomes `0.02`. If the nonlinear solver is moved into a subgroup below the `IndepVarComp`, the shift disappears and the derivative is exactly `44.0`.

### Expected behavior

`check_totals(method='cs')` should preserve the real values of independent/design sources and return the derivative at the declared real point. A disconnected output should not change that derivative.

### Actual behavior

The reconvergence nudge changes every solver-owned output, including no-op `IndepVarComp` outputs, so the complex solve runs at a translated real design. Changing an unrelated output norm changes the reported derivative.

### Environment

- OpenMDAO 3.45.0 from PyPI
- Binary64 / NumPy complex allocation via `setup(force_alloc_complex=True)`
- `NonlinearBlockGS`, `cs_reconverge=True`, `use_aitken=False`

### Relevant source locations

- `openmdao/approximation_schemes/complex_step.py`: `_run_point`
- `openmdao/solvers/nonlinear/nonlinear_block_gs.py`: `_iter_initialize`
- `openmdao/core/indepvarcomp.py`: `_solve_nonlinear`

Possible directions include restricting the nudge to true solver-state outputs, restoring independent outputs before each nonlinear sweep, or using a forced-iteration approach analogous to the Newton change in #2839. I have not tested or selected among these fixes.

machismo0311:
## Summary

`NonlinearBlockGS` with `cs_reconverge=True` nudges the output vector under complex step so that an already converged model still iterates and propagates the imaginary perturbation. The nudge is currently sized from the full output vector and added to every output. Outputs tagged `openmdao:indep_var` (IndepVarComp and auto-IVC outputs) are externally prescribed and are never recomputed by a Gauss-Seidel sweep, so the real nudge on them is never erased: the whole perturbed solve runs at a translated real design point and the complex step total is evaluated at the wrong base. Because the nudge is sized from the full vector, a disconnected independent output can also change an unrelated derivative.

With the reproducer from #3800 (`y = x**2`, `x = 22`), `check_totals(method='cs')` returns `44.000000096899946` instead of `44.0`; setting a disconnected `IndepVarComp` output to `1e8` changes it to `44.02`. The error is exactly `2 * (norm(outputs) * 1e-10)`, is independent of the complex step size, and disappears if the solver is moved into a subgroup below the IndepVarComp.

## Change

During complex step reconvergence, outputs tagged `openmdao:indep_var` are now excluded from both the real nudge and the norm used to size it. Solver-reachable outputs (explicit outputs and implicit states) keep the existing reconvergence behavior.

Both exclusions are needed. Masking only the assignment fixes the base point but still lets a disconnected independent value inflate the nudge and change how many sweeps the solver runs, which changes the derivative on coupled models. Masking only the norm leaves the base point translated.

## Why `openmdao:indep_var`

This is OpenMDAO's existing marker for externally provided independent outputs. It is applied unconditionally by `IndepVarComp` and to auto-IVC outputs, is documented as user-applicable for "IndepVarComp-like" outputs, and is already what `list_indep_vars`, `get_indep_vars`, and the inputs report use to mean "a value the user provides". No new metadata is introduced.

## Convergence behavior

One behavioral consequence should be stated plainly. On current master, a disconnected independent output's magnitude contributes to the reconvergence nudge, which can force extra NLBGS sweeps. On strongly coupled problems whose imaginary fixed-point iteration has not actually converged at the requested tolerance, those accidental extra sweeps can happen to bring the derivative closer to the fully converged value. For example, on a two-component cycle with contraction 0.9 at default `atol`/`rtol`, master runs 59 sweeps under complex step and returns a total with about 2% error, while this patch runs 29 sweeps and returns about 47% error; converging the imaginary iteration to tolerance on that problem takes roughly 219 sweeps, so neither result is a converged derivative at those settings. Master's smaller error there comes entirely from the unrelated independent value inflating the nudge.

After this change, how far the imaginary iteration converges is governed by the solver's `atol`, `rtol`, and `maxiter` and by the model's own states, not by the magnitude of unrelated independent outputs. Users who need tight complex step totals through a strongly coupled NLBGS should tighten the solver tolerances, on both master and this branch. Real (non-complex) solves, iteration counts, and recorded cases are unchanged; the modified code is only reachable under complex step with `cs_reconverge=True`.

## Tests

`TestNLBGSComplexStepReconverge` in `test_nonlinear_block_gs.py` covers: the exact #3800 reproducer (fails on master with the values above), disconnected-output invariance across several magnitudes, a vector-valued independent variable (every element protected), auto-IVC outputs, a coupled model that still reconverges (guards against removing the nudge), disconnected invariance on a coupled model (guards the norm masking), an implicit state under a nested Newton below the NLBGS, Aitken, repeated and interleaved cs/fd evaluations, and a direct check that the nudge leaves imaginary parts and independent reals untouched while shifting states by `norm(states) * 1e-10`. Eight of the ten fail on master; the coupled-reconvergence guard passes on master by design. The change was also validated under MPI (2 and 4 ranks, including a distributed IndepVarComp output with ranks owning zero elements) and against plausible broken variants (old behavior, inverted mask, deleted nudge, full-norm sizing, imaginary erasure, explicit-only or auto-only masking, first-element-only masking), each of which is caught. Targeted solver, derivative, IndepVarComp and auto-IVC suites pass; the full suite shows no failures attributable to this change.

## Scope

NLBGS only. FD behavior is unchanged (the nudge is gated on complex step). `NewtonSolver` is unchanged; it has used a forced single iteration under complex step since #2839 and does not nudge. `BroydenSolver` contains analogous nudge code and is intentionally left out of this PR; it should be handled separately. This PR does not implement the FD reconvergence discussed in #2929, but the retained nudge mechanism would carry over to that work unchanged.

## References

Fixes #3800


robfalck (approved):
This looks good, thank you for the contribution.

Kenneth-T-Moore (approved):

